PR #2901, opened 2021-07-08
[CIAPP] Support `GIT_URL_1` to extract repository url info in tests executed in Jenkins

drodriguezhdez:
Sometimes multiple repositories can be involved (e.g. forks). Jenkins can use the pattern `GIT_URL_N` with `N` {1,2,…}. 
By convention, `GIT_URL_1` should be the same as `GIT_URL`.


nachoBonafonte (approved):

tylerbenson (approved):
Does the PR title make sense as a line in the release notes?

drodriguezhdez:
> Does the PR title make sense as a line in the release notes?

probably it does not

PR #2903, opened 2021-07-09
Use DDAgentFeaturesDiscovery to enable/disable metrics aggregator

richardstartin:
This fixes a bug which would occur if the application started up before a trace agent with tracer metrics enabled, and the metric sink would fail validation, disabling metrics aggregation. The tracer would eventually receive a response from the trace agent once it would start up, and find that it supports metrics, setting a header to indicate that tracer metrics had already been computed, bypassing metrics aggregation in the trace agent. This is remedied by making `DDAgentFeaturesDiscovery` the golden source of trace agent feature availability, making inconsistency impossible.


mcculls (approved):

tylerbenson (approved):
In the description can you describe the sequence of events that caused problems and how this fixes it?

PR #2910, opened 2021-07-12
Fix/improve EventDispatcher subscriber lookup

cataphract:


dd-java-agent/appsec/src/jmh/java/datadog/appsec/benchmark/EventDispatcherBenchmark.java
```diff
@@ -0,0 +1,103 @@
...
+        if (iUnused < numUnusedSubscribers) {
+          iUnused++;
+          int i = iTotal++;
+          doSubscribe(unusedAddress, i);
+        }
+      }
+    }
+
+    private void doSubscribe(Address<?> address, int i) {
+      final OrderedCallback.Priority priority = OrderedCallback.Priority.values()[i % 4];
```
richardstartin, line 74: Watch out because `OrderedCallback.Priority.values()` allocates every time

richardstartin: And since only 2 bits are reserved for the priority, you might as well avoid an integer division (>30 cycles) and do `i & 3` here.

cataphract: This happens on setup phase of the benchmark so it doesn't affect the results, but good to know, I wasn't aware of that fact.

dd-java-agent/appsec/src/main/java/com/datadog/appsec/event/EventDispatcher.java
```diff
@@ -33,35 +39,39 @@ public EventDispatcher() {
...
-      int[] newArray = new int[newList.size()];
+      newList.sort(OrderedCallback.CallbackPriorityComparator.INSTANCE);
+      char[] newArray = new char[newList.size()];
       for (int i = 0; i < newArray.length; i++) {
-        newArray[i] = dataListenersIdx.indexOf(newList.get(i));
+        DataListener listener = newList.get(i);
+        char indexAndPriority =
+            (char)
+                (dataListenersIdx.indexOf(listener)
+                    | listener.getPriority().ordinal() << PRIORITY_SHIFT);
```
ValentinZakharov, line 73: I'm sorry, I can see you set priority bits, but it never actually use, because in `publishEvent` you just get rid of priority bits to get actual index. It seems, we can to not store priority bits.
Please, correct me, if I missed something.

cataphract: They are used in the bitset. We iterate the set bits with `nextSetBit(int)`, going from the least significant bits to the more significant bits, the bit indexes here being `listener_index | priority << 14`. This means that the highest priority listeners are found first, because `Priority.HIGHEST` contributes the bits `0 << 14` (that is, nothing), `Priority.HIGH` `1 << 14` and so on.

ValentinZakharov (approved):

PR #2917, opened 2021-07-13
cheaper capturing of non-async-propagating scopes

richardstartin:
This prevents from allocating a new `State` object when we won't capture the scope, and is a necessary step towards making `Continuation.activate()` calls monomorphic when restoring context (i.e. make `NoopContinuation` redundant).


devinsba (approved):

mcculls (approved):
makes sense

PR #2924, opened 2021-07-13
Add 'root-servlet' as service name to avoid lengthy tomcat naming for '/'

lizapressman:


dd-trace-api/src/main/java/datadog/trace/api/ConfigDefaults.java
```diff
@@ -22,6 +22,7 @@
   public static final String DEFAULT_DOGSTATSD_SOCKET_PATH = "/var/run/datadog/dsd.socket";
   public static final int DEFAULT_AGENT_TIMEOUT = 10; // timeout in seconds
   public static final String DEFAULT_SERVICE_NAME = "unnamed-java-app";
+  public static final String DEFAULT_SERVLET_ROOT_CONTEXT_SERVICE_NAME = "/root";
```
devinsba, line 25: I think I would call this `root-servlet`

dd-java-agent/instrumentation/tomcat-5.5/src/main/java/datadog/trace/instrumentation/tomcat/TomcatDecorator.java
```diff
@@ -62,9 +62,10 @@ public AgentSpan onRequest(
       String contextPath = request.getContextPath();
       String servletPath = request.getServletPath();
 
-      if (null != contextPath && !contextPath.isEmpty()) {
-        span.setTag("servlet.context", contextPath);
+      if (null == contextPath || contextPath.isEmpty()) {
+        contextPath = Config.get().getRootContextServiceName();
```
devinsba, line 66: I think the original code is still needed so the use of the Config as a fallback should be an else block

devinsba (changes requested):

dd-trace-api/src/main/java/datadog/trace/api/ConfigDefaults.java
lizapressman: Do you mean name it DEFAULT_ROOT_SERVLET_CONTEXT_SERVICE_NAME or set it to "/root-servlet"? I assume the first but just want to double check

devinsba: I mean the second. And I don't think the `/` is required but if it is then with it

dd-java-agent/instrumentation/tomcat-5.5/src/main/java/datadog/trace/instrumentation/tomcat/TomcatDecorator.java
lizapressman: I'm not sure that's right. I inverted the nots here so it changes the name of contextPath to "/root" or whatever the config is if it's empty and sets the tag no matter what. So the span tag either gets set to the contextPath that was taken from the request or if there was no context path in the request it grabs the config.

devinsba: Wow missed the change of !, my bad

dd-trace-api/src/main/java/datadog/trace/api/ConfigDefaults.java
lizapressman: so removing the / doesn't break anything but in general the servlet context seems to include the / in front:

Screen Shot 2021-07-14 at 3 13 27 PM

so removing it from the default service name makes it slightly inconsistent. Should I leave it for consistency or remove?

devinsba: If we are going to keep that consistency then we should set it to `/` here since that is most correct and then use the config value in `TagInterceptor.interceptServletContext`

lizapressman: Sorry, what do you mean by use the config value in TagInterceptor.interceptServletContext

lizapressman: Currently the service name is getting propagated correctly because I am updating the context path to the default and setting the tag to that context path. So the '/' trimming in TagInterceptor.interceptServletContext is trimming the value as if "/root" is just any typical context path it's given

lizapressman: Screen Shot 2021-07-14 at 3 28 50 PM

Heres what I'm seeing on the trace. service name is root-servlet and servlet context is /root-servlet

devinsba: Sorry I guess what I'm saying is, the correct value for the `servlet.context` would really be `/` in this case, but obviously that doesn't work once the tag gets to the interceptor because it won't affect the service name. So we would need to detect `/` as the tag value there and then substitute

lizapressman: ohhh so we want to see just "/" as context and service name should be "root-servlet"? I'll go ahead and modify that

devinsba: That makes the most sense to me as far as consistency goes. I don't think we want the context to be a path that doesn't exist

dd-trace-api/src/main/java/datadog/trace/api/ConfigDefaults.java
```diff
@@ -22,6 +22,7 @@
   public static final String DEFAULT_DOGSTATSD_SOCKET_PATH = "/var/run/datadog/dsd.socket";
   public static final int DEFAULT_AGENT_TIMEOUT = 10; // timeout in seconds
   public static final String DEFAULT_SERVICE_NAME = "unnamed-java-app";
+  public static final String DEFAULT_SERVLET_ROOT_CONTEXT_SERVICE_NAME = "/root-servlet";
```
devinsba, line 25: Because this is being used directly in `span.setServiceName(Config.get().getRootContextServiceName());` you'll want to drop the `/` from the beginning

devinsba (approved):
:+1: no need to wait for me to review again after the one quick change

richardstartin (approved):
👍🏻

PR #2925, opened 2021-07-14
namespace RunnableInstrumentation so it can be disabled in clojure apps

richardstartin:
This allows the `RunnableInstrumentation`, which is _mostly_ unnecessary now we instrument other things like `RunnableFuture`, to be disabled by users who understand the **potential** risk for broken traces, but also prevents us from transforming every `AFn` in clojure applications.

To disable the instrumentation, set `-Ddd.integration.runnable.enabled=false`


mcculls (approved):
👍

lizapressman (approved):

devinsba (approved):

antonmos:
Explanation of `potential risk` that i got from datadog support:

> The Runnable Instrumentation is an integration that our team is working on replacing in the future with instrumenting higher level APIs like Executors. In terms of the risk for broken traces - this would only be for customers using Kotlin Runnables and we will evaluate further instrumentation there in the future.

> Additionally, in a future release we do plan on having the Runnable instrumentation off by default, but before doing that we need to further understand the scope of this legacy instrumentation and the impact it’ll have on current customers.

PR #2926, opened 2021-07-14
prevent double registration of tracing interceptor when server builder is reused

richardstartin:
We have the same problem as OpenTelemetry had for server registration (https://github.com/open-telemetry/opentelemetry-java-instrumentation/pull/3544), but not for client registration where there was already a guard against double registration.

### DIFF
--- a/dd-java-agent/instrumentation/grpc-1.5/src/main/java/datadog/trace/instrumentation/grpc/server/GrpcServerBuilderInstrumentation.java
+++ b/dd-java-agent/instrumentation/grpc-1.5/src/main/java/datadog/trace/instrumentation/grpc/server/GrpcServerBuilderInstrumentation.java
@@ -3,6 +3,8 @@ package datadog.trace.instrumentation.grpc.server;
 import static datadog.trace.agent.tooling.bytebuddy.matcher.DDElementMatchers.extendsClass;
 import static datadog.trace.agent.tooling.bytebuddy.matcher.NameMatchers.named;
 import static datadog.trace.agent.tooling.bytebuddy.matcher.NameMatchers.namedOneOf;
+import static datadog.trace.bootstrap.CallDepthThreadLocalMap.incrementCallDepth;
+import static java.util.Collections.singletonMap;
 import static net.bytebuddy.matcher.ElementMatchers.isMethod;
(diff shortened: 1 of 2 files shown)


mcculls (approved):

dd-java-agent/instrumentation/grpc-1.5/src/main/java/datadog/trace/instrumentation/grpc/server/GrpcServerBuilderInstrumentation.java
```diff
@@ -61,16 +71,25 @@ public void adviceTransformations(AdviceTransformation transformation) {
...
     @Advice.OnMethodEnter(suppress = Throwable.class)
-    public static void onEnter(@Advice.This ServerBuilder<?> serverBuilder) {
-      int callDepth = CallDepthThreadLocalMap.incrementCallDepth(ServerBuilder.class);
-      if (callDepth == 0) {
-        serverBuilder.intercept(TracingServerInterceptor.INSTANCE);
+    public static boolean onEnter(@Advice.This ServerBuilder<?> serverBuilder) {
+      ContextStore<ServerBuilder, Boolean> interceptedStore =
+          InstrumentationContext.get(ServerBuilder.class, Boolean.class);
+      if (interceptedStore.get(serverBuilder) == null) {
+        int callDepth = incrementCallDepth(ServerBuilder.class);
```
devinsba, line 78: Can you add a test for why the call depth part is needed, from what I can tell the test you modified would pass fine without this. It looks to me like this protects against some sort of builder wrapping mechanism but that's not covered

richardstartin: Since the instrumentation only applies to a fixed set of implementations (which may change in the future to wrap one another), it's not possible to contrive.

devinsba: Ok so its a future proofing. Works for me

devinsba (approved):

PR #2927, opened 2021-07-14
only emit end checkpoints when a scope underwent a thread migration

richardstartin:
This fixes an issue where we see duplicate endTask events in akka-http, and will allow for other events to be eliminated (e.g. slick runnables).


devinsba (approved):

mcculls (approved):
Makes sense

PR #2941, opened 2021-07-19
Add printing mode for TimelineCheckpointer

luhenry:
This mode is used for generating test cases for the profiling backend.

The generated files are parsed for test cases, and the profiling backend
uses them to verify that it can reconstruct the timeline appropriately.
We need "real world" scenarios for the profiling backend to support
every possible cases of valid (or not so valid) sequence of checkpoints.

### DIFF
--- a/dd-java-agent/testing/src/main/groovy/datadog/trace/agent/test/TimelineCheckpointer.groovy
+++ b/dd-java-agent/testing/src/main/groovy/datadog/trace/agent/test/TimelineCheckpointer.groovy
@@ -3,6 +3,8 @@ package datadog.trace.agent.test
 import datadog.trace.api.Checkpointer
 import datadog.trace.api.DDId
 
+import java.io.File
+import java.io.PrintStream
 import java.util.concurrent.ConcurrentHashMap
 import java.util.concurrent.CopyOnWriteArrayList
 
@@ -55,7 +57,19 @@ class TimelineCheckpointer implements Checkpointer {
   }
 
   void printActivity() {
-    printTimeLine()
+    if (!encounterOrder.isEmpty()) {
+      if (Boolean.parseBoolean(System.getenv("TIMELINECHECKPOINTER_PRINT_PROFILING_TESTCASE"))) {
+        PrintStream out = new PrintStream(File.createTempFile("checkpointer-events", ".csv", null));
+        out.println("eventName,traceId,spanId,threadName")
+        for (Event event : encounterOrder) {
+          out.println(String.format("%s,%s,%s,%s", event.eventName, event.traceId, event.spanId, event.threadName))
+        }
+        out.flush()
+        out.close()
+      } else {
+        printTimeLine()
+      }
+    }
   }
 
   void printTimeLine() {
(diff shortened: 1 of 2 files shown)


dd-java-agent/testing/src/main/groovy/datadog/trace/agent/test/TimelineCheckpointer.groovy
```diff
@@ -55,7 +57,19 @@ class TimelineCheckpointer implements Checkpointer {
   }
 
   void printActivity() {
-    printTimeLine()
+    if (!encounterOrder.isEmpty()) {
+      if (Boolean.parseBoolean(System.getenv("TIMELINECHECKPOINTER_PRINT_PROFILING_TESTCASE"))) {
+        PrintStream out = new PrintStream(File.createTempFile("checkpointer-events", ".csv", null));
+        out.println("eventName,traceId,spanId,threadName")
+        for (Event event : encounterOrder) {
+          out.println(String.format("%s,%s,%s,%s", event.eventName, event.traceId, event.spanId, event.threadName))
```
richardstartin, line 65: I would add the raw flags here to avoid misinterpretation

richardstartin: Also, the thread name is great for the diagrams and I'm only aware of this happening for Kotlin coroutines, but some frameworks will mutate the thread name to reflect context, I think `Event` needs a thread id as well for your purposes.

dd-java-agent/testing/src/main/groovy/datadog/trace/agent/test/TimelineCheckpointer.groovy
```diff
@@ -55,7 +57,19 @@ class TimelineCheckpointer implements Checkpointer {
   }
 
   void printActivity() {
-    printTimeLine()
+    if (!encounterOrder.isEmpty()) {
+      if (Boolean.parseBoolean(System.getenv("TIMELINECHECKPOINTER_PRINT_PROFILING_TESTCASE"))) {
+        PrintStream out = new PrintStream(File.createTempFile("checkpointer-events", ".csv", null));
```
jbachorik, line 62: Perhaps using try-with-resources to make sure the output stream is closed in all circumstances would be good.

richardstartin: This is groovy :(

jbachorik: Dang, the diff looks like Java :)
Ok, then try/finally? Or that doesn't work as well? ;)

luhenry: Fixed.

dd-java-agent/testing/src/main/groovy/datadog/trace/agent/test/TimelineCheckpointer.groovy
luhenry: I added both.

richardstartin (approved):

jbachorik (approved):

PR #2944, opened 2021-07-20
Order property sources to be consistent and make external config highest priority

devinsba:


internal-api/src/main/java/datadog/trace/bootstrap/config/provider/ConfigProvider.java
```diff
@@ -198,10 +214,10 @@ public static ConfigProvider withPropertiesOverride(Properties properties) {
           new CapturedEnvironmentConfigSource());
     } else {
       return new ConfigProvider(
-          providedConfigSource,
           new SystemPropertiesConfigSource(),
           new EnvironmentConfigSource(),
           new PropertiesConfigSource(configProperties, true),
+          providedConfigSource,
```
devinsba, line 220: @tylerbenson Can you provide any context why this was highest priority?

luhenry (approved):

internal-api/src/main/java/datadog/trace/bootstrap/config/provider/ConfigProvider.java
tylerbenson: Well, the method name here is called `withPropertiesOverride`, so this change would be in conflict with the method name, but that isn't the case with `DDTracer.builder().withProperties(new Properties()).build()`.  I imagine this is rarely used, and only applies to `dd-trace-ot`, so I don't have a problem changing it (assuming the method name is also updated). Maybe consider marking as "breaking change"?

devinsba: I would really be nice to know how many customers are using this functionality..

I really don't think it's a good idea to let code overwrite configuration from the environment. The way many apps are deployed these days this would break assumptions of how things get configured. Though now that I write it down, the way we do our "collections" configs doesn't respect that at all

PR #2950, opened 2021-07-22
Google HTTP client checkpoints/propagation fix

richardstartin:
Despite using standard concurrency mechanisms, this instrumentation did custom context propagation, preventing it from getting out of the box checkpoint support.

Before:
```
Test worker:      |-startSpan/1-|-------------|-----------|-----------|
qtp1349869572-31: |-------------|-startSpan/2-|-endSpan/2-|-----------|
pool-1-thread-1:  |-------------|-------------|-----------|-endSpan/1-|
```

After:
```
Test worker:     |-startSpan/1-|-suspend/1-|----------|-------------|-----------|-----------|-----------|
pool-1-thread-1: |-------------|-----------|-resume/1-|-------------|-----------|-endSpan/1-|-endTask/1-|
qtp489592746-31: |-------------|-----------|----------|-startSpan/2-|-endSpan/2-|-----------|-----------|
```

The first commit is the minimal change required to fix, and the second simplifies the instrumentation so field injection isn't required.


dd-java-agent/instrumentation/google-http-client/src/main/java/datadog/trace/instrumentation/googlehttpclient/GoogleHttpClientInstrumentation.java
```diff
@@ -71,54 +60,47 @@ public void adviceTransformations(AdviceTransformation transformation) {
...
+        @Advice.This HttpRequest request, @Advice.Local("inherited") boolean inheritedScope) {
+      TraceScope activeScope = activeScope();
+      // detect if scope was propagated here by java-concurrent handling
+      // of async requests
+      if (activeScope instanceof AgentScope) {
+        AgentScope agentScope = (AgentScope) activeScope;
+        AgentSpan span = agentScope.span();
+        // reference equality to check this instrumentation created the span,
+        // not some other HTTP client
+        if (HTTP_REQUEST == span.getOperationName()) {
```
luhenry, line 73: Is this an optimisation to avoid creating another span or is this a correctness issue?

dd-java-agent/instrumentation/google-http-client/src/main/java/datadog/trace/instrumentation/googlehttpclient/GoogleHttpClientInstrumentation.java
```diff
@@ -71,54 +60,47 @@ public void adviceTransformations(AdviceTransformation transformation) {
...
-            span.setError(true);
-            span.setErrorMessage(response.getStatusMessage());
-          }
-
-          DECORATE.beforeFinish(span);
-          span.finish();
+        DECORATE.beforeFinish(span);
+        span.finish();
+      } finally {
+        if (!inheritedScope) {
```
luhenry, line 102: I assume that this was the issue you diagnosed on the call? You can't close the scope if you didn't create a new one at https://github.com/DataDog/dd-trace-java/pull/2950/files#diff-bc061095f361faeae35081cd6cd8842f50f89d122209329e8789ed6090be76c7R75, correct?

dd-java-agent/instrumentation/google-http-client/src/main/java/datadog/trace/instrumentation/googlehttpclient/GoogleHttpClientInstrumentation.java
```diff
@@ -127,36 +109,22 @@ public static void methodExit(
...
-    public static void methodEnter(@Advice.This final HttpRequest request) {
-      final AgentSpan span = startSpan(HTTP_REQUEST);
-
-      final ContextStore<HttpRequest, RequestState> contextStore =
-          InstrumentationContext.get(HttpRequest.class, RequestState.class);
-
-      final RequestState state = new RequestState(span);
-      contextStore.put(request, state);
+    public static AgentScope methodEnter(@Advice.This HttpRequest request) {
+      return activateSpan(DECORATE.prepareSpan(startSpan(HTTP_REQUEST), request));
```
luhenry, line 113: Why don't we do the same thing as https://github.com/DataDog/dd-trace-java/pull/2950/files#diff-bc061095f361faeae35081cd6cd8842f50f89d122209329e8789ed6090be76c7R65-R77 where we don't start a span and activate a scope if there is already a scope?

dd-java-agent/instrumentation/google-http-client/src/main/java/datadog/trace/instrumentation/googlehttpclient/GoogleHttpClientDecorator.java
```diff
@@ -29,6 +33,14 @@ protected URI url(final HttpRequest httpRequest) throws URISyntaxException {
     return new URI(fixedUrl);
   }
 
+  public AgentSpan prepareSpan(AgentSpan span, HttpRequest request) {
+    span.setMeasured(true);
+    DECORATE.afterStart(span);
+    DECORATE.onRequest(span, request);
+    propagate().inject(span, request, SETTER);
```
luhenry, line 40: Could you please elaborate on what this `propagate()` does?

luhenry (approved):

dd-java-agent/instrumentation/google-http-client/src/main/java/datadog/trace/instrumentation/googlehttpclient/GoogleHttpClientInstrumentation.java
richardstartin: We need to check the scope activated is the one we started on entry to `executeAsync`, because this could be called from anywhere. Reference equality means it's _the_ constant `HTTP_REQUEST` and not some user provided content (for whatever reason) or another instrumentation somehow calling `execute`.

dd-java-agent/instrumentation/google-http-client/src/main/java/datadog/trace/instrumentation/googlehttpclient/GoogleHttpClientInstrumentation.java
richardstartin: Yes, closing a scope twice leads to a negative reference count, which is an error.

dd-java-agent/instrumentation/google-http-client/src/main/java/datadog/trace/instrumentation/googlehttpclient/GoogleHttpClientInstrumentation.java
richardstartin: Because this method can't be called in many contexts - calling `executeAsync` starts a span for a new execution, but the method looks like this:

```java
    @Beta
    public Future<HttpResponse> executeAsync(Executor executor) {
        FutureTask<HttpResponse> future = new FutureTask(new Callable<HttpResponse>() {
            public HttpResponse call() throws Exception {
                return HttpRequest.this.execute();
            }
        });
        executor.execute(future);
        return future;
    }
```

which makes `HttpRequest.execute` ambiguous. Either it's a synchronous call from anywhere else, or it was triggered by `HttpRequest.executeAsync` and the span has already been started.

dd-java-agent/instrumentation/google-http-client/src/main/java/datadog/trace/instrumentation/googlehttpclient/GoogleHttpClientDecorator.java
richardstartin: This is a standard API for injecting headers (e.g. trace id, sampling priority) for distributed context propagation.

devinsba (approved):

tylerbenson (approved):

PR #2954, opened 2021-07-23
prevent leaking into flusher event loop

richardstartin:
Prevents propagation into this event loop:

```java
    private void runOnEventLoop() {
      assert eventLoop.inEventLoop();

      boolean didSomeWork = false;
      Write write;
      while ((write = writes.poll()) != null) {
        Channel channel = write.channel;
        channels.add(channel);
        channel.write(write.message, write.writePromise);
        didSomeWork = true;
      }

      for (Channel channel : channels) {
        channel.flush();
      }
      channels.clear();

      if (didSomeWork) {
        runsWithNoWork = 0;
      } else if (++runsWithNoWork > maxRunsWithNoWork) {
        // Prepare to stop
        running.set(false);
        // If no new writes have been enqueued since the previous line, we can return safely
        if (writes.isEmpty()) {
          return;
        }
        // Otherwise check if those writes have triggered a new run. If not, we need to do that
        // ourselves (i.e. not return yet)
        if (!running.compareAndSet(false, true)) {
          return;
        }
      }
      if (!eventLoop.isShuttingDown()) {
        eventLoop.schedule(this::runOnEventLoop, rescheduleIntervalNanos, TimeUnit.NANOSECONDS);
      }
    }
```

Leads to sensible checkpoints:

Before:

```
Test worker: |-startSpan/14-|-startSpan/15-|-suspend/15-|-----------|------------|------------|-----------|------------|------------|------------|--------------|------------|-----------|------------|------------|-----------|------------|------------|-----------|------------|------------|-----------|------------|------------|-----------|------------|-endSpan/14-|
s4-io-1:     |--------------|--------------|------------|-resume/15-|-suspend/15-|-endTask/15-|-resume/15-|-suspend/15-|-endTask/15-|-endSpan/15-|-startSpan/16-|-endSpan/16-|-resume/15-|-suspend/15-|-endTask/15-|-resume/15-|-suspend/15-|-endTask/15-|-resume/15-|-suspend/15-|-endTask/15-|-resume/15-|-suspend/15-|-endTask/15-|-resume/15-|-endTask/15-|------------|
```

After
```
Test worker: |-startSpan/14-|-startSpan/15-|-suspend/15-|-----------|------------|------------|--------------|------------|-endSpan/14-|
s4-io-1:     |--------------|--------------|------------|-resume/15-|-endTask/15-|-endSpan/15-|-startSpan/16-|-endSpan/16-|------------|
```

### DIFF
--- a/dd-java-agent/instrumentation/java-concurrent/src/main/java/datadog/trace/instrumentation/java/concurrent/AsyncPropagatingDisableInstrumentation.java
+++ b/dd-java-agent/instrumentation/java-concurrent/src/main/java/datadog/trace/instrumentation/java/concurrent/AsyncPropagatingDisableInstrumentation.java
@@ -74,7 +74,8 @@ public final class AsyncPropagatingDisableInstrumentation extends Instrumenter.T
             "io.grpc.internal.ServerImpl$ServerTransportListenerImpl",
             "okhttp3.ConnectionPool",
             "org.elasticsearch.transport.netty4.Netty4TcpChannel",
-            "org.springframework.cglib.core.internal.LoadingCache")
+            "org.springframework.cglib.core.internal.LoadingCache",
+            "com.datastax.oss.driver.internal.core.channel.DefaultWriteCoalescer$Flusher")
         .or(RX_WORKERS)
         .or(GRPC_MANAGED_CHANNEL)
         .or(REACTOR_DISABLED_TYPE_INITIALIZERS);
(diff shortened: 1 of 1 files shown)


luhenry:
IIUC, this avoids attributing code/cpu time to a span while the CPU wasn't executing anything related to the Span?

richardstartin:
Yes, scheduling of a duty cycle was happening in the context of an executing task, which would attribute running the duty cycle itself to whatever task was active at the time it was scheduled.

luhenry (approved):

richardstartin:
In the future it may be necessary for the span to propagate to `channel.write(write.message, write.writePromise);` so that  the associated work can be attributed to the span, at which point this will need modifying to target `eventLoop.schedule(this::runOnEventLoop, rescheduleIntervalNanos, TimeUnit.NANOSECONDS);` more specifically.

tylerbenson (approved):

tylerbenson (approved):

PR #2955, opened 2021-07-23
Add config option to disable context propagation for specified rabbit queues and exchanges

lizapressman:
Config to disable queues (from consumer side): `rabbit.propagation.disabled.queues`
Config to disable exchanges (from producer side): `rabbit.propagation.disabled.exchanges`

### DIFF
--- a/dd-java-agent/instrumentation/rabbitmq-amqp-2.7/src/main/java/datadog/trace/instrumentation/rabbitmq/amqp/TracedDelegatingConsumer.java
+++ b/dd-java-agent/instrumentation/rabbitmq-amqp-2.7/src/main/java/datadog/trace/instrumentation/rabbitmq/amqp/TracedDelegatingConsumer.java
@@ -15,6 +15,7 @@ import com.rabbitmq.client.AMQP;
 import com.rabbitmq.client.Consumer;
 import com.rabbitmq.client.Envelope;
 import com.rabbitmq.client.ShutdownSignalException;
+import datadog.trace.api.Config;
 import datadog.trace.api.DDTags;
 import datadog.trace.bootstrap.instrumentation.api.AgentScope;
 import datadog.trace.bootstrap.instrumentation.api.AgentSpan;
@@ -81,10 +82,21 @@ public class TracedDelegatingConsumer implements Consumer {
       final Context context =
           headers == null ? null : propagate().extract(headers, ContextVisitors.objectValuesMap());
 
-      final AgentSpan span =
-          startSpan(AMQP_COMMAND, context)
-              .setTag(MESSAGE_SIZE, body == null ? 0 : body.length)
-              .setMeasured(true);
+      final AgentSpan span;
+
+      // TODO: check dynamically bound queues
+      if (!Config.get().getRabbitPropagationDisabledQueues().contains(queue)) {
+        span =
+            startSpan(AMQP_COMMAND, context)
+                .setTag(MESSAGE_SIZE, body == null ? 0 : body.length)
+                .setMeasured(true);
+      } else {
+        span =
+            startSpan(AMQP_COMMAND, null)
+                .setTag(MESSAGE_SIZE, body == null ? 0 : body.length)
+                .setMeasured(true);
+      }
+
       CONSUMER_DECORATE.afterStart(span);
       CONSUMER_DECORATE.onDeliver(span, queue, envelope);
       final long spanStartTime = NANOSECONDS.toMillis(span.getStartTime());
--- a/dd-java-agent/instrumentation/rabbitmq-amqp-2.7/src/main/java/datadog/trace/instrumentation/rabbitmq/amqp/RabbitChannelInstrumentation.java
+++ b/dd-java-agent/instrumentation/rabbitmq-amqp-2.7/src/main/java/datadog/trace/instrumentation/rabbitmq/amqp/RabbitChannelInstrumentation.java
@@ -32,6 +32,7 @@ import com.rabbitmq.client.Consumer;
 import com.rabbitmq.client.GetResponse;
 import com.rabbitmq.client.MessageProperties;
 import datadog.trace.agent.tooling.Instrumenter;
+import datadog.trace.api.Config;
 import datadog.trace.bootstrap.CallDepthThreadLocalMap;
 import datadog.trace.bootstrap.instrumentation.api.AgentScope;
 import datadog.trace.bootstrap.instrumentation.api.AgentSpan;
@@ -162,7 +163,10 @@ public class RabbitChannelInstrumentation extends Instrumenter.Tracing {
         Map<String, Object> headers = props.getHeaders();
         headers = (headers == null) ? new HashMap<String, Object>() : new HashMap<>(headers);
 
-        propagate().inject(span, headers, SETTER);
+        if (Config.get().isRabbitPropagationEnabled()
+            && !Config.get().getRabbitPropagationDisabledExchanges().contains(exchange)) {
+          propagate().inject(span, headers, SETTER);
+        }
 
         props =
             new AMQP.BasicProperties(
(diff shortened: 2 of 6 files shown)


dd-java-agent/instrumentation/rabbitmq-amqp-2.7/src/main/java/datadog/trace/instrumentation/rabbitmq/amqp/TracedDelegatingConsumer.java
```diff
@@ -81,10 +82,21 @@ public void handleDelivery(
...
           headers == null ? null : propagate().extract(headers, ContextVisitors.objectValuesMap());
 
-      final AgentSpan span =
-          startSpan(AMQP_COMMAND, context)
-              .setTag(MESSAGE_SIZE, body == null ? 0 : body.length)
-              .setMeasured(true);
+      final AgentSpan span;
+
+      // TODO: check dynamically bound queues
+      if (!Config.get().getRabbitPropagationDisabledQueues().contains(queue)) {
```
tylerbenson, line 88: duplication could be reduced if you instead applied this check to the `context` var definition above.

Even better would be to put the config check into the constructor (since that's where `queue` is defined), then you have a simple boolean field.

dd-java-agent/instrumentation/rabbitmq-amqp-2.7/src/main/java/datadog/trace/instrumentation/rabbitmq/amqp/TracedDelegatingConsumer.java
```diff
@@ -81,10 +84,21 @@ public void handleDelivery(
...
           headers == null ? null : propagate().extract(headers, ContextVisitors.objectValuesMap());
 
-      final AgentSpan span =
-          startSpan(AMQP_COMMAND, context)
-              .setTag(MESSAGE_SIZE, body == null ? 0 : body.length)
-              .setMeasured(true);
+      final AgentSpan span;
+
+      // TODO: check dynamically bound queues
+      if (propagate) {
```
tylerbenson, line 90: the if block duplication can be removed by moving it to the context assignment:
`context = (headers == null || !propagate) ? null : ...`

dd-java-agent/instrumentation/rabbitmq-amqp-2.7/src/main/java/datadog/trace/instrumentation/rabbitmq/amqp/TracedDelegatingConsumer.java
```diff
@@ -79,12 +82,18 @@ public void handleDelivery(
...
-          headers == null ? null : propagate().extract(headers, ContextVisitors.objectValuesMap());
+          (headers == null || !propagate)
+              ? null
+              : propagate().extract(headers, ContextVisitors.objectValuesMap());
 
-      final AgentSpan span =
+      final AgentSpan span;
+
+      // TODO: check dynamically bound queues
+      span =
```
tylerbenson, line 92:
```suggestion
      final AgentSpan span =
```

dd-java-agent/instrumentation/rabbitmq-amqp-2.7/src/main/java/datadog/trace/instrumentation/rabbitmq/amqp/TracedDelegatingConsumer.java
```diff
@@ -79,12 +82,18 @@ public void handleDelivery(
...
       final Context context =
-          headers == null ? null : propagate().extract(headers, ContextVisitors.objectValuesMap());
+          (headers == null || !propagate)
+              ? null
+              : propagate().extract(headers, ContextVisitors.objectValuesMap());
 
-      final AgentSpan span =
+      final AgentSpan span;
+
+      // TODO: check dynamically bound queues
```
tylerbenson, line 91: What does this comment mean?  What is a "dynamically bound queue"?

tylerbenson (approved):

dd-java-agent/instrumentation/rabbitmq-amqp-2.7/src/main/java/datadog/trace/instrumentation/rabbitmq/amqp/TracedDelegatingConsumer.java
lizapressman: I wasn't entirely clear on that. Brian had me add that comment. @devinsba Do you mind explaining this better?

devinsba: The code as written supports queues and exchanges that are configured directly on the server...

but queues can create created and bound to an exchange using a combination of:

```
String queueName = channel.queueDeclare().getQueue()
channel.queueBind(queueName, exchangeName, routingKey)
```

Creates a new queue with a random name and binds it to an exchange, this case wouldn't be supported by what is currently written. We would need to capture the calls to queueBind and check the exchange name then dynamically add that queue name to the list of queues to ignore.

lizapressman: Should I update the comment to be more descriptive? Or maybe just throw a link to this slack thread in the TODO?

devinsba: Adding a link works for me

dd-java-agent/instrumentation/rabbitmq-amqp-2.7/src/main/java/datadog/trace/instrumentation/rabbitmq/amqp/TracedDelegatingConsumer.java
```diff
@@ -35,10 +36,12 @@
   private final String queue;
   private final Consumer delegate;
   private final boolean traceStartTimeEnabled;
+  private final boolean propagate;
 
   public TracedDelegatingConsumer(
       final String queue, final Consumer delegate, boolean traceStartTimeEnabled) {
     this.queue = queue;
+    this.propagate = !Config.get().getRabbitPropagationDisabledQueues().contains(queue);
```
devinsba, line 44: You didn't include `Config.get().isRabbitPropagationEnabled()` here, is that intentional?

lizapressman: That's actually a really good point that should've been addressed with kafka and jms. Not including it on the consumer side is consistent with both of those implementations but I'm not sure why. The reason I didn't include it in the consumer side for Kafka was because this config already existed there and was only being used on the producer side so I just didn't change it. So basically that isBlankPropagationEnabled config is only to enable or disable on the producer side and that's how I kept it. Do we want to change this though? Should this be a config that disables on both sides?

lizapressman: If so, I should probably change this for kafka and jms as well so that the configs are consistent across all the instrumentations

lizapressman: Isn't it the case though that if we break one side (consumer of producer), it technically breaks both? So it seems like it wouldn't be strictly necessary to check the consumer side as well, right?

devinsba: Right, ok, in the interest of consistency I think we can leave this as is.

devinsba (approved):

PR #2959, opened 2021-07-26
Add a checkpoint sequence validator

jbachorik:
This PR adds specialised validators for the checkpoint event sequences.

The validators are supposed to detect sequences which do not conform to the set of rules:
* for each involved thread the encountered span intervals (between `spanStart` and `spanEnd`) are not overlapping
* for each involved thread the initial checkpoint must be `startSpan` or `resume`
* for each involved thread the final checkpoint the final checkpoint must be either `endTask`, `endSpan` or `suspend`
* in the global span sequence (disregarding thread placements) 
   * a `resume` event must happen after `suspend`
   * a `suspend` event must happen after `startSpan` or `resume`
   * each `endTask` event must correspond either to `startSpan` or `resume` event
   * each `resume` event must correspond to a `suspend` event
   * after `endSpan` only `endTask` and `resume` events are allowed
   
By default, the validation is performed but the tests will not not be affected by the failing validation.
One needs to set `VALIDATE_CHECKPOINTS` environmental variable in order for the invalid checkpoint sequences to fail the tests


dd-java-agent/testing/src/main/groovy/datadog/trace/agent/test/checkpoints/TimelineCheckpointer.groovy
```diff
@@ -0,0 +1,55 @@
...
+class TimelineCheckpointer implements Checkpointer {
+
+  private static String[] emptySpaces = new String[128]
+
+  private final ConcurrentHashMap<DDId, List<Event>> spanEvents = new ConcurrentHashMap<>()
+  private final ConcurrentHashMap<String, List<Event>> threadEvents = new ConcurrentHashMap<>()
+  private final List<Event> orderedEvents = new CopyOnWriteArrayList<>()
+
+  @Override
+  void checkpoint(DDId traceId, DDId spanId, int flags) {
```
luhenry, line 18: Could we also capture the stacktrace for when the checkpoint is captured and display it when the sequence is invalid?

luhenry (approved):

dd-java-agent/testing/src/main/groovy/datadog/trace/agent/test/checkpoints/TimelineCheckpointer.groovy
jbachorik: While technically it is not difficult the stacktraces coming from the tests are very messy - they contain a ton of stack frames coming from spock, groovy and mockito which totally obscure the interesting ones. And there does not seem to be an easy way to filter them out in a way which would generate a comprehensible stacktrace.
We can return to this later when we decide to spend more time on figuring out how to properly represent the stacktrace in the context of a highly mocked test run.

dd-java-agent/testing/src/main/groovy/datadog/trace/agent/test/checkpoints/TimelineCheckpointer.groovy
```diff
@@ -0,0 +1,51 @@
...
+    def invalidEvents = TimelineValidator.validate(spanEvents, threadEvents, orderedEvents)
+    if (!invalidEvents.empty) {
+      System.err.println("=== Invalid checkpoint events encountered")
+      invalidEvents.each { System.err.println(it) }
+    }
+    TimelinePrinter.print(spanEvents, threadEvents, orderedEvents, invalidEvents)
+    TimelineExporter.export(orderedEvents)
+    System.err.println("")
+
+    // apparently gradle can not pass system property to spock tests - therefore using env variable instead
```
tylerbenson, line 40: Not true: https://github.com/DataDog/dd-trace-java/blob/master/dd-smoke-tests/dd-smoke-tests.gradle#L27

jbachorik: Ok, how do you do this from command line?
I mean, there does not seem to be a way to do `./gradlew -Dmy.property=true ...` which would transfer `my.property` value so I can get it by `System.getProperty("my.property")`, does it?

richardstartin:
A `suspend` can happen after `startSpan` or `resume`

richardstartin (approved):

dd-java-agent/testing/src/main/groovy/datadog/trace/agent/test/checkpoints/TimelineCheckpointer.groovy
tylerbenson: I think the property might go after the task list?
https://javabydeveloper.com/gradle-system-properties-via-command-line-step-by-step-example/

jbachorik: I think the important part is that it requires configuring the test task as
```
test{
    ...
    systemProperties = System.properties
}
```

Without this the system properties are not passed to the tests :/
I actually tried that and it simply does not work - unless I modify the test tasks.

PR #2960, opened 2021-07-26
refactor gRPC client instrumentation for better profiler integration

richardstartin:
This puts scope activations in the right places to record things like requests being serialized (a `grpc.message` span could be added for requests as well as responses now), and ensures scopes are open in various places to allow propagation to CPU intensive tasks, like deserialization and handling headers.


dd-java-agent/instrumentation/grpc-1.5/src/main/java/datadog/trace/instrumentation/grpc/client/ClientCallImplInstrumentation.java
```diff
@@ -0,0 +1,149 @@
...
+@AutoService(Instrumenter.class)
+public final class ClientCallImplInstrumentation extends Instrumenter.Tracing {
+
+  public ClientCallImplInstrumentation() {
+    super("grpc", "grpc-client");
+  }
+
+  @Override
+  public Map<String, String> contextStore() {
+    return Collections.singletonMap("io.grpc.ClientCall", AgentSpan.class.getName());
```
luhenry, line 37: Why `AgentSpan` and not `State`?

dd-java-agent/instrumentation/grpc-1.5/src/main/java/datadog/trace/instrumentation/grpc/client/ClientCallImplInstrumentation.java
```diff
@@ -0,0 +1,149 @@
...
+    return new String[] {
+      packageName + ".GrpcClientDecorator",
+      packageName + ".GrpcClientDecorator$1",
+      packageName + ".GrpcInjectAdapter"
+    };
+  }
+
+  public static final class Capture {
+    @Advice.OnMethodExit
+    public static void capture(
```
luhenry, line 67: Is that going to get called if there was an exception thrown?

dd-java-agent/instrumentation/grpc-1.5/src/main/java/datadog/trace/instrumentation/grpc/client/ClientCallImplInstrumentation.java
```diff
@@ -0,0 +1,149 @@
...
+import io.grpc.MethodDescriptor;
+import io.grpc.Status;
+import java.util.Collections;
+import java.util.Map;
+import net.bytebuddy.asm.Advice;
+import net.bytebuddy.description.type.TypeDescription;
+import net.bytebuddy.matcher.ElementMatcher;
+
+@AutoService(Instrumenter.class)
+public final class ClientCallImplInstrumentation extends Instrumenter.Tracing {
```
luhenry, line 29: I'm assuming that the `io.grpc.internal.ClientCallImpl` doesn't do anything asynchronous, correct?

dd-java-agent/instrumentation/grpc-1.5/src/main/java/datadog/trace/instrumentation/grpc/client/MessagesAvailableInstrumentation.java
```diff
@@ -0,0 +1,88 @@
...
+    public static void capture(@Advice.This Runnable task) {
+      AdviceUtils.capture(InstrumentationContext.get(Runnable.class, State.class), task, true);
+    }
+  }
+
+  public static final class ReceiveMessages {
+    @Advice.OnMethodEnter
+    public static AgentScope before() {
+      AgentSpan clientSpan = activeSpan();
+      if (clientSpan != null && clientSpan.getOperationName() == GRPC_CLIENT) {
```
luhenry, line 71: Is it because you need to check `clientSpan.getOperationName() == GRPC_CLIENT` that you can't use `AdvicdeUtils.startTaskScope`?

luhenry (approved):
Some questions for my general understanding, otherwise LGTM.

dd-java-agent/instrumentation/grpc-1.5/src/main/java/datadog/trace/instrumentation/grpc/client/ClientCallImplInstrumentation.java
richardstartin: No, but there won't be a call either in that case. It's just not possible to intercept exits from throwing constructor calls.

dd-java-agent/instrumentation/grpc-1.5/src/main/java/datadog/trace/instrumentation/grpc/client/MessagesAvailableInstrumentation.java
richardstartin: It's not the same operation, we need to create a `grpc.message` span here to cover receiving messages. Since this will execute inside a thread pool depending on the executor implementation (the executor may be direct) `startTaskScope` will sometimes have been called implicitly, as `ContextRunnable.run` calls `ContextRunnable.runInContext`

```java
public void messagesAvailable(final MessageProducer producer) {
            PerfMark.startTask("ClientStreamListener.messagesAvailable", ClientCallImpl.this.tag);
            final Link link = PerfMark.linkOut();

            try {
                final class MessagesAvailable extends ContextRunnable {
                    MessagesAvailable() {
                        super(ClientCallImpl.this.context);
                    }

                    public void runInContext() {
                        PerfMark.startTask("ClientCall$Listener.messagesAvailable", ClientCallImpl.this.tag);
                        PerfMark.linkIn(link);

                        try {
                            this.runInternal();
                        } finally {
                            PerfMark.stopTask("ClientCall$Listener.messagesAvailable", ClientCallImpl.this.tag);
                        }

                    }

                    private void runInternal() {
                        if (ClientStreamListenerImpl.this.exceptionStatus != null) {
                            GrpcUtil.closeQuietly(producer);
                        } else {
                            InputStream message;
                            try {
                                for(; (message = producer.next()) != null; message.close()) {
                                    try {
                                        ClientStreamListenerImpl.this.observer.onMessage(ClientCallImpl.this.method.parseResponse(message));
                                    } catch (Throwable var3) {
                                        GrpcUtil.closeQuietly(message);
                                        throw var3;
                                    }
                                }
                            } catch (Throwable var4) {
                                GrpcUtil.closeQuietly(producer);
                                ClientStreamListenerImpl.this.exceptionThrown(Status.CANCELLED.withCause(var4).withDescription("Failed to read message."));
                            }

                        }
                    }
                }

                ClientCallImpl.this.callExecutor.execute(new MessagesAvailable());
            } finally {
                PerfMark.stopTask("ClientStreamListener.messagesAvailable", ClientCallImpl.this.tag);
            }

        }
```
```java
abstract class ContextRunnable implements Runnable {

  private final Context context;

  public ContextRunnable(Context context) {
    this.context = context;
  }

  @Override
  public final void run() {
    Context previous = context.attach();
    try {
      runInContext();
    } finally {
      context.detach(previous);
    }
  }

  public abstract void runInContext();
}
```

dd-java-agent/instrumentation/grpc-1.5/src/main/java/datadog/trace/instrumentation/grpc/client/ClientCallImplInstrumentation.java
richardstartin: It does, optionally, but this instrumentation ensures there are scopes open whenever there is significant work submitted to `callExecutor` (i.e. sending, receiving, serializing/deserializing, headers) which means all the checkpoints built in to `java-concurrent` apply. It also means that, in the mean time, the scope events are meaningful.

dd-java-agent/instrumentation/grpc-1.5/src/main/java/datadog/trace/instrumentation/grpc/client/ClientCallImplInstrumentation.java
richardstartin: Because we don't need `State` - scopes are opened for all submissions to `callExecutor` which means that `java-concurrent` does all required capture/activation.

luhenry: Can we consider `State` an extension of some sort of `AgentSpan` then?

richardstartin: `State` just carries a `Continuation` which is just a reference to a `Scope` which can only be activated once. `State` only exists because we want to be able to do `getAndSet` operations on `Continuation` objects injected into carriers - it reduces the amount of generated code to add to injected classes at the cost of indirection and extra allocations. This instrumentation just needs to get the `AgentSpan` to the relevant places so it can be activated any number of times. All of the `State`/`Continuation` stuff is handled by java-concurrent when entering thread pools.

mcculls (approved):
Makes sense - appreciate the clear naming of the advice

dd-java-agent/instrumentation/grpc-1.5/src/test/groovy/GrpcStreamingTest.groovy
```diff
@@ -134,6 +134,8 @@ class GrpcStreamingTest extends AgentTestRunner {
             "$Tags.COMPONENT" "grpc-client"
             "$Tags.SPAN_KIND" Tags.SPAN_KIND_CLIENT
             "status.code" "OK"
+            "request.type" "example.Helloworld\$Response"
```
devinsba, line 137: Is this a typo? All the others have `example.Helloworld\$Request`

devinsba (approved):

dd-java-agent/instrumentation/grpc-1.5/src/test/groovy/GrpcStreamingTest.groovy
richardstartin: No! I spent quite a long time scratching my head about this, but here's the protobuf in the test, which I decided not to change:

```proto
service Greeter {
  rpc SayHello (Request) returns (Response) {
  }

  rpc Conversation (stream Response) returns (stream Response) {
  }

  rpc Ignore (Request) returns (Response) {
  }
}
```

devinsba: Ok, gross

dd-java-agent/instrumentation/grpc-1.5/src/main/java/datadog/trace/instrumentation/grpc/client/ClientCallImplInstrumentation.java
```diff
@@ -0,0 +1,149 @@
...
+        @Advice.This io.grpc.ClientCall<?, ?> call,
+        @Advice.Argument(0) MethodDescriptor<?, ?> method) {
+      AgentSpan span = DECORATE.startCall(method);
+      if (null != span) {
+        InstrumentationContext.get(ClientCall.class, AgentSpan.class).put(call, span);
+      }
+    }
+  }
+
+  public static final class Start {
```
tylerbenson, line 77: nit: The span is started in the constructor, so I think `Invoke` would make a better name.

tylerbenson (approved):

dd-java-agent/instrumentation/grpc-1.5/src/main/java/datadog/trace/instrumentation/grpc/client/ClientCallImplInstrumentation.java
richardstartin: The advice is applied to a method called `start`, so I think `Start` is the better name.

PR #2961, opened 2021-07-26
Request body interception for Servlet/Grizzly

cataphract:


dd-java-agent/appsec/src/main/java/com/datadog/appsec/gateway/GatewayBridge.java
```diff
@@ -86,6 +89,38 @@ public void init() {
     subscriptionService.registerCallback(Events.REQUEST_HEADER_DONE, new HeadersDoneCallback());
 
     subscriptionService.registerCallback(Events.REQUEST_URI_RAW, new RawURICallback());
+
+    subscriptionService.registerCallback(
+        Events.REQUEST_BODY_START,
+        (RequestContext ctx_, StoredBodySupplier supplier) -> {
```
dougqh, line 95: For now, I think this is fine.  However, I think we might want to avoid lambdas in general because they'll be for start-up.  We also want to avoid capturing lambdas to avoid extra allocation.

dd-java-agent/instrumentation/servlet/request-2/src/main/java/datadog/trace/instrumentation/servlet2/Servlet2Advice.java
```diff
@@ -52,6 +57,7 @@ public static AgentScope onEnter(
     final AgentSpan span = startSpan(SERVLET_REQUEST, extractedContext).setMeasured(true);
 
     DECORATE.afterStart(span);
+    request = DECORATE.wrapRequest(span, httpServletRequest);
```
dougqh, line 60: I thought we'd gotten rid of most the wrapping.  I know we had a few servers where wrapping ServletRequest or ServletResponse was causing problems.

dd-java-agent/instrumentation/servlet/request-2/src/main/java/datadog/trace/instrumentation/servlet2/Servlet2Instrumentation.java
```diff
@@ -42,9 +42,16 @@ public Servlet2Instrumentation() {
   @Override
   public String[] helperClassNames() {
     return new String[] {
+      "datadog.trace.instrumentation.servlet.http.common.ServletHttpServerDecorator",
       packageName + ".Servlet2Decorator",
       packageName + ".ServletRequestURIAdapter",
       packageName + ".HttpServletRequestExtractAdapter",
+      "datadog.trace.instrumentation.servlet.http.common.BodyCapturingHttpServletRequest",
```
dougqh, line 49: I think we need to think about separating about the appsec parts of instrumentation.
As is, this will cause us to inject more classes which will consume additional metaspace even when we're not using appsec.

dougqh: The wrappers also worry me.  Have we run the server framework test with these changes?  If not, we definitely should.

dougqh: I'm not sure how it fits with Config, but I simple solution would be only return the extra classes when appsec is enabled.  Although, then we need to careful about referring to those classes.

I think we probably just need to make a stronger separation between core tracing & app sec.  Same for profiling

dougqh: Wrappers also worry me.  I know there was at least one server that tries to cast directly to its servletrequest class.
We need to make sure to run all the server unit tests to make sure this is sound.

My other thought was to annotation any wrapper classes.  Then run a pessimistic static analysis over any supported libraries.  If we find a cast to any child class of the wrapped interface, then we can pessimistically assume we have a problem.

internal-api/src/main/java/datadog/trace/api/http/StoredCharBody.java
```diff
@@ -0,0 +1,110 @@
...
+
+  private char[] storedBody;
+  private int storedBodyLen;
+  private boolean bodyReadStarted = false;
+
+  public StoredCharBody(StoredBodyListener listener) {
+    this.listener = listener;
+  }
+
+  public synchronized void appendData(char[] chars, int start, int end) {
```
dougqh, line 24: I worry about the overhead of synchronization.  I assume this will be biased towards one thread, so it is probably fine.

internal-api/src/main/java/datadog/trace/api/http/StoredCharBody.java
```diff
@@ -0,0 +1,110 @@
...
+    if (this.listenerNotified.compareAndSet(false, true)) {
+      if (!bodyReadStarted) {
+        this.listener.onBodyStart(this);
+      }
+      this.listener.onBodyEnd(this);
+    }
+  }
+
+  @Override
+  public String get() {
```
dougqh, line 104: Would a CharSequence suffice?  Or, something that doesn't allocate every time?

internal-api/src/main/java/datadog/trace/api/http/StoredCharBody.java
```diff
@@ -0,0 +1,110 @@
...
+@SuppressWarnings("Duplicates")
+public class StoredCharBody implements StoredBodySupplier {
+  private static final int MIN_BUFFER_SIZE = 128; // chars
+  private static final int MAX_BUFFER_SIZE = 1024 * 1024; // 2 MB (char == 2 bytes)
+  private static final int GROW_FACTOR = 4;
+
+  private final AtomicBoolean listenerNotified = new AtomicBoolean();
+  private final StoredBodyListener listener;
+
+  private char[] storedBody;
```
dougqh, line 16: Why not use CharBuffer & ByteBuffer for this?

dd-java-agent/agent-bootstrap/src/main/java/datadog/trace/bootstrap/Agent.java
```diff
@@ -76,12 +76,14 @@
 
   private static boolean jmxFetchEnabled = true;
   private static boolean profilingEnabled = false;
+  public static boolean appSecEnabled = false;
```
dougqh, line 79: I hate that these fields are final, but that aside.  Why does this need to be public?

dd-java-agent/appsec/src/main/java/com/datadog/appsec/gateway/AppSecRequestContext.java
```diff
@@ -119,11 +123,25 @@ boolean isFinishedHeaders() {
...
+
   @Override
   public void close() {
     // currently no-op
   }
 
+  /* Should be accessible from the modules */
+
+  /** @return the portion of the body read so far, if any */
+  public Optional<String> getStoredRequestBody() {
```
dougqh, line 138: I'd prefer to avoid using Optional, since it creates extra allocation that we don't really need.
I also prefer to avoid the use of String as well.

dd-java-agent/instrumentation/servlet/request-2/src/main/java/datadog/trace/instrumentation/servlet2/Servlet2Advice.java
cataphract: It's true that, in principle, wrapping can cause problems here. These are the options I considered for instrumenting servlet (i.e., not considering here instrumenting details of specific servers):

1. instrument subclasses of `HttpServlet`, and wrap the `HttpServletRequest` argument on the `service` method (similar for filters); this wrapper than wraps the results of `getInputStream` / `getReader`. This is the option I went for.
2. instrument directly subclasses of `HttpServletRequest` and intercept calls to `getInputStream` / `getReader`; wrap the result.
3. Instrument directly `ServletInputStream` / `BufferedReader`.

I think we can discard 3. immediately; even if we could identify the buffered reader instances of interest, instrumenting `BufferedReader` would be too heavy-handed.

The reason I didn't go for 2. is because wrapping request objects is more common than wrapping servlets, and it is what we've been doing at Sqreen. If these methods themselves wrap the inputstream/reader for buffering the request entity, we may be tricked into double-wrapping ourselves, at least with the `instanceof` check we have. Now, perhaps this detection could be done with other means (maybe setting a request attribute once we've done some wrapping), or on balance 2. is still better than 1., but this was my thinking.

dd-java-agent/instrumentation/servlet/request-2/src/main/java/datadog/trace/instrumentation/servlet2/Servlet2Instrumentation.java
cataphract: > I think we need to think about separating about the appsec parts of instrumentation.
> As is, this will cause us to inject more classes which will consume additional metaspace even when we're not using appsec.

This flows from the design choice of having an appsec-agnostic "Instrumentation Gateway"; it means the appsec module will not be itself responsible for instrumentation. Still, in this design the extra instrumentation could be avoided if we require that the IG callbacks be set early in the application startup — then we can skip all the advice that exists in support of that callback.

cataphract: > My other thought was to annotation any wrapper classes. Then run a pessimistic static analysis over any supported libraries. If we find a cast to any child class of the wrapped interface, then we can pessimistically assume we have a problem.

I would think you'd need to check all the non-library code as well. A user could have setup a filter that wraps the request, and then downcast further down the line.

internal-api/src/main/java/datadog/trace/api/http/StoredCharBody.java
cataphract: Yeah, in principle you'll be reading the whole data in one thread so there is no contention. But you still need to guard against the possibility that stream is used in different threads. I couldn't think of a better way of optimizing the usual path, but I'm open to suggestions.

internal-api/src/main/java/datadog/trace/api/http/StoredCharBody.java
cataphract: The idea here is that this method is generally called only once: to get the final value that is then distributed to the appsec modules for inspection. It is possible, though, that some RASP appsec module is interested in all the data read so far, in which case, yes, we'd be building multiple (but distinct) strings. An option would be expose the storage directly, but this would be difficult to harmonize between the inputstream and reader paths.

dd-java-agent/agent-bootstrap/src/main/java/datadog/trace/bootstrap/Agent.java
cataphract: An artifact from an aborted change. I'll change it back.

internal-api/src/main/java/datadog/trace/api/http/StoredCharBody.java
cataphract: You mean as wrappers of the arrays? I need to grow them. I don't think I'd given it any thought, but it seems a bit of an overkill here. All I need is to keep an offset and advance, not marks limits and other misc stuff Char/ByteBuffer support.

dd-java-agent/appsec/src/main/java/com/datadog/appsec/gateway/AppSecRequestContext.java
cataphract: I'll make it nullable then. As for the `String`, I've commented elsewhere.

dd-java-agent/instrumentation/servlet/request-2/src/main/java/datadog/trace/instrumentation/servlet2/Servlet2Instrumentation.java
devinsba: We have a way to create instrumentation classes that are tied to specific features, perhaps we should create a new sub-class for app sec like the one for profiling: https://github.com/DataDog/dd-trace-java/blob/master/dd-java-agent/agent-tooling/src/main/java/datadog/trace/agent/tooling/Instrumenter.java#L383

dd-java-agent/instrumentation/servlet/request-2/src/main/java/datadog/trace/instrumentation/servlet2/Servlet2Advice.java
dougqh: I do agree that wrapping Request objects is likely to be safer than wrapping Servlets.  Wrapping only ServletInputStream would be strictly safer than wrapping both.

The proof is on the author to convince that wrapping is safe.  Our general rule is that you must run the server tests when you modify something that touches a server.  Since you touched the servlet instrumentation, you must run tests for as many servlet containers as possible, but I'm most interested in the ones that were troublesome previously.

I agree that instrumenting BufferedReader probably isn't viable.  However instrumenting the server specific subclasses of ServletInputStream might not be so bad.

dd-java-agent/instrumentation/servlet/request-2/src/main/java/datadog/trace/instrumentation/servlet2/Servlet2Instrumentation.java
dougqh: My requirement for AppSec is quite simple.  There must be a minimum of additional classes loaded when AppSec is disabled.  There must be zero additional classes injected when AppSec is enabled.

If the instrumentation gateway doesn't satisfy those requirements, then the gateway design must be rethought.

internal-api/src/main/java/datadog/trace/api/http/StoredCharBody.java
dougqh: It actually wouldn't be very hard to expose a view into the buffer and have that coexist with InputStream/Reader.

As long as the InputStream/Reader doesn't rewind, you just need to provide a view over the previously written bytes.
I cannot remember if there's a rewind capability, if there is then you might just need to version the buffer if it is rewound into a published segment.

In either case, this is pretty easy to support and reduces allocation considerably.

internal-api/src/main/java/datadog/trace/api/http/StoredCharBody.java
dougqh: I was just curious as to the reason.  Those classes are often intrinsified and often perform better than the handrolled version.

internal-api/src/main/java/datadog/trace/api/http/StoredByteBody.java
```diff
@@ -0,0 +1,179 @@
...
+public class StoredByteBody implements StoredBodySupplier {
+  private static final int MIN_BUFFER_SIZE = 128; // bytes
+  private static final int MAX_BUFFER_SIZE = 1024 * 1024; // 1 MB
+  private static final int GROW_FACTOR = 4;
+
+  private static final Charset UTF_8 = StandardCharsets.UTF_8;
+  private static final Charset ISO_8859_1 = StandardCharsets.ISO_8859_1;
+
+  private final StoredBodyListener listener;
+  private final AtomicBoolean listenerNotified = new AtomicBoolean();
```
dougqh, line 19: If notification is once and only once, you don't need an AtomicBoolean.   A volatile + synchronized would be sufficient and allocate less memory.

internal-api/src/main/java/datadog/trace/api/http/StoredCharBody.java
dougqh: Yes, I understand the multi-threading concern.  I'm just leery of adding additional locking.

dd-java-agent/instrumentation/servlet/request-2/src/main/java/datadog/trace/instrumentation/servlet2/Servlet2Advice.java
dougqh: You are also correct that checking the framework code for checkcast / instanceof is not sufficient to guarantee correctness.  Indeed worse case, the tracer should make a completely world assumption and should never wrap.

I don't think that's entirely practical.  However, we should endeavor to prove that we're not violating any casts inside the servers which we can relatively easily test and analyze.

dougqh (changes requested):
My 3 main concerns are...
1 - Injecting extra classes - strictly forbidden when AppSec is off
2 - Wrappers - we need to verify that these are safe
3 - Allocation - As far as I can tell, the allocation of the String can easily be eliminated

dd-java-agent/instrumentation/servlet/request-2/src/main/java/datadog/trace/instrumentation/servlet2/Servlet2Instrumentation.java
cataphract: @dougqh Please see https://github.com/DataDog/dd-trace-java/pull/2961/commits/ca9106bd519a81afc8f41397dcb4ad1a1718fa95 , which wraps only the is/reader. I've also used the strategy @devinsba suggested to avoid applying the instrumentation if appsec if off

internal-api/src/main/java/datadog/trace/api/http/StoredCharBody.java
cataphract: I can do this for char array variant, but not for the byte + charset variant. That is, unless I decode possibly the whole string and then discard it for every operation (for instance, to implement `CharSequence::getLength()`).

This is the current implementation for `StoredByteBody`:

```java
  @Override
  public synchronized String get() {
    if (this.storedBodyLen == 0) {
      return "";
    }
    if (this.charset != null) {
      return new String(this.storedBody, 0, this.storedBodyLen, this.charset);
    }

    if (isWellFormed(this.storedBody, 0, this.storedBodyLen)) {
      return new String(this.storedBody, 0, this.storedBodyLen, UTF_8);
    }

    return new String(this.storedBody, 0, this.storedBodyLen, ISO_8859_1);
  }
```

To keep this logic, the only possibility I see would be to keep instead of a `byte[]` a `char[]` and then:
1. keep the incomplete code unit sequences in a separate buffer. How big this buffer must be depends on the encoding, though.
2. if we don't have an encoding set, assume at first 7-bit ascii. When we find a sequence that looks like UTF-8, decode as UTF-8.
3. when we find a bad UTF-8 code unit sequence, go back, reencode everything as UTF-8 and interpret each individual byte as latin1.

internal-api/src/main/java/datadog/trace/api/http/StoredByteBody.java
cataphract: It turns out this is read in a synchronized method, so a plain boolean will do.

internal-api/src/main/java/datadog/trace/api/http/StoredCharBody.java
cataphract: Ok, I implemented something in those lines in https://github.com/DataDog/dd-trace-java/pull/2961/commits/de68f0ed928e65e6448dd52ccf7b575f891f0029

internal-api/src/main/java/datadog/trace/api/http/StoredCharBody.java
```diff
@@ -0,0 +1,136 @@
...
+      if (!bodyReadStarted) {
+        listener.onBodyStart(this);
+      }
+      listener.onBodyEnd(this);
+    }
+  }
+
+  @Override
+  public synchronized CharBuffer get() {
+    if (this.storedBodyLen == 0) {
```
dougqh, line 126: Yes, this looks better.  I'd still like to better understand how the call to native code is being done.

To keep overhead down, we don't to be copying from Java to native either.  We should be able to reduce that by use direct allocation.

dd-java-agent/appsec/src/main/java/com/datadog/appsec/gateway/AppSecRequestContext.java
```diff
@@ -28,11 +30,12 @@
   private final ConcurrentHashMap<Address<?>, Object> persistentData = new ConcurrentHashMap<>();
   private Collection<Attack010> collectedAttacks;
 
-  // assume this will always be accessed by the same thread
+  // assume these will always be accessed by the same thread
   private String savedRawURI;
   private CaseInsensitiveMap<List<String>> collectedHeaders = new CaseInsensitiveMap<>();
```
dougqh, line 35: final?

dd-java-agent/appsec/src/main/java/com/datadog/appsec/gateway/AppSecRequestContext.java
```diff
@@ -28,11 +30,12 @@
...
   private List<StringKVPair> collectedCookies = new ArrayList<>(4);
```
dougqh, line 36: final?

dd-java-agent/appsec/src/main/java/com/datadog/appsec/gateway/AppSecRequestContext.java
```diff
@@ -120,6 +123,10 @@ boolean isFinishedHeaders() {
     return collectedCookies;
   }
 
+  public void setStoredRequestBodySupplier(StoredBodySupplier storedRequestBodySupplier) {
```
dougqh, line 126: What's the threading model of this class?  I'm concerned by the amount of mutability.

dd-java-agent/appsec/src/main/java/com/datadog/appsec/gateway/GatewayBridge.java
```diff
@@ -87,6 +90,37 @@ public void init() {
...
+          AppSecRequestContext ctx = (AppSecRequestContext) ctx_;
+          ctx.setStoredRequestBodySupplier(supplier);
+          producerService.publishEvent(ctx, EventType.REQUEST_BODY_START);
+          return null;
+        });
+
+    subscriptionService.registerCallback(
```
dougqh, line 102: I think I need to understand the subscriptionService better.  I generally feel that generic callback systems are overly complicated for what instrumentation needs to be doing, but for now, I'm reserving judgment.

dd-java-agent/appsec/src/main/java/com/datadog/appsec/gateway/GatewayBridge.java
```diff
@@ -87,6 +90,37 @@ public void init() {
...
+        Events.REQUEST_BODY_DONE,
+        (RequestContext ctx_, StoredBodySupplier supplier) -> {
+          AppSecRequestContext ctx = (AppSecRequestContext) ctx_;
+          producerService.publishEvent(ctx, EventType.REQUEST_BODY_END);
+
+          if (rawRequestBodySubInfo == null) {
+            rawRequestBodySubInfo =
+                producerService.getDataSubscribers(KnownAddresses.REQUEST_BODY_RAW);
+          }
+          if (rawRequestBodySubInfo.isEmpty()) {
```
dougqh, line 112: All these checks seem easy to mess up to me, but if we're only added a few of these then I think it is probably fine.

dd-java-agent/instrumentation/servlet/src/main/java/datadog/trace/instrumentation/servlet/http/BufferedReaderWrapper.java
```diff
@@ -0,0 +1,107 @@
...
+
+  @Override
+  public String readLine() throws IOException {
+    String read = this.reader.readLine();
+    if (read == null) {
+      storedCharBody.maybeNotify();
+      return null;
+    }
+    storedCharBody.appendData(read);
+    storedCharBody.appendData('\n');
```
dougqh, line 48: I'm not sure just assuming '\n' is okay.  Traditionally, Windows would \r\n and old Macs would send \r\r.

dd-java-agent/instrumentation/servlet/src/main/java/datadog/trace/instrumentation/servlet/http/BufferedReaderWrapper.java
```diff
@@ -0,0 +1,107 @@
...
+    this.reader.close();
+    storedCharBody.maybeNotify();
+  }
+
+  @Override
+  public int read(CharBuffer target) throws IOException {
+    int initPos = target.position();
+    int read = this.reader.read(target);
+    if (read > 0) {
+      for (int i = initPos; i < initPos + read; i++) {
```
dougqh, line 88: This is likely to be inefficient.

CharBuffer supports some view operations & bulk transfer operations.  It would be better to use those.

dd-java-agent/instrumentation/servlet/src/main/java/datadog/trace/instrumentation/servlet/http/ServletInputStreamWrapper.java
```diff
@@ -0,0 +1,95 @@
...
+    storedByteBody.setCharset(charset);
+  }
+
+  @Override
+  public int read(byte[] b) throws IOException {
+    int numRead = is.read(b);
+    if (numRead > 0) {
+      storedByteBody.appendData(b, 0, numRead);
+    } else {
+      storedByteBody.maybeNotify();
```
dougqh, line 27: Why are we notifying on every write?

dougqh, line 27: So we're trying to notify on end?  numRead == 0 doesn't mean end?
It means no data available or the read was interrupted (potentially by kernel, etc).

If you want end, that's numRead == -1.  Out of curiosity, why not use servlet ReadListener?

dd-java-agent/appsec/src/main/java/com/datadog/appsec/gateway/AppSecRequestContext.java
```diff
@@ -20,6 +21,7 @@
 import org.slf4j.Logger;
 import org.slf4j.LoggerFactory;
 
+// XXX: this must be made thread-safe
```
dougqh, line 24: The lack of clear threading model here is concerning.  Regardless of the threading model, I think it would be better to reduce the mutability.

dd-java-agent/appsec/src/main/java/com/datadog/appsec/gateway/AppSecRequestContext.java
dougqh, line 126: Where do we set this?

dd-java-agent/appsec/src/main/java/com/datadog/appsec/gateway/GatewayBridge.java
```diff
@@ -44,6 +46,7 @@
 
   // subscriber cache
   private volatile EventProducerService.DataSubscriberInfo initialReqDataSubInfo;
+  private volatile EventProducerService.DataSubscriberInfo rawRequestBodySubInfo;
```
dougqh, line 49: As mentioned elsewhere, I think this subscriber system might be overkill an ill-performing.  The fact that we feel we need a cache for it -- only makes me more convinced that something simpler would be better.

dd-java-agent/instrumentation/servlet/src/main/java/datadog/trace/instrumentation/servlet/http/ServletRequestBodyInstrumentation.java
```diff
@@ -0,0 +1,172 @@
...
+        named("getReader")
+            .and(takesNoArguments())
+            .and(returns(named("java.io.BufferedReader")))
+            .and(isPublic()),
+        getClass().getName() + "$HttpServletGetReaderAdvice");
+  }
+
+  @Override
+  public String[] helperClassNames() {
+    return new String[] {
```
dougqh, line 63: So this is split into its own instrumentation -- good.
Why do we need the delegating listeners?  That seems rather heavy weight.

dd-java-agent/instrumentation/servlet/src/main/java/datadog/trace/instrumentation/servlet/http/IGDelegatingStoredBodyListener.java
```diff
@@ -0,0 +1,66 @@
...
+      bodyRawCallback = EMPTY_BODY_RAW_CALLBACK;
+    }
+    this.bodyRawCallback = bodyRawCallback;
+    this.ctx = ctx;
+  }
+
+  @Override
+  public void onBodyStart(final StoredBodySupplier storedByteBody) {
+    // adapt StoredBodySupplier to StoredBodySupplier
+    // due to how the module boundaries are done, StoredBodySupplier cannot implement
```
dougqh, line 56: I see, so this is the reason for the delegation.  I'm not opposed to pushing StoredBodySupplier to generally available location if need be.  I'd prefer that to wrappers on top of wrappers.

We could also use a more generic type that we push into core.  Mostly, I just think we talk through the whole notification system.

dd-java-agent/instrumentation/servlet/src/main/java/datadog/trace/instrumentation/servlet/http/IGDelegatingStoredBodyListener.java
```diff
@@ -0,0 +1,66 @@
...
+    // StoredBodySupplier
+    this.bodyStartCallback.apply(ctx, storedByteBody);
```
dougqh, line 58: Listeners on top of callbacks on top of listeners worries me because of both CPU & memory overhead

dd-java-agent/instrumentation/servlet/src/main/java/datadog/trace/instrumentation/servlet/http/ServletInputStreamWrapper.java
```diff
@@ -0,0 +1,95 @@
...
+
+  public void setCharset(Charset charset) {
+    storedByteBody.setCharset(charset);
+  }
+
+  @Override
+  public int read(byte[] b) throws IOException {
+    int numRead = is.read(b);
+    if (numRead > 0) {
+      storedByteBody.appendData(b, 0, numRead);
```
dougqh, line 25: If I remember correctly storedByteBody has a limit as to how much it accepts, shouldn't that be checked here.
Do you notify when you reach the limit or do you wait until the stream ends?

dd-java-agent/instrumentation/servlet/src/main/java/datadog/trace/instrumentation/servlet/http/ServletRequestBodyInstrumentation.java
dougqh: I saw elsewhere this was from StoredBodyListener-s being incompatible across modules.  I think that's the problem we actually need to solve -- that and just discussing the overall flow of data.

internal-api/src/main/java/datadog/trace/api/http/StoredByteBody.java
```diff
@@ -0,0 +1,206 @@
...
+      };
+
+  public static CharsetDecoder decoderFor(Charset charset) {
+    CharsetDecoder cd = (CharsetDecoder) DECODER_CACHE.forName(charset);
+    cd.onMalformedInput(CodingErrorAction.REPLACE);
+    cd.reset();
+    return cd;
+  }
+
+  private static ThreadLocal<CharsetEncoder> UTF8_ENCODER_CACHE =
```
dougqh, line 190: final

internal-api/src/main/java/datadog/trace/api/http/StoredByteBody.java
```diff
@@ -0,0 +1,206 @@
...
+}
+
+// adapted from sun.nio.cs
+class ThreadLocalCoders {
+  private static final int CACHE_SIZE = 3;
+
+  private abstract static class Cache {
+
+    // Thread-local reference to array of cached objects, in LRU order
+    private ThreadLocal<Object[]> cache = new ThreadLocal<>();
```
dougqh, line 131: final

dd-java-agent/instrumentation/servlet/src/main/java/datadog/trace/instrumentation/servlet/http/ServletRequestBodyInstrumentation.java
```diff
@@ -0,0 +1,172 @@
...
+      AgentSpan agentSpan = activeSpan();
+      if (agentSpan == null) {
+        return;
+      }
+      HttpServletRequest req = (HttpServletRequest) thiz;
+      Object alreadyWrapped = req.getAttribute("datadog.wrapped_request_body");
+      if (alreadyWrapped != null || is instanceof ServletInputStreamWrapper) {
+        return;
+      }
+      RequestContext requestContext = agentSpan.getRequestContext();
```
dougqh, line 91: So lack of a requestContext short circuits wrapping?  Good

dd-java-agent/instrumentation/servlet/src/main/java/datadog/trace/instrumentation/servlet/http/ServletRequestBodyInstrumentation.java
```diff
@@ -0,0 +1,172 @@
...
+      if (requestContext == null) {
+        return;
+      }
+      req.setAttribute("datadog.wrapped_request_body", Boolean.TRUE);
+
+      CallbackProvider cbp = AgentTracer.get().instrumentationGateway();
```
dougqh, line 97: Does this mean we're still wrapping when no rule is consuming the body?  If so, I think that's a problem.

dd-java-agent/instrumentation/servlet/src/main/java/datadog/trace/instrumentation/servlet/http/ServletRequestBodyInstrumentation.java
```diff
@@ -0,0 +1,172 @@
...
+
+      StoredByteBody storedByteBody = new StoredByteBody(listener, lengthHint);
+      ServletInputStreamWrapper servletInputStreamWrapper =
+          new ServletInputStreamWrapper(is, storedByteBody);
+
+      String encoding = req.getCharacterEncoding();
+      try {
+        if (encoding != null) {
+          Charset charset = Charset.forName(encoding);
+          servletInputStreamWrapper.setCharset(charset);
```
dougqh, line 119: Why not set the charSet in the ServletInputStreamWrapper constructor?
I don't see a data dependence that prevents it.

internal-api/src/main/java/datadog/trace/api/http/StoredCharBody.java
cataphract: Sorry, what native code are you referring to?

ValentinZakharov: I suppose, it's about send data to WAF via JNI.

dd-java-agent/instrumentation/servlet/src/main/java/datadog/trace/instrumentation/servlet/http/BufferedReaderWrapper.java
cataphract: It is true that assuming `\n` is problematic. Unfortunately, there are no better options, because the wrapped `BufferedReader` may have discarded a LF, CR, CR + LF, or nothing at all (EOF). We just can't know.

dd-java-agent/instrumentation/servlet/src/main/java/datadog/trace/instrumentation/servlet/http/ServletInputStreamWrapper.java
cataphract: I'm fixed the problems with the zeros.

As for the second point, that's only available for Servlet 3.1+. Plus the original implementation would do the notification as soon as the size limit was reached; it did not wait until the original stream reached the end.

dd-java-agent/appsec/src/main/java/com/datadog/appsec/gateway/AppSecRequestContext.java
cataphract: Originally this was only storing headers and request uri, which would presumably would be all delivered from the same thread (the same that created the object). With the request body changes, things got a bit more complicated, though since there are really no consumers of the request body now, it's not an immediate concern. I'll make some adjustments, though, really no reason to leave it in a semi-broken state.

dd-java-agent/appsec/src/main/java/com/datadog/appsec/gateway/AppSecRequestContext.java
cataphract: `GatewayBridge`, on the `REQUEST_BODY_START` callback.

dd-java-agent/appsec/src/main/java/com/datadog/appsec/gateway/GatewayBridge.java
cataphract: There are, alas, two subscription systems involved. The instrumentation gateway, and an appsec internal one. The appsec one allows modules to subscribe an arbitrary number of data "addresses", i.e., labels for data. This is what's being cached here, the set of subscribers that are interested in a at least one address in a particular bundle of data. The cache avoids our having to calculate this union set each time.

dd-java-agent/instrumentation/servlet/src/main/java/datadog/trace/instrumentation/servlet/http/ServletInputStreamWrapper.java
cataphract: `appendData` will be a noop if the limit was reached. Though I see that with the refactoring of `StoredByteBody`, it will still decode data that will then be discarded. I'll improve that a bit.

dd-java-agent/instrumentation/servlet/src/main/java/datadog/trace/instrumentation/servlet/http/ServletRequestBodyInstrumentation.java
cataphract: Indeed it does. We need to track the subscriptions across the two levels of listeners to prevent this, but I'll implement it.

cataphract: Indeed it does. We need to track the subscriptions across the two levels of listeners to prevent this, but I'll implement it.

dd-java-agent/instrumentation/servlet/src/main/java/datadog/trace/instrumentation/servlet/http/ServletRequestBodyInstrumentation.java
cataphract: Good point.

dd-java-agent/instrumentation/servlet/src/main/java/datadog/trace/instrumentation/servlet/http/ServletRequestBodyInstrumentation.java
cataphract: Actually the reason presented seems not to apply anymore since I moved some around modules. It would be possible to change StoredCharBody/StoredByteBody to, instead of receiving a listener, receive at least the `RequestContext`. It could also get the two callbacks directly, or get the InstrumentationGateway and look them up, or even get nothing else extra and lookup the InstrumentationGateway statically twice (for start or for end). In terms of prettiness of design, the current one is IMO the best as StoredCharBody/StoredByteBody need not known any details about the notification mechanism, but I understand the extra indirection has a cost. I can change it for the middle alternative, as the last one is a pain to unit test and requires looking up the agent tracer and IG  twice, at the cost of storing an extra field.

dd-java-agent/appsec/src/main/java/com/datadog/appsec/gateway/AppSecRequestContext.java
cataphract: I've made some changes that fix the problem with the request body supplier likely accessible from multiple threads in the future. As for the "initial data" fields (request uri, headers, ip), there is still no synchronization involved as I assume they all come from the same thread (or if they end up being processed in several threads, that the changes made in earlier rounds are published to later rounds of the http header processing). I think it's a reasonable assumption, though I may be wrong here for some of the supported web servers.

internal-api/src/main/java/datadog/trace/api/http/StoredByteBody.java
bantonsson: Still missing `final`

internal-api/src/main/java/datadog/trace/api/http/StoredCharBody.java
bantonsson: @dougqh The preparation of the parameters and the actual call to the native code is done by the `libsqreen` dependency, that lives in its own repo, and which currently serialize the data into a direct `ByteBuffer` that the native code access directly. A future optimization could definitely be to use a direct `ByteBuffer` for the message body and pass this to the native code without copying/serializing it, but that is a slightly larger undertaking.

internal-api/src/main/java/datadog/trace/api/http/StoredCharBody.java
bantonsson: If we in the future want to pass the message body directly to the `libsqreen` and native code, then we could do a direct `ByteBuffer` here with `MAX_SIZE` since the pages won't be committed until they are used.

bantonsson (approved):
LGTM

Apart from performance optimizations that would require changes in `libsqreen` I think that all bigger things have been addressed.

dd-java-agent/instrumentation/grizzly-http-2.3.20/src/main/java/datadog/trace/instrumentation/grizzlyhttp232/HttpHeaderFetchingHelper.java
```diff
@@ -0,0 +1,35 @@
...
+import java.lang.reflect.Field;
+import java.lang.reflect.UndeclaredThrowableException;
+import org.glassfish.grizzly.http.HttpHeader;
+import org.glassfish.grizzly.http.io.InputBuffer;
+
+/**
+ * Helper till we have support for invokedynamic-based caching on bytebuddy:
+ * https://github.com/raphw/byte-buddy/issues/1009
+ */
+public final class HttpHeaderFetchingHelper {
```
bantonsson, line 12: 👍

bantonsson (approved):

bantonsson:
@dougqh Would you mind taking another look at the changes made, and if they address your concerns?

dd-java-agent/appsec/src/main/java/com/datadog/appsec/event/EventDispatcher.java
```diff
@@ -193,6 +199,23 @@ public Flow publishDataEvent(
     return flow;
   }
 
+  @Override
+  public Collection<EventType> allSubscribedEvents() {
+    List<EventType> res = new ArrayList<>();
```
dougqh, line 204: I'm guessing this is set-up rarely enough that this isn't a big deal, but I'm still curious...

Is it worth sizing the res ArrayList to the values array.  How often are listeners empty?  We could always trim to capacity afterwards.

dd-java-agent/appsec/src/main/java/com/datadog/appsec/event/EventDispatcher.java
```diff
@@ -193,6 +199,23 @@ public Flow publishDataEvent(
...
+    return res;
+  }
+
+  @Override
+  public Collection<Address<?>> allSubscribedDataAddresses() {
+    return allSubscribedAddresses;
+  }
+
   private static <T> List<T> nullFilledList(int numElements) {
     ArrayList<T> ts = new ArrayList<>(numElements);
```
dougqh, line 220: I see this already here, but I'm still curious about a couple things.

Why do we need to null fill a List?
If the list doesn't need to grow, then this could be done more efficiently as.

return Arrays.asList(new Object[numElements]);

dd-java-agent/appsec/src/main/java/com/datadog/appsec/gateway/AppSecRequestContext.java
```diff
@@ -26,15 +27,17 @@
...
-  // assume this will always be accessed by the same thread
+  // assume these will always be written and read by the same thread
   private String savedRawURI;
-  private CaseInsensitiveMap<List<String>> collectedHeaders = new CaseInsensitiveMap<>();
+  private final CaseInsensitiveMap<List<String>> collectedHeaders = new CaseInsensitiveMap<>();
   private List<StringKVPair> collectedCookies = new ArrayList<>(4);
   private boolean finishedHeaders;
   private String ip;
 
+  private volatile StoredBodySupplier storedRequestBodySupplier;
```
dougqh, line 39: If we're assuming this is written and read by the same thread, why is this volatile?
Or, conversely why are the other fields not volatile?

I'm skeptical that the single thread assumption is sound.

dd-java-agent/appsec/src/main/java/com/datadog/appsec/gateway/GatewayBridge.java
dougqh: I see.  So why doesn't the appSec internal one, just present the full union of requested subscriptions?
Doesn't this mean we're recreating this union logic in every language?

dd-java-agent/appsec/src/main/java/com/datadog/appsec/gateway/GatewayBridge.java
```diff
@@ -279,4 +323,42 @@ private static int byteToDigit(byte b) {
...
+
+  private static class IGAppSecEventDependencies {
+    private static final Map<EventType, Collection<datadog.trace.api.gateway.EventType<?>>>
+        EVENT_DEPENDENCIES = new HashMap<>();
+
+    private static final Map<Address<?>, Collection<datadog.trace.api.gateway.EventType<?>>>
+        DATA_DEPENDENCIES = new HashMap<>();
+
+    static {
+      EVENT_DEPENDENCIES.put(EventType.REQUEST_BODY_START, l(Events.REQUEST_BODY_START));
```
dougqh, line 335: I don't particularly love Maps for these types of things because we're adding immoral memory footprint.
A static routine that does switch & return would be lower overhead.

At the very least, it would be good to size the Maps appropriately.

dd-java-agent/instrumentation/grizzly-http-2.3.20/src/main/java/datadog/trace/instrumentation/grizzlyhttp232/GrizzlyByteBodyInstrumentation.java
```diff
@@ -0,0 +1,203 @@
...
+  }
+
+  @Override
+  public void adviceTransformations(AdviceTransformation transformation) {
+    transformation.applyAdvice(
+        named("setInputBuffer")
+            .and(takesArguments(1))
+            .and(takesArgument(0, named("org.glassfish.grizzly.http.io.InputBuffer"))),
+        getClass().getName() + "$NIOInputStreamSetInputBufferAdvice");
+    /* we're assuming here none of these methods call the other instrumented methods */
```
dougqh, line 56: Okay, that's fine for now.

dd-java-agent/instrumentation/grizzly-http-2.3.20/src/main/java/datadog/trace/instrumentation/grizzlyhttp232/GrizzlyByteBodyInstrumentation.java
```diff
@@ -0,0 +1,203 @@
...
+      }
+      attributes.setAttribute("datadog.intercepted_request_body", Boolean.TRUE);
+
+      String lengthHeader = header.getHeader("content-length");
+      String encodingString = header.getCharacterEncoding();
+      if (encodingString != null) {
+        try {
+          charset = Charsets.lookupCharset(encodingString);
+        } catch (UnsupportedCharsetException use) {
+          // purposefully left blank
```
dougqh, line 99: We should make a way to report these unexpected exceptions, but that's a general tracer problem -- not for this PR.

dd-java-agent/instrumentation/grizzly-http-2.3.20/src/main/java/datadog/trace/instrumentation/grizzlyhttp232/GrizzlyByteBodyInstrumentation.java
```diff
@@ -0,0 +1,203 @@
...
+
+  static class NIOInputStreamReadAdvice {
+    @Advice.OnMethodExit(suppress = Throwable.class)
+    static void after(@Advice.This final NIOInputStream thiz, @Advice.Return int ret) {
+      StoredByteBody storedByteBody =
+          InstrumentationContext.get(NIOInputStream.class, StoredByteBody.class).get(thiz);
+      if (storedByteBody == null) {
+        return;
+      }
+      if (ret == -1) {
```
dougqh, line 118: I see, so instrumenting the individual read methods.  Nice.

dd-java-agent/instrumentation/servlet/src/main/java/datadog/trace/instrumentation/servlet/http/ServletInputStreamWrapper.java
dougqh: Okay

dd-java-agent/instrumentation/servlet/src/main/java/datadog/trace/instrumentation/servlet/http/ServletRequestBodyInstrumentation.java
dougqh: I fine with seeing how this pans out in time.

I don't necessarily find the indirection prettier myself.  I think all the components are still pretty aware of the notification system conceptually.  The indirection would only really be useful in my view if we were going to swap out components and dynamically wire things up.  Since we're not likely to do that, I don't see a great deal of value in it.

internal-api/src/main/java/datadog/trace/api/gateway/InstrumentationGateway.java
```diff
@@ -30,6 +33,13 @@ public InstrumentationGateway() {
     callbacks = new AtomicReferenceArray<>(MAX_EVENTS);
   }
 
+  // for tests
+  void reset() {
+    for (int i = 0; i < MAX_EVENTS; i++) {
```
dougqh, line 38: callbacks.size() will actually be less confusing for the optimizer.

internal-api/src/main/java/datadog/trace/api/http/StoredCharBody.java
dougqh: Yes, that's what I was thinking.  When I was prototyping reducing cost to native / embedded runtimes previously, I ended up using direct buffers to eliminate the copying.

internal-api/src/main/java/datadog/trace/api/http/StoredCharBody.java
dougqh: Alright, we can wait on that I suppose.  I'm a little worried we might design ourselves into a corner performance wise, but I guess we'll deal with that later on.

dougqh (approved):
I'm unblocking for now, but I do have one big concern still.

I'm confused by the threading model the AppRequestContext.  Specifically, the mix of non-volatile and volatile mutable fields.

I also have a several performance nits, but nothing, I think absolutely has to be fixed at this time.

dd-java-agent/appsec/src/main/java/com/datadog/appsec/event/EventDispatcher.java
cataphract: This is only called on initialization, to determine what hooks to use the instrumentation gateway side. But I guess giving the capacity to the constructor wouldn't hurt.

dd-java-agent/appsec/src/main/java/com/datadog/appsec/event/EventDispatcher.java
```diff
@@ -193,6 +199,23 @@ public Flow publishDataEvent(
...
     for (int i = 0; i < numElements; i++) {
```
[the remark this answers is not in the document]
cataphract: Turns out this method isn't called anymore.

dd-java-agent/appsec/src/main/java/com/datadog/appsec/gateway/AppSecRequestContext.java
cataphract: It's not necessarily read by the same thread. There is this public method:

```
  /** @return the portion of the body read so far, if any */
  public CharSequence getStoredRequestBody() { // ...
```

that could, at least in principle, be called from other threads that have access to the `AppSecReqContext`. The other fields:

```
  // assume these will always be written and read by the same thread
  private String savedRawURI;
  private final CaseInsensitiveMap<List<String>> collectedHeaders = new CaseInsensitiveMap<>();
  private List<StringKVPair> collectedCookies = new ArrayList<>(4);
  private boolean finishedHeaders;
  private String ip;
```

need not be volatile because they're only accessed through package-private methods which are called by IG hooks. Their purpose is to collect different pieces about the request so that they can be published and delivered in one go to the appsec bits (currently the WAF).

The assumption here is that the instrumentations calling the producer side of the IG provide all these different bits on the same thread (though I suppose the actual requirements are weaker; there could be some form of synchronization for us to accidentally piggyback on). If there is some http server for which this doesn't hold, this needs to be revisited.

dd-java-agent/appsec/src/main/java/com/datadog/appsec/gateway/GatewayBridge.java
cataphract: The problem with the switches is that they don't work with all the types. In particular, `Address` is not an enum. I'll adjust the capacity though

PR #2963, opened 2021-07-27
rework jetty-client

richardstartin:
Emits correct checkpoints to capture things like content serialisation. This preserves some of the callback behaviour, but it feels wrong to activate the parent at the time of the request for all callbacks. To prevent the child span from waiting on itself, I had to prevent `HttpClientTest` from waiting for its children. Removes some leaks in jetty.

The checkpoints are now correct
E.g. 
successful 
```
Test worker:              |-startSpan/27-|-startSpan/28-|-suspend/28-|-----------|------------|-endSpan/27-|
HttpClient@1909584114-56: |--------------|--------------|------------|-resume/28-|-endSpan/28-|------------|
```
connection refused:
```
Test worker:             |-startSpan/34-|-startSpan/35-|-suspend/35-|-----------|------------|-endSpan/34-|
HttpClient@608869962-55: |--------------|--------------|------------|-resume/35-|-endSpan/35-|------------|
```


dd-java-agent/instrumentation/jetty-client-9.1/src/main/java/datadog/trace/instrumentation/jetty_client/BufferingResponseListenerInstrumentation.java
```diff
@@ -0,0 +1,75 @@
...
+import java.util.HashMap;
+import java.util.Map;
+import net.bytebuddy.asm.Advice;
+import net.bytebuddy.description.type.TypeDescription;
+import net.bytebuddy.matcher.ElementMatcher;
+import org.eclipse.jetty.client.api.Request;
+import org.eclipse.jetty.client.api.Response;
+
+@AutoService(Instrumenter.class)
+public class BufferingResponseListenerInstrumentation extends Instrumenter.Tracing {
```
richardstartin, line 22: This is the internal class for handling marshalling of the buffer, only instrumenting this to emit checkpoints

dd-java-agent/instrumentation/jetty-client-9.1/src/main/java/datadog/trace/instrumentation/jetty_client/CallbackWrapper.java
```diff
@@ -0,0 +1,157 @@
...
+
+import datadog.trace.bootstrap.instrumentation.api.AgentScope;
+import datadog.trace.bootstrap.instrumentation.api.AgentSpan;
+import java.nio.ByteBuffer;
+import org.eclipse.jetty.client.api.Request;
+import org.eclipse.jetty.client.api.Response;
+import org.eclipse.jetty.client.api.Result;
+import org.eclipse.jetty.http.HttpField;
+
+public class CallbackWrapper implements Response.Listener, Request.Listener {
```
richardstartin, line 13: We wouldn't need this if we weren't associating callback executions with their grandparent, but we are, so we do.

dd-java-agent/instrumentation/jetty-client-9.1/src/main/java/datadog/trace/instrumentation/jetty_client/FutureResponseListenerInstrumentation.java
```diff
@@ -0,0 +1,82 @@
...
+import java.util.Map;
+import net.bytebuddy.asm.Advice;
+import net.bytebuddy.description.type.TypeDescription;
+import net.bytebuddy.matcher.ElementMatcher;
+import org.eclipse.jetty.client.api.Request;
+import org.eclipse.jetty.client.api.Response;
+import org.eclipse.jetty.client.api.Result;
+
+@AutoService(Instrumenter.class)
+public final class FutureResponseListenerInstrumentation extends Instrumenter.Tracing {
```
richardstartin, line 23: This is the main request scoped stateful future used internally.

dd-java-agent/instrumentation/jetty-client-9.1/src/main/java/datadog/trace/instrumentation/jetty_client/JettyAddListenerInstrumentation.java
```diff
@@ -33,42 +31,110 @@ public JettyAddListenerInstrumentation() {
...
+            .and(isPublic())
+            .and(named("onFailure"))
+            .and(takesArgument(0, named("org.eclipse.jetty.client.api.Request$FailureListener"))),
+        JettyAddListenerInstrumentation.class.getName() + "$WrapRequestFailureListener");
+    transformation.applyAdvice(
+        isMethod()
+            .and(isPublic())
+            .and(named("onComplete"))
+            .and(takesArgument(0, named("org.eclipse.jetty.client.api.Response$CompleteListener"))),
+        JettyAddListenerInstrumentation.class.getName() + "$WrapResponseCompleteListener");
```
richardstartin, line 67: Not all callbacks are traced, because we need to preserve the listener type in the advice, and it's time consuming. I don't see value in tracing all possible callbacks anyway.

dd-java-agent/instrumentation/jetty-client-9.1/src/main/java/datadog/trace/instrumentation/jetty_client/JettyClientInstrumentation.java
```diff
@@ -57,32 +68,33 @@ public void adviceTransformations(AdviceTransformation transformation) {
...
-      final AgentScope scope = activateSpan(span);
+      InstrumentationContext.get(Request.class, AgentSpan.class).put(request, span);
 
       DECORATE.afterStart(span);
       DECORATE.onRequest(span, request);
       propagate().inject(span, request, SETTER);
-      // Add listener to the front of the list.
-      listeners.add(0, new SpanFinishingCompleteListener(span));
-      return scope;
+      span.startThreadMigration();
```
richardstartin, line 86: This method is async, and we can't piggyback of use of standard concurrency primitives, so we need to mark that the span has left the thread manually (unless we want to instrument the various incarnations of jetty concurrency, but I don't see good ROI in doing that)

dd-java-agent/testing/src/main/groovy/datadog/trace/agent/test/base/HttpClientTest.groovy
```diff
@@ -279,7 +279,6 @@ abstract class HttpClientTest extends AgentTestRunner {
     def status = runUnderTrace("parent") {
       doRequest(method, server.address.resolve("/success"), ["is-dd-server": "false"], "") {
         runUnderTrace("child") {
-          blockUntilChildSpansFinished(1)
```
richardstartin, line 282: I don't understand why this is here, or why it isn't usually a problem. In this case, the child span was waiting on itself...

tylerbenson: This is supposed to be blocking until the http client span finishes and is helpful to ensure proper ordering of client/callback spans.

richardstartin: That's problematic if you want to profile the request, it means the callback to finish the span must be the first to be called on completion, which means you necessarily miss the rest. 

It feels a lot more natural for the child to be attributed to the request for jetty-client.

luhenry (approved):

dd-java-agent/instrumentation/jetty-client-9.1/jetty-client-9.1.gradle
```diff
@@ -6,12 +6,8 @@ muzzle {
...
     group = "org.eclipse.jetty"
     module = "jetty-client"
-    versions = "[9.1,10)"
-  }
-  fail {
-    group = "org.eclipse.jetty"
-    module = "jetty-client"
-    versions = "[,9.1)"
+    versions = "[9.0,)"
+    assertInverse = true
```
richardstartin, line 10: This extends the muzzle range by at least 3 major versions so far

dd-java-agent/testing/src/main/groovy/datadog/trace/agent/test/base/HttpClientTest.groovy
richardstartin: Completion callbacks now execute after the request span has finished, and the line above has been reinstated.

dd-java-agent/instrumentation/jetty-client-9.1/jetty-client-9.1.gradle
```diff
@@ -8,11 +8,6 @@ muzzle {
     module = "jetty-client"
     versions = "[9.1,10)"
   }
-  fail {
-    group = "org.eclipse.jetty"
-    module = "jetty-client"
-    versions = "[,9.1)"
-  }
```
tylerbenson, line 15: Why remove this?

richardstartin: because it passes for some versions

richardstartin: I've added it back, and if it doesn't pass unexpectedly I'll squash it into the commit so the change doesn't register.

tylerbenson: If earlier versions pass, then the supported version range should probably be updated in the module name.

richardstartin: If you want to bisect which minor version below 9.1 passes with a tool which doesn't respect gradle caching, be my guest 😆, but it feels pragmatic to me to leave the muzzle range at `[9.1,10)` (there's obviously some level of tolerance for muzzle ranges given that it doesn't work with libraries targeting JDK11+

richardstartin: I've dropped the commit because it passed unexpectedly for some version sub 9.1 as this instrumentation is now generalised, just not all the way back to 9.0.0

dd-java-agent/instrumentation/jetty-client-9.1/src/main/java/datadog/trace/instrumentation/jetty_client/JettyAddListenerInstrumentation.java
```diff
@@ -33,42 +31,110 @@ public JettyAddListenerInstrumentation() {
...
-        InstrumentationContext.get(Request.RequestListener.class, AgentSpan.class)
-            .put((Request.RequestListener) listener, parent);
+  public static class WrapRequestFailureListener {
+    @Advice.OnMethodEnter(suppress = Throwable.class)
+    public static void methodEnter(
+        @Advice.This Request request,
+        @Advice.Argument(value = 0, readOnly = false) Request.FailureListener listener) {
+      if (!(listener instanceof CallbackWrapper)) {
+        listener =
+            new CallbackWrapper(
```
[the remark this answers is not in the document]
richardstartin: I have read the code which is being instrumented.  `org.eclipse.jetty.client.HttpRequest` always wraps callbacks in an anonymous class. It is not possible to specify a downcast to an anonymous class.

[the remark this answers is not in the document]
tylerbenson: no, but we're taking something that implements one of the listener types and wrapping it in something that implements multiple types.  It's probably fine since you're checking inside the wrapper, but it will change the internal behavior of the callback execution.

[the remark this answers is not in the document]
richardstartin: Maybe I'm missing something, could you explain the domesday scenario wrapping `listener` leads to, given that it's wrapped in an anonymous type?

```java
    @Override
    public Request onComplete(final Response.CompleteListener listener)
    {
        this.responseListeners.add(new Response.CompleteListener()
        {
            @Override
            public void onComplete(Result result)
            {
                listener.onComplete(result);
            }
        });
        return this;
    }
```

dd-java-agent/instrumentation/java-concurrent/src/main/java/datadog/trace/instrumentation/java/concurrent/AsyncPropagatingDisableInstrumentation.java
```diff
@@ -111,9 +112,6 @@ public void adviceTransformations(AdviceTransformation transformation) {
         advice);
     transformation.applyAdvice(
         named("start").and(isDeclaredBy(named("rx.internal.util.ObjectPool"))), advice);
-    transformation.applyAdvice(
-        named("addConnection").and(isDeclaredBy(named("com.squareup.okhttp.ConnectionPool"))),
-        advice);
```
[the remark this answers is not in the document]
devinsba: Seems like maybe a bad rebase, since these were added in another one of your PRs

devinsba (approved):
Looks good to me once you have the rebase fixed up

dd-java-agent/instrumentation/java-concurrent/src/main/java/datadog/trace/instrumentation/java/concurrent/AsyncPropagatingDisableInstrumentation.java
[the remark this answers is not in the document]
richardstartin: Thanks for spotting this.

dd-java-agent/instrumentation/jetty-client-9.1/jetty-client-9.1.gradle
tylerbenson: I think the failure assertion is helpful, otherwise we end up with partial matching and an unexpected instrumentation state.

richardstartin: Please explain why this is important, bearing in mind that the versions where the boundary is blurred were released in 2013, and there isn't an instrumentation targeting older versions of jetty-client. Please explain what happens if somebody happens to be using 9.0.7.v20131107 (released November 2013) and the instrumentation activates, and if they are using 9.0.0.v20130308 (March 2013) and it doesn't activate? How does knowing the precise boundary help you unless you have an instrumentation overlap?

dd-java-agent/instrumentation/jetty-client-9.1/src/main/java/datadog/trace/instrumentation/jetty_client/JettyAddListenerInstrumentation.java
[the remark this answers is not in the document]
dougqh: Yes, the point about jetty itself using wrappers makes sense to me.
I also think Listeners are less likely to be type checked than core objects, so are generally safer.

@tylerbenson Is your concern that the incoming listener implements more types / less types?
I could see more types mattering if there are additional type checks further down, but yes, clarifying the concern would help.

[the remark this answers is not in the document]
tylerbenson: As I said, it's probably fine... the main point I was trying to get across is that the behavior of `org.eclipse.jetty.client.ResponseNotifier` will change slightly because it uses instanceof checks when iterating through the listeners.  Eg:
```
    public void notifySuccess(List<Response.ResponseListener> listeners, Response response)
    {
        for (Response.ResponseListener listener : listeners)
        {
            if (listener instanceof Response.SuccessListener)
                notifySuccess((Response.SuccessListener)listener, response);
        }
    }
```

I'm not concerned that it's breaking anything.  I'm more concerned about the muzzle issue than this.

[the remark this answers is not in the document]
richardstartin: I'm not concerned about that method. As I have tried to make clear, listeners are wrapped in an anonymous class of the appropriate type on registration in `HttpRequest` - defensive programming on the part of the framework authors so the interface can be relied on as a marker. This renders the concern moot, but even if this were not the case, here is the implementation of `notifySuccess`:

```java
    private void notifySuccess(Response.SuccessListener listener, Response response)
    {
        try
        {
            listener.onSuccess(response);
        }
        catch (Throwable x)
        {
            LOG.info("Exception while notifying listener " + listener, x);
        }
    }
```

If a `CallbackWrapper` which doesn't wrap a `Response.SuccessListener` somehow reaches here unwrapped, it will fail the instanceof check on the wrapped listener.

The question has to be what is the alternative? There is no testing of what happens when the callback is a lambda or method reference and field injection fails - the context store will fall back to a weak map. If it's a method reference of stateless lambda, the callback will be effectively a singleton, and this will lead to incorrectness. Field injection also requires that every implementation of the interfaces are found.

Regarding the muzzle range, I have asked for an explanation as to why it's important, as opposed to reiteration that the range has changed slightly. The muzzle story is already murky because 10+ is just ignored because these versions target JDK11 bytecode, so I'm struggling to understand the objection here.

dd-java-agent/instrumentation/jetty-client-9.1/jetty-client-9.1.gradle
dougqh: Yes, I think the value is negligible as well.  At some point, I think we just need to rethink how we handle multiple version of libraries (and JDKs), but not this for PR.

dougqh (approved):

dd-java-agent/instrumentation/jetty-client-9.1/jetty-client-9.1.gradle
tylerbenson: I went ahead and fixed the muzzle range check: [`24f14f8` (#2963)](https://github.com/DataDog/dd-trace-java/pull/2963/commits/24f14f88f5a5f73c19f86ce8791a946c4f703c98)

tylerbenson (approved):

PR #2965, opened 2021-07-28
Print stacktraces for spans which have invalid sequence of events

luhenry:
This allows to more specifically pinpoint where the events were created,
and diagnose why an event was wrongfully created or should have been
created.

### DIFF
--- a/dd-java-agent/testing/src/main/groovy/datadog/trace/agent/test/checkpoints/Event.groovy
+++ b/dd-java-agent/testing/src/main/groovy/datadog/trace/agent/test/checkpoints/Event.groovy
@@ -2,20 +2,23 @@ package datadog.trace.agent.test.checkpoints
 
 import static datadog.trace.api.Checkpointer.*
 import datadog.trace.api.DDId
+import java.util.stream.Collectors
 
 class Event {
   private final int flags
   private final long threadId
   private final String threadName
+  private final StackTraceElement[] stackTrace
   private final DDId traceId
   private final DDId spanId
 
-  Event(int flags, DDId traceId, DDId spanId, Thread thread) {
+  Event(int flags, DDId traceId, DDId spanId, Thread thread, StackTraceElement[] stackTrace) {
     this.flags = flags
     this.traceId = traceId
     this.spanId = spanId
     this.threadId = thread.id
     this.threadName = thread.name
+    this.stackTrace = stackTrace
   }
 
   String getName() {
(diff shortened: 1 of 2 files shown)


dd-java-agent/testing/src/main/groovy/datadog/trace/agent/test/checkpoints/TimelineCheckpointer.groovy
```diff
@@ -31,7 +33,9 @@ class TimelineCheckpointer implements Checkpointer {
     def invalidEvents = TimelineValidator.validate(spanEvents, threadEvents, orderedEvents)
     if (!invalidEvents.empty) {
       System.err.println("=== Invalid checkpoint events encountered")
-      invalidEvents.each { System.err.println(it) }
+      invalidEvents.stream().map { it.spanId }.distinct().forEach {
+        spanEvents.get(it).each { System.err.println(it) }
```
jbachorik, line 37: This will print the whole span sequence - that can be a lot of output. Is this desired/expected?

dd-java-agent/testing/src/main/groovy/datadog/trace/agent/test/checkpoints/Event.groovy
```diff
@@ -57,11 +60,15 @@ class Event {
...
   String toString() {
-    return "${name}/${spanId} (thread: ${threadName})\n"
+    return "${name}/${spanId} (thread: ${threadName})\n" +
+        String.join("\n",
+            stackTrace.stream()
+                .filter {
+                    !(it.className.startsWith("org.codehaus.groovy") || it.className.startsWith("groovy")) &&
+                    !(it.className.startsWith("org.spockframework")) }
+                .map { "  " + it.toString() }
+                .collect(Collectors.toList())) +
```
jbachorik, line 71: Could you use `Collectors.joining()` to avoid creating the temporary list?

dd-java-agent/testing/src/main/groovy/datadog/trace/agent/test/checkpoints/TimelineCheckpointer.groovy
luhenry: That's on purpose. This will allow us to understand better how we got to the current situation. Also, it will only print the events related to the faulty spans, not the correct ones.

richardstartin: I've been fixing these things for weeks now, and have a pretty good understanding of the instrumentations, it's always the suspends and resumes that diagnose the problem. The resume typically gets you the thread name which is a big hint, but you don't want the stack trace because it's going to be some runnable in a thread pool, and the suspend's stack trace tells you how it happened. So I would recommend parsimony here, or maybe have an ultra verbose mode which prints stack traces for everything.

PR #2967, opened 2021-07-28
Prevent DDSpan from escaping unwrapped with OT

tylerbenson:
Various methods in OTSpan are returning the underlying span object, not the wrapped one.  This is problematic because users expect to be able to cast back to an OT type.


devinsba (approved):

mcculls (approved):
Good catch

PR #2962, opened 2021-07-27
Mark tests which fail the checkpoint sequence validator

luhenry:
No tests should be added with this disable. This disable is only used to track which integrations needs fixing. Adding more means we are regressing the status quo.

PR #2968, opened 2021-07-28
Checkpoint validator fixes and improvements

jbachorik:
This change simplifies and fixes the sequence validator - instead of trying to assert the global sequence it will use per-thread sequence only. The reason is that the global sequence is impossible to check rigorously without false negatives. Restricting the sequence to one thread only the state machine becomes much simpler while still detecting the obvious errors.

The validation was split into 4 different sub-sections:
- thread sanity
- thread sequence correctness
- suspend-resume balance
- interval checks (proper start/end, no overlapping spans on a single thread)

This allowed to add granularity to what should be validated per test case. See eg. https://github.com/DataDog/dd-trace-java/pull/2968/files#diff-791c5ceeb56a85b4533527d498b452f9c6c8ec3995bfbb232476462abe1eb537R149 how the intervals checks would be disabled for vertx tests - the interplay of netty and vertx is causing very weird sequence of overlapping spans. The overlapping spans are a minor error and since the fix is very non-trivial we can mute the validator until we are able to properly fix the vertx instrumentation (if at all).

A more rigorous state machine test was added where all expected states and transitions are validated. This should provide more confidence than trying to mimic the real life sequences.

The last thing included here is taken from @luhenry PR for event stacktrace collection (https://github.com/DataDog/dd-trace-java/pull/2965) - I have extracted the relevant code and added it to this PR, mostly so that @luhenry does not need to  deal with merge conflicts in his PR.


luhenry (approved):
The idea would be to update https://github.com/DataDog/dd-trace-java/pull/2962 to use `CheckpointValidator.excludeValidations` in place of `IgnoreIf`?

jbachorik:
> The idea would be to update #2962 to use CheckpointValidator.excludeValidations in place of IgnoreIf?

Yes. Where it would make sense.

richardstartin (approved):

richardstartin:
@jbachorik does this contain a fix for the no suspend chronologically after a span end rule? That one is invalidating 5-10 instrumentations and this is a basic consequence of how async propagation works, so needs to be accepted.

dd-java-agent/testing/src/main/groovy/datadog/trace/agent/test/checkpoints/CheckpointValidator.groovy
```diff
@@ -0,0 +1,82 @@
...
+    def threadSequenceValidator = new ThreadSequenceValidator()
+    for (def event : events) {
+      if (!excludedValidations.contains(CheckpointValidationMode.SUSPEND_RESUME)) {
+        if (!suspendResumeValidator.onEvent(event)) {
+          invalidEvents.add([event, CheckpointValidationMode.SUSPEND_RESUME])
+        }
+      }
+      if (!excludedValidations.contains(CheckpointValidationMode.THREAD_SEQUENCE)) {
+        if (!threadSequenceValidator.onEvent(event)) {
+          invalidEvents.add([event, CheckpointValidationMode.THREAD_SANITY])
```
luhenry, line 67: Typo:
```suggestion
          invalidEvents.add([event, CheckpointValidationMode.THREAD_SEQUENCE])
```

luhenry (changes requested):

luhenry:
> @jbachorik does this contain a fix for the no suspend chronologically after a span end rule? That one is invalidating 5-10 instrumentations and this is a basic consequence of how async propagation works, so needs to be accepted.

If there is any work needed there, please let me update https://github.com/DataDog/dd-trace-java/pull/2962 first, it's getting increasingly frustrating to go through the 90+ test classes manually that are needs to be annotated differently.

luhenry (approved):

dd-java-agent/testing/src/main/groovy/datadog/trace/agent/test/checkpoints/CheckpointValidator.groovy
```diff
@@ -54,26 +54,24 @@ class CheckpointValidator {
...
-        }
-      }
-      if (!excludedValidations.contains(CheckpointValidationMode.THREAD_SEQUENCE)) {
-        if (!threadSequenceValidator.onEvent(event)) {
-          invalidEvents.add([event, CheckpointValidationMode.THREAD_SEQUENCE])
+    if (!excludedValidations.contains(CheckpointValidationMode.SEQUENCE)) {
+      def suspendResumeValidator = new SuspendResumeValidator()
+      def threadSequenceValidator = new ThreadSequenceValidator()
+      for (def event : events) {
+        if (!excludedValidations.contains(CheckpointValidationMode.SEQUENCE)) {
```
luhenry, line 61: Duplicate.

luhenry (changes requested):

dd-java-agent/testing/src/main/groovy/datadog/trace/agent/test/checkpoints/CheckpointValidator.groovy
```diff
@@ -58,13 +58,11 @@ class CheckpointValidator {
...
       def threadSequenceValidator = new ThreadSequenceValidator()
       for (def event : events) {
-        if (!excludedValidations.contains(CheckpointValidationMode.SEQUENCE)) {
-          if (!suspendResumeValidator.onEvent(event)) {
-            invalidEvents.add([event, CheckpointValidationMode.SEQUENCE])
-          }
-          if (!threadSequenceValidator.onEvent(event)) {
-            invalidEvents.add([event, CheckpointValidationMode.SEQUENCE])
-          }
+        if (!suspendResumeValidator.onEvent(event)) {
```
luhenry, line 61: I would have `if (!suspendResumeValidator.onEvent(event) || !threadSequenceValidator.onEvent(event))` to avoid adding duplicate events.

jbachorik: `invalidEvents` is a set so they will be deduplicated

jbachorik: But yes, could be simplified now that we don't use different validation modes checks.

luhenry (approved):

PR #2972, opened 2021-07-29
httpasyncclient checkpoints resilience

richardstartin:
Making the state machine resilient to redirects and failure is complicated, so this change cheats for finishing the span - it _always_ resumes before finishing.

The most interesting callbacks worth profiling are here:

```java
public void consumeContent(
        final InternalState state,
        final ContentDecoder decoder,
        final IOControl ioctrl) throws IOException {
    if (this.log.isDebugEnabled()) {
        this.log.debug("[exchange: " + state.getId() + "] Consume content");
    }
    if (state.getFinalResponse() != null) {
        final HttpAsyncResponseConsumer<?> responseConsumer = state.getResponseConsumer();
        responseConsumer.consumeContent(decoder, ioctrl);
    } else {
        final ByteBuffer tmpbuf = state.getTmpbuf();
        tmpbuf.clear();
        decoder.read(tmpbuf);
    }
}


public void produceContent(
        final InternalState state,
        final ContentEncoder encoder,
        final IOControl ioctrl) throws IOException {
    if (this.log.isDebugEnabled()) {
        this.log.debug("[exchange: " + state.getId() + "] produce content");
    }
    final HttpAsyncRequestProducer requestProducer = state.getRequestProducer();
    state.setRequestContentProduced();
    requestProducer.produceContent(encoder, ioctrl);
    if (encoder.isCompleted()) {
        requestProducer.resetRequest();
    }
}
```

The instrumentation  now ensures we resume and suspend around these methods, so that it doesn't matter how many redirects there are or if there is a failure, checkpoints always balance and span the most interesting work.

All tests validate after this change.

### DIFF
--- a/dd-java-agent/instrumentation/apache-httpasyncclient-4/src/main/java/datadog/trace/instrumentation/apachehttpasyncclient/TraceContinuedFutureCallback.java
+++ b/dd-java-agent/instrumentation/apache-httpasyncclient-4/src/main/java/datadog/trace/instrumentation/apachehttpasyncclient/TraceContinuedFutureCallback.java
@@ -29,15 +29,17 @@ public class TraceContinuedFutureCallback<T> implements FutureCallback<T> {
     this.delegate = delegate;
   }
 
-  public void responseReceived() {
-    // happens once the the response has been received,
-    // and captures activity like serialization which
-    // we want to notify profiling about
+  public void resume() {
     clientSpan.finishThreadMigration();
+  }
+
+  public void suspend() {
+    clientSpan.startThreadMigration();
   }
 
   @Override
   public void completed(final T result) {
+    resume();
     DECORATE.onResponse(clientSpan, context);
     DECORATE.beforeFinish(clientSpan);
     clientSpan.finish(); // Finish span before calling delegate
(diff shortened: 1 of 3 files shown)


tylerbenson (approved):

dd-java-agent/instrumentation/apache-httpasyncclient-4/src/main/java/datadog/trace/instrumentation/apachehttpasyncclient/TraceContinuedFutureCallback.java
```diff
@@ -29,15 +29,17 @@ public TraceContinuedFutureCallback(
...
 
-  public void responseReceived() {
-    // happens once the the response has been received,
-    // and captures activity like serialization which
-    // we want to notify profiling about
+  public void resume() {
     clientSpan.finishThreadMigration();
   }
 
+  public void suspend() {
```
devinsba, line 36: Do these 2 methods need to be public?

devinsba (approved):

dd-java-agent/instrumentation/apache-httpasyncclient-4/src/main/java/datadog/trace/instrumentation/apachehttpasyncclient/TraceContinuedFutureCallback.java
richardstartin: Yes, because it needs to be called by augmented code in several packages. Making these package private would not result in a verification error after applying the advice (as far as I'm aware) but would result in an `IllegalAccessError`. I think muzzle may also fail.

luhenry (approved):

PR #2974, opened 2021-07-29
Various muzzle fixes

tylerbenson:
Muzzle got broken when the byte buddy plugin changes were made that changed the compile directory to `raw`, causing the muzzle inspection to skip everything.

I also added an assert to the classloader creation to ensure that the instrumentation set is never empty.


mcculls (approved):
👍

devinsba (approved):

PR #2980, opened 2021-07-30
Add vertx smoke test

devinsba:


dd-smoke-tests/vertx-3.9/application/.editorconfig
```diff
@@ -0,0 +1,11 @@
+# EditorConfig is awesome: https://EditorConfig.org
```
tylerbenson, line 1: Seems like there's several files in here that are intended for a top level project.  Are they really needed here?

devinsba: The pattern I followed for this test is that it is a top level project. The jar is built using it's own gradle config and then used in the tests. I can remove this, the Readme, and the one test but otherwise the other files are all necessary

tylerbenson (approved):

dd-smoke-tests/vertx-3.9/application/.gitignore
```diff
@@ -0,0 +1,186 @@
+### Vert.x ###
+.vertx/
+
```
bantonsson, line 3: Is this huge `.gitignore` file really necessary?

dd-smoke-tests/vertx-3.9/application/gradlew
```diff
@@ -0,0 +1,185 @@
+#!/usr/bin/env sh
```
bantonsson, line 1: Can't we just add links to the scripts and `jar` file instead of copies?

bantonsson (approved):

dd-smoke-tests/vertx-3.9/application/gradlew
devinsba: It's a different version of gradle. I tried downgrading to our version but I couldn't make the build work because it's a different major version

dd-smoke-tests/vertx-3.9/application/.gitignore
devinsba: Probably not. I could at least prune out the stuff that we exclude at the root level

dd-smoke-tests/vertx-3.9/application/gradlew
bantonsson: I was kind of hoping that the gradle wrapper was actually just a wrapper that could download different versions of gradle.

devinsba: Oh yeah. Let me double check that and make the change if I can

bantonsson: Sorry, I meant to say that I checked, and the script is follow then symlinks to find the _correct_ `jar` file, so there can only be one 😞

devinsba: They were the exact same files, replacing with links

devinsba: Ok, yeah I see that now. I at least created a link to the jar because that file is the same

PR #2983, opened 2021-08-02
Improve the checkpoint validator output

jbachorik:
This increases the amount of information printed on validation failure - making it easier to debug and assert the failures.


dd-java-agent/testing/src/main/groovy/datadog/trace/agent/test/checkpoints/IntervalValidator.groovy
```diff
@@ -23,125 +25,126 @@ class IntervalValidator extends AbstractContextTracker {
...
+    closedIntervalsBySpan.put(interval.spanId, interval)
+
+    return result ? Result.OK : Result.FAILED
   }
 
   @Override
-  boolean endSequence() {
-    return openIntervalsByTime.findAll {!it.suspended}.empty
+  def endSequence() {
+    return openIntervalsByTime.findAll {!it.suspended}.empty ? Result.OK : Result.FAILED
```
bantonsson, line 148: Maybe not for this PR, but would it make sense to add the _failing_ open intervals to the message?

dd-java-agent/testing/src/main/groovy/datadog/trace/agent/test/checkpoints/SuspendResumeValidator.groovy
```diff
@@ -3,71 +3,86 @@ package datadog.trace.agent.test.checkpoints
...
     spanFinished = true
-    return spanStarted
+    return spanStarted ? Result.OK : Result.FAILED.withMessage("Attempting to end a non-existing span ${event?.spanId}")
   }
 
   @Override
-  boolean endSequence() {
-    return spanFinished && resumeCount == suspendCount
+  def endSequence() {
+    return spanFinished && resumeCount == suspendCount ? Result.OK : Result.FAILED
```
bantonsson, line 86: Same here. Maybe the diff would be useful in a message.

bantonsson (approved):

dd-java-agent/testing/src/main/groovy/datadog/trace/agent/test/checkpoints/IntervalValidator.groovy
jbachorik: I have an improvement for the interval validator in mind - so I will add that then.

jpbempel (approved):

PR #2987, opened 2021-08-04
Prevent DDSpan from escaping unwrapped with OpenTracing Via the Agent

tylerbenson:
This was previously fixed with `dd-trace-ot` in #2967, but I forgot to apply the same change to the agent.

Also make OTSpan copies more consistent.


devinsba (approved):
I feel like it would be nice if there weren't 3 copies of this but it's probably not avoidable, maybe we could get it down to 2 🤷 ?

mcculls (approved):

dougqh (approved):
For now the fix is fine.  I'm a bit concerned that such an obvious bug in the wrappers made it through review previously, but so be it.

I think this is indicative a general design failure with the wrappers.  I don't think they should have been in a common type hierarchy, then we wouldn't have had this problem.

bantonsson (approved):
Yeah it might be easier to guard against these types of things if they user needed to _convert_ between the types instead of just _casting_. Might be something to think about for a future API.

PR #2988, opened 2021-08-04
Allow `split-by-tag=servlet.context` to set service name.

tylerbenson:
IIRC this used to work, but stopped with the changes to remove the Rules a while back.

This should allow users that configure a service name to still have their war files treated as separate services.

### DIFF
--- a/dd-trace-core/src/test/groovy/datadog/trace/core/taginterceptor/TagInterceptorTest.groovy
+++ b/dd-trace-core/src/test/groovy/datadog/trace/core/taginterceptor/TagInterceptorTest.groovy
@@ -6,6 +6,7 @@ import datadog.trace.api.config.GeneralConfig
 import datadog.trace.api.env.CapturedEnvironment
 import datadog.trace.api.sampling.PrioritySampling
 import datadog.trace.bootstrap.instrumentation.api.AgentSpan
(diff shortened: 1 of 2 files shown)


dd-trace-core/src/test/groovy/datadog/trace/core/taginterceptor/TagInterceptorTest.groovy
```diff
@@ -178,6 +179,31 @@ class TagInterceptorTest extends DDCoreSpecification {
       .build()
   }
 
+  def "split-by-tags for servlet.context"() {
+    setup:
+    def tracer = createSplittingTracer(InstrumentationTags.SERVLET_CONTEXT)
+
+    when:
+    def span = tracer.buildSpan("some span")
```
dougqh, line 187: Was this test failing before?  I'm mostly wondering if this test is too artificial to really test what we're trying to test.

tylerbenson: yes, I confirmed that this test fails without my change to `TagInterceptor`.

mcculls (approved):

lpriima (approved):

bantonsson (approved):

PR #2989, opened 2021-08-04
Ignore exceptions thrown when canceling GRPC streaming

tylerbenson:
The stacktrace is unlikely to be useful and not something that should be treated as an error.

### DIFF
--- a/dd-java-agent/agent-bootstrap/src/main/java/datadog/trace/bootstrap/FieldBackedContextStore.java
+++ b/dd-java-agent/agent-bootstrap/src/main/java/datadog/trace/bootstrap/FieldBackedContextStore.java
@@ -69,6 +69,25 @@ public final class FieldBackedContextStore implements ContextStore<Object, Objec
     }
   }
 
+  @Override
+  public Object remove(Object key) {
+    if (key instanceof FieldBackedContextAccessor) {
+      final FieldBackedContextAccessor accessor = (FieldBackedContextAccessor) key;
+      Object existingContext = accessor.$get$__datadogContext$(storeId);
+      if (null != existingContext) {
+        synchronized (accessor) {
+          existingContext = accessor.$get$__datadogContext$(storeId);
+          if (null != existingContext) {
+            accessor.$put$__datadogContext$(storeId, null);
+          }
+        }
+      }
+      return existingContext;
+    } else {
+      return weakStore().remove(key);
+    }
+  }
+
   // only create WeakMap-based fall-back when we need it
   private volatile WeakMapContextStore weakStore;
   private final Object synchronizationInstance = new Object();
--- a/dd-java-agent/instrumentation/grpc-1.5/src/main/java/datadog/trace/instrumentation/grpc/client/ClientCallImplInstrumentation.java
+++ b/dd-java-agent/instrumentation/grpc-1.5/src/main/java/datadog/trace/instrumentation/grpc/client/ClientCallImplInstrumentation.java
@@ -109,7 +109,7 @@ public final class ClientCallImplInstrumentation extends Instrumenter.Tracing {
   public static final class Cancel {
     @Advice.OnMethodEnter
     public static void before(@Advice.This ClientCall<?, ?> call) {
-      AgentSpan span = InstrumentationContext.get(ClientCall.class, AgentSpan.class).get(call);
+      AgentSpan span = InstrumentationContext.get(ClientCall.class, AgentSpan.class).remove(call);
       if (null != span) {
         span.finish();
       }
@@ -139,7 +139,7 @@ public final class ClientCallImplInstrumentation extends Instrumenter.Tracing {
     @Advice.OnMethodExit(onThrowable = Throwable.class)
     public static void closeObserver(
         @Advice.This ClientCall<?, ?> call, @Advice.Argument(1) Status status) {
-      AgentSpan span = InstrumentationContext.get(ClientCall.class, AgentSpan.class).get(call);
+      AgentSpan span = InstrumentationContext.get(ClientCall.class, AgentSpan.class).remove(call);
       if (null != span) {
         DECORATE.onClose(span, status);
         span.finish();
--- a/dd-java-agent/instrumentation/grpc-1.5/src/main/java/datadog/trace/instrumentation/grpc/server/GrpcServerDecorator.java
+++ b/dd-java-agent/instrumentation/grpc-1.5/src/main/java/datadog/trace/instrumentation/grpc/server/GrpcServerDecorator.java
@@ -81,8 +81,8 @@ public class GrpcServerDecorator extends ServerDecorator {
     span.setTag("status.code", status.getCode().name());
     span.setTag("status.description", status.getDescription());
 
-    onError(span, status.getCause());
-    if (!status.isOk()) {
+    if (!status.isOk() && status != Status.CANCELLED) {
+      onError(span, status.getCause());
       span.setError(true);
     }
 
--- a/dd-java-agent/instrumentation/grpc-1.5/src/main/java/datadog/trace/instrumentation/grpc/server/TracingServerInterceptor.java
+++ b/dd-java-agent/instrumentation/grpc-1.5/src/main/java/datadog/trace/instrumentation/grpc/server/TracingServerInterceptor.java
@@ -18,6 +18,7 @@ import io.grpc.ServerCall;
 import io.grpc.ServerCallHandler;
 import io.grpc.ServerInterceptor;
 import io.grpc.Status;
+import java.util.concurrent.CancellationException;
 
 public class TracingServerInterceptor implements ServerInterceptor {
 
@@ -124,6 +125,9 @@ public class TracingServerInterceptor implements ServerInterceptor {
       try (final AgentScope scope = activateSpan(span)) {
         delegate().onCancel();
         span.setTag("canceled", true);
+      } catch (CancellationException e) {
+        // No need to report an exception or mark as error that it was canceled.
+        throw e;
       } catch (final Throwable e) {
         DECORATE.onError(span, e);
         throw e;
--- a/dd-java-agent/instrumentation/grpc-1.5/src/main/java/datadog/trace/instrumentation/grpc/server/TracingServerInterceptor.java
+++ b/dd-java-agent/instrumentation/grpc-1.5/src/main/java/datadog/trace/instrumentation/grpc/server/TracingServerInterceptor.java
@@ -19,7 +19,6 @@ import io.grpc.ServerCallHandler;
 import io.grpc.ServerInterceptor;
 import io.grpc.Status;
 import java.util.concurrent.CancellationException;
-import java.util.concurrent.atomic.AtomicBoolean;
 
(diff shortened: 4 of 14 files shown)


mcculls (approved):

bantonsson (approved):

charliegracie (approved):

dd-java-agent/instrumentation/grpc-1.5/src/main/java/datadog/trace/instrumentation/grpc/server/GrpcServerDecorator.java
```diff
@@ -75,8 +75,8 @@ public AgentSpan onClose(final AgentSpan span, final Status status) {
     span.setTag("status.code", status.getCode().name());
     span.setTag("status.description", status.getDescription());
 
-    onError(span, status.getCause());
-    if (!status.isOk()) {
```
tylerbenson, line 79: I think the canceled span was still getting marked as an error in the first commit because of this.

dd-java-agent/agent-bootstrap/src/main/java/datadog/trace/bootstrap/FieldBackedContextStore.java
```diff
@@ -69,6 +69,25 @@ public Object putIfAbsent(final Object key, final Factory<Object> contextFactory
     }
   }
 
+  @Override
+  public Object remove(Object key) {
+    if (key instanceof FieldBackedContextAccessor) {
+      final FieldBackedContextAccessor accessor = (FieldBackedContextAccessor) key;
+      Object existingContext = accessor.$get$__datadogContext$(storeId);
```
dougqh, line 76: I'd prefer to see this remove implemented as method on FieldBacckedContextProvider.

dd-java-agent/instrumentation/grpc-1.5/src/main/java/datadog/trace/instrumentation/grpc/server/TracingServerInterceptor.java
```diff
@@ -34,34 +35,45 @@ private TracingServerInterceptor() {}
 
     final Context spanContext = propagate().extract(headers, GETTER);
     final AgentSpan span = startSpan(GRPC_SERVER, spanContext).setMeasured(true);
+    /**
+     * grpc has quite a few states that can finish the span. rather than track down the correct
+     * combination of them to exclusively finish the span, just guard against double finish.
+     */
+    final AtomicBoolean finishSpan = new AtomicBoolean(true);
```
dougqh, line 42: Why not just push this into the Span?  Then it can be reused by any instrumentation and can provide sanity checking in a strict mode for all instrumentations.

Also, AtomicBoolean is probably overkill.  A volatile field with some synchronization would have a smaller memory footprint.

dd-java-agent/instrumentation/grpc-1.5/src/main/java/datadog/trace/instrumentation/grpc/server/TracingServerInterceptor.java
```diff
@@ -73,37 +85,45 @@ public void close(final Status status, final Metadata trailers) {
...
-      DECORATE.afterStart(span);
-      try (AgentScope scope = activateSpan(span)) {
+      DECORATE.afterStart(msgSpan);
+      try (AgentScope scope = activateSpan(msgSpan)) {
         delegate().onMessage(message);
       } catch (final Throwable e) {
-        DECORATE.onError(span, e);
-        DECORATE.beforeFinish(this.span);
-        this.span.finish();
+        // I'm not convinced we should actually be finishing the span here...
```
dougqh, line 117: How do we answer that question?  Do we need more tests?

dougqh: Yes, why are we touching this.span?  In onMessage, I'd really only expect us to touch the msgSpan.

dd-java-agent/instrumentation/grpc-1.5/src/main/java/datadog/trace/instrumentation/grpc/client/ClientCallImplInstrumentation.java
```diff
@@ -139,7 +139,7 @@ public static void after(@Advice.Enter AgentScope scope) {
     @Advice.OnMethodExit(onThrowable = Throwable.class)
     public static void closeObserver(
         @Advice.This ClientCall<?, ?> call, @Advice.Argument(1) Status status) {
-      AgentSpan span = InstrumentationContext.get(ClientCall.class, AgentSpan.class).get(call);
+      AgentSpan span = InstrumentationContext.get(ClientCall.class, AgentSpan.class).remove(call);
```
dougqh, line 142: I like the idea of adding a remove.  I'm surprised we didn't already have one.

dougqh:
I wonder if we're solving this in the right place.

It seems like we have another situation where instrumentations are violating the contract with the core tracer.
I think we need the core to have a strict mode that that fails when a Span is mutated after finish or is double finished.

This could even be best effort using a mutable unsynchronized field.  That won't guarantee notification on failure but it will catch many errors -- that's how collection mutability checks work.

dd-java-agent/instrumentation/grpc-1.5/src/main/java/datadog/trace/instrumentation/grpc/server/TracingServerInterceptor.java
tylerbenson: This is how it was working before.  I didn't actually change that behavior. More tests probably wouldn't hurt... I just don't know enough about normal grpc operation to determine realistic test scenarios.

dd-java-agent/instrumentation/grpc-1.5/src/main/java/datadog/trace/instrumentation/grpc/client/ClientCallImplInstrumentation.java
bantonsson: I think it's called `put(call, null)`.

dd-java-agent/instrumentation/grpc-1.5/src/main/java/datadog/trace/instrumentation/grpc/server/TracingServerInterceptor.java
bantonsson: That is a really good question. Looking at this a bit more closely, we do have a mechanism to ignore double `finish` in `DDSpan`, but it's too late for this use case. We could probably just expose the `isFinished` from the `DDSpan` and use that to check, or if there is actually a real race, and `check` followed by `change` is not good enough, then change the finish mechanism in the AgentSpan to be something like:

```
if (span.startFinish()) {
  DECORATE.onError(span, e);
  DECORATE.beforeFinish(span);
  span.endFinish();
}
```

If we're still using the duration as the finished marker, then we would miss the time between `finishStart` and `finishEnd`, but maybe that's ok, or we could put a `1 ns` placeholder there and overwrite it in the `finishEnd`.

The normal `finish()` methods could just do call `finishStart` and `finishEnd` until they get changed.

dd-java-agent/instrumentation/grpc-1.5/src/main/java/datadog/trace/instrumentation/grpc/client/ClientCallImplInstrumentation.java
tylerbenson: That doesn't work because put returns void. If you prefer I could make it return the previous value instead.

bantonsson: Sorry I wasn't clear @tylerbenson, I didn't mean that we could use `put(call, null)` in the new way, only that it was the way to _remove_ things before. I think the `remove` addition makes sense.

dd-java-agent/agent-bootstrap/src/main/java/datadog/trace/bootstrap/FieldBackedContextStore.java
dougqh: To clarify, I don't necessarily think it is wrong to have remove on FieldBackedContextStore.
But I think it would be better as a thin wrapper around a thin method on FieldBackedContextAccessor -- which is presumably the case for get and put.

That would provide better encapsulation and potentially allow for more efficient implementation.  While I don't necessarily mind the externally locking here, it does feel a bit inconsistent.

dd-java-agent/instrumentation/grpc-1.5/src/main/java/datadog/trace/instrumentation/grpc/server/TracingServerInterceptor.java
dougqh: On some level, I think it is fine to keep the change set minimal for the bug fix.
However, I'm wondering if we're just doing something wrong here.  Did you look at Richard's grpc reworking PR?

dd-java-agent/agent-bootstrap/src/main/java/datadog/trace/bootstrap/FieldBackedContextStore.java
tylerbenson: set and put are both implemented in bytecode by `FieldBackedContextInjector`, but they don't have the synchronization like putIfAbsent, and which remove also needs. I still think it makes sense to keep remove here because of this.

dd-java-agent/instrumentation/grpc-1.5/src/main/java/datadog/trace/instrumentation/grpc/server/TracingServerInterceptor.java
tylerbenson: [His PR](#2960) was merged and this change is done on top of it. I don't think it had any impact on what this is trying to fix if that's what you mean.

dd-java-agent/agent-bootstrap/src/main/java/datadog/trace/bootstrap/FieldBackedContextStore.java
bantonsson: @dougqh I agree with @tylerbenson that this is the level/place where `remove` should be implemented right now. Neither `get` nor `put` does any synchronization or atomic updates for the injected fields, which `putIfAbsent` and `remove` needs. If we should change the behavior  of `get` and `put` or push down the atomics is a different discussion.

dd-java-agent/instrumentation/grpc-1.5/src/main/java/datadog/trace/instrumentation/grpc/server/TracingServerInterceptor.java
tylerbenson: Ok, I've implemented this suggestion with slightly different naming.  I didn't worry about the added duration between finish and publish.  LMK what you think.

bantonsson (approved):

charliegracie (approved):

PR #2875, opened 2021-07-01
Capture exceptions passed through our Netty handlers

tylerbenson:
This is not likely to be very helpful since most frameworks should handle the exceptions before this point, but we should capture it just in case.

TODO: try to identify some tests that exercises this behavior.

PR #2991, opened 2021-08-05
Ensure netty.client.request spans are finished

charliegracie:
If the client times out or another exception is thrown after
the request is sent the span will not be closed

### DIFF
--- a/dd-java-agent/instrumentation/netty-4.1/src/main/java/datadog/trace/instrumentation/netty41/client/HttpClientResponseTracingHandler.java
+++ b/dd-java-agent/instrumentation/netty-4.1/src/main/java/datadog/trace/instrumentation/netty41/client/HttpClientResponseTracingHandler.java
@@ -42,4 +42,26 @@ public class HttpClientResponseTracingHandler extends ChannelInboundHandlerAdapt
       ctx.fireChannelRead(msg);
     }
   }
+
+  @Override
+  public void exceptionCaught(ChannelHandlerContext ctx, Throwable cause) throws Exception {
+    final Attribute<AgentSpan> parentAttr = ctx.channel().attr(CLIENT_PARENT_ATTRIBUTE_KEY);
+    parentAttr.setIfAbsent(noopSpan());
+    final AgentSpan parent = parentAttr.get();
+    final AgentSpan span = ctx.channel().attr(SPAN_ATTRIBUTE_KEY).getAndSet(parent);
+    if (span != null) {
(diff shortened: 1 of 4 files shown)


devinsba (approved):

tylerbenson:
Similar to my changes on #2875.  I'll close that PR and let you apply them here.

lpriima (approved):

charliegracie:
> Similar to my changes on #2875. I'll close that PR and let you apply them here.

I applied the changes to netty-3.8 and netty-4.0 HttpClientResponseTracingHandler. HttpClientRequestTracingHandler appear to already handle exceptions properly. Currently, only 4.1 is being tested since it is the only version which had a direct NettyReactor test.

Both HttpServerRequestTracingHandler and HttpServerResponseTracingHandler appear to already handle the exceptions properly.

tylerbenson (approved):

dd-java-agent/instrumentation/netty-4.1/src/main/java/datadog/trace/instrumentation/netty41/client/HttpClientResponseTracingHandler.java
```diff
@@ -42,4 +42,26 @@ public void channelRead(final ChannelHandlerContext ctx, final Object msg) {
...
+      // client span won't be finished with a proper response, so we should finish the span here.
+      try (final AgentScope scope = activateSpan(span)) {
+        DECORATE.onError(span, cause);
+        DECORATE.beforeFinish(span);
+        span.finish();
+      }
+    }
+    // We want the callback in the scope of the parent, not the client span
+    try (final AgentScope scope = activateSpan(parent)) {
+      scope.setAsyncPropagation(true);
```
tylerbenson, line 63: I assume this was copy/pasted from somewhere else. I don't think this should be needed anymore.

charliegracie: Yeah I copied it. To be clear it is the `scope.setAsyncPropagation(true);` line you are talking about right?

tylerbenson: yes.  there are several in this PR.  not a big deal to leave them... more of an FYI.

PR #2993, opened 2021-08-06
Limit tomcat latest test to 10.0.+

tylerbenson:
10.1+ seems will require Java 11 which is not compatible with our build currently.


mcculls (approved):
+1

btw, you can use this technique to add a latestDepTest for Java11:

https://github.com/DataDog/dd-trace-java/pull/2929/files#diff-7838da4953eebbfc448dc465a78444ff86106fe0ecd96bff1d8ba2ec21db93c1R79

(note the circleCI config still needs updating to run these on CI)

lpriima (approved):

bantonsson (approved):
There were some more places that pulled in the Java 11 version of Tomcat

tylerbenson (commented):
Thanks for the additional changes @bantonsson.

PR #2995, opened 2021-08-09
Checkpoint Sampling

jbachorik:
WIP


dd-trace-core/jfr-openjdk/src/main/java/datadog/trace/core/jfr/openjdk/CheckpointEvent.java
```diff
@@ -26,17 +23,9 @@
   @Label("Flags")
   private final int flags;
 
-  @Label("Thread CPU Time")
-  private final long cpuTime;
```
mar-kolya, line 30: Is this the long term intention? Are we no longer planning to ship this data ever? :)

internal-api/src/main/java/datadog/trace/api/Checkpointer.java
```diff
@@ -25,11 +27,10 @@
...
-   * @param traceId the traceId of the trace the span belongs to.
-   * @param spanId the span's identifier
+   * @param span the span
    * @param flags a description of the event
    */
-  void checkpoint(DDId traceId, DDId spanId, int flags);
+  void checkpoint(AgentSpan span, int flags);
 
-  void onRootSpanPublished(String route, DDId traceId);
+  void onRootSpan(String route, DDId traceId, boolean published);
```
mar-kolya, line 35: Maybe worth documenting this method

internal-api/src/jmh/java/datadog/trace/bootstrap/instrumentation/api/URIUtilsDecode.java
```diff
@@ -14,6 +14,7 @@
 
   @Benchmark
   public void decode(Blackhole blackhole) {
+    // remove me
```
mar-kolya, line 17: Is this intentional?

internal-api/internal-api-8/src/test/java/datadog/trace/api/sampling/AdaptiveSampler7Test.java
```diff
@@ -0,0 +1,522 @@
...
+ * Test various hand crafted scenarios of events coming in different patterns. Test both, the
+ * isolated single threaded execution as well as events arriving on concurrent threads.
+ *
+ * <p>The test supports 'benchmark' mode to explore the reliability boundaries where all test cases
+ * can be run multiple times - the number of iteration is passed in in {@literal
+ * com.datadog.profiling.exceptions.test-iterations} system property.
+ */
+@ExtendWith(MockitoExtension.class)
+class AdaptiveSampler7Test {
+  // delete me
```
mar-kolya, line 45: Should this be acted upon? :)

dd-trace-core/jfr-openjdk/src/main/java/datadog/trace/core/jfr/openjdk/EndpointEvent.java
```diff
@@ -18,8 +18,12 @@
   @Label("Trace Id")
   private final long traceId;
 
-  public EndpointEvent(final String endpoint, final long traceId) {
+  @Label("Kept")
```
mar-kolya, line 21: It would be nice to document this field since meaning here may be unclear

Also, above the `endpoint` field - could you please change `@Label` there to `endpoint`?

dd-trace-core/jfr-openjdk/src/main/java/datadog/trace/core/jfr/openjdk/CheckpointEvent.java
mar-kolya: If that's the case - should we also drop the `CPU` flag?

jbachorik: IMO, this is the long term intention. 
The price of obtaining that information is prohibitively high especially in async based applications running on a cgroups container (which would be the majority of deployments nowadays, I guess).

I would prefer keeping the flag for now - it provides differentiation between the span and task level events.
I mean, we can rename the constant but conceptually the flag is still valid.

internal-api/src/jmh/java/datadog/trace/bootstrap/instrumentation/api/URIUtilsDecode.java
jbachorik: Definitely not. 
Removing the comment - it seems out of place.

internal-api/internal-api-8/src/test/java/datadog/trace/api/sampling/AdaptiveSampler7Test.java
jbachorik: Yeah, already doing that :)

internal-api/src/main/java/datadog/trace/api/Checkpointer.java
jbachorik: I can take a stab but it would just a reflection of the current state - I am not fully aware of the original intentions.

dd-trace-core/jfr-openjdk/src/main/java/datadog/trace/core/jfr/openjdk/EndpointEvent.java
jbachorik: Would the label change affect the backend processing?
If there is a slight chance of messing that up I would rather defer the rename to a separate PR.
If there is no interconnection with the backend I can do the rename.

jbachorik: Regarding the `kept` field - do you think something like `sampled` would convey the purpose better?

internal-api/src/main/java/datadog/trace/api/Checkpointer.java
mar-kolya: 'Current state' is all we can hope for :)

dd-trace-core/jfr-openjdk/src/main/java/datadog/trace/core/jfr/openjdk/EndpointEvent.java
mar-kolya: > If there is a slight chance of messing that up
I think this is just a human readable thing... @hdafgard would correct me if I'm wrong.

mar-kolya: > Regarding the kept field - do you think something like sampled would convey the purpose better?

I do not have a strong opinion on this

jbachorik: Ok

internal-api/src/main/java/datadog/trace/api/Checkpointer.java
jbachorik: Done

dd-trace-core/jfr-openjdk/src/main/java/datadog/trace/core/jfr/openjdk/CheckpointEvent.java
jbachorik: Since it is not obvious whether the flag should be kept or removed I am proposing to keep it for now and do the removal as a separate PR if we come to the conclusion that it is indeed the thing we want to do.

dd-trace-core/jfr-openjdk/src/main/java/datadog/trace/core/jfr/openjdk/CheckpointSummaryEvent.java
```diff
@@ -0,0 +1,32 @@
...
+@StackTrace(false)
+public class CheckpointSummaryEvent extends Event {
+  @Label("Dropped")
+  private final long dropped;
+
+  @Label("Emitted")
+  private final long emitted;
+
+  @Label("Rate Limit")
+  private final int rate;
```
mar-kolya, line 25: should this be called `rateLimit` instead?

dd-trace-core/jfr-openjdk/src/main/java/datadog/trace/core/jfr/openjdk/EndpointEvent.java
mar-kolya: Actually... maybe you are right - `kept` sounds a bit... strange - `sampled` would be more conventional

dd-trace-core/src/main/java/datadog/trace/core/CoreTracer.java
```diff
@@ -656,18 +657,19 @@ void write(final List<DDSpan> trace) {
 
       DDSpan spanToSample = rootSpan == null ? writtenTrace.get(0) : rootSpan;
       spanToSample.forceKeep(forceKeep);
-      if (forceKeep || sampler.sample(spanToSample)) {
+      boolean published = forceKeep || sampler.sample(spanToSample);
+      if (published) {
```
mar-kolya, line 661: this seems unnecessary

dd-trace-core/src/main/java/datadog/trace/core/DDSpan.java
```diff
@@ -169,6 +172,42 @@ public boolean isForceKeep() {
...
+  @Override
+  public Boolean isEmittingCheckpoints() {
+    /*
+    The decision to emit checkpoints is made at the local root span level.
+    This is to ensure consistency in the emitted checkpoints where the whole
+    local root span subtree must either be fully covered or no checkpoints should
+    be emitted at all.
+     */
+    AgentSpan rootSpan = getLocalRootSpan();
+    return (rootSpan != null && this != rootSpan)
```
mar-kolya, line 206: Can you do 

```
// rootSpan will always be an instance of DDSpan
    DDSpan span = (DDSpan) rootSpan;
```

like above and read the field directly instead of recursive getter call?

internal-api/internal-api-8/src/main/java/datadog/trace/api/sampling/AdaptiveSampler.java
```diff
@@ -214,4 +241,13 @@ public void run(final AdaptiveSampler target) {
...
   }
+
+  // access for tests
+  long testCount() {
+    return countsRef.get().testCount();
+  }
+
+  long sampleCount() {
+    return countsRef.get().sampleCount();
+  }
```
mar-kolya, line 252: Wouldn't you be able to test this class by its behaviour without exposing counters?

dd-trace-core/jfr-openjdk/src/main/java/datadog/trace/core/jfr/openjdk/JFRCheckpointer.java
```diff
@@ -3,33 +3,145 @@
...
+import org.slf4j.LoggerFactory;
 
-public final class JFRCheckpointer implements Checkpointer {
+public class JFRCheckpointer implements Checkpointer {
+  private static final Logger log = LoggerFactory.getLogger(JFRCheckpointer.class);
 
-  private static final boolean RECORD_CPU_TIME =
-      ConfigProvider.createDefault()
-          .getBoolean(ProfilingConfig.PROFILING_CHECKPOINTS_RECORD_CPU_TIME, false);
+  private static final int MASK = ~CPU;
```
mar-kolya, line 20: TBH the need for the `CPU` flag is less and less clear now - given how it is being used :)

mar-kolya (approved):
Some minor comments, otherwise seems reasonable

internal-api/internal-api-8/src/main/java/datadog/trace/api/sampling/AdaptiveSampler.java
jbachorik: That is a bit problematic. For that I would need to expose the `Counts` inner class and make it possible to inject the actual implementation into `AdaptiveSampler`. So, instead of 2 read-only package private method I would end up with  package private `Counts` class and a special constructor allowing to set up the `Counts` instance so I can mock it.

dd-trace-core/jfr-openjdk/src/main/java/datadog/trace/core/jfr/openjdk/JFRCheckpointer.java
jbachorik: Yeah. I kind of agree. But I really would prefer to tackle that in a separate work unit :)

dd-trace-core/jfr-openjdk/src/main/java/datadog/trace/core/jfr/openjdk/CheckpointSummaryEvent.java
jbachorik: Yeah, sounds right

dd-trace-core/src/main/java/datadog/trace/core/CoreTracer.java
jbachorik: Ah, some left overs from debugging times ...
Will revert.

jbachorik: Actually, no. It should stay as it is.
`published` is used on L671 to set the flag of the endpoint event.

mar-kolya: ah, missed that

dd-trace-core/src/main/java/datadog/trace/core/DDSpan.java
jbachorik: Ok, I will need more explanation here.

I checked `getLocalRootSpan()` and it defers to `context.getTrace().getRootSpan()`- I don't see any field where I could get that information directly.
Or is this not what you meant?

PR #2998, opened 2021-08-10
Ensure all "client" and a few "server" spans are marked as "measured"

tylerbenson:
Rather than doing this for each client integration, I moved the logic to the `ClientDecorator`.  I removed the duplicate asserts from the relevant instrumentation. I also added asserts for it to a bunch of tests to ensure I didn't miss something.

While doing this, I also discovered a few server spans (rmi and grpc) that weren't measured that probably should be. I included those changes here too.

### DIFF
--- a/dd-java-agent/agent-bootstrap/src/main/java/datadog/trace/bootstrap/instrumentation/rmi/RmiServerDecorator.java
+++ b/dd-java-agent/agent-bootstrap/src/main/java/datadog/trace/bootstrap/instrumentation/rmi/RmiServerDecorator.java
@@ -1,5 +1,6 @@
 package datadog.trace.bootstrap.instrumentation.rmi;
 
+import datadog.trace.bootstrap.instrumentation.api.AgentSpan;
 import datadog.trace.bootstrap.instrumentation.api.InternalSpanTypes;
 import datadog.trace.bootstrap.instrumentation.api.UTF8BytesString;
 import datadog.trace.bootstrap.instrumentation.decorator.ServerDecorator;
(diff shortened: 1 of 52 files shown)


dd-java-agent/agent-bootstrap/src/main/java/datadog/trace/bootstrap/instrumentation/rmi/RmiServerDecorator.java
```diff
@@ -23,4 +24,10 @@ protected CharSequence spanType() {
   protected CharSequence component() {
     return RMI_SERVER;
   }
+
+  @Override
+  public AgentSpan afterStart(final AgentSpan span) {
+    span.setMeasured(true);
```
bantonsson, line 30: This seems unrelated to the other _client_ changes since it's in `RmiServerDecorator`.

dd-java-agent/instrumentation/grpc-1.5/src/main/java/datadog/trace/instrumentation/grpc/server/GrpcServerDecorator.java
```diff
@@ -59,6 +59,12 @@ protected CharSequence component() {
     return COMPONENT_NAME;
   }
 
+  @Override
+  public AgentSpan afterStart(final AgentSpan span) {
+    span.setMeasured(true);
```
bantonsson, line 64: Same thing here with changes to a server decorator.

bantonsson (approved):
👍 with only minor questions about why the two server decorators were changed.

dd-java-agent/agent-bootstrap/src/main/java/datadog/trace/bootstrap/instrumentation/rmi/RmiServerDecorator.java
tylerbenson: Oops, you're right.  I forgot intermingled a couple of server changes here.  Would you prefer I update the title/description or separate to a different PR?

charliegracie (approved):

PR #3000, opened 2021-08-11
Unmask some tests that pass the checkpoint sequence validator

luhenry:


devinsba (approved):
You requested review but it's in draft so not sure if you really wanted review yet 🤷

dd-java-agent/instrumentation/scala-promise/scala-promise-2.10/src/forkedTest/groovy/Scala210PromiseForkedTests.groovy
```diff
@@ -53,10 +50,9 @@ class Scala210PromiseCompletionPriorityThreadPoolForkedTest extends ScalaPromise
     return ExecutionContext$.MODULE$.fromExecutorService(Executors.newCachedThreadPool())
   }
 
-  @Override
   def setup() {
     CheckpointValidator.excludeValidations_DONOTUSE_I_REPEAT_DO_NOT_USE(
-      CheckpointValidationMode.SUSPEND_RESUME,
+      CheckpointValidationMode.INTERVALS,
```
jbachorik, line 55: Why is the mode changed here? Did it fail on `INTERVALS` check?

luhenry: I made sure to have the same failure mode covered for all `scala-promise-2.*` tests. There is quite some flakiness on these so I'll check with CI if we should keep both or just `INTERVALS`.

dd-java-agent/testing/src/main/groovy/datadog/trace/agent/test/AgentTestRunner.groovy
```diff
@@ -166,10 +166,13 @@ abstract class AgentTestRunner extends DDSpecification implements AgentBuilder.L
...
-    TEST_CHECKPOINTER.clear()
-    new MockUtil().detachMock(STATS_D_CLIENT)
-    new MockUtil().detachMock(TEST_CHECKPOINTER)
+    try {
+      TEST_CHECKPOINTER.publish()
+    } finally {
+      TEST_CHECKPOINTER.clear()
+      new MockUtil().detachMock(STATS_D_CLIENT)
+      new MockUtil().detachMock(TEST_CHECKPOINTER)
+    }
```
jbachorik, line 175: 💯

PR #3001, opened 2021-08-12
API design for suppressing spans from being created for a scope.

tylerbenson:
Actual implementation of span suppression to follow in separate PR.

### DIFF
--- a/dd-trace-core/src/main/java/datadog/trace/core/CoreTracer.java
+++ b/dd-trace-core/src/main/java/datadog/trace/core/CoreTracer.java
@@ -142,7 +142,7 @@ public class CoreTracer implements AgentTracer.TracerAPI {
 
   private final InstrumentationGateway instrumentationGateway;
 
-  private final ThreadLocal<Set<CharSequence>> spanSuppression =
+  private static final ThreadLocal<Set<CharSequence>> SPAN_SUPPRESSION =
       new ThreadLocal<Set<CharSequence>>() {
         @Override
         protected Set<CharSequence> initialValue() {
@@ -601,12 +601,12 @@ public class CoreTracer implements AgentTracer.TracerAPI {
 
   @Override
   public SuppressScope suppress(final CharSequence spanName) {
-    boolean added = spanSuppression.get().add(spanName);
+    boolean added = SPAN_SUPPRESSION.get().add(spanName);
     if (added) {
       return new SuppressScope() {
         @Override
         public void close() {
-          spanSuppression.get().remove(spanName);
+          SPAN_SUPPRESSION.get().remove(spanName);
         }
       };
     }
--- a/dd-trace-core/src/main/java/datadog/trace/core/CoreTracer.java
+++ b/dd-trace-core/src/main/java/datadog/trace/core/CoreTracer.java
@@ -28,6 +28,7 @@ import datadog.trace.bootstrap.instrumentation.api.AgentScopeManager;
 import datadog.trace.bootstrap.instrumentation.api.AgentSpan;
 import datadog.trace.bootstrap.instrumentation.api.AgentTracer;
 import datadog.trace.bootstrap.instrumentation.api.ScopeSource;
+import datadog.trace.bootstrap.instrumentation.api.SuppressScope;
 import datadog.trace.common.metrics.MetricsAggregator;
 import datadog.trace.common.sampling.PrioritySampler;
 import datadog.trace.common.sampling.Sampler;
@@ -49,12 +50,14 @@ import java.math.BigInteger;
 import java.util.ArrayList;
 import java.util.Collection;
 import java.util.Comparator;
+import java.util.HashSet;
 import java.util.LinkedHashMap;
 import java.util.List;
 import java.util.Map;
 import java.util.Properties;
 import java.util.ServiceConfigurationError;
 import java.util.ServiceLoader;
+import java.util.Set;
 import java.util.SortedSet;
 import java.util.concurrent.ConcurrentSkipListSet;
 import java.util.concurrent.TimeUnit;
@@ -138,6 +141,14 @@ public class CoreTracer implements AgentTracer.TracerAPI {
   private final HttpCodec.Extractor extractor;
 
   private final InstrumentationGateway instrumentationGateway;
+
+  private final ThreadLocal<Set<CharSequence>> spanSuppression =
+      new ThreadLocal<Set<CharSequence>>() {
+        @Override
+        protected Set<CharSequence> initialValue() {
+          return new HashSet<>();
+        }
+      };
 
   @Override
   public TraceScope.Continuation capture() {
@@ -586,6 +597,22 @@ public class CoreTracer implements AgentTracer.TracerAPI {
   @Override
   public AgentPropagation propagate() {
     return this;
+  }
+
+  @Override
+  public SuppressScope suppress(final CharSequence spanName) {
+    boolean added = spanSuppression.get().add(spanName);
+    if (added) {
+      return new SuppressScope() {
+        @Override
+        public void close() {
+          spanSuppression.get().remove(spanName);
+        }
+      };
+    }
+    // this spanName is already being suppressed.
+    // rely on the previous invocation to close suppression.
+    return AgentTracer.NoopSuppressScope.INSTANCE;
   }
 
   @Override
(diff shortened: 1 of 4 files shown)


dd-trace-core/src/main/java/datadog/trace/core/CoreTracer.java
```diff
@@ -588,6 +599,22 @@ public AgentPropagation propagate() {
     return this;
   }
 
+  @Override
+  public SuppressScope suppress(final CharSequence spanName) {
```
jbachorik, line 603: Would the use case of suppressing multiple spans in one `SupressScope` be valid?

tylerbenson: Are you suggesting maybe use a var-arg instead/in addition to?

internal-api/src/main/java/datadog/trace/bootstrap/instrumentation/api/AgentTracer.java
```diff
@@ -112,6 +116,8 @@ private AgentTracer() {}
 
     AgentPropagation propagate();
 
+    SuppressScope suppress(CharSequence spanName);
```
dougqh, line 119: I don't particularly want to create another object for this.  I think we actually should get away from creating objects for Scopes as well.

dd-trace-core/src/main/java/datadog/trace/core/CoreTracer.java
```diff
@@ -588,6 +599,22 @@ public AgentPropagation propagate() {
...
+    boolean added = SPAN_SUPPRESSION.get().add(spanName);
```
dougqh, line 604: I don't believe CharSequence supports a meaningful hashCode.  To put it in a HashSet, you probably have to String.valueOf.

internal-api/src/main/java/datadog/trace/bootstrap/instrumentation/api/AgentTracer.java
tylerbenson: This allows for the try-with-resources style call and simplified cleanup.  You would rather have it return a boolean?

dd-trace-core/src/main/java/datadog/trace/core/CoreTracer.java
jbachorik: No suggestion - just figuring out the use case.

dd-trace-core/src/main/java/datadog/trace/core/CoreTracer.java
```diff
@@ -139,6 +142,14 @@ public int compare(final TraceInterceptor o1, final TraceInterceptor o2) {
 
   private final InstrumentationGateway instrumentationGateway;
 
+  private static final ThreadLocal<Set<CharSequence>> SPAN_SUPPRESSION =
```
dougqh, line 145: I'm not sure Tracer is the right place for this.

Given that is contextual, I think this should be placed closer to the other scope management.

dd-trace-core/src/main/java/datadog/trace/core/CoreTracer.java
```diff
@@ -588,6 +599,22 @@ public AgentPropagation propagate() {
     return this;
   }
 
+  @Override
+  public SuppressScope suppress(final CharSequence spanName) {
+    boolean added = SPAN_SUPPRESSION.get().add(spanName);
+    if (added) {
+      return new SuppressScope() {
```
dougqh, line 606: I think it would be fine to just let the caller do unsuppress(name).  We don't need an extra object for this.

dd-trace-core/src/test/groovy/datadog/trace/core/CoreTracerTest.groovy
```diff
@@ -375,6 +376,40 @@ class CoreTracerTest extends DDCoreSpecification {
...
+    scope = tracer.suppress("suppressedSpan")
+
+    then: "valid scope"
+    !(scope instanceof AgentTracer.NoopSuppressScope)
+
+    and: "duplicate calls are not no-ops"
+    tracer.suppress("suppressedSpan") instanceof AgentTracer.NoopSuppressScope
+
+    when:
+    def span = tracer.buildSpan("suppressedSpan").start()
```
dougqh, line 402: We'll need a lot more testing here.

We need to show how suppression interacts with...
- activation
- child span creation
- baggage

internal-api/src/main/java/datadog/trace/bootstrap/instrumentation/api/AgentTracer.java
dougqh: Yes, I know it supports try-with-resources clean-up, but that adds significant overhead that we don't need.  We also don't need a boolean.

All we need is...
TraceContext.suppress(operationName);
try {
  ...
} finally {
  TracerContext.unsuppress(operationName);
}

dd-trace-core/src/main/java/datadog/trace/core/CoreTracer.java
```diff
@@ -139,6 +142,14 @@ public int compare(final TraceInterceptor o1, final TraceInterceptor o2) {
 
   private final InstrumentationGateway instrumentationGateway;
 
+  private static final ThreadLocal<Set<CharSequence>> SPAN_SUPPRESSION =
+      new ThreadLocal<Set<CharSequence>>() {
+        @Override
+        protected Set<CharSequence> initialValue() {
+          return new HashSet<>();
```
dougqh, line 149: This is probably oversized for this use case.  I honestly wonder if we should use a flat sized set until we reach some limit.

dd-trace-core/src/main/java/datadog/trace/core/CoreTracer.java
dougqh: There are a couple use cases, we're playing with addressing.

In one situation, we get _extra_ spans because of iterating over message objects multiple times.
This can happen Kafka Streams for complex queries.
This can happen with Spring when iterates over the collection once before then handing it to the Kafka listener.

In those situations, we want to suppress the spans from those initial iterations for the complex cases.  But in the simpler cases, where the object is iterated just once we want the spans.

I think we'll have to hook this mechanism up to one of those cases before we know if this is really the right solution.

The other case that we're thinking about is nested server spans or nested client spans.  The client case is easier to understand, so I'll use it as the example.

If you consider, jms on top of sqs on top of http.  The end user typically only needs to see layer with which they are interacting.  In this example, jms.

dougqh (commented):
I'm quite convinced we need span suppression, but I less convinced of the driving mechanism.  And I think we'll only know that when we try in the context of one of our problems.

I think we might be more successful separating out the span portion -- which I think should be less controversial.

My thinking is that we just add a suppress() method onto SpanBuilder and then work through the details of activation, child creation, baggage, and context propagation.

dd-trace-core/src/main/java/datadog/trace/core/CoreTracer.java
tylerbenson: Our implementation of `UTF8BytesString` does.  Since this is intended as an internal API, perhaps I should use that type specifically instead?

dd-trace-core/src/main/java/datadog/trace/core/CoreTracer.java
tylerbenson: done.

dd-trace-core/src/main/java/datadog/trace/core/CoreTracer.java
tylerbenson: I set it to size 4 for now to account for the load factor.

internal-api/src/main/java/datadog/trace/bootstrap/instrumentation/api/AgentTracer.java
tylerbenson: done.

dd-trace-core/src/main/java/datadog/trace/core/CoreTracer.java
tylerbenson: I switched to `UTF8BytesString` specifically. We can relax that in the future if needed.

dd-trace-core/src/main/java/datadog/trace/core/CoreTracer.java
tylerbenson: I put it in here because it will also need to be referenced by the SpanBuilder for the suppression.

dougqh (changes requested):
This looks more efficient than the previous version, but I still think we need a demonstration of usefulness before we merge this.

tylerbenson:
Closing for now as I think we have a better work around.

PR #3007, opened 2021-08-16
Suppress extra span from Kafka Streams internal queuing

tylerbenson:
Internally `StreamTask` iterates through each `ConsumerRecord` and adds them to the queue.  This iteration triggers an extra span from the kafka instrumentation.  While evaluating implementation of #3001, I realized we could achieve similar behavior by unwrapping the Iterable.  This relies on specific behavior between the kafka and kafka streams instrumentation, but accomplishes the removal of the extra span.

### DIFF
--- a/dd-java-agent/instrumentation/kafka-streams-0.11/src/latestDepTest/groovy/KafkaStreamsTest.groovy
+++ b/dd-java-agent/instrumentation/kafka-streams-0.11/src/latestDepTest/groovy/KafkaStreamsTest.groovy
@@ -58,7 +58,7 @@ class KafkaStreamsTest extends AgentTestRunner {
         void onMessage(ConsumerRecord<String, String> record) {
           // ensure consistent ordering of traces
           // this is the last processing step so we should see 2 traces here
-          TEST_WRITER.waitForTraces(3)
+          TEST_WRITER.waitForTraces(2)
           TEST_TRACER.activeSpan().setTag("testing", 123)
           records.add(record)
         }
@@ -77,7 +77,7 @@ class KafkaStreamsTest extends AgentTestRunner {
       .mapValues(new ValueMapper<String, String>() {
         @Override
         String apply(String textLine) {
-          TEST_WRITER.waitForTraces(2) // ensure consistent ordering of traces
+          TEST_WRITER.waitForTraces(1) // ensure consistent ordering of traces
           TEST_TRACER.activeSpan().setTag("asdf", "testing")
           return textLine.toLowerCase()
         }
@@ -102,7 +102,7 @@ class KafkaStreamsTest extends AgentTestRunner {
     received.value() == greeting.toLowerCase()
     received.key() == null
 
-    assertTraces(4) {
+    assertTraces(3) {
       trace(1) {
         // PRODUCER span 0
         span {
@@ -111,30 +111,12 @@ class KafkaStreamsTest extends AgentTestRunner {
           resourceName "Produce Topic $STREAM_PENDING"
           spanType "queue"
           errored false
+          measured true
           parent()
           tags {
             "$Tags.COMPONENT" "java-kafka"
             "$Tags.SPAN_KIND" Tags.SPAN_KIND_PRODUCER
             defaultTags()
-          }
-        }
-      }
-      trace(1) {
-        // CONSUMER span 0
-        span {
-          serviceName "kafka"
-          operationName "kafka.consume"
-          resourceName "Consume Topic $STREAM_PENDING"
-          spanType "queue"
-          errored false
-          childOf trace(0)[0]
-          tags {
-            "$Tags.COMPONENT" "java-kafka"
-            "$Tags.SPAN_KIND" Tags.SPAN_KIND_CONSUMER
-            "$InstrumentationTags.PARTITION" { it >= 0 }
-            "$InstrumentationTags.OFFSET" 0
-            "$InstrumentationTags.RECORD_QUEUE_TIME_MS" { it >= 0 }
-            defaultTags(true)
           }
         }
       }
@@ -148,6 +130,7 @@ class KafkaStreamsTest extends AgentTestRunner {
           resourceName "Consume Topic $STREAM_PENDING"
           spanType "queue"
           errored false
+          measured true
           childOf trace(0)[0]
 
           tags {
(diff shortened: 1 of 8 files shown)


dd-java-agent/instrumentation/kafka-streams-0.11/src/latestDepTest/groovy/KafkaStreamsTest.groovy
```diff
@@ -119,25 +120,6 @@ class KafkaStreamsTest extends AgentTestRunner {
           }
         }
       }
-      trace(1) {
-        // CONSUMER span 0
-        span {
-          serviceName "kafka"
-          operationName "kafka.consume"
-          resourceName "Consume Topic $STREAM_PENDING"
```
tylerbenson, line 127: You can see the extra span being removed here in the test.

dd-java-agent/instrumentation/kafka-streams-0.11/src/main/java/datadog/trace/instrumentation/kafka_streams/KafkaStreamTaskInstrumentation.java
```diff
@@ -0,0 +1,61 @@
...
+    @Advice.OnMethodEnter(suppress = Throwable.class)
+    public static void onEnter(
+        @Advice.Argument(value = 1, readOnly = false) Iterable<ConsumerRecord<?, ?>> records) {
+      // This method adds the records to a queue, so we want to bypass the kafka instrumentation
+      // since the resulting spans are very short and uninteresting.
+      // KafkaStreamsProcessorInstrumentation will create a new span instead.
+
+      // Expecting a TracingList because TaskManager.addRecordsToTasks calls records(partition).
+      if (records instanceof TracingList) {
+        records = ((TracingList) records).getDelegate();
```
tylerbenson, line 57: Here's the unwrapping.

dougqh: Couldn't we just skip wrapping in the first place?

I guess that has the disadvantage of requiring additional state somewhere that the TracingList can access.

The downside of this approach is that it assumes that the incoming records object is a TracingList.  It doesn't work if addRecords is passed something from partitions or another Iterable wrapped around the TracingList.

tylerbenson: To clarify, you mean adding unwrapping support for each of `{TracingIterable, TracingIterator, TracingList, TracingListIterator}`?  I don't have a problem with that.

tylerbenson: I added unwrapping to both `TracingIterable` and `TracingList` since the method signature requires a `Iterable`.  I don't think the corresponding `Iterator` classes need to be handled.

dd-java-agent/instrumentation/kafka-streams-0.11/src/main/java/datadog/trace/instrumentation/kafka_streams/KafkaStreamTaskInstrumentation.java
```diff
@@ -0,0 +1,62 @@
...
+  public ElementMatcher<TypeDescription> typeMatcher() {
+    return named("org.apache.kafka.streams.processor.internals.StreamTask");
+  }
+
+  @Override
+  public String[] helperClassNames() {
+    return new String[] {
+      "datadog.trace.instrumentation.kafka_clients.KafkaDecorator",
+      "datadog.trace.instrumentation.kafka_clients.TextMapExtractAdapter",
+      "datadog.trace.instrumentation.kafka_clients.TracingIterableDelegator",
```
bantonsson, line 32: Wouldn't it be enough to just have `TracingIterableDelegator` as a helper class in this instrumentation?

tylerbenson: Yep, good call.

bantonsson (approved):

dd-java-agent/instrumentation/kafka-streams-0.11/src/latestDepTest/groovy/KafkaStreamsTest.groovy
```diff
@@ -58,7 +58,7 @@ class KafkaStreamsTest extends AgentTestRunner {
         void onMessage(ConsumerRecord<String, String> record) {
           // ensure consistent ordering of traces
           // this is the last processing step so we should see 2 traces here
-          TEST_WRITER.waitForTraces(3)
```
dougqh, line 61: So previously, we knowingly had double sets of spans?

dd-java-agent/instrumentation/kafka-streams-0.11/src/latestDepTest/groovy/KafkaStreamsTest.groovy
dougqh: I'm surprised that we knowingly had extra spans previously.

That reinforces my general belief that our current assertions are really just inefficient copy & paste -- and not really true verification.

dd-java-agent/instrumentation/kafka-streams-0.11/src/latestDepTest/groovy/KafkaStreamsTest.groovy
```diff
@@ -102,7 +102,7 @@ class KafkaStreamsTest extends AgentTestRunner {
     received.value() == greeting.toLowerCase()
```
dougqh, line 102: So we're still only testing the simple test cases?  No grouping?

Since we're trying to fix bugs, I think I'm _okay_ with that because this is a meaningful step forward.
However, I do want to see much comprehensive tests added almost immediately in another PR.

devinsba (approved):

devinsba:
Given the 2 solutions, I think I personally would lean toward not wrapping at all in the streams context but I think it probably doesn't matter. Though if one seems easier to maintain across versions I think I would lean toward that

PR #3009, opened 2021-08-17
Better support for reactor netty 1.0.x connect handling

bantonsson:


dd-java-agent/instrumentation/reactor-netty-1/src/test/groovy/ReactorNettyHttpClientTest.groovy
```diff
@@ -0,0 +1,85 @@
...
+        Mono<Integer> apply(HttpClientResponse httpClientResponse, ByteBufMono byteBufMono) {
+          Mono.just(httpClientResponse.status().code())
+        }
+      })
+      .doOnSuccess({ if (callback) { callback.run() } })
+  }
+
+  @Override
+  int doRequest(String method, URI uri, Map<String, String> headers, String body, Closure callback) {
+    createRequest(method, uri, headers, body, callback).block()
```
tylerbenson, line 40: What's the motivation for moving this to another method? seems like it could be inlined.

tylerbenson (approved):

charliegracie (approved):

dd-java-agent/instrumentation/reactor-netty-1/src/test/groovy/ReactorNettyHttpClientTest.groovy
bantonsson: Thanks. That is just residue after my different attempts at coming up with test cases before I found the resolve address one. I'll clean it up.

PR #3010, opened 2021-08-17
[AppSec] Added smoke test to validate WAF

ValentinZakharov:
AppSec smoke test implemented based on SpringBoot smoke test.
Added AppSec report simulator to track if the malicious request reported correctly.
Thу test is a sequence of 200 malicious requests. Each request will be detected by AppSec and should be sent 200 events to the backend with corresponding information about the attack. To validate AppSec we are checking events reported by AppSec.

Since the event dispatch mechanism has a timeout (minimum and maximum time between dispatches), I added the `appsec.report.timeout` configuration option to force the maximum interval between dispatches. Thus, for tests, we set the value `-Dappsec.report.timeout=0`, which allows perform tests fast as possible (event commitment happens almost instantly).

To use the additional option, I had to modify `Config.java`, but after switching to AppSec Span Interface, I need to get rid of this option.

### DIFF
--- a/dd-java-agent/appsec/src/main/java/com/datadog/appsec/report/ReportStrategy.java
+++ b/dd-java-agent/appsec/src/main/java/com/datadog/appsec/report/ReportStrategy.java
@@ -2,6 +2,8 @@ package com.datadog.appsec.report;
 
 import com.datadog.appsec.report.raw.events.AppSecEvent100;
 import com.datadog.appsec.util.JvmTime;
+import datadog.trace.api.Config;
+import java.util.concurrent.TimeUnit;
 import javax.annotation.Nonnull;
 
 public interface ReportStrategy {
@@ -10,8 +12,10 @@ public interface ReportStrategy {
   boolean shouldFlush(@Nonnull AppSecEvent100 event);
 
   class Default implements ReportStrategy {
-    private static final long MIN_INTERVAL_NANOS = 5L * 1000L * 1000L * 1000L;
-    private static final long MAX_INTERVAL_NANOS = 60L * 1000L * 1000L * 1000L;
+    private static final long MIN_INTERVAL_NANOS =
+        TimeUnit.SECONDS.toNanos(Config.get().getAppSecReportMinTimeout());
+    private static final long MAX_INTERVAL_NANOS =
+        TimeUnit.SECONDS.toNanos(Config.get().getAppSecReportMaxTimeout());
     private static final int MAX_QUEUED_ITEMS = 50;
 
     private final JvmTime jvmTime;
--- a/dd-smoke-tests/appsec/springboot/springboot.gradle
+++ b/dd-smoke-tests/appsec/springboot/springboot.gradle
@@ -1,0 +1,30 @@
+plugins {
+  id "com.github.johnrengelman.shadow"
+}
+
+ext {
+  minJavaVersionForTests = JavaVersion.VERSION_1_8
+}
+
+apply from: "$rootDir/gradle/java.gradle"
+description = 'SpringBoot Smoke Tests.'
+
+// The standard spring-boot plugin doesn't play nice with our project
+// so we'll build a fat jar instead
+jar {
+  manifest {
+    attributes('Main-Class': 'datadog.smoketest.appsec.springboot.SpringbootApplication')
+  }
+}
+
+dependencies {
+  implementation group: 'org.springframework.boot', name: 'spring-boot-starter-web', version: '2.6.0'
+
+  testImplementation project(':dd-smoke-tests:appsec')
+}
+
+tasks.withType(Test).configureEach {
+  dependsOn "shadowJar"
+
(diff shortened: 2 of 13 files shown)


dd-smoke-tests/appsec/springboot/springboot.gradle
```diff
@@ -0,0 +1,25 @@
...
+// The standard spring-boot plugin doesn't play nice with our project
+// so we'll build a fat jar instead
+jar {
+  manifest {
+    attributes('Main-Class': 'datadog.smoketest.appsec.springboot.SpringbootApplication')
+  }
+}
+
+dependencies {
+  implementation group: 'org.springframework.boot', name: 'spring-boot-starter-web', version: '1.5.18.RELEASE'
```
devinsba, line 16: This is a really old version of spring-boot, can we at least bump up into the 2.x line of releases? I don't have data on what versions our customers use but a lot of the customer issues recently have at least been 2.x

devinsba (commented):
I'm interested in opinions from others on the spring-boot version.

As far as a smoke test goes, at some point it would probably be good to have an app that includes spring-security to ensure there aren't interoperability issues

dd-smoke-tests/appsec/springboot/springboot.gradle
bantonsson: Yes, everything below `2.3.x` is considered EOL by Spring.

dd-java-agent/appsec/src/main/java/com/datadog/appsec/report/ReportStrategy.java
```diff
@@ -10,8 +11,10 @@
   boolean shouldFlush(@Nonnull AppSecEvent100 event);
 
   class Default implements ReportStrategy {
-    private static final long MIN_INTERVAL_NANOS = 5L * 1000L * 1000L * 1000L;
-    private static final long MAX_INTERVAL_NANOS = 60L * 1000L * 1000L * 1000L;
+    private static final long MIN_INTERVAL_NANOS =
+        Config.get().getAppSecReportMinTimeout() * 1000L * 1000L * 1000L;
+    private static final long MAX_INTERVAL_NANOS =
+        Config.get().getAppSecReportMaxTimeout() * 1000L * 1000L * 1000L;
```
bantonsson, line 17: Could switch this to `TimeUnit.SECONDS.toNanos(...)`?

dd-smoke-tests/appsec/springboot/src/test/groovy/datadog/smoketest/appsec/SpringBootSmokeTest.groovy
```diff
@@ -0,0 +1,57 @@
...
+
+import okhttp3.Request
+import spock.util.concurrent.PollingConditions
+
+import java.util.concurrent.TimeUnit
+
+class SpringBootSmokeTest extends AbstractAppSecServerSmokeTest {
+
+  // Estimate for the amount of time instrumentation, plus request, plus some extra
+  public static final int TIMEOUT_SECS = 30
```
bantonsson, line 11: The comment makes no sense to me, since it's the total time it would take from the request is **done** to the AppSec attack has been reported.

internal-api/src/main/java/datadog/trace/api/Config.java
```diff
@@ -858,6 +861,10 @@ private Config(final ConfigProvider configProvider) {
     appSecEnabled = configProvider.getBoolean(APPSEC_ENABLED, DEFAULT_APPSEC_ENABLED);
     appSecRulesFile = configProvider.getString(APPSEC_RULES_FILE, null);
 
+    // Default AppSec report timeout min=5, max=60
+    appSecReportMaxTimeout = configProvider.getInteger(APPSEC_REPORT_TIMEOUT_SEC, 30);
```
bantonsson, line 865: Comment says 60, but code says 30

bantonsson (changes requested):
Looks good. Only minor things.

dd-java-agent/appsec/src/main/java/com/datadog/appsec/report/ReportStrategy.java
ValentinZakharov: Switched

dd-smoke-tests/appsec/springboot/src/test/groovy/datadog/smoketest/appsec/SpringBootSmokeTest.groovy
ValentinZakharov: Ah, I'm sorry, blind copy-paste. Removed redundant code.

internal-api/src/main/java/datadog/trace/api/Config.java
ValentinZakharov: Old comment - fixed.

bantonsson: Well, the comment might be old, but the tests seem to think that it should be 60 seconds since this fails `subsequent event of same type should wait 1 min`

bantonsson (approved):

PR #3011, opened 2021-08-17
Ratpack request body interception

cataphract:


dd-java-agent/instrumentation/ratpack-1.5/src/main/java8/datadog/trace/instrumentation/ratpack/RatpackBodyReadStreamAdvice.java
```diff
@@ -0,0 +1,89 @@
...
+    }
+  }
+
+  public static class ByteBufIntoByteBufferCallback
+      implements StoredByteBody.ByteBufferWriteCallback {
+    ByteBuf byteBuf;
+
+    @Override
+    public void put(ByteBuffer undecodedData) {
+      byteBuf.readBytes(undecodedData);
```
bantonsson, line 86: I tried to understand the doc for `readBytes` to see if this is _safe_, but it simply makes no sense to me. It says that `Throws: IndexOutOfBoundsException – if dst.remaining() is greater than this.readableBytes`. Why would you throw an exception if there is more space than you need? So I guess that's why the use of `limit` in `appendData`. I think needs to be a comment in `ByteBufferWriteCallback` about the _requirements_ described in `appendData` as well.

Another approach would be to send in the length into `put` and have it return the remaining bytes, and do the `limit` dance inside the `put` method?

internal-api/src/main/java/datadog/trace/api/http/StoredByteBody.java
```diff
@@ -56,6 +56,38 @@ public synchronized void appendData(byte[] bytes, int start, int end) {
     storedCharBody.maybeNotifyStart();
   }
 
+  /**
+   * Writes either nothing or exactly <code>len</code> bytes directly in <code>undecodedData</code>,
```
bantonsson, line 60: To me it looks like the code could write parts of the data into `undecodedData`.

dd-java-agent/instrumentation/ratpack-1.5/src/main/java/datadog/trace/instrumentation/ratpack/RatpackTypedDataInstrumentation.java
```diff
@@ -0,0 +1,52 @@
...
+  @Override
+  public Map<String, String> contextStore() {
+    return Collections.singletonMap(
+        "ratpack.http.internal.ByteBufBackedTypedData", Boolean.class.getName());
+  }
+
+  @Override
+  public String[] helperClassNames() {
+    return new String[] {
+      "datadog.trace.instrumentation.ratpack.RatpackRequestBodyCallGetTextAdvice"
```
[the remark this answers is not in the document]
cataphract: Right, I would expect that `RatpackForkedHttpServerTest` would fail in this scenario, though, since it calls TypedData::getBytes(), which is instrumented by `RatpackRequestBodyCallGetTextAdvice`, which is in its turn depends on the undeclared `GetTextCharSequenceSupplier`. I'll need to check why it passes nonetheless.

dd-java-agent/instrumentation/ratpack-1.5/src/main/java8/datadog/trace/instrumentation/ratpack/RatpackBodyReadStreamAdvice.java
```diff
@@ -0,0 +1,89 @@
...
+    final StoredByteBody byteBody =
+        StoredBodyFactories.maybeCreateForByte(null, requestBody.getContentLength());
+    if (byteBody == null) {
+      return;
+    }
+
+    publisher = new RequestBodyCollectionPublisher(byteBody, publisher);
+  }
+
+  public static class RequestBodyCollectionPublisher implements TransformablePublisher<ByteBuf> {
```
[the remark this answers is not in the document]
cataphract: Does it matter if they're static inner classes though? The classes can be loaded independently of the outer class, right?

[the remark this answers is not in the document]
bantonsson: Yes they can, but the risk of inadvertently using something in the outer class is pretty big.

dd-java-agent/instrumentation/ratpack-1.5/src/main/java/datadog/trace/instrumentation/ratpack/RatpackTypedDataInstrumentation.java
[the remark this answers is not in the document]
bantonsson: I think this is because of how the tests are set up, and that those classes are reachable in the tests. The `muzzle` task fails though, since there are things it can't find.

dd-java-agent/instrumentation/ratpack-1.5/src/main/java8/datadog/trace/instrumentation/ratpack/RatpackBodyReadStreamAdvice.java
cataphract: > Why would you throw an exception if there is more space than you need?

I found it strange as well, it could just advance the position of the ByteBuffer or even return the number of written bytes, but it's what was available.

> Another approach would be to send in the length into put and have it return the remaining bytes, and do the limit dance inside the put method?
I think on the whole this would be less efficient. The comparison of `remaining()` and `left` would have to be done on the put method (to adjust the limit if necessary) and the caller would also need to compare `remaining()` with the passed in length. So there would be some extra duplication.

internal-api/src/main/java/datadog/trace/api/http/StoredByteBody.java
cataphract: The loop only completes if `len` bytes have asked from the callback. It is correct that on each one iteration of the loop only a part may get written to `undecodedData`, but across all iterations the full amount must have been written.

bantonsson: Ok, fine that also means that we can try to write up to `len - 1` more data than `MAX_BUFFER_SIZE` and it will be dropped on the floor, if I read the code correctly.

cataphract: Yes, even the full amount can be dropped because `storedCharBody.isLimitReached()` can return false and yet there be enough uncommitted data in `undecodedData` already to fill it up (we can't know if we reached the limit until this undecoded data is converted into `char`s). But is your observation about the wording of the comment or the behavior?

bantonsson: About both actually. I don't understand what the word _written_ means in this context when things can be dropped. That aside, I think that it would be good to stop trying to _write_ as soon as the buffer reaches max capacity.

bantonsson (approved):

ValentinZakharov (approved):

PR #3013, opened 2021-08-17
Refactor Kafka Stream instrumentation

tylerbenson:
The big benefit with this change is it avoids using `activeSpan`/`activeScope` to close the span/scope in `StopSpanAdvice`.
Because we are compiling against the early version instead of the late version, I'm using method handles to access the headers rather than forking the advice.


dd-java-agent/instrumentation/kafka-streams-0.11/src/test/groovy/KafkaStreamsTest.groovy
```diff
@@ -129,10 +129,11 @@ class KafkaStreamsTest extends AgentTestRunner {
           childOf trace(0)[0]
 
           tags {
-            "$Tags.COMPONENT" "java-kafka"
+            "$Tags.COMPONENT" "java-kafka-streams"
```
devinsba, line 132: If we are changing this, can we drop the java prefix?

devinsba (approved):

lpriima (approved):

dd-java-agent/instrumentation/kafka-streams-0.11/src/test/groovy/KafkaStreamsTest.groovy
tylerbenson: Not opposed to it, but I'd rather have that be a separate change across multiple integrations. I thought this change here was important to distinguish the span from the regular kafka span.

bantonsson (approved):

PR #3014, opened 2021-08-18
Use Thread as key for timeline as the same name can be attributed to multiple threads

luhenry:


dd-java-agent/testing/src/main/groovy/datadog/trace/agent/test/checkpoints/Event.groovy
```diff
@@ -15,6 +16,7 @@ class Event {
     this.flags = flags
     this.traceId = traceId
     this.spanId = spanId
+    this.thread = thread
     this.threadId = thread.id
     this.threadName = thread.name
```
jbachorik, line 21: Theoretically, we don't need to keep these values separately any more. They are readily available from the `thread`

jbachorik (approved):

tylerbenson (approved):

devinsba (approved):
Don't forget to add the testing label before merging

charliegracie (approved):

PR #3019, opened 2021-08-19
Kafka Streams: remove span/scope from task

tylerbenson:
When finishing the span, remove it to allow it to be GC'd and not linger.  Don't think it's causing any problems... just more sanitary.

Also fix a flake in the test.

### DIFF
--- a/dd-java-agent/instrumentation/kafka-streams-0.11/src/main/java/datadog/trace/instrumentation/kafka_streams/KafkaStreamTaskInstrumentation.java
+++ b/dd-java-agent/instrumentation/kafka-streams-0.11/src/main/java/datadog/trace/instrumentation/kafka_streams/KafkaStreamTaskInstrumentation.java
@@ -166,7 +166,8 @@ public class KafkaStreamTaskInstrumentation extends Instrumenter.Tracing {
     @Advice.OnMethodExit(onThrowable = Throwable.class, suppress = Throwable.class)
     public static void stop(
         @Advice.Thrown final Throwable throwable, @Advice.This StreamTask task) {
-      AgentScope scope = InstrumentationContext.get(StreamTask.class, AgentScope.class).get(task);
+      AgentScope scope =
+          InstrumentationContext.get(StreamTask.class, AgentScope.class).remove(task);
       if (scope != null) {
         AgentSpan span = scope.span();
         CONSUMER_DECORATE.onError(span, throwable);
--- a/dd-java-agent/instrumentation/kafka-streams-0.11/src/latestDepTest/groovy/KafkaStreamsTest.groovy
+++ b/dd-java-agent/instrumentation/kafka-streams-0.11/src/latestDepTest/groovy/KafkaStreamsTest.groovy
@@ -120,6 +120,7 @@ class KafkaStreamsTest extends AgentTestRunner {
           }
         }
       }
+      def producerSpan = null
       trace(2) {
         sortSpansByStart()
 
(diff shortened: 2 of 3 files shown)


bantonsson (approved):
Makes sense to me, since the next call to `process` will set the next scope in the `StreamTask` anyway.

dd-java-agent/instrumentation/kafka-streams-0.11/src/main/java/datadog/trace/instrumentation/kafka_streams/KafkaStreamTaskInstrumentation.java
```diff
@@ -166,7 +166,8 @@ public static void start(
     @Advice.OnMethodExit(onThrowable = Throwable.class, suppress = Throwable.class)
     public static void stop(
         @Advice.Thrown final Throwable throwable, @Advice.This StreamTask task) {
-      AgentScope scope = InstrumentationContext.get(StreamTask.class, AgentScope.class).get(task);
+      AgentScope scope =
```
dougqh, line 169: One question, that I suppose was more related to the prior PR.

While I _think_ it was probably generally a good idea to track scopes / span on the work object rather than the context stack, it does raise a question for me.

In this context, the only way I can see the wrong thing being on the context stack with prior implementation is if we were already crossing scope close -- which makes me worried that we're still masking a deeper problem.

Are we running strict scope close checking for kafka streams?  If not, we should.
I'd also like to see us add tests that involve `@Trace` annotations, since that seems to a significant source of bugs.

tylerbenson: > Are we running strict scope close checking for kafka streams? If not, we should.

I believe so.  We aren't overriding the `useStrictTraceWrites` method in the test and it defaults to true.

> I'd also like to see us add tests that involve @Trace annotations, since that seems to a significant source of bugs.

I can see how the `@Trace` might mess things up with the prior solution since we were using `activeScope` and `activeSpan`, but that is no longer the case here.  Do you have specific examples of where it was messing things up before and maybe I can try to reproduce and compare?

mcculls (approved):

charliegracie (approved):

dd-java-agent/instrumentation/kafka-streams-0.11/src/main/java/datadog/trace/instrumentation/kafka_streams/KafkaStreamTaskInstrumentation.java
dougqh: Good to know we're using strict mode, I'm guessing our tests aren't thorough enough to expose the problem.
I don't have specific examples -- you can consult the issues and  check with @mcculls.

But my point is that we know somehow activeScope was returning the wrong thing previously.  
The main way that would happen is crossed scopes.

The fix here is an improvement to resiliency but is almost certainly masking another problem.  There needs to be a greater effort to recreate that issue.

tylerbenson: I think most of the complaints about this issue were from prior to #2951. Since then we've had several confirmations that things are much better. I can't prove it, but I'm optimistic that the issue is solved.

PR #3021, opened 2021-08-20
[AppSec] Fixed request body capture

ValentinZakharov:


cataphract (approved):
Makes sense. I remember the current code was a conscious choice at the time, but became out-of-date when I decided to check for both callbacks before collecting the body.

bantonsson (approved):
So could this have been an issue if there was only a subscriber to `KnownAddresses.REQUEST_BODY_RAW` an no subscriber to `EventType.REQUEST_BODY_START`?

cataphract:
> So could this have been an issue if there was only a subscriber to `KnownAddresses.REQUEST_BODY_RAW` an no subscriber to `EventType.REQUEST_BODY_START`?

Yes, exactly.

PR #3023, opened 2021-08-20
Complete appsec Attack report

cataphract:


bantonsson (changes requested):
I think sending the whole `AgentSpan` exposes too many things.

dd-java-agent/agent-bootstrap/src/main/java/datadog/trace/bootstrap/instrumentation/decorator/HttpServerDecorator.java
```diff
@@ -198,23 +209,28 @@ private static void onRequestEndForInstrumentationGateway(@Nonnull final AgentSp
     CallbackProvider cbp = AgentTracer.get().instrumentationGateway();
     RequestContext requestContext = span.getRequestContext();
     if (cbp != null && requestContext != null) {
-      Function<RequestContext, Flow<Void>> callback = cbp.getCallback(Events.REQUEST_ENDED);
+      BiFunction<RequestContext, AgentSpan, Flow<Void>> callback =
+          cbp.getCallback(Events.REQUEST_ENDED);
```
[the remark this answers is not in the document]
bantonsson: @cataphract That will do for now. The issue is also pulling out the map of tags, which will create a copy every time, but that can be optimized later.

bantonsson (approved):

bantonsson:
The failing tests are due to a broken merge on `master`. Rebase and it should be gone.

dd-java-agent/agent-bootstrap/src/main/java/datadog/trace/bootstrap/instrumentation/decorator/HttpServerDecorator.java
```diff
@@ -198,23 +210,28 @@ private static void onRequestEndForInstrumentationGateway(@Nonnull final AgentSp
...
       }
     }
   }
 
-  private static void onRequestIpForInstrumentationGateway(
-      @Nonnull final AgentSpan span, @Nonnull final String ip) {
+  private static Flow<Void> callIGCallbackSocketAddress(
+      @Nonnull final AgentSpan span, @Nonnull final String ip, final int port) {
     CallbackProvider cbp = AgentTracer.get().instrumentationGateway();
+    if (cbp == null) {
```
ValentinZakharov, line 224: Should we also check if `span` is not null?

cataphract: It's supposed tot be guaranteed non-null, as the annotation `@NonNull` indicates. If you check the caller, you see that `span` must be non-null for the code to even reach this place.

dd-java-agent/agent-bootstrap/src/main/java/datadog/trace/bootstrap/instrumentation/decorator/HttpServerDecorator.java
```diff
@@ -92,8 +94,9 @@ public AgentSpan onRequest(
       }
     }
 
+    String method = null;
```
ValentinZakharov, line 97: I'm sorry, I don't really understand, why did you move `method` declaration and null assignment outside the `if` block as it was initially?

cataphract: I guess an artifact of an earlier change I walked back from. I'll change it back.

internal-api/src/main/java/datadog/trace/api/gateway/IGSpanInfo.java
```diff
@@ -0,0 +1,12 @@
+package datadog.trace.api.gateway;
+
+import datadog.trace.api.DDId;
+import java.util.Map;
+
+public interface IGSpanInfo {
```
ValentinZakharov, line 6: I'm sorry, I don't get why did you decide to use separated interface? It seems, we can use `AgentSpan` directly to get same functionallity.

dd-java-agent/agent-bootstrap/src/main/java/datadog/trace/bootstrap/instrumentation/decorator/HttpServerDecorator.java
```diff
@@ -171,17 +176,24 @@ public AgentSpan onResponse(final AgentSpan span, final RESPONSE response) {
   //    return super.onError(span, throwable);
   //  }
 
-  private static void onRequestForInstrumentationGateway(
-      @Nonnull final AgentSpan span, @Nonnull final URIDataAdapter url) {
+  private static void callIGCallbackURI(
```
ValentinZakharov, line 179: Here we use two callbacks in order, for `URI` and `METHOD`. May be it make sense to commit everything with some single callback to reduce amount of IG calls?
Unless you foresee the need for a separate commit for other web servers in the future.

dd-java-agent/agent-bootstrap/src/main/java/datadog/trace/bootstrap/instrumentation/decorator/HttpServerDecorator.java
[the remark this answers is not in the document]
ValentinZakharov: @bantonsson I got your point, but still don't like idea populate interfaces. If we follow this way, we may need (in future) add more extra interfaces for another specific cases.
May be it makes sense to have some read-only interface for all available fields (ex: ImmutableSpan) with getters only. WDYT?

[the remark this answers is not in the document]
bantonsson: @ValentinZakharov I think that is exactly what this interface is, only that we don't expose all the things prematurely.

dd-java-agent/agent-bootstrap/src/main/java/datadog/trace/bootstrap/instrumentation/decorator/HttpServerDecorator.java
cataphract: I've merged them together.

ValentinZakharov (approved):

PR #3028, opened 2021-08-25
Fix checkpoint interval validator

jbachorik:
This change will prevent intermittent test failures due to wrong interval checks.

### DIFF
--- a/dd-java-agent/testing/src/main/groovy/datadog/trace/agent/test/checkpoints/IntervalValidator.groovy
+++ b/dd-java-agent/testing/src/main/groovy/datadog/trace/agent/test/checkpoints/IntervalValidator.groovy
@@ -6,7 +6,7 @@ class IntervalValidator extends AbstractValidator {
     final Event startEvent
     final int startTick
     int endTick = Integer.MAX_VALUE
-    boolean suspended = false
+    boolean inactive = false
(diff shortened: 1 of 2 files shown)


dd-java-agent/testing/src/main/groovy/datadog/trace/agent/test/checkpoints/IntervalValidator.groovy
```diff
@@ -57,8 +57,7 @@ class IntervalValidator extends AbstractValidator {
   }
 
   def endTask(def spanId) {
-    tick++
-    return Result.OK
+    return suspendSpan(spanId)
```
luhenry, line 60:
```suggestion
    return deactivateSpan(spanId)
```

luhenry (changes requested):

dd-java-agent/testing/src/main/groovy/datadog/trace/agent/test/checkpoints/IntervalValidator.groovy
jbachorik: Fixed.

luhenry (approved):

tylerbenson (approved):

PR #3036, opened 2021-08-26
Retry for uploads failing with 5xx errors

luhenry:
The approach is similar to how it's done in dd-trace-go.

### DIFF
--- a/dd-java-agent/agent-profiling/profiling-uploader/src/main/java/com/datadog/profiling/uploader/ProfileUploader.java
+++ b/dd-java-agent/agent-profiling/profiling-uploader/src/main/java/com/datadog/profiling/uploader/ProfileUploader.java
@@ -40,6 +40,7 @@ import java.util.List;
 import java.util.Map;
 import java.util.concurrent.ExecutorService;
 import java.util.concurrent.SynchronousQueue;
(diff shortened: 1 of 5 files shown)


dd-java-agent/agent-profiling/profiling-uploader/src/main/java/com/datadog/profiling/uploader/ProfileUploader.java
```diff
@@ -158,7 +166,7 @@ public ProfileUploader(final Config config) throws IOException {
     // Reusing connections causes non daemon threads to be created which causes agent to prevent app
     // from exiting. See https://github.com/square/okhttp/issues/4029 for some details.
     final ConnectionPool connectionPool =
-        new ConnectionPool(MAX_RUNNING_REQUESTS, 1, TimeUnit.SECONDS);
+        new ConnectionPool(MAX_RUNNING_REQUESTS, uploadPeriod * 2, TimeUnit.SECONDS);
```
jbachorik, line 169: How is increasing the keepalive duration affecting the problem described in the above comment?
By reading that comment it almost sounds like we might be blocking the host application from exiting until the connection is released (eg. 120s here). Is it still the case?

luhenry:
Shelving that for now.

PR #3049, opened 2021-08-31
Rework lettuce reactive client instrumentation

jbachorik:
The Lettuce reactive client is based on [Reactive Streams](http://reactives-treams.org)

The whole command execution process is driven by an associated `io.lettuce.core.RedisPublisher#RedisSubscription` instance which creates a `reactor.core.publisher.Mono` (for a singleton command) or `reactor.core.publisher.Flux` (for a command set) instance. 

Once the subscription is created the client will subscribe to it and by doing this will trigger  the reactive command processing.
As a part of the subscription process a special `io.lettuce.core.RedisPublisher.SubscriptionCommand` wrapper command instance is created which will be handed over for transmission to Redis in `io.lettuce.core.RedisPublisher.RedisSubscription#dispatchCommand()` method. At exactly this point the associated span can be marked for migration as the Redis response (if any) may be processed by a different thread.

When the response is received back from Redis server the subscription command is connected with the incoming data and `io.lettuce.core.RedisPublisher.SubscriptionCommand#complete()` method is invoked to start processing the response. At this point the command span migration can be finalized as the span processing resumes on this particular thread.

When it is a set of commands that is being executed we want to have the information about the number of processed commands - this can be achieved by hooking into `io.lettuce.core.RedisPublisher.RedisSubscription#onNext(Object)` method, incrementing the tracked count until the associated subscription command has been cancelled. In order to propagate the parent span active at the time of issuing the Redis command it is necessary to instrument
`io.lettuce.core.AbstractRedisReactiveCommands`, intercepting all `createMono` and `create*Flux` methods - capturing the parent span when `reactor.core.publisher.Mono` or `reactor.core.publisher.Flux` instance is created and reactivating each time they are subscribed.


dd-java-agent/instrumentation/lettuce-5/src/main/java/datadog/trace/instrumentation/lettuce5/LettuceReactiveClientInstrumentation.java
```diff
@@ -0,0 +1,133 @@
...
+import datadog.trace.bootstrap.instrumentation.api.AgentSpan;
+import java.util.HashMap;
+import java.util.Map;
+import net.bytebuddy.description.method.MethodDescription;
+import net.bytebuddy.description.type.TypeDescription;
+import net.bytebuddy.matcher.ElementMatcher;
+import net.bytebuddy.matcher.ElementMatchers;
+
+/**
+ * The Lettuce reactive client is based on <a href=http://reactives-treams.org">Reactive
```
devinsba, line 22:
```suggestion
 * The Lettuce reactive client is based on <a href="http://reactives-treams.org">Reactive
```
Githubs code highlighter apparently isn't cool with that

devinsba:
This looks good to me. I think we might want to generalize some of this into the reactor instrumentation at some point but I like starting with the more specific instrumentation here

dd-java-agent/instrumentation/lettuce-5/src/main/java8/datadog/trace/instrumentation/lettuce5/rx/LettuceFluxCreationAdvice.java
```diff
@@ -15,22 +14,11 @@ public static RedisCommand extractCommandName(
...
-        new LettuceFluxTerminationRunnable(command, finishSpanOnClose);
-    publisher = publisher.doOnSubscribe(handler.getOnSubscribeConsumer());
-    // don't register extra callbacks to finish the spans if the command being instrumented is one
-    // of those that return
-    // Mono<Void> (In here a flux is created first and then converted to Mono<Void>)
-    if (!finishSpanOnClose) {
-      publisher = publisher.doOnEach(handler);
-      publisher = publisher.doOnCancel(handler);
-    }
+  public static void afterCreateMono(@Advice.Return(readOnly = false) Flux<?> publisher) {
```
devinsba, line 20: nit: `afterCreateFlux`

dd-java-agent/instrumentation/lettuce-5/src/main/java8/datadog/trace/instrumentation/lettuce5/rx/RedisChannelHandlerAdvice.java
```diff
@@ -0,0 +1,27 @@
...
+
+  @Advice.OnMethodExit(suppress = Throwable.class)
+  public static void afterDispatch(@Advice.Argument(0) RedisCommand command) {
+
+    ContextStore<RedisCommand, Subscription> ctx1 =
+        InstrumentationContext.get(RedisCommand.class, Subscription.class);
+    Subscription subscription = ctx1 != null ? ctx1.get(command) : null;
+    if (subscription != null) {
+      ContextStore<Subscription, AgentSpan> ctx2 =
+          InstrumentationContext.get(Subscription.class, AgentSpan.class);
```
devinsba, line 20: The instrumentation context lookups in this advice don't match either of the ones that are declared in the instrumentation class so I suspect this advice is unnecessary if things work without it ever doing a migration here

```
store.put("org.reactivestreams.Subscription", packageName + ".rx.RedisSubscriptionState");
store.put("io.lettuce.core.protocol.RedisCommand", AgentSpan.class.getName());
```

dd-java-agent/instrumentation/lettuce-5/src/main/java8/datadog/trace/instrumentation/lettuce5/rx/RedisChannelHandlerAdvice.java
```diff
@@ -0,0 +1,27 @@
+package datadog.trace.instrumentation.lettuce5.rx;
+
+import datadog.trace.bootstrap.ContextStore;
+import datadog.trace.bootstrap.InstrumentationContext;
+import datadog.trace.bootstrap.instrumentation.api.AgentSpan;
+import io.lettuce.core.protocol.RedisCommand;
+import net.bytebuddy.asm.Advice;
+import org.reactivestreams.Subscription;
+
+public class RedisChannelHandlerAdvice {
```
devinsba, line 10: Scratch that, this advice is never used

devinsba (approved):
Looks good once you delete the advice that isn't used

dd-java-agent/instrumentation/lettuce-5/src/main/java8/datadog/trace/instrumentation/lettuce5/rx/LettuceFluxCreationAdvice.java
jbachorik: done

dd-java-agent/instrumentation/lettuce-5/src/main/java8/datadog/trace/instrumentation/lettuce5/rx/RedisChannelHandlerAdvice.java
jbachorik: and it's gone

PR #3052, opened 2021-09-02
Fix Hibernate checkpoints sequence

jbachorik:
Current state of the instrumentation will open a new span when a hibernate session is created. This span is closed when the session is closed. From the checkpoint PoV this has the consequence of attributing any work done between opening and closing the session (with the closing possibly happening on a completely different thread) would be attributed to hibernate, if not enclosed in some other span.

While it would be possible to pepper the checkpoint sequence with suspends/resumes which would put the session span to 'sleep' immediately after it has been created and then wake it up just for the time while an operation is being executed on behalf of that session this would create quite a lot of checkpoints without any additional value.

Therefore I am proposing to enhance the internal API for starting a span to specify a flag to disable checkpoint generation - this would be used for the session span here, making totally transparent from the checkpoint PoV.

Unfortunately, fixing this issue unearthed a problem in the checkpoint interval validator and fixing that problem, in turn, showed more places in integrations where I had to put the checkpoint validation excludes to make the tests passing. All of those places are in integrations we already know have some issues with checkpoints so the overall state is not changing at all.


bantonsson (approved):
I think this approach makes sense. I would like to disable checkpoints for synthetic queue spans, that I'm working on right now as well.

luhenry (approved):

PR #3053, opened 2021-09-03
[AppSec] Update libddwaf and use standard appsec logging

cataphract:


dd-java-agent/appsec/src/main/java/com/datadog/appsec/util/StandardizedLogging.java
```diff
@@ -0,0 +1,231 @@
...
+    logger.error(
+        "AppSec could not load libddwaf native library, as a result, "
+            + "AppSec could not start. No security activities will be collected. "
+            + "Please contact support at https://docs.datadoghq.com/help/ for help. "
+            + "Host information: operating_system: {}, libc: {}, arch: {}, runtime: {} {}",
+        System.getProperty("os.name"),
+        libc,
+        System.getProperty("os.arch"),
+        System.getProperty("java.vm.vendor"),
+        System.getProperty("java.vm.version"));
```
bantonsson, line 41: So for the _Tracer Telemetry_ project, we've settled on the `java.version` property for the `runtime_version` tag, since the internal build number doesn't really add any useful information.

Not important now, but some of this could probably be moved into `datadog.trace.api.Platform` so we use the same names and properties when we display information to the user.

dd-java-agent/appsec/src/main/java/com/datadog/appsec/util/StandardizedLogging.java
```diff
@@ -0,0 +1,231 @@
...
+   *
+   * Cannot be implemented without cooperation from libddwaf
+   */
+
+  /*
+   * D1: Loaded rules:
+   *     <for (ruleName,ruleAddresses) in valid_rules:>
+   *       <ruleName> on addresses <ruleAddresses>"
+   *
+   * Cannot be implemented with cooperation from libddwaf
```
bantonsson, line 91: _with_ or _without_?

internal-api/src/main/java/datadog/trace/api/Config.java
```diff
@@ -2060,6 +2067,11 @@ public String toString() {
         + grpcIgnoredOutboundMethods
         + ", configProvider="
         + configProvider
+        + ", appSecEnabled="
+        + appSecEnabled
+        + ", appSecRules=\'"
```
bantonsson, line 2072: Extra `\`?

internal-api/src/main/java/datadog/trace/api/Config.java
```diff
@@ -67,6 +67,7 @@
 import static datadog.trace.api.IdGenerationStrategy.RANDOM;
 import static datadog.trace.api.Platform.isJavaVersionAtLeast;
 import static datadog.trace.api.config.AppSecConfig.APPSEC_ENABLED;
+import static datadog.trace.api.config.AppSecConfig.APPSEC_RULES;
```
ValentinZakharov, line 70: I'm sorry `APPSEC_RULES` and `appSecRules` makes me confused, because it's not rules as string, but filename. Can we use `FILE` postfix to make it more clear?
Ex: `APP_SEC_RULES_FILE`, `appSecRulesFile`

dd-java-agent/appsec/src/main/java/com/datadog/appsec/util/StandardizedLogging.java
```diff
@@ -0,0 +1,231 @@
...
+  public static void rulesFileInvalid(Logger logger, String filename, RulesInvalidReason reason) {
+    logger.error(
+        "AppSec could not read the rule file {} as it was invalid: {}. "
+            + "AppSec will not run any protections in this application.",
+        filename,
+        reason);
+  }
+
+  /*
+   * E1: Some rules are invalid in <paht_to_rules_file>:
```
ValentinZakharov, line 79: typo `path`
```suggestion
   * E1: Some rules are invalid in <path_to_rules_file>:
```

dd-java-agent/appsec/src/main/java/com/datadog/appsec/gateway/GatewayBridge.java
```diff
@@ -150,6 +150,11 @@ public void init() {
         });
   }
 
+  public void stop() {
```
ValentinZakharov, line 153: It seems, `stop()` is never called.
Should we define `GatewayBridge` as static field in `AppSecSystem` and call stop in `AppSecSystem.stop()`?
Accounting fact that we can't reset IG, I think, it still make sense to call this `stop()` method.

dd-java-agent/appsec/src/main/java/com/datadog/appsec/gateway/AppSecRequestContext.java
```diff
@@ -48,6 +49,9 @@ public void addAll(DataBundle newData) {
       if (prev != null) {
         log.warn("Illegal attempt to replace context value for {}", entry.getKey());
       }
+      if (log.isDebugEnabled()) {
```
ValentinZakharov, line 52: Do we really need to check if debug enabled? 
I was thinking, logger will suppress debug messages if level is not `DEBUG`

dd-java-agent/appsec/src/main/java/com/datadog/appsec/config/AppSecConfigServiceImpl.java
```diff
@@ -95,8 +117,50 @@ public void init() {
...
+      throw fnfe;
+    } catch (IOException ioe) {
+      StandardizedLogging.rulesFileInvalid(log, filename, INVALID_JSON_FILE);
+      throw ioe;
+    }
+  }
+
+  private static int countRules(Map<String, Object> config) {
+    Object waf = config.get("waf");
+    if (waf == null || !(waf instanceof Map)) {
```
ValentinZakharov, line 156: I believe we can remove null-check. The `instanceof` operation returns false in case if operand is null.

dd-java-agent/appsec/src/main/java/com/datadog/appsec/gateway/GatewayBridge.java
cataphract: I'm still not sure how to do this. It cannot be shutdown while the application is running because the IG is not resettable. Maybe registering a shutdown hook: this could be used to flush the events to the agent, but currently it's not possible because they are sent in an executor running in another thread and there are shutdown hooks that cancel everything. So right now, there's nothing useful we can do on a stop().

dd-java-agent/appsec/src/main/java/com/datadog/appsec/gateway/AppSecRequestContext.java
cataphract: Yes but this way we avoid the unnecessary call to entry.get() and `StandardizedLogging.addressPushed`. Since DEBUG logging is usually off it should pay off.

ValentinZakharov (approved):

bantonsson (approved):

PR #3058, opened 2021-09-08
introduce CWS integration

safchain:
This PR introduces a way to integrate trace/span with Cloud Workload Security event. 

This is done by registering a memory area to the eBPF part of CWS which contains mapping between Native ThreadId and Span/Trace. The memory address is registered using a special ioctl syscall that the CWS eBPF programs catch. 

Every time a span is created the mapping is updated.

When a security event is emitted this area is read by the eBPF program so that CWS can link spanId/traceId with the event. 

This feature is by default disabled


dd-cws-tls/src/main/java/datadog/cws/tls/CwsErpcTls.java
```diff
@@ -0,0 +1,139 @@
...
+        CLibrary Instance = (CLibrary) Native.load("c", CLibrary.class);
+
+        NativeLong gettid();
+    }
+
+    private int getTID() {
+        Integer thread = threadCache.get(Thread.currentThread());
+        if (thread == null) {
+            int threadId = CLibrary.Instance.gettid().intValue();
+            threadCache.put(Thread.currentThread(), threadId);
```
bantonsson, line 46: I think that the `tid` could be a modeled as a java `ThreadLocal<Integer>` instead of a `ConcurrentLinkedHashMap<Thread, Integer>` to avoid contention and locking.

safchain: Thanks Björn, I made the recommended modification.

PR #3073, opened 2021-09-16
Fix mule-4 integration for checkpoints

luhenry:


dd-java-agent/instrumentation/java-concurrent/src/main/java/datadog/trace/instrumentation/java/concurrent/RunnableFutureInstrumentation.java
```diff
@@ -64,15 +65,28 @@ public void adviceTransformations(AdviceTransformation transformation) {
...
+        isConstructor()
+            .and(
+                isDeclaredBy(
+                    named("java.util.concurrent.FutureTask")
+                        .or(
+                            nameEndsWith(
+                                "com.google.common.util.concurrent.TrustedListenableFutureTask"))))
+            .and(takesArgument(0, named(Runnable.class.getName()))),
+        getClass().getName() + "$ConstructWithRunnable");
+    // transformation.applyAdvice(isConstructor(), getClass().getName() + "$Construct");
```
jbachorik, line 89: Should be removed before the final merge

dd-java-agent/instrumentation/mule-4/src/main/java/datadog/trace/instrumentation/mule4/HttpExtensionClientInstrumentation.java
```diff
@@ -0,0 +1,65 @@
...
+  @Override
+  public ElementMatcher<TypeDescription> typeMatcher() {
+    return namedOneOf(
+        "org.mule.extension.http.internal.request.client.HttpExtensionClient");
+  }
+
+  @Override
+  public Map<ExcludeType, ? extends Collection<String>> excludedClasses() {
+    Map<ExcludeType, Collection<String>> exclude = new EnumMap<>(ExcludeType.class);
+    exclude.put(COMPLETABLE_FUTURE, Arrays.asList("java.util.concurrent.CompletableFutureWrapper"));
```
jbachorik, line 48: The actual `CompletableFutureWrapper` class is in a different package. Will it get 'automagically' relocated to `java.util.concurrent`?
Also, how does this exclude work? AFAICU, the wrapper will not be instrumented but the wrapped completable future delegate still will so we would end up in the instrumented code anyway. Or is this not how the exclude works?

dd-java-agent/instrumentation/java-concurrent/src/main/java/datadog/trace/instrumentation/java/concurrent/RunnableFutureInstrumentation.java
```diff
@@ -64,15 +65,28 @@ public void adviceTransformations(AdviceTransformation transformation) {
...
+    transformation.applyAdvice(
+        isConstructor()
+            .and(
+                isDeclaredBy(
+                    named("java.util.concurrent.FutureTask")
+                        .or(
+                            nameEndsWith(
+                                "com.google.common.util.concurrent.TrustedListenableFutureTask"))))
+            .and(takesArgument(0, named(Runnable.class.getName()))),
+        getClass().getName() + "$ConstructWithRunnable");
```
jbachorik, line 88: Can you move this block right after L73 and add a short explanatory comment why seemingly same transformation is used - eg. to distinguish between `Runnable` and `Callable` argument?

dd-java-agent/instrumentation/mule-4/src/main/java8/datadog/trace/instrumentation/mule4/BoundedLocalCacheAdvice.java
```diff
@@ -0,0 +1,17 @@
...
+public class BoundedLocalCacheAdvice {
+  @Advice.OnMethodEnter(suppress = Throwable.class)
+  public static void onEnter() {
+    CallDepthThreadLocalMap.incrementCallDepth(ForkJoinPool.class);
+  }
+
+  @Advice.OnMethodExit(suppress = Throwable.class, onThrowable = Throwable.class)
+  public static void onExit() {
+    CallDepthThreadLocalMap.decrementCallDepth(ForkJoinPool.class);
+  }
```
devinsba, line 16: Is it not possible to use the existing java-concurrent exclusions to suppress these?

bantonsson: This is special, since it's this particular instance of `UniWhenComplete` that needs to be suppressed and not all instances of `UniWhenComplete`.

devinsba: :+1:

dd-java-agent/instrumentation/mule-4/mule-4.gradle
```diff
@@ -14,7 +14,7 @@ muzzle {
     assertInverse = true
   }
 
-  fail {
+  pass {
```
jbachorik, line 17: Why is the muzzle condition changed here?

bantonsson: It shouldn't be changed (discussed in slack). That's why the `muzzle` task is failing the build, since the build now says that it should allow the instrumentation to be active for all versions from `3.2.0` and upwards, which it shouldn't.

luhenry: It was failing before already, so this change had no impact on the `muzzle` task passing or not (probably a different failure though).

bantonsson: I can say with 100% certainty that all the failures from `3.2.0` and up to but not including `4.0.0` that are identical to this one is because of this change:
```
> Task :dd-java-agent:instrumentation:mule-4:muzzle-AssertPass-org.mule-mule-core-3.2.0 FAILED
FAILED MUZZLE VALIDATION: datadog.trace.instrumentation.mule4.EventContextInstrumentation mismatches:
-- datadog.trace.instrumentation.mule4.EventContextCreationAdvice:16 Missing class org.mule.runtime.api.event.EventContext
-- datadog.trace.instrumentation.mule4.EventContextCreationAdvice:36 Missing class org.mule.runtime.core.internal.event.DefaultEventContext
```

luhenry:
@bantonsson now that I reverted the change to `muzzle`, it's failing with the following
```
> Task :dd-java-agent:instrumentation:mule-4:muzzle-AssertFail-org.mule-mule-core-3.2.1 FAILED
MUZZLE PASSED DefaultSchedulerInstrumentation BUT FAILURE WAS EXPECTED
```
How do I get that fixed?

jbachorik:
> @bantonsson now that I reverted the change to `muzzle`, it's failing with the following
> 
> ```
> > Task :dd-java-agent:instrumentation:mule-4:muzzle-AssertFail-org.mule-mule-core-3.2.1 FAILED
> MUZZLE PASSED DefaultSchedulerInstrumentation BUT FAILURE WAS EXPECTED
> ```
> 
> How do I get that fixed?

I guess you will need to make the instrumentation more specific - something is getting activated for mule 3.2.1 even though it should not - usually it is a more generic type/method matching which will fit also older classes.

bantonsson:
Exactly. Here is an example of a method added to the advice in the hazelcast `ClientInvocationInstrumentation` that makes sure the instrumentation is only applied in the range `[3.6,3.9)`. The method doesn't need to do anything useful as long as it _pulls in_ types that will narrow down the versions.

```
    public static void muzzleCheck(
        // Moved in 4.0
        ClientMapProxy proxy,
        // New in 3.6
        DiscoveryStrategy strategy,
        // Renamed in 3.9
        ClientNonSmartInvocationServiceImpl invocationService) {
      strategy.start();
      proxy.getServiceName();
      invocationService.start();
    }
'''

File: dd-java-agent/instrumentation/java-concurrent/java-completablefuture/src/main/java8/java/util/concurrent/CompletableFutureAdvice.java [hunk 6]
```diff
@@ -16,11 +17,17 @@
   public static final class UniConstructor {
     @Advice.OnMethodExit(suppress = Throwable.class)
     public static void afterInit(@Advice.This UniCompletion zis) {
-      TraceScope scope = activeScope();
-      if (zis.isLive() && scope != null) {
-        ContextStore<UniCompletion, ConcurrentState> contextStore =
-            InstrumentationContext.get(UniCompletion.class, ConcurrentState.class);
-        ConcurrentState.captureScope(contextStore, zis, scope);
+      try {
+        if (CallDepthThreadLocalMap.incrementCallDepth(CompletableFuture.class) == 0) {
```
bantonsson, line 21: It's quite possible that this could be changed to handle `noopSpan()` properly, but that's for another PR.

bantonsson (changes requested):
It's only the `UniConstructor` that needs a new mechanism right now. The other ones can use the existing way of wrapping the method in a `noopSpan()`.

PR #3077, opened 2021-09-17
Fix scope activation in vertx-redis client

jbachorik:
In this PR the main point of change is actually activating the span started in `RedisAPICallAdvice` and dealing with the `Future` handler in `RedisAPIImplSendAdvice` where we are hooking into `RedisAPIImpl.send()` method to be able to get our hands on the `Future` instance we need to set the handler to.

In that advice we will try to add the handler if the future is not already completed and simply run the handler if the future is completed. However, this still does not prevent double firing because the future may be completed in between two checks for being complete - therefore the handler is enhanced to ignore the second attempt to run the handling logic.

### DIFF
--- a/dd-java-agent/instrumentation/vertx-redis-client-3.9/src/main/java8/datadog/trace/instrumentation/vertx_redis_client/RedisAPICallAdvice.java
+++ b/dd-java-agent/instrumentation/vertx-redis-client-3.9/src/main/java8/datadog/trace/instrumentation/vertx_redis_client/RedisAPICallAdvice.java
@@ -1,11 +1,14 @@
 package datadog.trace.instrumentation.vertx_redis_client;
 
+import static datadog.trace.bootstrap.instrumentation.api.AgentTracer.activateSpan;
 import static datadog.trace.bootstrap.instrumentation.api.AgentTracer.activeSpan;
 import static datadog.trace.bootstrap.instrumentation.api.AgentTracer.captureSpan;
 import static datadog.trace.instrumentation.vertx_redis_client.VertxRedisClientDecorator.DECORATE;
 
 import datadog.trace.bootstrap.CallDepthThreadLocalMap;
+import datadog.trace.bootstrap.InstrumentationContext;
 import datadog.trace.bootstrap.instrumentation.api.AgentSpan;
+import datadog.trace.bootstrap.instrumentation.api.AgentTracer;
 import datadog.trace.context.TraceScope;
 import io.vertx.core.AsyncResult;
 import io.vertx.core.Handler;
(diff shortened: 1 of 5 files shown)


dd-java-agent/instrumentation/vertx-redis-client-3.9/src/main/java8/datadog/trace/instrumentation/vertx_redis_client/RedisAPICallAdvice.java
```diff
@@ -123,18 +135,26 @@ public static boolean beforeCall(
...
   public static void afterCall(
-      @Advice.Thrown final Throwable throwable, @Advice.Enter final boolean decrement) {
+      @Advice.Thrown final Throwable throwable,
+      @Advice.This final RedisAPI self,
+      @Advice.Enter final boolean decrement) {
     if (decrement) {
       CallDepthThreadLocalMap.decrementCallDepth(RedisAPI.class);
     }
+
+    AgentTracer.activeScope().close();
```
luhenry, line 154: Should we store the scope in a "local" variable?

dd-java-agent/instrumentation/vertx-redis-client-3.9/src/test/groovy/VertxRedisAPITestBase.groovy
```diff
@@ -167,8 +158,8 @@ class VertxRedisAPIRedisConnectionForkedTest extends VertxRedisAPITestBase {
 
   @Override
   def setup() {
-    CheckpointValidator.excludeValidations_DONOTUSE_I_REPEAT_DO_NOT_USE(
-      CheckpointValidationMode.INTERVALS,
-      CheckpointValidationMode.THREAD_SEQUENCE)
+    //    CheckpointValidator.excludeValidations_DONOTUSE_I_REPEAT_DO_NOT_USE(
```
luhenry, line 161: Remove

luhenry (changes requested):

dd-java-agent/instrumentation/vertx-redis-client-3.9/src/main/java8/datadog/trace/instrumentation/vertx_redis_client/RedisAPICallAdvice.java
jbachorik: Yeah. I thought about that but that would have required wrapping the scope and the boolean state into a new object. So, I went for less allocations ...
But definitely can be changed if the allocations are acceptable and we want more safety (eg. not closing some stray scope by chance).

devinsba: I think you can use https://javadoc.io/doc/net.bytebuddy/byte-buddy/1.10.4/net/bytebuddy/asm/Advice.Local.html to declare the scope variable

jbachorik: done

dd-java-agent/instrumentation/vertx-redis-client-3.9/src/test/groovy/VertxRedisAPITestBase.groovy
jbachorik: gone

luhenry (approved):

devinsba (approved):

PR #3083, opened 2021-09-21
Remove trailing slash from HTTP resource name

charliegracie:
2 URLs http://localhost/foo and http://localhost/foo/ should
yield the same resource name


devinsba:
I wonder if this is somewhat context specific, for instance the following scenario:

JSP at `/api/people/index.jsp` which the servlet container also makes available at `/api/people/`

REST Endpoint at `/api/people`

Maybe this needs to be configurable with a flag?

PR #3090, opened 2021-09-23
Move AppSec HTTP header extraction to HttpServerDecorator

bantonsson:
This PR moves the AppSec events out from the `HttpCodec.Extractor` and into the `HttpServerDecorator` so we can ensure that there is always a `Span` available that the `RequestContext` can be associated with. This touches a lot of files in a mechanical way, since the extraction and span creation has moved into the `HttpServerDecorator`.

### DIFF
--- a/internal-api/src/main/java/datadog/trace/bootstrap/instrumentation/api/AgentTracer.java
+++ b/internal-api/src/main/java/datadog/trace/bootstrap/instrumentation/api/AgentTracer.java
@@ -88,16 +88,16 @@ public class AgentTracer {
     return get().noopSpan();
   }
 
-  private static final TracerAPI DEFAULT = new NoopTracerAPI();
+  public static final TracerAPI NOOP_TRACER = new NoopTracerAPI();
 
-  private static volatile TracerAPI provider = DEFAULT;
+  private static volatile TracerAPI provider = NOOP_TRACER;
 
   public static boolean isRegistered() {
-    return provider != DEFAULT;
+    return provider != NOOP_TRACER;
   }
 
   public static synchronized void registerIfAbsent(final TracerAPI tracer) {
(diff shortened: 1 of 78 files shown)


internal-api/src/main/java/datadog/trace/bootstrap/instrumentation/api/AgentTracer.java
```diff
@@ -88,16 +88,16 @@ public static AgentSpan noopSpan() {
     return get().noopSpan();
   }
 
-  private static final TracerAPI DEFAULT = new NoopTracerAPI();
+  public static final TracerAPI NOOP_TRACER = new NoopTracerAPI();
```
devinsba, line 91: Can groovy not get access to this when it was private?

devinsba (approved):
Not for this PR but I think we need to consider the naming of the decorators if we start consolidating the logic like this.

Maybe Adapter or Tracer would be more appropriate than Decorator

internal-api/src/main/java/datadog/trace/bootstrap/instrumentation/api/AgentTracer.java
bantonsson: Yeah probably it can, and that's one of the things I dislike about `groovy` 😉 Do you want me to change it back?

devinsba: It probably doesn't matter given what it is but I think that because internal-api is exposed through dd-trace-ot we need to pay attention to what we make public

bantonsson: Yeah I agree. It would probably be good to at least have separate packages for the internal things.

mcculls (approved):

PR #3092, opened 2021-09-24
Fix spring-webflux-5 integration for checkpoints

luhenry:


dd-java-agent/instrumentation/netty-3.8/src/main/java/datadog/trace/instrumentation/netty38/client/HttpClientRequestTracingHandler.java
```diff
@@ -72,6 +72,10 @@ public void writeRequested(final ChannelHandlerContext ctx, final MessageEvent m
         span.finish();
         throw throwable;
       }
+
+      if (scope.isAsyncPropagating()) {
```
jbachorik, line 76: Is this check necessary? `AgentTracer.activateSpan(span)` is forcing async propagation so when you do this check for any scope that has just been created by calling that method it will yield `true`.

dd-java-agent/instrumentation/spring-webflux-5/src/main/java8/datadog/trace/instrumentation/springwebflux/server/AdviceUtils.java
```diff
@@ -102,26 +70,33 @@ public SpanFinishingSubscriber(CoreSubscriber<? super T> subscriber, AgentSpan s
       this.subscriber = subscriber;
       this.span = span;
       this.context = subscriber.currentContext().put(AgentSpan.class, span);
+
+      span.startThreadMigration();
```
jbachorik, line 74: This behaviour should be described in the class javadoc. It might not be obvious that creating a subscriber instance would initiate the span thread migration.

luhenry: This `SpanFinishingSubscriber` is only used in `spring-webflux-5`, and only from `finishSpanIfPresent` in the parent class.

dd-java-agent/instrumentation/spring-webflux-5/src/main/java8/datadog/trace/instrumentation/springwebflux/server/AdviceUtils.java
```diff
@@ -102,26 +70,33 @@ public SpanFinishingSubscriber(CoreSubscriber<? super T> subscriber, AgentSpan s
...
+        span.startThreadMigration();
       }
     }
 
     @Override
     public void onNext(T t) {
       try (AgentScope scope = activateSpan(span)) {
+        span.finishThreadMigration();
         subscriber.onNext(t);
+        span.startThreadMigration();
```
devinsba, line 92: Could you put a javadoc here to explain why we do the sequence this way.

devinsba (approved):
I appreciate that you fixed all of the netty versions while you were in here, thanks

luhenry:
I'll add a bunch of Javadoc on all the span.finishThreadMigration and span.startThreadMigration before merging.

dd-java-agent/instrumentation/spring-webflux-5/src/main/java8/datadog/trace/instrumentation/springwebflux/server/AdviceUtils.java
jbachorik: Still, it would be good to have at least a short notice there when someone has to come back to this instrumentation in the future. No need to write a whole novel - a simple one liner would suffice :)

PR #3094, opened 2021-09-24
Add an option to disable endpoint collection for profiles

mar-kolya:
`profiling.endpoint.collection.enabled`
Enabled by default (but only active if checkpoint events are enabled)

### DIFF
--- a/dd-trace-api/src/main/java/datadog/trace/api/config/ProfilingConfig.java
+++ b/dd-trace-api/src/main/java/datadog/trace/api/config/ProfilingConfig.java
@@ -51,6 +51,9 @@ public final class ProfilingConfig {
       "profiling.checkpoints.record.cpu.time";
   public static final String PROFILING_CHECKPOINTS_RATE_LIMIT = "profiling.checkpoints.rate.limit";
   public static final int PROFILING_CHECKPOINTS_RATE_LIMIT_DEFAULT = 50000;
+  public static final String DD_PROFILING_ENDPOINT_COLLECTION_ENABLED =
+      "profiling.endpoint.collection.enabled";
+  public static final boolean DD_PROFILING_ENDPOINT_COLLECTION_ENABLED_DEFAULT = true;
 
   // Not intended for production use
   public static final String PROFILING_AGENTLESS = "profiling.agentless";
--- a/dd-trace-core/jfr-openjdk/src/main/java/datadog/trace/core/jfr/openjdk/JFRCheckpointer.java
+++ b/dd-trace-core/jfr-openjdk/src/main/java/datadog/trace/core/jfr/openjdk/JFRCheckpointer.java
@@ -26,6 +26,7 @@ public class JFRCheckpointer implements Checkpointer {
   private final LongAdder emitted = new LongAdder();
   private final LongAdder dropped = new LongAdder();
   private final int rateLimit;
+  private final boolean isEndpointCollectionEnabled;
 
   public JFRCheckpointer() {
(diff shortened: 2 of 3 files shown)


devinsba (approved):

charliegracie (approved):

luhenry (approved):

dd-trace-core/jfr-openjdk/src/main/java/datadog/trace/core/jfr/openjdk/JFRCheckpointer.java
```diff
@@ -26,6 +26,7 @@
   private final LongAdder emitted = new LongAdder();
   private final LongAdder dropped = new LongAdder();
   private final int rateLimit;
+  private final boolean isEndpointCollectionEnabled;
```
luhenry, line 29: Add the method `isEndpointCollectionEnabled()` to `Config` like all other configurations.

dd-trace-api/src/main/java/datadog/trace/api/config/ProfilingConfig.java
```diff
@@ -51,6 +51,9 @@
       "profiling.checkpoints.record.cpu.time";
   public static final String PROFILING_CHECKPOINTS_RATE_LIMIT = "profiling.checkpoints.rate.limit";
   public static final int PROFILING_CHECKPOINTS_RATE_LIMIT_DEFAULT = 50000;
+  public static final String DD_PROFILING_ENDPOINT_COLLECTION_ENABLED =
```
luhenry, line 54: Rename it to `PROFILING_ENDPOINT_COLLECTION_ENABLED`

luhenry (changes requested):

dd-trace-core/jfr-openjdk/src/main/java/datadog/trace/core/jfr/openjdk/JFRCheckpointer.java
jbachorik: Here I am not totally sure. 
AFAIK, `Config` is deprecated and developers should be discouraged from adding more stuff in there.

jbachorik (approved):
Approving with the condition of renaming the config key to be in line with the other config keys.

dd-trace-core/jfr-openjdk/src/main/java/datadog/trace/core/jfr/openjdk/JFRCheckpointer.java
mar-kolya: Yeah, looks like 'current' approach is to not use `Config` anymore

dd-trace-api/src/main/java/datadog/trace/api/config/ProfilingConfig.java
mar-kolya: Renamed

luhenry (approved):

dd-trace-core/jfr-openjdk/src/main/java/datadog/trace/core/jfr/openjdk/JFRCheckpointer.java
luhenry: It turns out this is now the only use of `ConfigProvider` anywhere else than `Config`, so the current way seems to still be to use `Config` for it. I indeed don't understand why it's marked as `@Deprecated`, but it's not so much in practice.

PR #3098, opened 2021-09-28
Add config knob to capture stacktraces on CheckpointEvent

luhenry:
This is not intended for production use, only for debugging purposes.


jbachorik:
Overall, I am not totally convinced that we should make it easier for the users to turn on the event stacktrace collection.
Mostly because of the possibility of performance degradation as well as unexpected profile size increase.
AFAIK, it is still possible to enable the stacktrace collection via the JFP overrides file - and it should only be used for on-site debugging under the direct supervision of our supporting engineers when the customer is clearly communicated all possible consequences.

PR #3103, opened 2021-09-29
Add type parameter to RequestContext

bantonsson:
This PR adds a type parameter to the `RequestContext` in the `InstrumentationGateway` and separates the actual user data from the `RequestContext` carrier. This is in preparation for adding more functionality to the `RequestContext`.


ValentinZakharov (approved):

dd-java-agent/appsec/src/main/java/com/datadog/appsec/gateway/GatewayBridge.java
```diff
@@ -65,12 +67,13 @@ public GatewayBridge(
   }
 
   public void init() {
+    Events<AppSecRequestContext> events = Events.get();
```
cataphract, line 70: do you mean to use the static `EVENTS`?

cataphract (approved):

dd-java-agent/appsec/src/main/java/com/datadog/appsec/gateway/GatewayBridge.java
bantonsson: Nope, I want this to be typed to `AppSecRequestContext` to avoid casts in the rest of the code.

PR #3106, opened 2021-09-30
[AppSec] Fixed event types

ValentinZakharov:


utils/test-utils/src/main/groovy/datadog/trace/test/util/DDSpecification.groovy
```diff
@@ -80,6 +80,7 @@ abstract class DDSpecification extends Specification {
       configConstructor.setAccessible(true)
       isConfigInstanceModifiable = true
     } catch (ClassNotFoundException e) {
+      e.printStackTrace()
```
bantonsson, line 83: Leftover? It is printed in one of the if statement branches below.

bantonsson (changes requested):
👍 Only a superfluous `printStackTrace` left in the code.

utils/test-utils/src/main/groovy/datadog/trace/test/util/DDSpecification.groovy
ValentinZakharov: Fixed

bantonsson (approved):

PR #3108, opened 2021-10-02
[AppSec] Fixed NPE caused by absent address values

ValentinZakharov:
AppSec can produce NPE if web server provide null values for addresses.
To avoid this, we will ignore addresses with null values.


dd-java-agent/appsec/src/main/java/com/datadog/appsec/gateway/AppSecRequestContext.java
```diff
@@ -45,12 +45,18 @@
   // to be called by the Event Dispatcher
   public void addAll(DataBundle newData) {
     for (Map.Entry<Address<?>, Object> entry : newData) {
-      Object prev = persistentData.putIfAbsent(entry.getKey(), entry.getValue());
+      Address<?> address = entry.getKey();
+      Object value = entry.getValue();
+      if (value == null) {
+        log.warn("Address {} ignored, because contains null value.", address);
```
bantonsson, line 51: I'm wondering if all these logs should really be warnings.

bantonsson (approved):

PR #3110, opened 2021-10-04
Simplify mongo instrumentation and fix checkpoint emission

jbachorik:
This PR is composed of two parts:
- Fixing the instrumentation to properly suspend/resume spans on thread migrations
- Modifying the project structure to allow passing muzzle even with the new instrumentation

### Checkpoint fix
The main idea here is to intercept `InternalStreamConnection.sendMessageAsync()` method which will send a mongo command to server and the response can then be processed by any thread. Therefore a 'suspend span' checkpoint is necessary there.
There does not seem to be a clear cut-point where we could inject 'resume span' - there are several candidates but they are not sufficiently specific and are missing the distinguishing information which is expensive or impossible to recover.
With this in mind the current approach resorts to resuming a span right before finishing it - if that span was suspended before.

### Project structure modifications
Shared types are extracted to `common` project. That project does not contain any instrumentation, just helper classes.

In `driver-3.1` project contains the generic 'suspend-span' instrumentation as well as instrumentations specific to mongo 3.1-3.3. The thing is that in mongo 3.4 they changed the `BsonWriter` interface, which we are using, in a way that is not binary compatible so we need to have two separate implementations of that interface. 
The unfortunate consequence is that in order to provide such implementation we need to pull in conflicting dependencies - forcing us to split the instrumentation into two projects where in the `driver-3.4` project the instrumentation definition is basically a copy of what is in `driver-3.1`.
Also, it is not possible to have an 'inverse' muzzle check where one would deactivate advice in case such check would pass - and this would be needed to 'turn-off' the `driver-3.1` instrumentation when `driver-3.4` becomes active.
In order to overcome this the `MongoCommandListener` type we are using to register a custom mongo command listener supports the concept of 'priority' where a listener with higher priority will replace any other listener (with lower priority). So, when the `driver-3.4` is active it will effectively supersede the `driver-3.1` instrumentation.

Yet another 'split' instrumentation is `bson-document`. Since the version range for muzzle check 'pass' in `driver-3.1` project is larger than what the bson document instrumentation can handle it needs to be moved to a separate project which can have its muzzle check set up properly.

And then there are test only projects. They don't contain any instrumentation and their only purpose is to test the existing instrumentation in the context of each particular mongo client library.

The driver instrumentation projects are containing the default tests using `mongo-driver-java` client libraries. For the `mongodb-driver-core` libraries there are `driver-*-core-test` projects. 
The `mongodb-driver-sync` library is tested in `driver-3.10-sync-test` and `driver-4.0-test` projects.
The `mongodb-driver-async` library is tested in `driver-3.3-async-test` project.
The `mongodb-driver-reactivestreams` library is tested in `driver-4.0-test` project.


dd-java-agent/instrumentation/mongo/common/src/main/java/datadog/trace/instrumentation/mongo/Context.java
```diff
@@ -1,6 +1,6 @@
 package datadog.trace.instrumentation.mongo;
 
-final class Context {
+public final class Context {
```
devinsba, line 3: Looks like this could be reverted, everything else is in the same package. I know some muzzle things don't like package-private though so if that's the case then ignore me

devinsba (approved):

dd-java-agent/instrumentation/mongo/common/src/main/java/datadog/trace/instrumentation/mongo/Context.java
jbachorik: I will try that. I moved things around a lot and this might be just a result of a particular move when muzzle was refusing to cooperate but may be just fine now :)

jbachorik: Seems to be working fine.

luhenry (approved):

dd-java-agent/instrumentation/mongo/common/src/main/java/datadog/trace/instrumentation/mongo/MongoCommandListener.java
```diff
@@ -0,0 +1,178 @@
...
+   */
+  public static MongoCommandListener tryRegister(
+      MongoCommandListener listener, List<CommandListener> listeners) {
+    if (listeners.isEmpty()) {
+      listeners.add(listener);
+      return listener;
+    }
+
+    ListIterator<CommandListener> iterator = listeners.listIterator();
+    while (iterator.hasPrevious()) {
```
charliegracie, line 99: I believe this loop should be `while(iterator.hasNext()) {` and change the code inside the loop to get the `next()` instead of previous. AFAIK `hasPrevious()` will answer false if there have been no calls to `next()`.

PR #3131, opened 2021-10-11
[WIP] Instrumentation for Http4S Blaze Server 0.21

bantonsson:


veysiertekin:
Hi @bantonsson , any ETA for this integration?

bantonsson:
@veysiertekin Not really. I've stopped working on it since properly supporting context propagation for cats-effect 2 is way too complicated. Maybe we can do a simpler `experimantal` version like #3244 for this version as well.

PR #2482, opened 2021-03-11
Allow gRPC users to ignore specific outbound RPC calls

richardstartin:
By setting `-Ddd.trace.grpc.ignored.outbound.methods=<comma separated list of full method names>` users can prevent tracing of specific outbound requests.

PR #3162, opened 2021-10-25
Allow gRPC users to ignore specific inbound RPC calls

dev-66:
This PR is based on the work for #2482. `-Ddd.trace.grpc.ignored.inbound.methods=<comma separated list of full method names>` is used to prevent tracing of specific inbound requests.  This can be used to ignore gRPC health checks.


devinsba (approved):
Thanks for the contribution

bantonsson (approved):
👍

PR #3172, opened 2021-10-29
Azure app services extension support

randomanderson:
This pull request add support for the java tracer to run in Azure App Services via an extension https://github.com/DataDog/datadog-aas-extension/tree/master/java

**Support "named pipes" as the transport**
To accomplish this a new `SocketFactory` is used with okhttp. Additionally, I cleaned up some of the usages of OkHttpUtils. There was a bug where tracer metrics did not use unix domain sockets. Unfortunately, because `DDAgentWriter` is still part of the public API, it couldn't be cleaned up further.

**Add launcher for trace agent and dogstatsd**
In AAS, with the extension, the tracer is responsible for launching the trace agent and dogstatsd process. `ProcessSupervisor` relaunches the process if they ever die

**Accept "1" as "true" for configuration**
In AAS, as well as other tracer languages, setting an environment variable to "1" is commonly used to signify true.

**Create parent folder for tracer logs**
Previously, if a parent folder did not exist, the tracer would simply not log anything. With this PR the tracer attempts to create the parent folders.

**Add local root tags**
A set of Azure specific tags are added


bantonsson (approved):

randomanderson:
@bantonsson The last commit added a default service name. Not sure if you want to look it over because I don't like sneaking in extra commits after an approval

bantonsson (approved):
👍

PR #3174, opened 2021-11-01
Add database client by host split by instance type suffix

mohammedalics:
#### Given: 
- `DB_CLIENT_HOST_SPLIT_BY_INSTANCE` is set to `true` and
-  The database name is the same as the service name (for example `jdbc:postgresql://db.example.com:5432/myservice` gives database name `myservice`)

#### Expected: 
Two services with the same name `myservice` but different types. one of type web and another of type db. 

#### Actual:
Found only one service with name `myservice` of type web (and no database service). All database traces are mapped to the web service. 

#### Proposed fix:
Introduce a new boolean variable `DB_CLIENT_HOST_SPLIT_BY_INSTANCE_TYPE_SUFFIX` to control adding the database type as a suffix to the database service name in case of `DB_CLIENT_HOST_SPLIT_BY_INSTANCE` sets to `true`. For instance `myservice-postgresql`

### DIFF
--- a/dd-java-agent/agent-bootstrap/src/main/java/datadog/trace/bootstrap/instrumentation/decorator/DatabaseClientDecorator.java
+++ b/dd-java-agent/agent-bootstrap/src/main/java/datadog/trace/bootstrap/instrumentation/decorator/DatabaseClientDecorator.java
@@ -41,7 +41,10 @@ public abstract class DatabaseClientDecorator<CONNECTION> extends ClientDecorato
       span.setTag(Tags.DB_INSTANCE, instanceName);
 
       if (instanceName != null && Config.get().isDbClientSplitByInstance()) {
-        span.setServiceName(Config.get().getDbClientSplitByInstancePrefix() + instanceName);
+        span.setServiceName(
+            Config.get().isDbClientSplitByInstanceTypeSuffix()
+                ? instanceName + "_" + dbType()
+                : instanceName);
       }
 
       CharSequence hostName = dbHostname(connection);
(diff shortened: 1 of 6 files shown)


dd-java-agent/agent-bootstrap/src/main/java/datadog/trace/bootstrap/instrumentation/decorator/DatabaseClientDecorator.java
```diff
@@ -41,7 +41,10 @@ public AgentSpan onConnection(final AgentSpan span, final CONNECTION connection)
       span.setTag(Tags.DB_INSTANCE, instanceName);
 
       if (instanceName != null && Config.get().isDbClientSplitByInstance()) {
-        span.setServiceName(instanceName);
+        span.setServiceName(
+            Config.get().isDbClientSplitByInstanceTypeSuffix()
+                ? instanceName + "_" + dbType()
```
devinsba, line 46:
```suggestion
                ? instanceName + "-" + dbType()
```

Please update to a hyphen

devinsba (changes requested):

dd-java-agent/agent-bootstrap/src/main/java/datadog/trace/bootstrap/instrumentation/decorator/DatabaseClientDecorator.java
```diff
@@ -41,7 +41,10 @@ public AgentSpan onConnection(final AgentSpan span, final CONNECTION connection)
       span.setTag(Tags.DB_INSTANCE, instanceName);
 
       if (instanceName != null && Config.get().isDbClientSplitByInstance()) {
-        span.setServiceName(instanceName);
+        span.setServiceName(
+            Config.get().isDbClientSplitByInstanceTypeSuffix()
+                ? instanceName + "-" + dbType()
+                : instanceName);
```
bantonsson, line 47: I'm fine with this as long as we're aware that there is no caching for this, while for example the `JDBCDecorator` caches the `instanceName` to avoid recreating the strings.

bantonsson (approved):

devinsba:
We'd like to add some tests before merging. I can take a look this afternoon, unless you have time to add a couple before I get to it

devinsba:
@mohammedalics can you enable modifications from maintainers: 
image

devinsba:
Thanks @mohammedalics

mohammedalics:
@devinsba Do you have an idea when this change will be released?

devinsba:
Looks like there will be a release over the next couple days assuming we don't run into any roadblocks. Keep an eye out for 0.90.0

PR #3175, opened 2021-11-01
Add tests using Play 2.6+ error handler

devinsba:
Add tests to prove that we cover the error handler for play 2.6+


dd-java-agent/testing/src/main/groovy/datadog/trace/agent/test/base/HttpServerTest.groovy
```diff
@@ -855,7 +855,7 @@ abstract class HttpServerTest<SERVER> extends WithHttpServer<SERVER> {
   }
 
   void controllerSpan(TraceAssert trace, ServerEndpoint endpoint = null) {
-    def exception = endpoint == CUSTOM_EXCEPTION ? InputMismatchException : expectedExceptionType()
+    def exception = endpoint == CUSTOM_EXCEPTION ? expectedCustomExceptionType() : expectedExceptionType()
```
bantonsson, line 858: 👍

bantonsson (approved):

PR #3195, opened 2021-11-11
Add config knob for CPU sampling engine in async-profiler

luhenry:


jbachorik (dismissed):

dd-java-agent/agent-profiling/profiling-auxiliary-async/src/main/java/com/datadog/profiling/auxiliary/async/AuxiliaryAsyncProfiler.java
```diff
@@ -248,26 +249,17 @@ String cmdStartProfiling(Path file) throws IllegalStateException {
...
-      if (profilingModes.contains(ProfilingMode.MEMLEAK)) {
-        // if combined with memleak profiling set the memleak interval
-        cmd.append(",memleak=").append(getMemleakInterval()).append('b');
-      }
-    } else if (profilingModes.contains(ProfilingMode.MEMLEAK)) {
-      // only memleak profiling is enabled
-      cmd.append(",event=memleak,memleak=").append(getMemleakInterval()).append('b');
+      // cpu profiling is enabled. This will try to use perf and fallback on
+      // itimer if perf is not available.
+      cmd.append(",event=cpu,interval=").append(getCpuInterval());
```
devinsba, line 254: I noticed the `.append('b')` and m were removed in this change, just wanted to make sure that's intentional

devinsba (approved):

dd-java-agent/agent-profiling/profiling-auxiliary-async/src/main/java/com/datadog/profiling/auxiliary/async/AuxiliaryAsyncProfiler.java
jbachorik: Oh, good catch. We need those postfixes such that we don't have to write unit conversion routines.
It is easier to have everything in bytes('b') and milliseconds('m').

jbachorik (changes requested):
Please, add back the unit postfixes.

jbachorik (approved):

dd-trace-api/src/main/java/datadog/trace/api/config/ProfilingConfig.java
```diff
@@ -53,6 +53,8 @@
   public static final String PROFILING_ASYNC_ALLOC_INTERVAL = "profiling.async.alloc.interval";
   public static final int PROFILING_ASYNC_ALLOC_INTERVAL_DEFAULT = 256 * 1024;
   public static final String PROFILING_ASYNC_CPU_ENABLED = "profiling.async.cpu.enabled";
+  public static final String PROFILING_ASYNC_CPU_MODE = "profiling.async.cpu.mode";
+  public static final String PROFILING_ASYNC_CPU_MODE_DEFAULT = "itimer";
```
jbachorik, line 57: This is inconsistent with the comment at https://github.com/DataDog/dd-trace-java/pull/3195/files#diff-49635e629af4a1c72f11d4e54984a895cf058f1bd535193e3d0433c7b394330dR252

Is the 'cpu' profiling mode not working?

luhenry: Test suite is passing locally with `cpu` (aka perf with fallback to itimer), but rather than shipping a version of dd-trace-java which may potentially be broken, I rather add a config knob that we can easily set in staging for dogfooding. And once the tradeoff is better understood, we can switch the default.

luhenry: And I've fixed the comment at https://github.com/DataDog/dd-trace-java/pull/3195/files#diff-49635e629af4a1c72f11d4e54984a895cf058f1bd535193e3d0433c7b394330dR252

PR #3196, opened 2021-11-11
[CIAPP] CI Visibility Mode v1: Instrumentation refactor

drodriguezhdez:
In this PR, we refactor the Instrumentation for CI Visibility (JUnit4, JUnit5, and TestNG) to use a dedicated `Instrumenter` subclass called `Instrumenter.CiVisibility`.

### DIFF
--- a/dd-java-agent/testing/src/main/groovy/datadog/trace/agent/test/base/TestFrameworkTest.groovy
+++ b/dd-java-agent/testing/src/main/groovy/datadog/trace/agent/test/base/TestFrameworkTest.groovy
@@ -15,8 +15,6 @@ abstract class TestFrameworkTest extends AgentTestRunner {
   void configurePreAgent() {
     super.configurePreAgent()
     injectSysConfig("dd.civisibility.enabled", "true")
-    injectSysConfig("dd.integration.junit.enabled", "true")
-    injectSysConfig("dd.integration.testng.enabled", "true")
   }
 
   void testSpan(TraceAssert trace, int index, final String testSuite, final String testName, final String testStatus, final Map<String, String> testTags = null, final Throwable exception = null) {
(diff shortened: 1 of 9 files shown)


dd-java-agent/testing/src/main/groovy/datadog/trace/agent/test/base/TestFrameworkTest.groovy
```diff
@@ -14,8 +14,7 @@ abstract class TestFrameworkTest extends AgentTestRunner {
   @Override
   void configurePreAgent() {
     super.configurePreAgent()
-    injectSysConfig("dd.integration.junit.enabled", "true")
-    injectSysConfig("dd.integration.testng.enabled", "true")
+    injectSysConfig("dd.civisibility.enabled", "true")
```
devinsba, line 17: I don't love this property name, considering this is java could we add a hyphen `dd.ci-visibility.enabled`? Or is this the same as other languages will have ie: `DD_CIVISIBILITY_ENABLED`

drodriguezhdez: I think we are already using `dd.civisibility.enabled` in other languages, so I'd say we want consistency here.

devinsba: Agreed, just wanted to check since it's a really awkward word to see and type.

bantonsson:
I think is going in the right direction. I'm wondering if there should be a `project/ci-visibility` branch where all the changes go, much like `project/appsec-inband-poc` and other projects before that, so things can be merged and built up without affecting `master` until the change is _complete_?

drodriguezhdez:
> I think is going in the right direction. I'm wondering if there should be a project/ci-visibility branch where all the changes go, much like project/appsec-inband-poc and other projects before that, so things can be merged and built up without affecting master until the change is complete?

Sure, I created a branch called `project/ci-visibility` from this one. The new PR link: https://github.com/DataDog/dd-trace-java/pull/3199

Thanks!!

PR #3206, opened 2021-11-17
Reworked java version parser

ValentinZakharov:
Customer wasn't able to run tracer with AdoptOpenJDK version `11.0.9.1+1` because of java version parser wasn't able to recognize version consist from 5 numbers.
Proposed solution uses custom parser spliterator to recognize numbers in given string.


internal-api/src/main/java/datadog/trace/api/Platform.java
```diff
@@ -15,45 +17,52 @@
...
+    List<Integer> results = new LinkedList<>();
+
+    byte[] arr = str.getBytes();
+    int len = str.length();
+
+    int value = 0;
+    for (int i = 0; i < len; i++) {
+      byte ch = arr[i];
+      // digits 0..9 -> code 48..57
+      if (ch >= 48 && ch <= 57) {
```
bantonsson, line 31: I would use `ch - '0' >= 0 && ch - '0' <= '9'` for readability, or simply have `int digit = ch - '0'` and use that in the code after.

bantonsson (approved):
Only minor nitpick

internal-api/src/main/java/datadog/trace/api/Platform.java
ValentinZakharov: Fixed. 
I used first suggestion, because of `int digit = ch - '0'` will perform subtraction every iteration even when it's not needed.

PR #3216, opened 2021-11-19
[AppSec] Fixed tags reported by AppSec

ValentinZakharov:


dd-java-agent/appsec/src/main/java/com/datadog/appsec/gateway/GatewayBridge.java
```diff
@@ -375,6 +378,18 @@ private static int byteToDigit(byte b) {
     return -1;
   }
 
+  /** Read 'DD_TAGS' environment variable and split it to Set with comma or space delimiter */
+  private static Set<String> getEnvTags() {
```
bantonsson, line 382: I find it weird that this is not available in the `Config` and if it isn't, then maybe it should be?

cataphract (commented):
@ValentinZakharov Is the problem here that `Config.getGlobalTags()` has too much stuff and `spanInfo.getTags() even more? It would indeed seem that the better course of action would be to add that to `Config`,  otherwise sources of configuration other than environment variables will not be used. Also, it's strange to have this `IGSpanInfo` with a `getTags` that is never used by appsec.

internal-api/src/main/java/datadog/trace/api/Config.java
```diff
@@ -617,6 +617,7 @@ private Config(final ConfigProvider configProvider) {
     {
       final Map<String, String> tags = new HashMap<>(configProvider.getMergedMap(GLOBAL_TAGS));
       tags.putAll(configProvider.getMergedMap(TAGS));
+      tags.putAll(parseStringIntoMapOfNonEmptyStrings(System.getenv(propertyNameToEnvironmentVariableName(TAGS))));
```
bantonsson, line 620: I don't see why you would need to do this? The line above already goes through the `env` and `properties`

bantonsson (changes requested):
I'm still not sure what you are trying to fix. Maybe a description in the PR of what is wrong and what is expected would help.

internal-api/src/main/java/datadog/trace/api/Config.java
ValentinZakharov: Fixed.
I was wrong with expected format of `DD_TAGS`.

bantonsson:
@ValentinZakharov I can see what you are doing here, but I still don't know why you are doing the change. What is is that AppSec expects? A user could use the tracing API (ot or dd) to override a global tag in a span, and what would you expect to have in that field then?

ValentinZakharov:
I additionally clarified with the collect engineer. The AppSec Events don't care about any overrides in the span. The goal is to propagate the tags manually configured by customer using `DD_TAGS` env variable.

bantonsson (approved):

PR #3227, opened 2021-11-30
Set the hard limit for number of checkpoints

jbachorik:
This change adds a hard-limit for the number of checkpoints emitted in one recording.
It is meant to be the last-resort emergency break for non-standard situations (eg. a never ending local root span) where the rate limiting does not work because of the per-local-root-span granularity (a local root span is either sampled out or the checkpoints must be all generated).

Of course, this will mean broken recordings - we might have non-closed intervals etc. To signal this possibility the checkpoint summary event now contains a flag for the truncated checkpoints which might be used to process the data more leniently or not process it at all.


dd-trace-core/jfr-openjdk/src/main/java/datadog/trace/core/jfr/openjdk/JFRCheckpointer.java
```diff
@@ -142,12 +159,19 @@ private void tryEmitCheckpoint(final AgentSpan span, final int flags) {
...
       (usually 'start span') is emitted.
       Things might become problematic when child spans start emitting checkpoints before the
-      local root span has been created but that is obvously invalid behaviour, breaking the whole tracer.
+      local root span has been created but that is obviously invalid behaviour, breaking the whole tracer.
       Therefore, we can afford to omit proper synchronization here and rely only on the span internal
       checkpoint emission flag being properly published (eg. via 'volatile').
        */
       // if the flag hasn't been set yet consult the sampler
-      checkpointed = sampler.sample();
+      if (sampleCount <= samplerConfig.sampleLimit) {
```
mar-kolya, line 167: Should it be `recordingSampleLimit` and `recordingSampleCount` to make intent more clear?

dd-trace-core/jfr-openjdk/src/main/java/datadog/trace/core/jfr/openjdk/JFRCheckpointer.java
```diff
@@ -81,6 +88,10 @@ public int hashCode() {
   private final boolean isEndpointCollectionEnabled;
   private final SamplerConfig samplerConfig;
 
+  private volatile long recordingSampleCount = 0L;
+  private final AtomicLongFieldUpdater<JFRCheckpointer> recordingSampleCountUpdater =
```
luhenry, line 92: I'm curious whether this adds any measurable overhead as it's a thread synchronisation point on emitting events.

luhenry (approved):

dd-trace-core/jfr-openjdk/src/main/java/datadog/trace/core/jfr/openjdk/JFRCheckpointer.java
jbachorik: Yes, there will definitely be a slight overhead. Although not full synchronization, we will have to pay for memory barriers.

PR #3230, opened 2021-12-01
Add ability to submit metrics via named pipes

randomanderson:
On Windows, allows metrics to be submitted via a named pipe. Most of the changes here are plumbing the named pipe config option through to the dogstatsd client builder.

I also added retry logic to the connection because there is often a race between the start of the dogstatsd server and the tracer starting.

This PR relies on https://github.com/DataDog/java-dogstatsd-client/pull/169 being merged and released (does not compile until it is released)

There are no automated tests in this PR because currently neither the build, nor the testsuite can run on Windows. It would be quite a large effort to get the testsuite working on windows.


bantonsson (approved):

devinsba (approved):
Looks good to me, apparently CircleCI does have windows support, I think it would be great to get a sense for how many windows users we have, if it's a large enough proportion then we should prioritize getting tests working

randomanderson:
@devinsba It was fairly simple to add Windows to the JMXFetch PR: https://github.com/DataDog/java-dogstatsd-client/pull/169/files#diff-78a8a19706dbd2a4425dd72bdab0502ed7a2cef16365ab7030a5a0588927bf47 .

The work is really getting the build to work on Windows.

PR #3234, opened 2021-12-02
Switch to profiling V2.4 format uploads

flodav:
This PR add supports for the V2.4 format uploads. It allows to send requests compliant with a new format used by Datadog profiling intake.

The new format can be activated by enabling `-Ddd.profiling.format.v4.enabled=true` flag, the new format is disabled by default.

The new format will supersede the current one for which the support will be removed afterwards.

### DIFF
--- a/dd-java-agent/agent-profiling/profiling-uploader/src/main/java/com/datadog/profiling/uploader/ProfileUploader.java
+++ b/dd-java-agent/agent-profiling/profiling-uploader/src/main/java/com/datadog/profiling/uploader/ProfileUploader.java
@@ -16,6 +16,8 @@
 package com.datadog.profiling.uploader;
 
 import static datadog.common.socket.SocketUtils.discoverApmSocket;
+import static datadog.trace.api.config.ProfilingConfig.DEFAULT_PROFILING_FORMAT_V4_ENABLED;
+import static datadog.trace.api.config.ProfilingConfig.PROFILING_FORMAT_V4_ENABLED;
 import static datadog.trace.util.AgentThreadFactory.AgentThread.PROFILER_HTTP_DISPATCHER;
 
 import com.datadog.profiling.controller.RecordingData;
@@ -27,6 +29,7 @@ import datadog.common.socket.NamedPipeSocketFactory;
 import datadog.common.socket.UnixDomainSocketFactory;
 import datadog.trace.api.Config;
 import datadog.trace.api.IOLogger;
+import datadog.trace.bootstrap.config.provider.ConfigProvider;
 import datadog.trace.util.AgentProxySelector;
 import datadog.trace.util.AgentThreadFactory;
 import java.io.File;
@@ -140,8 +143,14 @@ public final class ProfileUploader {
   private final CompressionType compressionType;
   private final boolean useV4Format;
 
-  public ProfileUploader(final Config config) throws IOException {
-    this(config, new IOLogger(log), ContainerInfo.get().getContainerId(), TERMINATION_TIMEOUT);
+  public ProfileUploader(final Config config, final ConfigProvider configProvider)
+      throws IOException {
+    this(
+        config,
+        configProvider,
+        new IOLogger(log),
+        ContainerInfo.get().getContainerId(),
+        TERMINATION_TIMEOUT);
   }
 
   /**
@@ -150,6 +159,7 @@ public final class ProfileUploader {
    */
   ProfileUploader(
       final Config config,
+      final ConfigProvider configProvider,
       final IOLogger ioLogger,
       final String containerId,
       final int terminationTimeout)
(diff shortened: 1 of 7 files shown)


dougqh:
Can we please add a better description & title to this PR?  WIP doesn't really tell me what this PR is aiming to do.

internal-api/src/main/java/datadog/trace/api/Config.java
```diff
@@ -413,6 +415,7 @@
   private final boolean profilingExcludeAgentThreads;
   private final boolean profilingHotspotsEnabled;
   private final boolean profilingUploadSummaryOn413Enabled;
+  private final boolean profilingFormatV4Enabled;
```
jbachorik, line 418: AFAIK, the `Config` class is marked as deprecated.
It would be preferable not to add more stuff to deprecated classes.
Can you avoid using `Config` and use `ConfigProvider` instead?

dd-trace-api/src/main/java/datadog/trace/api/ConfigDefaults.java
```diff
@@ -74,6 +74,7 @@
   static final boolean DEFAULT_PROFILING_AGENTLESS = false;
   static final boolean DEFAULT_PROFILING_LEGACY_TRACING_INTEGRATION = true;
   static final boolean DEFAULT_PROFILING_UPLOAD_SUMMARY_ON_413 = false;
+  static final boolean DEFAULT_PROFILING_FORMAT_V4_ENABLED = false;
```
jbachorik, line 77: Can you move this definition to `ProfilingConfig`?
Having the default value close to the key definition is rather nice (at least IMO)

dd-java-agent/agent-profiling/profiling-uploader/src/main/java/com/datadog/profiling/uploader/ProfileUploader.java
```diff
@@ -268,45 +301,82 @@ public void shutdown() {
...
+    os.append("\"version\":\"" + V4_VERSION + "\"");
+    os.append("}");
+    return os.toString().getBytes();
   }
 
-  private void makeUploadRequest(
-      @Nonnull final RecordingType type,
-      @Nonnull final RecordingData data,
-      @Nonnull Runnable onCompletion) {
+  private MultipartBody makeUploadRequestV4(
```
jbachorik, line 317: I wonder - if there is a real chance of having the format changing in the future, while we would also need to keep the backward compatibility, it might be cleaner (in the long run) to refactor the class such that there would be one `ProfileUploader` implementation per the format version.

flodav: The idea is to remove the current format once the new one has been validated and proved to be working across the whole ingestion pipeline. We don't want to keep backward compatibility since it will prevent EP team to remove track-specific intakes.

dd-java-agent/agent-profiling/profiling-uploader/src/main/java/com/datadog/profiling/uploader/ProfileUploader.java
```diff
@@ -268,45 +316,82 @@ public void shutdown() {
...
-   * Note that this method is only visible for testing and should not be used from outside this
-   * class.
-   */
-  OkHttpClient getClient() {
-    return client;
+  private byte[] createEvent(@Nonnull final RecordingData data) {
+    StringBuilder os = new StringBuilder();
+    os.append("{");
+    os.append("\"attachments\":[\"" + V4_ATTACHMENT_FILENAME + "\"],");
+    os.append("\"tags_profiler\":\"" + tags.stream().collect(Collectors.joining(",")) + "\",");
```
mar-kolya, line 323: maybe just use `String.join()`?

mar-kolya: Also, it looks like you could precompute this string instead of joining on every upload

mar-kolya (approved):
Seems resonable, but we should test this before it gets merged

dd-java-agent/agent-profiling/profiling-uploader/src/main/java/com/datadog/profiling/uploader/ProfileUploader.java
pmbauer: fyi: as of JDK9, there is no advantage to using `StringBuilder` over basic `+` as long as the string construction isn't in a loop.  The compiler takes care of things.

pmbauer: details here https://bauer.codes/notes/Java+String+Concatenation

and here
https://richardstartin.github.io/posts/perf-myths-and-continuous-profiling#true-once-prefer-stringbuilder-to-string-concatenation

and here

PR #3235, opened 2021-12-02
Propagate trace context for Mongo Reactive Streams

charliegracie:
Mongo reactive streams uses its own work queue so trace context
does not propagate. This change propagtes the trace context via
the callback

Tests were modified to include the failing case


dd-trace-api/src/main/java/datadog/trace/api/config/GeneralConfig.java
```diff
@@ -50,5 +50,7 @@
   public static final String AZURE_APP_SERVICES = "azure.app.services";
   public static final String INTERNAL_EXIT_ON_FAILURE = "trace.internal.exit.on.failure";
 
+  public static final String MONGO_REACTIVESTREAMS_PROPAGATE_ALL_SINGLERESULTCALLBACK = "mongo.reactivestreams.propagate.all.singleresultcallback";
```
devinsba, line 53: I know we aren't consistent with this but I think it's easier to spot typos if we change the property like this:
```suggestion
  public static final String MONGO_REACTIVESTREAMS_PROPAGATE_ALL_SINGLERESULTCALLBACK = "mongo.reactivestreams.propagate.all.single-result-callback";
```

devinsba (approved):

dd-trace-api/src/main/java/datadog/trace/api/config/GeneralConfig.java
bantonsson: I agree that we should make the settings more readable. Also, the long line is too long...

bantonsson (approved):

dd-java-agent/instrumentation/mongo/driver-4.0/src/main/java/datadog/trace/instrumentation/mongo/InternalStreamConnectionInstrumentation.java
```diff
@@ -0,0 +1,96 @@
...
+        isMethod()
+            .and(named("readAsync"))
+            .and(takesArgument(1, named("com.mongodb.internal.async.SingleResultCallback"))),
+        packageName + ".Arg1Advice");
+
+    // THESE COULD END WITH AN EXCEPTION AND THE callback.onResult NOT CALLED so the continuation is
+    // not cancelled/activated
+    // FIXED in:
+    // https://github.com/mongodb/mongo-java-driver/pull/783
+    // https://github.com/mongodb/mongo-java-driver/commit/0eac1f09b9006899b2aed677dbcfdfe0ce94ab45
```
charliegracie, line 42: I believe this was the issue causing the intermittent failures in CI. The failures only ever happened when testing the 4.0.0 release. When testing the latestDepTest it always passed. @bantonsson and @devinsba what are your thoughts on the fix? See comments in CallbackWrapper.

dd-java-agent/instrumentation/mongo/driver-4.0/src/main/java/datadog/trace/instrumentation/mongo/CallbackWrapper.java
```diff
@@ -0,0 +1,39 @@
+package datadog.trace.instrumentation.mongo;
+
+import static datadog.trace.bootstrap.instrumentation.api.AgentTracer.activeScope;
+
+import com.mongodb.internal.async.SingleResultCallback;
+import datadog.trace.bootstrap.instrumentation.api.AgentScope;
+
+public class CallbackWrapper<T> implements SingleResultCallback<Object> {
+  private final AgentScope.Continuation continuation;
```
charliegracie, line 9: I have debated setting this with an atomic or using [State](https://github.com/DataDog/dd-trace-java/blob/master/dd-java-agent/agent-bootstrap/src/main/java/datadog/trace/bootstrap/instrumentation/java/concurrent/State.java).

dd-java-agent/instrumentation/mongo/driver-4.0/src/main/java/datadog/trace/instrumentation/mongo/CallbackWrapper.java
```diff
@@ -0,0 +1,39 @@
...
+    AgentScope scope = continuation.activate();
+    try {
+      wrapped.onResult(result, t);
+    } finally {
+      scope.close();
+    }
+  }
+
+  public void cancel() {
+    continuation.cancel();
```
charliegracie, line 29: In the current implementation of the mongo code I could not attempt to cancel the continuation if it has been activated but I am worried about future changes. The tradeoff is to use an atomic to set/get the continuation which will be a slight performance loss but it will be much more future proof.

dd-java-agent/instrumentation/mongo/driver-4.0/src/main/java/datadog/trace/instrumentation/mongo/InternalStreamConnectionInstrumentation.java
charliegracie: I have re-run the testing 5 times now and none of the Mongo reactive tests have failed. Previous to this fix CI would get at least one failure every 2 times a complete set of testing was executed.

dd-java-agent/instrumentation/mongo/driver-4.0/src/main/java/datadog/trace/instrumentation/mongo/CallbackWrapper.java
bantonsson: Hey, I'm not really sure where this possible race is, and calling `cancel()` on an activated `Continuation` would only result in debug logging. Is the idea to clear out the `Continuation` when you use it in some way? That would probably be fine, and I wouldn't worry too much about the overhead.

bantonsson (approved):

dd-java-agent/instrumentation/mongo/driver-4.0/src/main/java/datadog/trace/instrumentation/mongo/CallbackWrapper.java
charliegracie: I was worried about the debug logging of calling `cancel()` after it was activated. 

If we thought it was bad form to call `cancel()` after the `Continuation` was activated I was going to clear out the continuation once it was cancelled/activated.

bantonsson: I think it is bad from to call `cancel` after activation, so if you think that is a possible scenario, then it's best to clear it out.

charliegracie: My latest commit addresses this.

dd-java-agent/instrumentation/mongo/driver-4.0/src/main/java/datadog/trace/instrumentation/mongo/CallbackWrapper.java
```diff
@@ -4,29 +4,52 @@
...
-    continuation.cancel();
+  private void cancel() {
+    AgentScope.Continuation continuation = getAndResetContinuation();
+    if (null != continuation) {
+      continuation.cancel();
+    }
+  }
+
+  private AgentScope.Continuation getAndResetContinuation() {
+    AgentScope.Continuation continuation = CONTINUATION.get(this);
```
bantonsson, line 47: Won't this do more checks than a normal `AgentScope.Continuation continuation = this.continuation`? Maybe they're optimized away in the end.

charliegracie: I am not sure they will get optimized away due to the atomic in between them. Let me see if I can remove the extra check by changing the code.

bantonsson: @charliegracie Maybe I misunderstood your reply, but I was just thinking that you should use a normal read of the `this.continuation` variable instead of doing `CONTINUATION.get(this)` which does type checks and other things.

charliegracie: Sorry @bantonsson I misunderstood your comment and thought you were talking about the null check.

Did you mean change the code like this:
```
  private AgentScope.Continuation getAndResetContinuation() {
    AgentScope.Continuation continuation = this.continuation;
    if (continuation != null) {
      if (CONTINUATION.compareAndSet(this, continuation, null)) {
        return continuation;
      }
    }
    return null;
  }
  ```

bantonsson: Yes, that was the idea.

charliegracie: changed and pushed

PR #3238, opened 2021-12-06
Compile code and tests with the correct JDK

bantonsson:


bantonsson:
Back to the drawing board 😞

PR #3243, opened 2021-12-07
[AppSec] Enabled WAF stateful mode (WAF Additive API)

ValentinZakharov:
Changed the way we are calling WAF to enforce stateful mode (using WAF Additive API).
Since we are calling WAF several times at different stages of request processing, we need to keep state of WAF (collected data) between calls. 
For example, we will call WAF three times:
1. At the beginning, when parsed URI and http headers;
2. When request body read;
3. When response status available.

For each stage, we need to keep information about perviously collected data related to current request. This requirement based on fact that some of attacks could be detected by compounded data received at different stages.
Enforce using Additive API, allows to share collected data between WAF calls and will be bounded by request context lifetime.

Basically, we are creating WAF Additive context, during the first WAF call and dispose it when `REQUEST_END` event happens. Given WAF Additive context is kept with the `AppSecRequestContext` and used to process only the data related to current request.

### DIFF
--- a/dd-java-agent/appsec/appsec.gradle
+++ b/dd-java-agent/appsec/appsec.gradle
@@ -134,6 +134,7 @@ ext {
   excludedClassesBranchCoverage = [
     'com.datadog.appsec.gateway.GatewayBridge',
     'com.datadog.appsec.event.data.Address',
+    'com.datadog.appsec.powerwaf.PowerWAFModule.PowerWAFEventsCallback',
     // assert never fails
     'com.datadog.appsec.util.StandardizedLogging',
     'com.datadog.appsec.util.AbortStartupException',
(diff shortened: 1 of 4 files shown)


bantonsson (commented):
This looks good, but I would really like there to be some other cleanup mechanism as well, i.e. max outstanding requests and/or time since request start, to clean up the `Additive`, so we don't have unbounded growth.

bantonsson (approved):
So I thought a bit more about this, and as long as we don't hold on to the `AppSecRequestContext` in any other place, I think that automatic cleanup should work fine. It's a bit unfortunate that the PowerWaf code is using finalizers, but that is not for this PR to clean up.

ValentinZakharov:
But, you are correct, in case if `REQUEST_END` event will not be triggered in some instrumentation, this will lead us to memory leak (because of no explicit call `additive.close()`

bantonsson:
From reading the PowerWaf java bridge code, I think that as long as nobody holds on to the `AppSecRequestContext`, it will be freed by the finalizer in the `Additive`. This is not ideal, but better than always leaking when there is no `REQUEST_END`.

PR #3246, opened 2021-12-07
Make it easier to correlate startup debug log lines

devinsba:
My goal is to make it easier to see all lines in debug logs related to a specific instrumentation or instrumenter class. Being able to correlate lines easily is great when we work with spans and traces, so working with the matchers and transformers should be as easy

TODO:

AgentInstaller$TransformLoggingListener

### DIFF
--- a/dd-java-agent/agent-tooling/src/main/java/datadog/trace/agent/tooling/Instrumenter.java
+++ b/dd-java-agent/agent-tooling/src/main/java/datadog/trace/agent/tooling/Instrumenter.java
@@ -163,10 +163,12 @@ public interface Instrumenter {
     @Override
     public final AgentBuilder instrument(final AgentBuilder parentAgentBuilder) {
       if (!isEnabled()) {
-        log.debug(
-            "Disabled - instrumentation.names=[{}] instrumentation.class={}",
-            Strings.join(",", instrumentationNames),
-            getClass().getName());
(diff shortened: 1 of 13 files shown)


bantonsson (approved):
Great cleanup!

dd-java-agent/agent-tooling/src/main/java/datadog/trace/agent/tooling/Instrumenter.java
```diff
@@ -162,7 +163,10 @@ private void lazyInit() {
     @Override
     public final AgentBuilder instrument(final AgentBuilder parentAgentBuilder) {
       if (!isEnabled()) {
-        log.debug("Instrumentation {} is disabled", this);
+        log.debug(
```
charliegracie, line 166: I might wrap this code with `if (log.isDebugEnabled()) {` in case someone disables a lot of instrumentations to save the String creation from `Strings.join`. I am not sure if this is performance critical enough to really matter though.

charliegracie (approved):

dd-java-agent/agent-tooling/src/main/java/datadog/trace/agent/tooling/Instrumenter.java
devinsba: Thanks. I missed this one

PR #3259, opened 2021-12-12
[AppSec] Event reporting via Span tags

ValentinZakharov:
# What Does This Do
Removed legacy AppSec event reporting mechanism and replaced with reporting via Span tags.
When attack detected, event stored as a json string in top span in the tag `_dd.appsec.json`.

# Motivation
The legacy implementation was unreliable and had lack of consistency. Many private betta customers were complaining about problems when traces were received by backend, but at the same time there was no AppSec events sent (for example, if the client was using an outdated version of the agent). Also, there were difficulties in synchronization between traces and AppSec events on the backend side. Both problems happened because traces and AppSec events were sent by different communication channels. That is, the data was not sent up-to-date (asynchronously) or might not be sent at all due to some kind of communication issues.
To solve this problem, it was decided to rework AppSec events reporting in a way where both traces and AppSec events are transmitted through same communication channel and at the same time.

# Additional Notes
To dispatch the AppSec event, a json serialized `AppSecEvent100` object is used. However, according to backend requirements, we must pass the list of AppSec events, so for this purpose, the `AppSecEventWraper` object is used, which will support triggers (see schemas [app-sec-event-1.0.0.json](https://github.com/DataDog/dd-trace-java/blob/69e360d54a09e435fb86a763935dc2097725f2e0/dd-java-agent/appsec/src/main/schemas/raw/events/app-sec-event-1.0.0.json) and [app-sec-event-wrapper.json](https://github.com/DataDog/dd-trace-java/blob/69e360d54a09e435fb86a763935dc2097725f2e0/dd-java-agent/appsec/src/main/schemas/raw/events/app-sec-event-wrapper.json))

In addition, despite the fact that there is a special `TraceSegmet.setTagTop(...)` method for passing hidden data, I deliberately used the `TraceSegmet.setDataTop(...)` method, because I believe that `setDataTop` hides the full tag name (add `_dd. ***. json`) thereby complicating the search and understanding of the code (for correct understanding, it is required that the engineer knows exactly how the `setDataTop` works and how it differs from `setTagTop`)


dd-java-agent/appsec/src/main/java/com/datadog/appsec/gateway/GatewayBridge.java
```diff
@@ -89,19 +87,29 @@ public void init() {
...
+                      .withTriggers(new ArrayList<>(collectedEvents))
+                      .build();
 
-          for (AppSecEvent100 event : collectedEvents) {
-            EventEnrichment.enrich(event, spanInfo, ctx);
+              JsonAdapter<AppSecEventWrapper> adapter =
+                  AppSecEventSerializer.getAppSecEventAdapter();
+              String json = adapter.toJson(wrapper);
+
+              traceSeg.setTagTop("_dd.appsec.json", json);
```
bantonsson, line 110: The whole point of `setDataTop("appsec", lazyJsonThingy)` is twofold:

1. Hide the ugly implementation detail that the information is currently sent as a special tag! This is not something that `AppSec` need to know, but it's a contract between the serialization of the trace, and the backend.
2. Make the JSON string creation lazy, i.e. happen at the point of writing the trace, which happens on another thread, and not here at the point of reporting, which is happening on the application thread.

dd-java-agent/appsec/src/test/groovy/com/datadog/appsec/gateway/GatewayBridgeSpecification.groovy
```diff
@@ -93,7 +91,10 @@ class GatewayBridgeSpecification extends DDSpecification {
     then:
     1 * mockAppSecCtx.transferCollectedEvents() >> [event]
     1 * mockAppSecCtx.close()
-    1 * reportService.reportEvents([event], traceSegment)
+    //1 * reportService.reportEvents([event], traceSegment)
```
bantonsson, line 94: Remove this?

bantonsson (changes requested):
I think the changes look good apart from using `setTagTop` instead of `setDataTop`. See comment.

dd-java-agent/appsec/src/main/java/com/datadog/appsec/gateway/GatewayBridge.java

dd-java-agent/appsec/src/main/java/com/datadog/appsec/gateway/GatewayBridge.java
```diff
@@ -89,19 +85,25 @@ public void init() {
...
+              traceSeg.setTagTop("appsec.event", true);
 
-          for (AppSecEvent100 event : collectedEvents) {
-            EventEnrichment.enrich(event, spanInfo, ctx);
+              AppSecEventWrapper wrapper =
+                  new AppSecEventWrapper.AppSecEventWrapperBuilder()
+                      .withTriggers(new ArrayList<>(collectedEvents))
+                      .build();
+
+              traceSeg.setDataTop("appsec", wrapper);
```
bantonsson, line 104: Will the `toString()` method of the `AppSecEventWrapper` produce JSON? That is the contract in the doc for `setDataTop`

dd-java-agent/appsec/src/main/java/com/datadog/appsec/report/AppSecEventWrapper.java
```diff
@@ -0,0 +1,43 @@
...
+
+  private static final JsonAdapter<AppSecEventWrapper> ADAPTER =
+      new Moshi.Builder().build().adapter(AppSecEventWrapper.class);
+
+  private Collection<AppSecEvent100> triggers;
+  private final String json;
+
+  public AppSecEventWrapper(Collection<AppSecEvent100> events) {
+    this.triggers = events;
+    this.json = ADAPTER.toJson(this);
```
bantonsson, line 19: This is no longer _lazy_. It creates the JSON during construction. Remove the `final` from the `json` field and set it in the `toString()` if it is empty.

bantonsson (approved):

PR #3263, opened 2021-12-13
Use USER_(KEEP/REJECT) for RuleBasedSampler

ygree:
# What Does This Do

Mark the ruler based sample's decisions with user priority.

# Motivation

Decisions that are originated from any user configuration to the `RuleBasedSampler` use priority values -1 (for forced rejection) and 2 (for forced keeping). This means that `RuleBasedSampler` sampling configurations will always be strictly respected by the whole Datadog processing pipeline.

# Additional Notes

See https://github.com/DataDog/dd-trace-rb/pull/1769

### DIFF
--- a/dd-trace-core/src/main/java/datadog/trace/common/sampling/RuleBasedSampler.java
+++ b/dd-trace-core/src/main/java/datadog/trace/common/sampling/RuleBasedSampler.java
@@ -101,19 +101,19 @@ public class RuleBasedSampler<T extends CoreSpan<T>> implements Sampler<T>, Prio
       if (matchedRule.sample(span)) {
         if (rateLimiter.tryAcquire()) {
           span.setSamplingPriority(
-              PrioritySampling.SAMPLER_KEEP,
+              PrioritySampling.USER_KEEP,
               SAMPLING_RULE_RATE,
               matchedRule.getSampler().getSampleRate());
         } else {
           span.setSamplingPriority(
-              PrioritySampling.SAMPLER_DROP,
+              PrioritySampling.USER_DROP,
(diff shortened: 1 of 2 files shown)


dd-trace-core/src/main/java/datadog/trace/common/sampling/RuleBasedSampler.java
```diff
@@ -101,19 +101,19 @@ public void setSamplingPriority(final T span) {
...
         if (rateLimiter.tryAcquire()) {
           span.setSamplingPriority(
-              PrioritySampling.SAMPLER_KEEP,
+              PrioritySampling.USER_KEEP,
               SAMPLING_RULE_RATE,
               matchedRule.getSampler().getSampleRate());
         } else {
           span.setSamplingPriority(
-              PrioritySampling.SAMPLER_DROP,
+              PrioritySampling.USER_DROP,
```
ygree, line 109: When the rate limiter exceeds the limit should it be marked with `SAMPLE_DROP` or `USER_DROP` for the `RuleBasedSampler`?

bantonsson: I think that since the rate limiter is part of the user configurable rules for the `RuleBasedSampler`, it should set `USER_DROP`.

bantonsson (commented):
👍

The build fails on `codenarc` violations which are not auto corrected by the formatter 😞. To just check for that locally run `./gradlew :some-project-here:check -PskipTests`

ygree:
Are there any objections against merge this PR?

charliegracie (approved):

PR #3268, opened 2021-12-15
Add v0.5 trace span decoder for testing

bantonsson:
# What Does This Do

This adds a decoder for the `v0.5` Datadog agent message format that can be used to verify the traces and spans that we send from the instrumented application.

# Motivation

Allows us to do more realistic testing of the full Java Tracer, and is a prerequisite to be able to do proper smoke testing of trace and span contents for things like AppSec inband reporting over traces.

# Additional Notes

This is a bare minimum decoder and helpers to unblock AppSec inband reporting. Future work would be to support `v0.4` messages and create proper matchers to make testing easy.


gradle/java.gradle
```diff
@@ -1,379 +1,2 @@
-import java.time.Duration
```
bantonsson, line 1: Don't know why GitHub doesn't show this as a rename and change, since 90% of the file is moved unchanged to `gradle/java_no_deps.gradle`

gradle/java_no_deps.gradle
```diff
@@ -0,0 +1,371 @@
+import java.time.Duration
```
bantonsson, line 1: This is the old `gradle/java.gradle` with only the dependencies removed.

ValentinZakharov (approved):

ValentinZakharov:
Is there any message format difference between `v0.4` and `v0.5` ?
Can we use same decoder for `v0.4`?

bantonsson:
We can't use the same decoder. The `v0.5` message is dictionary based and the `v0.4` is not.

mcculls (approved):

PR #3272, opened 2021-12-15
[AppSec] Fixed rules file parser

ValentinZakharov:
# What Does This Do
Fixed AppSec rules file parser.
Added support fo both rules formats `1.x` and `2.x`

# Motivation
Using rules file, downloaded from dashboard, caused an exception at AppSec startup, because of inconsistent rules file format. This been preventing AppSec to protect customers application, when they are trying to use external rules file.

# Additional Notes

### DIFF
--- a/dd-java-agent/appsec/src/main/java/com/datadog/appsec/config/AppSecConfig.java
+++ b/dd-java-agent/appsec/src/main/java/com/datadog/appsec/config/AppSecConfig.java
@@ -1,44 +1,52 @@
 package com.datadog.appsec.config;
 
-import com.datadog.appsec.util.Generated;
 import com.squareup.moshi.JsonAdapter;
 import com.squareup.moshi.Moshi;
+import java.io.IOException;
 import java.util.List;
 import java.util.Map;
 import java.util.Objects;
 
-public class AppSecConfig {
+public interface AppSecConfig {
 
-  private static final JsonAdapter<AppSecConfig> ADAPTER =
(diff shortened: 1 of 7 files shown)


bantonsson (dismissed):

dd-java-agent/appsec/src/main/java/com/datadog/appsec/config/AppSecConfig.java
```diff
@@ -58,20 +65,74 @@ public String getName() {
...
-    if (this == o) return true;
-    if (o == null || getClass() != o.getClass()) return false;
-    AppSecConfig config = (AppSecConfig) o;
-    return Objects.equals(version, config.version)
-        && Objects.equals(rules, config.rules)
-        && Objects.equals(rawConfig, config.rawConfig);
+  class AppSecConfigV1 implements AppSecConfig {
+
+    private String version;
+    private List<Rule> events;
```
bantonsson, line 72: Why is the `List<Rule>` named `events`, and the getter named `getRules`?

ValentinZakharov: In rules format `v1.x` json field called `events`, and in `v2.x` it was renamed to `rules`, but both contains security rules. So to unify usage, I created `AppSecConfig` interface, which provide method `getRules`.
Indeed, this confusing a bit, but current implementation seems to me the most simplest.
We will get rid of `v1.x` because it will be deprecated soon.

bantonsson (approved):

PR #3286, opened 2021-12-21
[AppSec] Collect and report pre-defined HTTP request headers

ValentinZakharov:
# What Does This Do
Implemented code to collect and report pre-defined HTTP request headers

# Motivation

# Additional Notes


dd-java-agent/appsec/src/main/java/com/datadog/appsec/gateway/HeadersCollector.java
```diff
@@ -0,0 +1,58 @@
...
+
+import com.datadog.appsec.event.data.CaseInsensitiveMap;
+import java.util.ArrayList;
+import java.util.Arrays;
+import java.util.List;
+
+class HeadersCollector {
+
+  private final CaseInsensitiveMap<List<String>> headers = new CaseInsensitiveMap<>();
+  private final CaseInsensitiveMap<String> allowedHeaders = new CaseInsensitiveMap<>();
```
bantonsson, line 11: Why is there a difference in types between `headers` and allowed headers?

dd-java-agent/appsec/src/main/java/com/datadog/appsec/gateway/HeadersCollector.java
```diff
@@ -0,0 +1,58 @@
...
+          "accept",
+          "accept-encoding",
+          "accept-language");
+
+  void addHeader(String name, String value) {
+    if (name == null || value == null) {
+      return;
+    }
+
+    List<String> strings = headers.computeIfAbsent(name, h -> new ArrayList<>(1));
```
bantonsson, line 39: I genuinely dislike the `CaseInsensitiveMap` trying to magically hide the fact that it just stores the keys converted to lowercase. It has weird implications, like this `computeIfAbsent` converting `name` to lowercase twice. Once for `get` and once for `put`.

dd-java-agent/appsec/src/main/java/com/datadog/appsec/gateway/HeadersCollector.java
```diff
@@ -0,0 +1,58 @@
...
+    strings.add(value);
+
+    if (headersPassList.stream().anyMatch(name::equalsIgnoreCase)) {
```
bantonsson, line 42: I think that converting the `name` to lowercase and then doing a map lookup could be faster. Also if the types for `headers` and `allowedHeaders` were the same, they could point to the same list, and you would only need to do the check if the header wasn't found in the `headers` map.

bantonsson (changes requested):

dd-java-agent/appsec/src/main/java/com/datadog/appsec/gateway/HeadersCollector.java
ValentinZakharov: `headers` it's all headers which used for analysis by WAF (i.e each record contains List of values)
`allowedHeaders` it's headers we are going to report to backend (only allowed headers) and they are represented as string.

bantonsson: Yes, but that doesn't explain why the `allowedHeaders` are eagerly concatenated into a string, when they for key `x` are exactly the same as `headers` for key `x`.

dd-java-agent/appsec/src/main/java/com/datadog/appsec/gateway/GatewayBridge.java
```diff
@@ -95,9 +92,28 @@ public void init() {
...
               traceSeg.setDataTop("appsec", wrapper);
+
+              ctx.getRequestHeaders()
+                  .forEach(
+                      (name, value) -> {
+                        if (AppSecRequestContext.HEADERS_ALLOW_LIST.contains(name)) {
+                          int size = value.size();
+                          if (size > 0) {
+                            String v;
+                            if (size == 1) {
```
bantonsson, line 107: I would move this optimization into `Strings.join`, and change this whole bit to:

```
String v = Strings.join(",", value);
if (!v.isEmpty()) {
  traceSeg.setTagTop("http.request.headers." + name, v);
}
```

internal-api/src/main/java/datadog/trace/util/Strings.java
```diff
@@ -107,22 +108,38 @@ public static String getInnerClassName(final String className) {
...
+      return "";
     }
-    // truncate to remove the last joiner
-    if (sb.length() > 0) {
-      sb.setLength(sb.length() - joiner.length());
+
+    // first element
+    CharSequence first = it.next();
+    if (!it.hasNext()) {
+      return first.toString();
```
bantonsson, line 124: 👍

bantonsson (approved):

PR #3296, opened 2021-12-29
[AppSec] Report metrics for web span only

ValentinZakharov:
# What Does This Do
AppSec reports metrics/events for span with `web` type.  

# Motivation
Several customers reported that database services were showing up in their list of AppSec services (which is obviously wrong as we only support web services as of now).

# Additional Notes

### DIFF
--- a/dd-java-agent/appsec/src/main/java/com/datadog/appsec/gateway/GatewayBridge.java
+++ b/dd-java-agent/appsec/src/main/java/com/datadog/appsec/gateway/GatewayBridge.java
@@ -84,19 +84,25 @@ public class GatewayBridge {
 
           TraceSegment traceSeg = ctx_.getTraceSegment();
 
+          // AppSec report metric and events for web span only
           if (traceSeg != null) {
+            traceSeg.setTagTop("_dd.appsec.enabled", 1);
+            traceSeg.setTagTop("_dd.runtime_family", "jvm");
+
             Collection<AppSecEvent100> collectedEvents = ctx.transferCollectedEvents();
             // If detected any events - mark span at appsec.event
-            if (!collectedEvents.isEmpty() && spanInfo != null) {
+            if (!collectedEvents.isEmpty()) {
               // Keep event related span, because it could be ignored in case of
               // reduced datadog sampling rate.
               traceSeg.setTagTop(DDTags.MANUAL_KEEP, true);
               traceSeg.setTagTop("appsec.event", true);
-              traceSeg.setTagTop("actor.ip", ctx.getPeerAddress());
+              traceSeg.setTagTop("network.client.ip", ctx.getPeerAddress());
 
+              // Report AppSec events via "_dd.appsec.json" tag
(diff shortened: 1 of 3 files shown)


dd-java-agent/appsec/src/main/java/com/datadog/appsec/gateway/GatewayBridge.java
```diff
@@ -84,19 +84,25 @@ public void init() {
 
           TraceSegment traceSeg = ctx_.getTraceSegment();
 
-          if (traceSeg != null) {
+          // AppSec report metric and events for web span only
+          if (traceSeg != null && spanInfo != null && "web".equals(spanInfo.getSpanType())) {
+            traceSeg.setTagTop("_dd.appsec.enabled", 1);
+            traceSeg.setTagTop("_dd.runtime_family", "jvm");
```
bantonsson, line 90: Is there a SPEC for this? Can't this be handled in the backend? I mean AppSec is technically enabled even if there are no web requests processed. One of the downsides to this is that we can't see if AppSec has been enabled if the web framework is not yet supported.

ValentinZakharov: Here is the [AppSec tracking requirements](https://datadoghq.atlassian.net/wiki/spaces/APS/pages/2137128985/Tracking+AppSec+libraries+usage+in+beta).
It seems, there is no way to see if AppSec enabled i case of not supported framework.

bantonsson: Ok, fine. But this still feels a bit weird.

bantonsson (approved):

dd-java-agent/appsec/src/main/java/com/datadog/appsec/gateway/GatewayBridge.java
bantonsson: Woke up this morning with a feeling that this was wrong, and it is.

1. We should only get these callbacks for web requests, so you don't have to filter for the `spanType`.
2. You're setting the top tags based on the span type of the current span, which doesn't necessarily map 1:1.

I think it's fine to always set the top tags here, since we should only get callbacks for web requests. The problem with the other solution was that all local roots regardless of type got the tags.

bantonsson (changes requested):

ValentinZakharov:
You remark makes sense - I removed filter by spanType (tests are passed correctly)

dd-java-agent/appsec/src/main/java/com/datadog/appsec/gateway/GatewayBridge.java
ValentinZakharov: You remark makes sense - I removed filter by spanType (tests are passed correctly)

bantonsson (approved):

PR #3298, opened 2022-01-01
[AppSec] Added response status code analysis

ValentinZakharov:
# What Does This Do
Implemented support for HTTP response code analysis - required for Rule 404.
Reading and analysis of the response status happens at the beginning of http response, BEFORE actual data will be committed.

# Motivation

# Additional Notes
Status code type in KnownAddresses changed to String, because of WAF currently able to process only strings.

### DIFF
--- a/dd-java-agent/agent-bootstrap/src/main/java/datadog/trace/bootstrap/instrumentation/decorator/HttpServerDecorator.java
+++ b/dd-java-agent/agent-bootstrap/src/main/java/datadog/trace/bootstrap/instrumentation/decorator/HttpServerDecorator.java
@@ -193,10 +193,10 @@ public abstract class HttpServerDecorator<REQUEST, CONNECTION, RESPONSE, CARRIER
       if (null != cbp) {
         RequestContext<Object> ctx = span.getRequestContext();
         if (ctx != null) {
-          BiConsumer<RequestContext<Object>, Integer> addrCallback =
+          BiFunction<RequestContext<Object>, Integer, Flow<Void>> addrCallback =
               cbp.getCallback(EVENTS.responseStarted());
           if (null != addrCallback) {
-            addrCallback.accept(ctx, status);
+            addrCallback.apply(ctx, status);
           }
         }
       }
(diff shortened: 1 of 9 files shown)


dd-java-agent/appsec/src/main/java/com/datadog/appsec/event/data/KnownAddresses.java
```diff
@@ -51,10 +51,10 @@
   Address<List<StringKVPair>> REQUEST_COOKIES = new Address<>("server.request.cookies");
 
   /** Same as server transport related field. */
-  Address<String> REQUEST_TRANSPORT = new Address<String>("server.request.transport");
+  Address<String> REQUEST_TRANSPORT = new Address<>("server.request.transport");
 
   /** status code of HTTP response */
-  Address<Integer> RESPONSE_STATUS = new Address<>("server.response.status");
+  Address<String> RESPONSE_STATUS = new Address<>("server.response.status");
```
cataphract, line 57: Any reason you change this to a String? Is it because of the bug on the waf where an integer status is ignored? If so, it would warrant a comment.

dd-java-agent/agent-bootstrap/src/main/java/datadog/trace/bootstrap/instrumentation/decorator/HttpServerDecorator.java
```diff
@@ -193,10 +193,10 @@ public AgentSpan onResponse(final AgentSpan span, final RESPONSE response) {
       if (null != cbp) {
         RequestContext<Object> ctx = span.getRequestContext();
         if (ctx != null) {
-          BiConsumer<RequestContext<Object>, Integer> addrCallback =
```
cataphract, line 196: You add a flow result here, but it is ignored in `HttpServerDecorator`. Not sure if this is intended.

cataphract (approved):

dd-java-agent/agent-bootstrap/src/main/java/datadog/trace/bootstrap/instrumentation/decorator/HttpServerDecorator.java
ValentinZakharov: The flow deliberately added for the future implementation of the blocking (however, result value is not currently used yet).

PR #3305, opened 2022-01-07
Reenable jdk.OldObjectSample event based on the version

luhenry:
# What Does This Do

The bug affecting the OpenJDK [1] was fixed in JDK 17.0.3 and JDK 18.

[1] https://bugs.openjdk.java.net/browse/JDK-8277919

# Motivation

We had to disable capturing the event in 0.91.0 because of a bug discovered in the OpenJDK. Now that we know which version has the right fix, we can enable it again.

# Additional Notes


dd-java-agent/agent-profiling/profiling-controller-openjdk/src/main/java/com/datadog/profiling/controller/openjdk/OpenJdkController.java
```diff
@@ -112,18 +113,17 @@ public OpenJdkController(final Config config)
...
-      // }
+      if (!Boolean.parseBoolean(recordingSettings.get("jdk.OldObjectSample#enabled"))) {
+        if (((isJavaVersion(11) && isJavaVersionAtLeast(11, 0, 12))
+            || (isJavaVersion(15) && isJavaVersionAtLeast(15, 0, 4))
+            || (isJavaVersion(16) && isJavaVersionAtLeast(16, 0, 2))
+            || (isJavaVersion(17) && isJavaVersionAtLeast(17, 0, 3))
+            || isJavaVersionAtLeast(18))) {
+          // It was enabled based on JDK version so disabled by override file
+          log.warn(
+              "The OldObjectSample JFR event is disabled with the override file but enabled with the config.");
```
jbachorik, line 124:
```suggestion
              "The OldObjectSample JFR event is disabled by the override file but enabled by the config.");
```

jbachorik (approved):

PR #3308, opened 2022-01-11
[AppSec] Support InputStream interception in Servlet 3.1+

cataphract:
# What Does This Do

Supports request body interception (when `ServletRequest#getInputStream` is used, rather than `getReader`) in Servlet 3.1.

# Motivation

Servlet 3.1 added 3 new abstract methods in `ServletInputStream`. Our implementation doesn't implement these, causing errors at runtime. This PR implements the remaining methods if servlet 3.1 is detected.

### DIFF
--- a/dd-java-agent/instrumentation/servlet/src/main/java/datadog/trace/instrumentation/servlet/http/ServletInputStreamWrapper.java
+++ b/dd-java-agent/instrumentation/servlet/src/main/java/datadog/trace/instrumentation/servlet/http/ServletInputStreamWrapper.java
@@ -1,16 +1,69 @@
 package datadog.trace.instrumentation.servlet.http;
 
+import static net.bytebuddy.matcher.ElementMatchers.named;
+import static net.bytebuddy.matcher.ElementMatchers.takesArguments;
+
 import datadog.trace.api.http.StoredByteBody;
 import java.io.IOException;
+import java.lang.invoke.MethodHandle;
+import java.lang.invoke.MethodHandles;
+import java.lang.invoke.MethodType;
 import javax.servlet.ServletInputStream;
+import net.bytebuddy.ByteBuddy;
+import net.bytebuddy.dynamic.DynamicType;
+import net.bytebuddy.dynamic.scaffold.subclass.ConstructorStrategy;
+import net.bytebuddy.implementation.MethodDelegation;
+import net.bytebuddy.matcher.ElementMatchers;
 
 public class ServletInputStreamWrapper extends ServletInputStream {
-  final ServletInputStream is;
+  public final ServletInputStream is;
   final StoredByteBody storedByteBody;
+
+  private static final MethodHandle NEW_WRAPPER;
+
+  static {
+    Class<? extends ServletInputStream> finalClass = ServletInputStreamWrapper.class;
+    try {
+      Class<?> readListenerClass =
+          Class.forName(
+              "javax.servlet.ReadListener", true, ServletInputStreamWrapper.class.getClassLoader());
+      DynamicType.Unloaded<ServletInputStreamWrapper> wrapperSubCls =
+          new ByteBuddy()
+              .subclass(
+                  ServletInputStreamWrapper.class, ConstructorStrategy.Default.IMITATE_SUPER_CLASS)
+              .method(named("isFinished").and(ElementMatchers.takesNoArguments()))
+              .intercept(MethodDelegation.toField("is"))
+              .method(named("isReady").and(ElementMatchers.takesNoArguments()))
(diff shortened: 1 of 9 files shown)


dd-java-agent/instrumentation/servlet/request-3.1/src/main/java/datadog/trace/instrumentation/servlet31/ServletRequestGetInputStreamInstrumentation.java
```diff
@@ -0,0 +1,130 @@
...
+            .and(takesNoArguments())
+            .and(returns(named("javax.servlet.ServletInputStream")))
+            .and(isPublic()),
+        getClass().getName() + "$HttpServletGetInputStreamAdvice");
+  }
+
+  @Override
+  public String[] helperClassNames() {
+    return new String[] {
+      "datadog.trace.instrumentation.servlet31.ServletInputStreamWrapper",
```
devinsba, line 63: We usually use the `packageName` helper variable here to make refactors easier
```suggestion
      packageName + ".ServletInputStreamWrapper",
```

cataphract:
@DataDog/apm-java I'm not sure how to handle the muzzle failure. I could limit the servlet version I test against to servlet 3.0, but I do intend the instrumentations other than `ServletRequestGetInputStreamInstrumentation` to apply to servlet 3.1+. Is the only option to split this into a new module for servlet 2.3-3.0 just with this instrumentation?

dd-java-agent/instrumentation/servlet/request-3.1/request-3.1.gradle
```diff
@@ -0,0 +1,28 @@
+muzzle {
+  pass {
+    group = "javax.servlet"
+    module = 'javax.servlet-api'
+    versions = "[3.1,)"
```
charliegracie, line 5: Do you want this to apply to servlet 4.0+? If not change it to "[3.1,4.0)"

dd-java-agent/instrumentation/servlet/src/main/java/datadog/trace/instrumentation/servlet/http/ServletInputStreamWrapper.java
```diff
@@ -1,18 +1,71 @@
...
+import static net.bytebuddy.matcher.ElementMatchers.named;
+import static net.bytebuddy.matcher.ElementMatchers.takesArguments;
+
 import datadog.trace.api.http.StoredByteBody;
 import java.io.IOException;
+import java.lang.invoke.MethodHandle;
+import java.lang.invoke.MethodHandles;
+import java.lang.invoke.MethodType;
 import javax.servlet.ServletInputStream;
+import net.bytebuddy.ByteBuddy;
```
devinsba, line 12: I don't think this is desirable, this is going to cause ByteBuddy classes to be loaded onto any classloader that this helper class gets loaded onto. I wonder if this transformer can instead live in a Instrumenter class in some way

bantonsson: Yeah, pulling in `ByteBuddy` here is not something that we do. I would just pull out 3 method handles for the 3 methods in question, and implement the methods in the wrapper (without the `@Overrides` annotation) and then only call the method handle if it's not `null`.

bantonsson (commented):
Looks good as soon as the `muzzle` thing is fixed.

dd-java-agent/instrumentation/servlet/src/main/java/datadog/trace/instrumentation/servlet/http/ServletInputStreamWrapper.java
cataphract: Unless I'm missing something, I can't implement the methods because one of those takes a parameter of a type that only exists on servlet 3.1.

Since @devinsba 's comment yesterday I tried conditionally instrumenting the wrapper class, but it turned out it doesn't work because it would change the structure of the wrapper (plus I needed a new exception in the ignore filter).

What I'm going for in attempt number 4 is to try to transform the wrapper before it is injected...

bantonsson: My bad 🤦‍♂️

cataphract: I've pushed a new implementation.

ValentinZakharov (approved):

PR #3311, opened 2022-01-13
Add InstrumentationGateway callback for Http response headers

bantonsson:
# What Does This Do

This adds an InstrumentationGateway callback for Http response headers. It does not connect that callback to the AppSec code.

# Motivation

Attack rules for AppSec need to be able to check the Http response headers as well

# Additional Notes

Since there is a overlap in the request and response header processing for the `AgentPropagation.ContextVisitor<>` implementations for most frameworks, the previous `RequestExtractAdapter` classes where refactored and renamed `ExtractAdapter`, and even though the diff isn't that big, the files now show up as _deleted_ and _added_.

The other change is adding response headers to all `HttpServerTest` subclasses for all the different frameworks, which sometimes lead to indentation changes in some DSLs, which also look huge, even though only a few lines were added.


dd-java-agent/agent-bootstrap/src/main/java/datadog/trace/bootstrap/instrumentation/decorator/HttpServerDecorator.java
```diff
@@ -28,7 +28,7 @@
 import org.slf4j.Logger;
 import org.slf4j.LoggerFactory;
 
-public abstract class HttpServerDecorator<REQUEST, CONNECTION, RESPONSE, CARRIER>
+public abstract class HttpServerDecorator<REQUEST, CONNECTION, RESPONSE, REQUEST_CARRIER>
```
ValentinZakharov, line 31: Does it make sense to correct names? I feels a bit confused, because we have `REQUEST` and `RESPONSE` but they are unrelated to each other?
```suggestion
public abstract class HttpServerDecorator<REQUEST, CONNECTION, RESPONSE_CARRIER, REQUEST_CARRIER>
```

bantonsson: That is not correct, since `REQUEST, CONNECTION, RESPONSE` are what the are named, but the `CARRIER` is the type that has the http **_request_** headers on it, which sometimes is different from the `REQUEST`. That is why I tried to make it clearer by naming it `REQUEST_CARRIER`.

charliegracie (approved):

dd-java-agent/agent-bootstrap/src/main/java/datadog/trace/bootstrap/instrumentation/decorator/HttpServerDecorator.java
```diff
@@ -189,17 +191,7 @@ public AgentSpan onResponse(final AgentSpan span, final RESPONSE response) {
...
-        RequestContext<Object> ctx = span.getRequestContext();
-        if (ctx != null) {
-          BiFunction<RequestContext<Object>, Integer, Flow<Void>> addrCallback =
-              cbp.getCallback(EVENTS.responseStarted());
-          if (null != addrCallback) {
-            addrCallback.apply(ctx, status);
-          }
-        }
-      }
+      callIGCallbackResponseAndHeaders(span, response, status);
```
ygree, line 194: Nice to have it in a self descriptive method

ygree (approved):
Looks good! Thanks for additional notes on the PR.

PR #3319, opened 2022-01-17
Small fixes to logging

charliegracie:
# What Does This Do
Small cleanup of some of our logging code

# Motivation
Do not create objects for no reason and make the code more readable

# Additional Notes
I had a much larger change which wrapped all logging with the appropriate `isXXXEnabled()` but it seemed like overkill at this point. With this change there are no String concatenations left and all of the `log.XXX(String, Object ...)` in hot paths are properly wrapped.

### DIFF
--- a/dd-java-agent/agent-bootstrap/src/main/java/datadog/trace/bootstrap/Agent.java
+++ b/dd-java-agent/agent-bootstrap/src/main/java/datadog/trace/bootstrap/Agent.java
@@ -514,7 +514,9 @@ public class Agent {
   /** Enable JMX based system access provider once it is safe to touch JMX */
   private static synchronized void initializeJmxSystemAccessProvider(
       final ClassLoader classLoader) {
-    log.debug("Initializing JMX system access provider for " + classLoader.toString());
+    if (log.isDebugEnabled()) {
+      log.debug("Initializing JMX system access provider for " + classLoader.toString());
+    }
     try {
       final Class<?> tracerInstallerClass =
           classLoader.loadClass("datadog.trace.core.util.SystemAccess");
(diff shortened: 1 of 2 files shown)


dd-java-agent/agent-bootstrap/src/main/java/datadog/trace/bootstrap/Agent.java
```diff
@@ -514,7 +514,9 @@ private static synchronized void registerDeadlockDetectionEvent(URL bootstrapUrl
   /** Enable JMX based system access provider once it is safe to touch JMX */
   private static synchronized void initializeJmxSystemAccessProvider(
       final ClassLoader classLoader) {
-    log.debug("Initializing JMX system access provider for " + classLoader.toString());
+    if (log.isDebugEnabled()) {
+      log.debug("Initializing JMX system access provider for " + classLoader.toString());
+    }
```
devinsba, line 519: Would there be any value to switching to the following like is done in the other file?

```suggestion
    log.debug("Initializing JMX system access provider for {}", classLoader);
```

devinsba (approved):

dd-java-agent/agent-bootstrap/src/main/java/datadog/trace/bootstrap/Agent.java
charliegracie: It would make the code feel more like the other code which could be a benefit. I would still keep it wrapped with the if statement to save the call before the code get JIT compiled. In theory we should wrap all of the the logging calls to remove the call from interpreted calls when Debug is not enabled.

ygree (approved):
LGTM

dd-java-agent/agent-bootstrap/src/main/java/datadog/trace/bootstrap/Agent.java
charliegracie: I will just leave it as is

PR #3322, opened 2022-01-18
Remove uses of Objects.hash()

charliegracie:
# What Does This Do
Remove all instances of `Objects.hash()` since it is poorly optimized by the JVM. 

# Motivation
Calls to `Objects.hash()` will allocate an array to hold the parameters and all primitive parameters will be autoboxed. There are some caches for primitives in the JVM but they are limited and common values for things like a web server port will not be in the cache. This causes unneeded object allocations. 

# Additional Notes
I will add some JMH results comparing the results for an inline version of `Objects.hash()` (like this change) and calling `Objects.hash()` for different sets of parameters.

### DIFF
--- a/dd-java-agent/appsec/src/main/java/com/datadog/appsec/report/AppSecEventWrapper.java
+++ b/dd-java-agent/appsec/src/main/java/com/datadog/appsec/report/AppSecEventWrapper.java
@@ -27,12 +27,12 @@ public class AppSecEventWrapper {
     if (this == o) return true;
     if (o == null || getClass() != o.getClass()) return false;
     AppSecEventWrapper that = (AppSecEventWrapper) o;
-    return Objects.equals(triggers, that.triggers) && Objects.equals(json, that.json);
+    return Objects.equals(triggers, that.triggers);
   }
 
   @Override
   public int hashCode() {
-    return Objects.hash(triggers, json);
+    return triggers.hashCode();
   }
 
   @Override
--- a/dd-trace-core/jfr-openjdk/src/main/java/datadog/trace/core/jfr/openjdk/JFRCheckpointer.java
+++ b/dd-trace-core/jfr-openjdk/src/main/java/datadog/trace/core/jfr/openjdk/JFRCheckpointer.java
@@ -57,7 +57,14 @@ public class JFRCheckpointer implements Checkpointer, ProfilingListener<Profilin
 
     @Override
     public int hashCode() {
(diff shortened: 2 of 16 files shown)


dd-trace-core/jfr-openjdk/src/main/java/datadog/trace/core/jfr/openjdk/JFRCheckpointer.java
```diff
@@ -57,7 +57,14 @@ public boolean equals(Object o) {
 
     @Override
     public int hashCode() {
-      return Objects.hash(windowSize, samplesPerWindow, averageLookback, budgetLookback);
+      // Removed Objects.hash() since it will allocate an array for its parameters
```
ygree, line 60: Nice observation!

dd-java-agent/appsec/src/main/java/com/datadog/appsec/report/AppSecEventWrapper.java
```diff
@@ -32,7 +32,12 @@ public boolean equals(Object o) {
 
   @Override
   public int hashCode() {
-    return Objects.hash(triggers, json);
+    // TODO if toString() is called after hashcode the value for hashCode will change!!!
```
charliegracie, line 35: I am hoping this can just be removed along with the `.equals()` since if they are used before and then after a call to `toString()` they will provide different results

bantonsson: Yes, this is not used, and can be removed.

bantonsson (approved):

dd-java-agent/appsec/src/main/java/com/datadog/appsec/report/AppSecEventWrapper.java
charliegracie: thanks

charliegracie: I failed tests when I removed it so I modified my change to just add a comment describing that hashCode is technically incorrect

bantonsson:
You could just remove the json string from the equals and hashCode.

dd-trace-core/jfr-openjdk/src/main/java/datadog/trace/core/jfr/openjdk/JFRCheckpointer.java
charliegracie: thanks

PR #3325, opened 2022-01-19
Remove jcenter from projects where it is possible

bantonsson:
# What Does This Do

Removes the JCenter repository in all places where it's possible

# Motivation

Make our JCenter dependencies explicit and minimize impact of JCenter service disruptions

# Additional Notes


bantonsson:
This passed with a clean build without any caches.

mcculls (approved):
+1 ... thanks for tracking down which dependencies we get from jcenter and listing them in the comments

devinsba (approved):

PR #3327, opened 2022-01-19
[AppSec] Provide actor.ip in the trace

cataphract:
# What Does This Do

It determines the value of the `actor.ip` tag and includes it (when there is an appsec event).

# Motivation

This is currently done on the backend, but will need to be moved to the tracer so that blocking decisions can be made on the same request.

# Additional Notes

See "Client IP addresses resolution" on the wiki and "AppSec Span tags RFC" on docs.

### DIFF
--- a/dd-java-agent/appsec/src/main/java/com/datadog/appsec/AppSecSystem.java
+++ b/dd-java-agent/appsec/src/main/java/com/datadog/appsec/AppSecSystem.java
@@ -59,7 +59,8 @@ public class AppSecSystem {
 
     EventDispatcher eventDispatcher = new EventDispatcher();
     sco.createRemaining(config);
-    GatewayBridge gatewayBridge = new GatewayBridge(gw, eventDispatcher);
+    GatewayBridge gatewayBridge =
+        new GatewayBridge(gw, eventDispatcher, config.getAppSecIpAddrHeader());
 
     loadModules(eventDispatcher);
     gatewayBridge.init();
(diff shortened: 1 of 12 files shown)


dd-java-agent/appsec/src/main/java/com/datadog/appsec/AppSecSystem.java
```diff
@@ -59,7 +59,8 @@ private static void doStart(SubscriptionService gw, SharedCommunicationObjects s
 
     EventDispatcher eventDispatcher = new EventDispatcher();
     sco.createRemaining(config);
-    GatewayBridge gatewayBridge = new GatewayBridge(gw, eventDispatcher);
+    GatewayBridge gatewayBridge =
+        new GatewayBridge(gw, eventDispatcher, config.getAppSecIpAddrHeader());
```
ValentinZakharov, line 63: You pass `IpAddrHeader` as method argument, but we can get value directly in `GatewayBridge` from config. Is there any specific reason for that, except testing purpose?

dd-java-agent/appsec/src/main/java/com/datadog/appsec/gateway/GatewayBridge.java
```diff
@@ -97,23 +102,29 @@ public void init() {
...
               traceSeg.setTagTop("appsec.event", true);
-              traceSeg.setTagTop("network.client.ip", ctx.getPeerAddress());
+              String peerAddress = ctx.getPeerAddress();
+              traceSeg.setTagTop("network.client.ip", peerAddress);
+
+              Map<String, List<String>> requestHeaders = ctx.getRequestHeaders();
+              InetAddress inferredAddr =
+                  ClientIpAddressResolver.doResolve(this.ipAddrHeader, requestHeaders);
+              traceSeg.setTagTop(
+                  "actor.ip", inferredAddr != null ? inferredAddr.getHostAddress() : peerAddress);
```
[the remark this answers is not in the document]
cataphract: Yes, I think you're right.

dd-java-agent/appsec/src/main/java/com/datadog/appsec/AppSecSystem.java
cataphract: That's it.

ValentinZakharov (approved):

bantonsson (approved):

PR #3328, opened 2022-01-20
Really remove jcenter from the default repositories

bantonsson:
# What Does This Do

Really removes jcenter from the default repositories, which was missed in #3325 due to a bad merge.


bantonsson:
Clean build without cache worked fine.

hisener (approved):

mcculls (approved):

PR #3330, opened 2022-01-21
Some gradle cleanup in preparation for 7.x

bantonsson:
# What Does This Do

* Changes all compile/runtime configurations to implementation/runtimeOnly
* Switches to the new supported play framework plugin

# Motivation

These constructs are no longer supported on gradle 7.x and would fail the build

# Additional Notes

The biggest blocker now is the muzzle `instrumentation` plugin that is beyond my current understanding of gradle, as well as the `:dd-java-agent:appsec:weblog:weblog-spring-app` that uses old versions of Spring and the Spring Boot plugin.


charliegracie (approved):

charliegracie:
Sigh was trying to ask a question and missed clicked and approved :(

devinsba (approved):

mcculls (approved):
👍

PR #3333, opened 2022-01-24
[AppSec] RateLimit reports

cataphract:
# What Does This Do

It limits the amount of times/sec that appsec will report events. This defaults to 100 reports/s

# Additional Notes

See Google docs "RFC: public beta AppSec rate limiter"


bantonsson:
Is there a particular reason that the existing [SimpleRateLimiter](https://github.com/DataDog/dd-trace-java/blob/master/dd-trace-core/src/main/java/datadog/trace/core/util/SimpleRateLimiter.java) wasn't used?

cataphract:
> Is there a particular reason that the existing [SimpleRateLimiter](https://github.com/DataDog/dd-trace-java/blob/master/dd-trace-core/src/main/java/datadog/trace/core/util/SimpleRateLimiter.java) wasn't used?

I was not aware of its existence. Maybe it does the job, as the RFC mentions a moving sum but gives leeway to simplify if technical reasons call for it. I'd still prefer to keep this implementation, which approximates a moving sum, because it's more likely to avoid flaky systems tests. Besides the lack of "smoothing", as it calls it, `SimpleRateLimiter` also doesn't seem completely correct because `lastRefillTime` can be read and `tokens` read and decremented while another thread is between the update to `lastRefillTime` and to `tokens`.

dd-java-agent/appsec/src/test/groovy/com/datadog/appsec/gateway/RateLimiterSpecification.groovy
```diff
@@ -0,0 +1,155 @@
+package com.datadog.appsec.gateway
+
+import spock.lang.Specification
+
+class RateLimiterSpecification extends Specification {
+
+  interface NanoProvider {
```
bantonsson, line 7: This concept is already available and named `TimeSource`. It could have been used in the `RateLimiter` directly.

dd-java-agent/appsec/src/main/java/com/datadog/appsec/gateway/RateLimiter.java
```diff
@@ -0,0 +1,68 @@
...
+  }
+
+  public final boolean isThrottled() {
+    long curSec = getNanoTime();
+    long storedState;
+    long newState;
+
+    do {
+      storedState = this.state.get();
+      int storedCurCount = (int) (storedState & 0xFFFFFF);
```
bantonsson, line 30: Could we name these masks and constants?

dd-java-agent/appsec/src/main/java/com/datadog/appsec/gateway/RateLimiter.java
```diff
@@ -0,0 +1,68 @@
+package com.datadog.appsec.gateway;
+
+import java.util.concurrent.atomic.AtomicLong;
+
+public class RateLimiter {
```
bantonsson, line 5: Could you please add some comments to the class and code explaining the _smoothing_ over the interval, and any assumptions about what would happen if there was no calls for the `2^16 (+-1)` seconds limit time imposed by the storage format.

dd-java-agent/appsec/src/main/java/com/datadog/appsec/gateway/RateLimiter.java
```diff
@@ -0,0 +1,68 @@
...
+
+    do {
+      storedState = this.state.get();
+      int storedCurCount = (int) (storedState & 0xFFFFFF);
+      int storedPrevCount = (int) ((storedState & 0xFFFFFF000000L) >> 24);
+      int storedCurSec16 = (int) (storedState >>> 48);
+
+      int curSec16 = curSecond16bit(curSec);
+      int diff = (curSec16 - storedCurSec16) & 0xFFFF;
+      if (diff == 0 || diff == 0xFFFF /* -1 in ring Z_{0x10000} */) {
```
bantonsson, line 36: Why is `-1`, aka `0xFFFF`, treated specially here? I don't get the comment. I thought that either you're in this second, `diff == 0`, or you're in the previous second, `diff == 1`, or you simple don't care and reset.

bantonsson:
I agree that there are _holes_ in the `SimpleRateLimiter`, including the fact that recursive calls to `tryAcquire` are not bounded in stack if the scheduling gods are against us.

It's fine to have a _better_ implementation, but I would like it to have documentation.

dd-java-agent/appsec/src/main/java/com/datadog/appsec/gateway/RateLimiter.java
cataphract: It's for the case when thread T1 is at the end of the second when it reads the time. Then it's possible that:
1) another thread T2 reads the beginning of the next second and writes the state before T1 has a chance to read it, or, more likely,
2) T2 wins the race the write the state; CAS will then fail on T1. Because the clock is read only once, no matter the amount of failures, T1 will always be behind on the "current second".

bantonsson: Oh, I missed that the clock is only read once. So in essence, if the thread is starved (contention and scheduling) so the gap between `curSec16` and `storedCurSec16` becomes  larger than `-1` i.e. `-2`, then we will move the `storedCurrSec` backwards and reset the `count`?

cataphract: Yes, in theory that's possible, if somehow the method 1) took more than 1 second to run and 2) the limit was never reached during the several attempts (because if the limit is reached there is no update)

bantonsson: I don't see this as a theoretical issue. On an under provisioned kubernetes pod, which seems to be way more common than one would hope, one thread of several not being allowed to run for upwards of hundreds of ms, isn't that strange. That combined with some contention during the updates, and you have your 1+ second. Why can't you just reread the clock to make it less likely?

dd-java-agent/appsec/src/main/java/com/datadog/appsec/gateway/RateLimiter.java
cataphract: I've added a comment on the class.

dd-java-agent/appsec/src/main/java/com/datadog/appsec/gateway/RateLimiter.java
cataphract: The motivation was that reading the clock is a relatively expensive operation even if vDSO can prevent a syscall. WDYT instead giving 8 extra bits to the timestamp, subtract the time read at construction, and just assume it doesn't wrap around?

bantonsson: I understand the motivation for not reading the clock more than necessary, but maybe it's a bit premature? So regardless of how many bits you use for the clock, what should you do if it's taken more than 2 seconds trying to update the atomic value?

cataphract: OK, I'm rereading the clock now.

bantonsson: I agree that the expanded time is enough to make the wrap around collision very unlikely. Let's just leave things as is for now.

bantonsson (approved):

PR #3335, opened 2022-01-25
[AppSec] Fixed double status assignment

ValentinZakharov:
# What Does This Do
Removed code duplication

# Motivation
Customers discovered in logs warning message `Illegal attempt to replace context value for Address{key='server.response.status'}`

# Additional Notes


bantonsson (approved):

cataphract (approved):
I think there's a fair risk of this class of bug reoccurring if `AppSecRequestContext` is not split as mentioned in the comment at its top. Perhaps we should prioritize it.

PR #3341, opened 2022-01-26
[Appsec] request.path_params for spring webmvc

cataphract:
# What Does This Do

Collects webmvc path and matrix template parameters and publishes it to the instrumentation gateway. This will trigger the WAF (once more...); unfortunately I don't think there is any way to consolidate the WAF runs.


dd-java-agent/instrumentation/spring-webmvc-3.1/src/main/java/datadog/trace/instrumentation/springweb/TemplateVariablesUrlHandlerInstrumentation.java
```diff
@@ -0,0 +1,80 @@
...
+import datadog.trace.bootstrap.instrumentation.api.AgentTracer;
+import java.util.Map;
+import javax.servlet.http.HttpServletRequest;
+import net.bytebuddy.asm.Advice;
+import net.bytebuddy.description.type.TypeDescription;
+import net.bytebuddy.matcher.ElementMatcher;
+
+/** Obtain template and matrix variables for AbstractUrlHandlerMapping */
+@AutoService(Instrumenter.class)
+public class TemplateVariablesUrlHandlerInstrumentation extends Instrumenter.AppSec {
```
devinsba, line 27: I think this could be accomplished without two new instrumentation classes. We have `HandlerMappingResourceNameFilter` that we add to capture the spring path match before any other processing is done on a request. I think you could leverage that to get the variables out that you need

devinsba: You might be seeing that `PathMatchingHttpServletRequestWrapper` blocks the modifications. You could add your two fields there also in order to capture them. This is in place in order for us to capture the matcher state before auth filters etc.

cataphract: OK, this strategy works for `RequestMappingInfoHandlerMapping` (see last commit), but not for the attribute injection done by `UriTemplateVariablesHandlerInterceptor` (done by `AbstractUrlHandlerMapping` and its subclasses like `SimpleUrlHandlerMapping`).

cataphract: I'm reverting to the original strategy, because `PathMatchingHttpServletRequestWrapper` does not work reliably. On the smoke tests, which use spring web 5.3.13, `HandlerMappingResourceNameFilter` throws inside `findMapping` because its saved mapping inside `handlerMappings` (of type `org.springframework.web.servlet.mvc.method.annotation.RequestMappingHandlerMapping`) cannot work without the attribute `org.springframework.web.util.ServletRequestPathUtils.PATH` set, and `DispatcherServlet` sets this attribute further downstream from our filter.

ValentinZakharov (approved):

PR #3342, opened 2022-01-27
Make SimpleRateLimiter more robust

bantonsson:
# What Does This Do

This moves the count and time of the `SimpleRateLimiter` into the same atomic long to avoid a race between updating the `tokens` and `lastRefillTime` in the previous implementation (thanks @cataphract for noticing).

# Motivation

There was a race between reading/updating `tokens` and `lastRefillTime` since they were in separate atomics. The code was also doing a recursive call which, although highly unlikely, could lead to large stack depths during heavy contention.

# Additional Notes

The implementation assumes that an uptime of more than 68 years is not going to happen. The rate limit is also changed from `long` to `int` since the only place that uses it has an `int` limit.


dd-trace-core/src/main/java/datadog/trace/core/util/SimpleRateLimiter.java
```diff
@@ -2,50 +2,76 @@
...
+  protected SimpleRateLimiter(int rate, TimeSource timeSource) {
     this.timeSource = timeSource;
-
-    capacity = Math.max(1, rate);
-
-    tokens = new AtomicLong(capacity);
-
-    lastRefillTime = new AtomicLong(timeSource.getNanoTime());
+    this.startNanos = timeSource.getNanoTime();
+    capacity = Math.min(Math.max(1, rate), Integer.MAX_VALUE - 1);
```
cataphract, line 25: Not sure why you need `capacity` to be a long if you're limiting it to 0x7FFFFFFE. You could use also the full 32-bit, with some changes below to avoid wrap arounds and sign extensions.

dd-trace-core/src/main/java/datadog/trace/core/util/SimpleRateLimiter.java
```diff
@@ -2,50 +2,76 @@
...
+    return (int) (timeAndCount >> 32);
+  }
+
+  private static int getStoredCount(long timeAndCount) {
+    return (int) (timeAndCount & Integer.MAX_VALUE);
+  }
 
-    return MathUtils.boundedDecrement(tokens, 0);
+  private static long combineSecondsAndCount(int seconds, int count) {
+    return ((long) (seconds & Integer.MAX_VALUE)) << 32 | (count & Integer.MAX_VALUE);
```
cataphract, line 75: `count & Integer.MAX_VALUE` does nothing, since `count` is necessarily a positive int.

cataphract (approved):

mcculls (approved):
neat

dd-trace-core/src/main/java/datadog/trace/core/util/SimpleRateLimiter.java
bantonsson: You're right, `capacity` doesn't need to be a `long`. I'll change it, and also use the full `31` bits.

dd-trace-core/src/main/java/datadog/trace/core/util/SimpleRateLimiter.java
bantonsson: Sure, I'll remove it.

PR #3353, opened 2022-01-31
[AppSec] Collect and report pre-defined HTTP response headers

ValentinZakharov:
# What Does This Do
Implemented code to collect and report pre-defined HTTP response headers

# Motivation

# Additional Notes

### DIFF
--- a/dd-java-agent/appsec/src/main/java/com/datadog/appsec/gateway/AppSecRequestContext.java
+++ b/dd-java-agent/appsec/src/main/java/com/datadog/appsec/gateway/AppSecRequestContext.java
@@ -54,7 +54,7 @@ public class AppSecRequestContext implements DataBundle, Closeable {
   private final Map<String, List<String>> requestHeaders = new LinkedHashMap<>();
   private final Map<String, List<String>> responseHeaders = new LinkedHashMap<>();
   private List<StringKVPair> collectedCookies = new ArrayList<>(4);
-  private boolean finishedHeaders;
+  private boolean finishedRequestHeaders;
   private String peerAddress;
   private int peerPort;
 
@@ -110,6 +110,7 @@ public class AppSecRequestContext implements DataBundle, Closeable {
   }
 
   @Override
+  @SuppressWarnings("unchecked")
   public <T> T get(Address<T> addr) {
     return (T) persistentData.get(addr);
   }
@@ -149,8 +150,8 @@ public class AppSecRequestContext implements DataBundle, Closeable {
   }
 
   void addRequestHeader(String name, String value) {
-    if (finishedHeaders) {
-      throw new IllegalStateException("Headers were said to be finished before");
+    if (finishedRequestHeaders) {
+      throw new IllegalStateException("Request headers were said to be finished before");
     }
 
     if (name == null || value == null) {
@@ -160,6 +161,18 @@ public class AppSecRequestContext implements DataBundle, Closeable {
     List<String> strings =
         requestHeaders.computeIfAbsent(name.toLowerCase(), h -> new ArrayList<>(1));
     strings.add(value);
+  }
+
+  void finishRequestHeaders() {
+    this.finishedRequestHeaders = true;
+  }
+
+  boolean isFinishedRequestHeaders() {
+    return finishedRequestHeaders;
+  }
+
+  Map<String, List<String>> getRequestHeaders() {
+    return requestHeaders;
   }
 
   void addResponseHeader(String name, String value) {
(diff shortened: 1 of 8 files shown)


cataphract (commented):
Looks good, but it's missing tests: unit, integration, and perhaps also smoke if there are rules that use the response headers.

dd-java-agent/appsec/src/main/java/com/datadog/appsec/gateway/AppSecRequestContext.java
```diff
@@ -162,39 +164,51 @@ void addRequestHeader(String name, String value) {
...
+    strings.add(value);
+  }
+
   Map<String, List<String>> getResponseHeaders() {
     return responseHeaders;
   }
 
+  void addCookie(StringKVPair cookie) {
+    if (finishedRequestHeaders) {
+      throw new IllegalStateException("Headers were said to be finished before");
```
bantonsson, line 207: Use _request headers_ in the log message here as well?

dd-java-agent/appsec/src/test/groovy/com/datadog/appsec/gateway/GatewayBridgeSpecification.groovy
```diff
@@ -148,10 +150,10 @@ class GatewayBridgeSpecification extends DDSpecification {
...
   void 'bridge can collect headers'() {
     when:
-    headerCB.accept(ctx, 'header1', 'value 1.1')
-    headerCB.accept(ctx, 'header1', 'value 1.2')
-    headerCB.accept(ctx, 'Header1', 'value 1.3')
-    headerCB.accept(ctx, 'header2', 'value 2')
+    reqHeaderCB.accept(ctx, 'header1', 'value 1.1')
+    reqHeaderCB.accept(ctx, 'header1', 'value 1.2')
+    reqHeaderCB.accept(ctx, 'Header1', 'value 1.3')
+    reqHeaderCB.accept(ctx, 'header2', 'value 2')
```
bantonsson, line 156: Shouldn't there be a test for the response headers as well?

dd-java-agent/appsec/src/main/java/com/datadog/appsec/gateway/AppSecRequestContext.java
```diff
@@ -162,39 +164,51 @@ void addRequestHeader(String name, String value) {
...
+    return finishedResponseHeaders;
   }
 
   Map<String, List<String>> getRequestHeaders() {
     return requestHeaders;
   }
 
+  void addResponseHeader(String name, String value) {
+    if (finishedResponseHeaders) {
+      throw new IllegalStateException("Resopnse headers were said to be finished before");
```
bantonsson, line 189: Typo in `Resopnse`

bantonsson (commented):
Looks good with some minor fixes.

dd-java-agent/appsec/src/main/java/com/datadog/appsec/gateway/AppSecRequestContext.java
ValentinZakharov: Fixed

dd-java-agent/appsec/src/test/groovy/com/datadog/appsec/gateway/GatewayBridgeSpecification.groovy
ValentinZakharov: Added tests

dd-java-agent/appsec/src/main/java/com/datadog/appsec/gateway/AppSecRequestContext.java
ValentinZakharov: Fixed

bantonsson: It feels weird that the fix to this was to remove the check... 🤔

cataphract (approved):
LGTM, the only thing is that it's missing a unit test that covers `GatewayBridge.java`, lines 137-143.

dd-java-agent/appsec/src/main/java/com/datadog/appsec/gateway/AppSecRequestContext.java
ValentinZakharov: When I was thinking about AppSec logic flow, I realized, that we don't need to send response headers to WAF for analysis, so we don't need to check when headers are committed and ready for submitting. Right now, the only case is to report collected headers when request ended and I assume, that after request has been ended, there will be no new response headers anymore. Please correct me if I'm wrong, but it seems, there is no sense to "finalize" response headers collection.

cataphract: It would be preferable to include support for publishing `KnownAddress.RESPONSE_HEADERS_NO_COOKIES` when the response headers are finished. We already have a mechanism to skip the subscription of an IG event in case there are no subscribers to the corresponding address.

ValentinZakharov: Added publishing response headers

bantonsson (approved):

PR #3355, opened 2022-02-01
Instrumentation JFR events

ygree:
# What Does This Do

Adds JFR class transformation events to measure instrumentation.

# Motivation

This provides information about instrumentation time per class.

# Additional Notes

It works for Java 8+:

Screen Shot 2022-02-01 at 09 42 19


dd-java-agent/agent-tooling/src/main/java8/datadog/trace/agent/tooling/bytebuddy/ClassTransformationEvent.java
```diff
@@ -0,0 +1,25 @@
+package datadog.trace.agent.tooling.bytebuddy;
+
+import jdk.jfr.Category;
+import jdk.jfr.Event;
+import jdk.jfr.Label;
+import jdk.jfr.Name;
+
+@Category({"Data Dog Tracer", "Class Transformation"})
```
jbachorik, line 8:
```suggestion
@Category({"Datadog", "Tracer", "Class Transformation"})
```

We already have a bunch of events in `Datadog` category so it would be nice to have this one there as well.

dd-java-agent/agent-tooling/src/main/java11/datadog/trace/agent/tooling/bytebuddy/DDJava9ClassFileTransformer.java
```diff
@@ -59,12 +59,22 @@ public DDJava9ClassFileTransformer(final ResettableClassFileTransformer classFil
...
+      event.begin();
+      byte[] result =
+          classFileTransformer.transform(
+              module,
+              classLoader,
+              internalClassName,
+              classBeingRedefined,
+              protectionDomain,
+              classFileBuffer);
+      event.setTransformed(result != null);
```
jbachorik, line 74: Would it make sense to track also the bytecode size? Before and after transformation?

ygree: That's a good idea! Thanks!

dd-java-agent/agent-tooling/src/main/java8/datadog/trace/agent/tooling/bytebuddy/ClassTransformationEvent.java
ygree: Good point! Thanks!

dd-java-agent/agent-tooling/src/main/java8/datadog/trace/agent/tooling/bytebuddy/ClassTransformationEvent.java
```diff
@@ -0,0 +1,42 @@
...
+
+  protected void afterClassTransformation(byte[] classfileBuffer) {
+    if (classfileBuffer == null) {
+      classSizeAfterTransformation = classSizeBeforeTransformation;
+      transformed = false;
+    } else {
+      classSizeAfterTransformation = classfileBuffer.length;
+      transformed = classSizeBeforeTransformation != classSizeAfterTransformation;
+    }
+    classSizeDifference = classSizeAfterTransformation - classSizeBeforeTransformation;
```
jbachorik, line 40: You can wrap this in `if (shouldCommit()) { ...}` to avoid paying the cost if the event is eg. disabled in the settings.

ygree: Good point! Thanks!

dd-java-agent/agent-tooling/src/main/java/datadog/trace/agent/tooling/AgentTooling.java
```diff
@@ -57,6 +57,19 @@ private static TransformerDecorator loadTranformerDecorator() {
       } catch (Throwable e) {
         log.warn("Problem loading Java9 Module support, falling back to legacy transformer", e);
       }
+    } else if (Platform.isJavaVersionAtLeast(8)) {
```
jbachorik, line 60: Probably not something very common but it would be good to be on the safe side and check for `Platform.isJavaVersionAtLeast(8, 272, 0)` to make sure the JFR classes are available.

We could avoid raising the exception in that case.

ygree: Added. Thanks!

devinsba (approved):

mcculls (changes requested):
Just want to check - is this PR just for benchmarking or is the plan to merge it into master at some point?

If the intent is to merge it into master then I'd like there to be a feature-flag that allows us to switch back to the previous transformers without JFR event collection. This would help us check relative performance of this feature and provide a way to workaround unforeseen issues in the field.

ygree:
> Just want to check - is this PR just for benchmarking or is the plan to merge it into master at some point?
> 
> If the intent is to merge it into master then I'd like there to be a feature-flag that allows us to switch back to the previous transformers without JFR event collection. This would help us check relative performance of this feature and provide a way to workaround unforeseen issues in the field.

Thanks, Stuart! I agree. The plan is to add a feature flag or even two: one for transformation and another for matchers (they are more expensive because of the number of matching events could be millions). I'll mark this PR as draft for now.

jbachorik:
> > Just want to check - is this PR just for benchmarking or is the plan to merge it into master at some point?
> > If the intent is to merge it into master then I'd like there to be a feature-flag that allows us to switch back to the previous transformers without JFR event collection. This would help us check relative performance of this feature and provide a way to workaround unforeseen issues in the field.
> 
> Thanks, Stuart! I agree. The plan is to add a feature flag or even two: one for transformation and another for matchers (they are more expensive because of the number of matching events could be millions). I'll mark this PR as draft for now.

Uff. Millions of events can turn out to be rather problematic as they will increase the size of the collected JFR recording significantly. That could affect our transfer and storage costs.

Would sampling of be suitable for matcher events? Something similar to what we do for exception in order to make sure we don't blow up the recording.

ygree:
> > > Just want to check - is this PR just for benchmarking or is the plan to merge it into master at some point?
> > > If the intent is to merge it into master then I'd like there to be a feature-flag that allows us to switch back to the previous transformers without JFR event collection. This would help us check relative performance of this feature and provide a way to workaround unforeseen issues in the field.
> > 
> > 
> > Thanks, Stuart! I agree. The plan is to add a feature flag or even two: one for transformation and another for matchers (they are more expensive because of the number of matching events could be millions). I'll mark this PR as draft for now.
> 
> Uff. Millions of events can turn out to be rather problematic as they will increase the size of the collected JFR recording significantly. That could affect our transfer and storage costs.
> 
> Would sampling of be suitable for matcher events? Something similar to what we do for exception in order to make sure we don't blow up the recording.

I don't think we will have instrumentation JFR events enabled by default, especially matching events that affect startup time drastically. At the moment we are using it to figure out how the startup can be optimized.

ygree:
Superseded by #3546

PR #3357, opened 2022-02-01
Some gradle cleanup in preparation for 7.x (take 2)

bantonsson:
# What Does This Do

This is take 2 of #3330

* Adds a minimal test that check that the `dd-trace-ot` dependency pulls in `dd-trace-api` as a dependency.
* Changes all compile/runtime configurations to implementation/runtimeOnly
* Switches to the new supported play framework plugin

# Motivation

These constructs are no longer supported on gradle 7.x and would fail the build

# Additional Notes

The biggest blocker now is the muzzle `instrumentation` plugin that is beyond my current understanding of gradle, as well as the `:dd-java-agent:appsec:weblog:weblog-spring-app` that uses old versions of Spring and the Spring Boot plugin.

### DIFF
--- a/gradle/publish.gradle
+++ b/gradle/publish.gradle
@@ -37,7 +37,7 @@ publishing {
             pom.withXml { xml ->
               def dependenciesNode = xml.asNode().appendNode('dependencies')
 
-              project.configurations.compile.allDependencies.each {
+              project.configurations.api.allDependencies.each {
                 if ((it instanceof ProjectDependency) || !(it instanceof SelfResolvingDependency)) {
                   def dependencyNode = dependenciesNode.appendNode('dependency')
                   dependencyNode.appendNode('groupId', it.group)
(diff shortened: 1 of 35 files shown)


gradle/publish.gradle
```diff
@@ -37,7 +37,7 @@ publishing {
             pom.withXml { xml ->
               def dependenciesNode = xml.asNode().appendNode('dependencies')
 
-              project.configurations.compile.allDependencies.each {
+              project.configurations.api.allDependencies.each {
```
bantonsson, line 40: This is the reason that the `pom` file was broken. Our own build messes it up.

ygree: Great finding!

ygree (approved):
It's great now we have this check that all transitive dependencies are available.

mcculls (approved):
LGTM - regarding the `instrumentation` plugin, I can look into making that work with Gradle 7

PR #3358, opened 2022-02-01
[AppSec] Expose tomcat/jetty parsed urlencoded and spring converted body

cataphract:
# What Does This Do

Instruments tomcat and jetty to get the result of their parsing urlencoded request bodies.
Also instruments Spring WebMVC's `HttpMessageConverters` so that when they return a map, iterable, or string, this is exposed to appsec as a converted request body. The IG callback can be called multiple times, but the appsec module only considers the first object published.

# Motivation

Instrumenting the servlet API is not option due to its mixing the query parameters (which we parse ourselves) and the body parameters.

# Additional Notes

The fact that all instrumentations in a module must rise and fall together for muzzle purposes complicates the implementation, as the ranges for the appsec instrumentation don't match those of the current instrumentations. So new modules were created specifically for appsec. Some adjustments to muzzle were required too.


dd-java-agent/instrumentation/jetty-appsec-7/src/main/java/datadog/trace/instrumentation/jetty70/RequestExtractParametersInstrumentation.java
```diff
@@ -0,0 +1,60 @@
...
+
+  public static class ExtractParametersAdvice {
+    @Advice.OnMethodEnter(suppress = Throwable.class)
+    static int before() {
+      return CallDepthThreadLocalMap.incrementCallDepth(Request.class);
+    }
+
+    @Advice.OnMethodExit(suppress = Throwable.class)
+    static void after(@Advice.Enter final int depth) {
+      if (depth > 0) {
```
bantonsson, line 54: So won't this risk messing up the counts, in the case of exceptions being thrown in the surrounded method? I think you need to add an `onThrowable = Throwable.class` to ensure it get's run every time.

dd-java-agent/agent-tooling/src/main/java/datadog/trace/agent/tooling/muzzle/IReferenceMatcher.java
```diff
@@ -0,0 +1,41 @@
+package datadog.trace.agent.tooling.muzzle;
+
+import java.util.ArrayList;
+import java.util.List;
+
+public interface IReferenceMatcher {
```
bantonsson, line 6: I've been thinking about this one a fair bit, and I think you're right that this is the least invasive way to get  the extended manual matching ability.

bantonsson (approved):
This is quite a big one, but digestible commit by commit. Only minor comments, and one question about exceptions.

ValentinZakharov (approved):

PR #3371, opened 2022-02-07
Skip scala.collection and akka/akka-http classes

ygree:
# What Does This Do
'scala/collection' saves about 0.5s skipping ~470 classes
'akka', 'akka.http', 'akka.stream' saves about 0.4s skipping ~1200 classes

# Motivation
We don't instrument Scala collection classes and only a few Akka classes.

# Additional Notes

Specific JFR events allowing this type of analysis implemented in [this PR](https://github.com/DataDog/dd-trace-java/pull/3355/).

`no-optimization` - base line
`skip-akka` - skips Akka classes that are not instrumented earlier
`skip-scala` - skips Scala Collection classes
`skip-scala-akka` - skips both Akka and Scala Collection classes
`without-has-super` - excludes possibly expensive instrumentation that make use of `hasSuperType/Class`: jakarta-rs-annotations, jax-rs-annotations, rediscala, servlet/request, trace-annotation

image

image

image

image


devinsba (approved):

devinsba (approved):

charliegracie (approved):
I am curious if you have any deeper insights on the wins. From the PR description:
```
'scala/collection' saves about 0.5s skipping ~470 classes
'akka', 'akka.http', 'akka.stream' saves about 0.4s skipping ~1200 classes
```

Skipping the `scala.collection.*` classes saves more than all of the `akka` classes combined even though it is ~1/3 of the number of classes. Are the collection classes generally larger (bytecode size)? Do they have deeper class chain hierarchies? Is there something else different about them?

This is a great change I am just curious if there is something we can take away about the cost of matching based on characteristics of the classes.

bantonsson:
That is a very good point  @charliegracie. My gut feeling is that the scala colllections all implement quite a few interfaces that extende interfaces that extend interfaces... But it would be good to do some proper analysis of it.

bantonsson:
Yeah, it was just as fun as I remembered. The class `scala.collection.immutable.List` in Scala 2.12 has a hierarchy of 28 distinct classes and interfaces in the end, but they are reached via several classed and interfaces implementing the same interfaces, which is a hierarchy of 228 classes and interfaces. I think it's even worse in Scala 2.11.

devinsba:
@charliegracie that makes me wonder which matchers were executing for those classes. The has interface matchers, has super type, and has annotation matchers we know are bad for performance. I wonder if we should add shortcuts to those matchers in the appropriate place to ignore classes in the stdlibs and libraries that definitely won't have them. IE: `org.springframework` classes will most likely never have JAX-RS annotations added to them.

ygree:
> I am curious if you have any deeper insights on the wins. From the PR description:
> 
> ```
> 'scala/collection' saves about 0.5s skipping ~470 classes
> 'akka', 'akka.http', 'akka.stream' saves about 0.4s skipping ~1200 classes
> ```
> 
> Skipping the `scala.collection.*` classes saves more than all of the `akka` classes combined even though it is ~1/3 of the number of classes. Are the collection classes generally larger (bytecode size)? Do they have deeper class chain hierarchies? Is there something else different about them?
> 
> This is a great change I am just curious if there is something we can take away about the cost of matching based on characteristics of the classes.

Currently it's only possible to analyze total instrumentation time per class. We will need to add more JFR event, possibly per matcher / instrumenter to find hot spots.

ygree:
> Yeah, it was just as fun as I remembered. The class `scala.collection.immutable.List` in Scala 2.12 has a hierarchy of 28 distinct classes and interfaces in the end, but they are reached via several classes and interfaces implementing the same interfaces, which is a hierarchy of 228 classes and interfaces. I think it's even worse in Scala 2.11.

I've run experiment removing those instrumentation parts that get use of hasSuper. It improves instrumentation time but only about a half comparing to the scala/collection and akka class filtering.

Are there any other possibly expensive matchers other than hasSuper?

ygree:
> @charliegracie that makes me wonder which matchers were executing for those classes. The has interface matchers, has super type, and has annotation matchers we know are bad for performance. I wonder if we should add shortcuts to those matchers in the appropriate place to ignore classes in the stdlibs and libraries that definitely won't have them. IE: `org.springframework` classes will most likely never have JAX-RS annotations added to them.

I think we will need to extend our instrumentation so we can collect this type of data. Currently it's a bit of a pain because of Java 7 dependency that doesn't come with JFR support.

I ran an experiment removing possibly expensive instrumentation that make use of hasSuper but the improvement was only about a half comparing to early filtering of scala/collection and akka classes that we don't instrument.

ygree:
> > @charliegracie that makes me wonder which matchers were executing for those classes. The has interface matchers, has super type, and has annotation matchers we know are bad for performance. I wonder if we should add shortcuts to those matchers in the appropriate place to ignore classes in the stdlibs and libraries that definitely won't have them. IE: `org.springframework` classes will most likely never have JAX-RS annotations added to them.
> 
> I think we will need to extend our instrumentation so we can collect this type of data. Currently it's a bit of a pain because of Java 7 dependency that doesn't come with JFR support.
> 
> I ran an experiment removing possibly expensive instrumentation that make use of hasSuper but the improvement was only about a half comparing to early filtering of scala/collection and akka classes that we don't instrument.

@devinsba I've analyzed matcher performance with additional custom JFR events and I can confirm that the most expensive matcher types `has-interface` and `has-super-type` among other matchers contribute the most of time spent of checking scala.collection classes. I think that in my previous experiment (without-has-super see pictures in the PR description) I only disabled known expensive matchers and there could still be expensive matchers which accumulate in this time difference with and without early scala.collection filtering. Below is the graph showing time and number for each matcher type.

`no-optimization` Top Target Time shows a lot of scala.collection classes.
image

`skip-scala-collection` Now Top Target Time drops and there are no scala.collection classes.
image

devinsba:
I'm surprised `isAnnotatedWith` doesn't cause a large amount of overhead. Maybe this is because it is nested within `safeHasSuperType`

ygree:
> I'm surprised `isAnnotatedWith` doesn't cause a large amount of overhead. Maybe this is because it is nested within `safeHasSuperType`

@devinsba You're right, this report doesn't provide enough granularity to see ByteBuddy matchers. I was only able to apply this monitoring on high level HasSuperMethodMatcher, NameMatchers, SafeHasSuperTypeMatcher at the moment.

PR #3374, opened 2022-02-08
Upgrade gradle to 7.4

bantonsson:


devinsba (approved):

bric3:
On my repos I noticed the wrapper jar also changed in 7.4

bantonsson:
@bric3 Yes, the wrapper jar changes for every version, but it's generally not necessary to upgrade it.

bric3:
The wrapper should be fairly stable, I believe. Indeed unless there's a specific bugfix it might not be necessary. But gradle will update the wrapper anyway on the first run.

PR #3377, opened 2022-02-08
Update the async profiler packaging to the new artifact name

jbachorik:
# What Does This Do
Updates the shadowJar configuration to the new async profiler artifact name

# Motivation
The datadog specific async profiler artifact was renamed to reduce the potential confusion. However, the shadow gradle plugin needs to match the jars and since the jar name is now different no async profiler content gets included.

# Additional Notes


luhenry (approved):
We can remove https://github.com/DataDog/dd-trace-java/pull/3377/files#diff-de02c761438b3711658972e77f24b9a6fb4846e338343c8f62f712cd729f56e7R29-R50 with that code

PR #3384, opened 2022-02-10
grizzly/jersey/resteasy parsed bodies

cataphract:
# What Does This Do

Instruments jersey, grizzly (in a path taken e.g. by Glassfish) and resteasy so as to capture parsed urlencoded bodies as well as as arbitrary objects from the message body readers of the JAX-RS implementations. These arbitrary objects are converted by inspection via reflection.


devinsba:
The name of this is somewhat confusing at first look. Would it make sense to rename things to make it clear that the focus here is content of type `form-urlencoded`

dd-java-agent/appsec/src/main/java/com/datadog/appsec/event/data/DataBundle.java
```diff
@@ -12,4 +15,75 @@
 
   // nonnull iif hasAddress(addr) is true
   <T> T get(Address<T> addr);
+
+  static DataBundle unionOf(DataBundle db1, DataBundle db2) {
```
bantonsson, line 19: I can't see that this is used in this PR, but only in #3403

dd-java-agent/appsec/src/main/java/com/datadog/appsec/event/data/ObjectIntrospection.java
```diff
@@ -0,0 +1,120 @@
...
+import java.util.List;
+import java.util.Map;
+
+public final class ObjectIntrospection {
+  private static final int MAX_DEPTH = 20;
+  private static final int MAX_ELEMENTS = 256;
+
+  private ObjectIntrospection() {}
+
+  public static Object convert(Object obj) {
```
bantonsson, line 18: Convert to what? Some comments would be nice.

dd-java-agent/instrumentation/resteasy-appsec/src/main/java/datadog/trace/instrumentation/resteasy/DecodedFormParametersInstrumentation.java
```diff
@@ -0,0 +1,140 @@
...
+  }
+
+  public static class GetDecodedFormParametersAdvice {
+    @Advice.OnMethodEnter(suppress = Throwable.class)
+    static boolean before(
+        @Advice.FieldValue("decodedFormParameters") final MultivaluedMap<String, String> map) {
+      if (map == null) {
+        return true; // proceed
+      }
+      return false;
```
bantonsson, line 114: This naming confuses me a bit. When the 'map' is empty when we enter the `getDecodedFormParameters` method, we should try to `proceed` when we exit the method. I assume this is to try to only call the callback at most once, but the word 'proceed' is so very generic.

bantonsson (approved):
👍 Only some minor comments

dd-java-agent/appsec/src/main/java/com/datadog/appsec/event/data/DataBundle.java
cataphract: Yes, I put it in this changeset by mistake, but I guess no harm done.

dd-java-agent/appsec/src/main/java/com/datadog/appsec/event/data/ObjectIntrospection.java
cataphract: I'm adding a comment.

PR #3385, opened 2022-02-10
Fix computation of Heap profiler capacity

luhenry:


dd-java-agent/agent-profiling/profiling-auxiliary-async/src/main/java/com/datadog/profiling/auxiliary/async/AuxiliaryAsyncProfiler.java
```diff
@@ -105,6 +98,10 @@ public AuxiliaryImplementation provide(ConfigProvider configProvider) {
     if (instance != null) {
       FlightRecorder.addPeriodicEvent(AsyncProfilerConfigEvent.class, this::emitConfiguration);
     }
+
+    this.memleakIntervalDefault =
+        Math.max(0, (int) ManagementFactory.getMemoryMXBean().getHeapMemoryUsage().getMax())
+            / getMemleakCapacity();
```
jbachorik, line 104: This might fail due to division by 0 - `getMemleakCapacity()` may return 0.

luhenry: Fixed with `Math.max(1, ...)`

jbachorik (approved):

PR #3395, opened 2022-02-15
[AppSec] Fixed double calling

ValentinZakharov:
# What Does This Do
Added checkers to avoid double calling of AppSec.

# Motivation
Some combinations of instrumentations can cause events’ reduplication. Customer reported the application built based on Ratpack produces exceptions. In sample RatPack app, I discovered there are `HttpServerDecorator.onRequest()` and  `HttpServerDecorator.onResponse()`, are called twice per request. First time from `netty-4.1` instrumentation and the second time from `ratpack-1.5`. Overall this looks reasonable, but in terms of AppSec, doubling events results in redundant analysis iteration (calling WAF).

# Additional Notes

### DIFF
--- a/dd-java-agent/appsec/src/main/java/com/datadog/appsec/AppSecModule.java
+++ b/dd-java-agent/appsec/src/main/java/com/datadog/appsec/AppSecModule.java
@@ -13,6 +13,8 @@ public interface AppSecModule {
 
   String getName();
 
+  String getInfo();
+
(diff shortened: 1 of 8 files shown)


devinsba:
This looks like a great workaround. I'll make sure you are a reviewer when we make progress on the underlying issue

devinsba (approved):

bantonsson (approved):

PR #3396, opened 2022-02-16
Make mongo tests setup more robust

bantonsson:
# What Does This Do

Tries to make test setup more robust by waiting for the actual `setup` span to be reported. Also upgrades the async test for `mongodb-driver-reactivestreams` to run against `4.0.1` as a minimum, since there was a race condition bug in earlier versions, that sometimes triggered during testing.

# Motivation

Flaky tests are flaky.

# Additional Notes

Has run this commit fully 5+ times in CI without any failures in the mongo tests. Not sure that is enough testing, but at least it doesn't seem to get worse 😉


mcculls (approved):
:+1:

PR #3403, opened 2022-02-21
Appsec: basic grpc support

cataphract:
Basic implementation of interception of server received GRPC messages and their delivery to the instrumentation gateway and the appsec subsystem. Because there might me multiple messages per grpc server span (which creates an appsec request context), the grpc message is not saved along with other appsec data, and the WAF run is one-shot as well (no accumulation of data).

A limitation is that these messages are all reported in the grpc server span and for long running requests there may be many of them.

### DIFF
--- a/dd-java-agent/instrumentation/grpc-1.5/src/main/java/datadog/trace/instrumentation/grpc/server/TracingServerInterceptor.java
+++ b/dd-java-agent/instrumentation/grpc-1.5/src/main/java/datadog/trace/instrumentation/grpc/server/TracingServerInterceptor.java
@@ -1,5 +1,6 @@
 package datadog.trace.instrumentation.grpc.server;
 
+import static datadog.trace.api.gateway.Events.EVENTS;
 import static datadog.trace.bootstrap.instrumentation.api.AgentTracer.activateSpan;
 import static datadog.trace.bootstrap.instrumentation.api.AgentTracer.propagate;
 import static datadog.trace.bootstrap.instrumentation.api.AgentTracer.startSpan;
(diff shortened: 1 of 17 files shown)


dd-smoke-tests/appsec/springboot-grpc/src/test/groovy/datadog/smoketest/appsec/SpringBootWithGRPCAppSecTest.groovy
```diff
@@ -0,0 +1,53 @@
...
+    assert springBootShadowJar != null
+
+    List<String> command = [
+      javaPath(),
+      *defaultJavaProperties,
+      *defaultAppSecProperties,
+      '-agentlib:jdwp=transport=dt_socket,server=y,suspend=y,address=0.0.0.0:5005',
+      "-jar",
+      springBootShadowJar,
+      "--server.port=${httpPort}"
```
bantonsson, line 20: I think this should be `--grpc.server.port=...`

bantonsson (commented):
👍 only two questions...

ValentinZakharov (approved):

dd-java-agent/instrumentation/grpc-1.5/src/main/java/datadog/trace/instrumentation/grpc/server/TracingServerInterceptor.java
```diff
@@ -186,10 +225,97 @@ public void onReady() {
...
+      }
+    }
+  }
+
+  private static void callIGCallbackGrpcMessage(@Nonnull final AgentSpan span, Object obj) {
+    if (obj == null) {
+      return;
+    }
+
+    CallbackProvider cbp = tracer().instrumentationGateway();
```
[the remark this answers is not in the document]
cataphract: Yeah, good catch.

dd-smoke-tests/appsec/springboot-grpc/src/test/groovy/datadog/smoketest/appsec/SpringBootWithGRPCAppSecTest.groovy
cataphract: This is the port for web server; the request goes through http. `server.port` is the correct one.

bantonsson (approved):

PR #3404, opened 2022-02-22
Use jdk.ObjectAllocationSample when available for Allocation profiling

luhenry:
# What Does This Do

Enable the Allocation profiler by default JDK 16 and later.

# Motivation

When using JDK 16+, jdk.ObjectAllocationSample provides low-cost
allocation sampling. We then can afford to enable Allocation profiling
by default on these versions.

# Additional Notes


dd-java-agent/agent-profiling/profiling-controller-openjdk/src/main/java/com/datadog/profiling/controller/openjdk/OpenJdkController.java
```diff
@@ -68,80 +67,61 @@ public OpenJdkController(final Config config)
...
-            || (isJavaVersion(15) && isJavaVersionAtLeast(15, 0, 4))
-            || (isJavaVersion(16) && isJavaVersionAtLeast(16, 0, 2))
-            || (isJavaVersion(17) && isJavaVersionAtLeast(17, 0, 3))
-            || isJavaVersionAtLeast(18))) {
-          // It was enabled based on JDK version so disabled by override file
-          log.warn(
-              "The OldObjectSample JFR event is disabled by the override file but enabled by the config.");
-        }
-      }
+    if (isOldObjectSampleEnabledInConfig(configProvider)) {
```
jbachorik, line 104: I know this hasn't changed but still makes me wonder whether we should rather do the check from L70-75 after the config had been applied? To prevent the users to shoot their legs off  on JDKs we know are causing troubles.

luhenry: How about we add a warning message at the end checking whether the `OldObjectSample` is enabled on a JDK version where it isn't "available". Something like "The OldObjectSample JFR is enabled on a JDK version that doesn't have a performant implementation".

jbachorik (approved):

PR #3409, opened 2022-02-23
[Appsec] Move to libsqreen w/ finalizers; add guardrails

cataphract:
# What Does This Do

Moves to libsqreen to version 3.0.8, which removes `finalize()` implementations. Adds extra checks to make it more likely that `Additive` objects are closed and don't leak memory.

# Motivation

So that we don't pay the performance penalty of having `finalize()` methods.

# Additional Notes

Perhaps different strategies for calling `close()` when the root span finishes are more appropriate, like using `TraceInterceptor`.

### DIFF
--- a/dd-java-agent/appsec/src/main/java/com/datadog/appsec/gateway/AppSecRequestContext.java
+++ b/dd-java-agent/appsec/src/main/java/com/datadog/appsec/gateway/AppSecRequestContext.java
@@ -96,6 +96,13 @@ public class AppSecRequestContext implements DataBundle, Closeable {
   }
 
   public void setAdditive(Additive additive) {
+    Additive curAdditive = this.additive;
+    // the better check would be if the curAdditive has been closed,
+    // but this is behind a private field
+    if (curAdditive != null && additive != null) {
+      log.warn("Replacing WAF object. This is a bug");
+      curAdditive.close();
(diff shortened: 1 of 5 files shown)


dd-java-agent/appsec/src/main/java/com/datadog/appsec/gateway/AppSecRequestContext.java
```diff
@@ -96,6 +96,11 @@ public Additive getAdditive() {
   }
 
   public void setAdditive(Additive additive) {
+    Additive curAdditive = this.additive;
+    if (curAdditive != null) {
+      log.warn("Replacing WAF object. This is a bug");
+      curAdditive.close();
+    }
```
bantonsson, line 103: Since the `EventType.REQUEST_END` callback will call `setAdditive(null)`, won't this always log when there was an additive, even if it has been closed?

cataphract: Absolutely right, I should have checked if the parameter is not null. I didn't see the warning in the logs, but I realize it's because the smoke tests don't merge the stdout of the test process and of the launched process of the server.

bantonsson (approved):

PR #3412, opened 2022-02-24
ensure priority sampling is sent in case of partial flush

amarziali:
# What Does This Do

A large amount of java trace chunks received by the backend are missing priority. No span of the chunk has the tag span.Metrics["_sampling_priority_v1"].

This results in incomplete traces and the new ingestion page visualization will display them with ingestion_reason:unknown. I

This seems to be due to the pending trace buffer that's flushing partial traces. It takes priority from the first slice assuming it's a root, instead of using the related getter that's taking the same from the root context.

This use case is commonly frequent when java applications creates lots of children spans. For instance, JPA repository call with many to many eager fetch may easily run into this issue. The issue is also easy to reproduce with spring petclinic demo app.

In this PR I'm providing a fix and related unit test.

# Testing

I build a simple spring boot application exposing a rest service who's looping on a repository call a large number of time (13k) in order to generate child spans. So for each call we expect 1 root for the http call, 1 child for the rest controller and then 13k *2 (1 for the spring data repository call and 1 for the sql call). So in total 13 *2 + 2 = 26002 spans per call.

Then I do some load testing with apache benchmark in order to generate some traffic

This test case is done one time with the official agent (latest version) and one time with the patched one.

I used following sys properties: -javaagent:/path/to/dd-java-agent.jar -Ddd.profiling.enabled=false -XX:FlightRecorderOptions=stackdepth=256 -Ddd.logs.injection=true -Ddd.trace.sample.rate=1 -Ddd.service=tracer-ko -Ddd.env=staging

The service name `tracer-ko` has been used to test bad behaviour and `tracer-ok` for the patched one.

So with upatched tracer:

Traces are shown broken and tagged with unknown ingestion reason (missing priority)

image

As well span list looks incomplete (< 26k)
image

With patched tracer:

Traces look complete

image

And the distribution of ingested traces (grouped by service) with ingestion_reason to unknown shows that only the service tracer-ko reported them.

image

# Additional notes

It relates to Ticket APMS-6717.


devinsba:
Thanks for taking a look and proposing a fix. To fix the smoke test failures, please take a look at the payload files in `dd-smoke-tests/springboot-openliberty/snapshots`. One of the failure results I've pasted below:

```
At expected span:
{"name" "spring.handler"
 "service" "smoke-test-java-app"
 "resource" "DemoController.connectFinal"
 "type" "web"
 "error" 0
 "span_id" 1
 "trace_id" 0
 "parent_id" 0
 "start" 1618011719625753238
 "duration" 76667015
 "meta" {"env" "smoketest"
         "language" "jvm"
         "version" "99"
         "thread.name" "Default Executor-thread-2"
         "component" "spring-web-controller"
         "span.kind" "server"}
 "metrics" {"_dd.measured" 1
            "thread.id" 44}}
Received actual span:
{"name" "spring.handler"
 "service" "smoke-test-java-app"
 "resource" "DemoController.connectFinal"
 "type" "web"
 "error" 0
 "span_id" 6006327373674933785
 "trace_id" 3208154529880048784
 "parent_id" 8588968414339109057
 "start" 1645711580899140276
 "duration" 133681986
 "meta" {"env" "smoketest"
         "language" "jvm"
         "version" "99"
         "thread.name" "Default Executor-thread-15"
         "component" "spring-web-controller"
         "span.kind" "server"}
 "metrics" {"_dd.measured" 1
            "_sampling_priority_v1" 1
            "thread.id" 63}}
Span data mismatch.
❌ Key 'metrics._sampling_priority_v1' not expected.
```

bantonsson (approved):
👍

dd-smoke-tests/springboot-openliberty/snapshots/datadog.smoketest.SpringBootOpenLibertySnapshotTest.exception404.snap
```diff
@@ -63,7 +64,8 @@
                    "version" "99"
                    "thread.name" "Default Executor-thread-13"
                    "component" "java-web-servlet-response"}
-           "metrics" {"thread.id" 62}}
+           "metrics" {"_sampling_priority_v1" 1
```
devinsba, line 67: Should this really be set on every single span? I understand when it's a partial flush it probably should but when we do a normal write wouldn't it be more efficient to only put it on the entry span?

devinsba (approved):
Just a clarifying question

dd-smoke-tests/springboot-openliberty/snapshots/datadog.smoketest.SpringBootOpenLibertySnapshotTest.exception404.snap
bantonsson: That is a very good point indeed.

amarziali: Yep indeed we can add priority on first and last span only on each chunk. the last because it seems (according to @raphaelgavache ) is exploited by the old version of the agent. What do you think about this tradeoff @devinsba and @bantonsson ?

bantonsson: That sounds like a good tradeoff.

devinsba: Agreed

dougqh: Yes, that sounds fine to me.  Thanks for making the change -- what we've been missing on our end is a specification that tells us which spans need this information.

charliegracie (approved):
Approved based on the moving the priority to the first and last span of each chunk

dd-smoke-tests/springboot-openliberty/snapshots/datadog.smoketest.SpringBootOpenLibertySnapshotTest.exception404.snap
amarziali: I did the changes and added a test to check expected behaviour. Could you please have a look again. It should be good . Thanks

bantonsson (approved):
👍

PR #3436, opened 2022-03-10
Make TraceInterceptor application more robust

bantonsson:
# What Does This Do

Make TraceInterceptor application more robust to exceptions thrown in the interceptors, by trying to continue with processing and not dropping the full trace.

# Motivation

A badly behaving TraceInterceptor could lead to no trace information being sent to the backend.

# Additional Notes

The exception being thrown is logged at a rate of once per minute, to not flood the logs.


gradle/java_no_deps.gradle
```diff
@@ -24,7 +24,7 @@ tasks.register('forkedTest', Test) {
 
 test {
   exclude("**/*ForkedTest*")
-  dependsOn "forkedTest"
+  finalizedBy "forkedTest"
```
bantonsson, line 27: Not strictly related, but I got tired of the forked tests being run before the test I was trying to run.

dd-trace-core/src/main/java/datadog/trace/civisibility/CiVisibilityTraceInterceptor.java
```diff
@@ -27,11 +27,12 @@
     final DDSpan spanToCheck = null == localRootSpan ? firstSpan : localRootSpan;
 
     // If root span does not have type == "test", we drop the full trace.
-    if (!TEST_TYPE.contentEquals(spanToCheck.getType())) {
```
bantonsson, line 30: This threw an exception in the `CiVisibilityTraceInterceptorTest."discard a trace that does not come from a test"`, which also dropped the whole trace before, but now we need to treat 'null' separately.

dd-trace-core/src/main/java/datadog/trace/core/CoreTracer.java
```diff
@@ -683,6 +684,8 @@ public AgentSpan noopSpan() {
     return extractor.extract(carrier, getter);
   }
 
+  private final RatelimitedLogger rlLog = new RatelimitedLogger(log, 1, TimeUnit.MINUTES);
```
devinsba, line 687: I wonder if this should be 1 per trace interceptor instead of 1 for the whole instance. Potentially a user could have errors in a second interceptor that they don't ever see because the first one would steal the log write

devinsba (approved):

dd-trace-core/src/main/java/datadog/trace/core/CoreTracer.java
bantonsson: Yes that's correct, but seeing any issues in the `TraceInterceptor` would lead us to enable debug, or at least say that they should clean up the broken `TraceInterceptor` and _voila_, they would now see the new broken `TraceInterceptor`.

charliegracie (approved):

PR #3445, opened 2022-03-15
Options to specify which GRPC status codes are errors

charliegracie:
# What Does This Do
Allows users to specify which GRPC status codes are errors

# Motivation
This is a follow on to previous work to ignore cancelled status codes for GRPC servers. With the option users can now decide which status codes to treat as errors

# Additional Notes


devinsba (approved):
Looks good

PR #3447, opened 2022-03-16
Add resource name priority for changes made through dd-trace-ot

devinsba:
# What Does This Do

Adds a priority level for resource names set by users using dd-trace-ot.

# Motivation

A customer issue where instrumentation was overwriting a resource name that they were adding to the root span

# Additional Notes

### DIFF
--- a/dd-trace-ot/src/test/groovy/datadog/opentracing/OTSpanTest.groovy
+++ b/dd-trace-ot/src/test/groovy/datadog/opentracing/OTSpanTest.groovy
@@ -1,0 +1,47 @@
+package datadog.opentracing
+
+import datadog.trace.api.interceptor.MutableSpan
+import datadog.trace.bootstrap.instrumentation.api.ResourceNamePriorities
+import datadog.trace.test.util.DDSpecification
+import io.opentracing.Scope
+import io.opentracing.Span
+import io.opentracing.util.GlobalTracer
+import spock.lang.Shared
+
+class OTSpanTest extends DDSpecification {
+  @Shared
+  DDTracer tracer = DDTracer.builder().build()
+
+  def setup() {
+    GlobalTracer.register(tracer)
+  }
+
+  def "test resource name assignment through MutableSpan casting"() {
+    given:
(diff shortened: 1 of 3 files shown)


dd-trace-ot/src/test/groovy/datadog/opentracing/OTSpanTest.groovy
```diff
@@ -0,0 +1,47 @@
...
+  @Shared
+  DDTracer tracer = DDTracer.builder().build()
+
+  def setup() {
+    GlobalTracer.register(tracer)
+  }
+
+  def "test resource name assignment through MutableSpan casting"() {
+    given:
+    OTSpan testSpan = tracer.buildSpan("parent").withResourceName("test-resource").start() as OTSpan
```
devinsba, line 21: Not sure if I should apply this priority to the builder also. The customer issue was instrumentation overwriting their resource name since it had a higher priority. I can't think of a scenario where a span would be created in opentracing but modified by autoinstrumentation.

internal-api/src/main/java/datadog/trace/bootstrap/instrumentation/api/ResourceNamePriorities.java
```diff
@@ -7,4 +7,5 @@
   public static final byte HTTP_FRAMEWORK_ROUTE = 3;
   public static final byte HTTP_SERVER_CONFIG_PATTERN_MATCH = 4;
   public static final byte TAG_INTERCEPTOR = 5;
+  public static final byte MANUAL_INSTRUMENTATION = 5;
```
devinsba, line 10: Not sure if this should be higher priority than tag interceptor or not

bantonsson: I think that it's fine that they have the same priority, since they serve the same purpose. Only user code should set the tag to trigger the interceptor.

bantonsson (commented):
👍

bantonsson (approved):

PR #3448, opened 2022-03-16
Added extra parsing for property/env

ValentinZakharov:
# What Does This Do
Added extra parsing properties/env variables, in case if customer tries to enable feature(modules) with "0", "1".

# Motivation
In case, if try to enable module with value `1`, it will be displayed as enabled, but nothing in practice module will not be started.
For example if customer tries to enable AppSec with environment variable `DD_APPSEC_ENABLED=1`, the AppSec module will not be loaded, however in logs we will be able to see:
`DATADOG TRACER CONFIGURATION {... "appsec_enabled":true ...}`

# Additional Notes
The problem exist because of property/env reading happens in two places: in `Agent::isFeatureEnabled(...)` and in `Config` but with the different logic.


bantonsson:
I think it would make more sense to make checking in `isFeatureEnabled` work the same way as `booleanValueOf` in the `ConfigConverter`

bantonsson (approved):

PR #3450, opened 2022-03-16
WIP Use spring-boot to test all servlet implementations

devinsba:
# What Does This Do

Not the final state of this PR. The first commit is for discussion purposes, after that I will be making a spring-webmvc test set in the tomcat and jetty instrumentations in order to verify the integrated state of those instrumentations

# Motivation

# Additional Notes


dd-java-agent/instrumentation/spring-webmvc-3.1/spring-webmvc-3.1.gradle
```diff
@@ -39,9 +39,11 @@ dependencies {
     exclude(module: 'jetty-server') // incompatable servlet api
   }
 
-  // Include servlet instrumentation for verifying the tomcat requests
+  // Include instrumentation for verifying the tomcat requests
   testImplementation project(':dd-java-agent:instrumentation:servlet')
+  testImplementation project(':dd-java-agent:instrumentation:servlet:request-2')
   testImplementation project(':dd-java-agent:instrumentation:servlet:request-3')
+  testImplementation project(':dd-java-agent:instrumentation:tomcat-5.5')
```
devinsba, line 46: I added all the potential instrumentations so that we get the real world view of things

dd-java-agent/instrumentation/spring-webmvc-3.1/src/test/groovy/test/boot/SecurityConfig.java
```diff
@@ -0,0 +1,49 @@
+package test.boot;
```
devinsba, line 1: Changed to a java file because autocomplete is already hard enough with `WebSecurityConfigurerAdapter` without including groovys type inference

dd-java-agent/instrumentation/spring-webmvc-3.1/src/test/groovy/test/boot/SpringBootBasedTest.groovy
```diff
@@ -148,6 +148,8 @@ class SpringBootBasedTest extends HttpServerTest<ConfigurableApplicationContext>
     if (endpoint == REDIRECT) {
       // Spring is generates a RenderView and ResponseSpan for REDIRECT
       return super.spanCount(endpoint) + 1
+    } else if (endpoint == NOT_FOUND) {
+      return super.spanCount(endpoint) + 2
```
devinsba, line 152: We found 2 more spans by adding tomcat. We had the same problem with tomcat not including the servlet instrumentation recently

dd-java-agent/instrumentation/spring-webmvc-3.1/src/test/groovy/test/boot/WithBasePathTest.groovy
```diff
@@ -0,0 +1,23 @@
...
+  }
+
+  @Override
+  SpringApplication application() {
+    return new SpringApplication(Config, AppConfig, SecurityConfig, AuthServerConfig, TestController)
+  }
+
+  @Override
+  Map<String, Serializable> expectedExtraServerTags(ServerEndpoint endpoint) {
+    ["servlet.path": "/spring-servlet", "servlet.context": "/$servletContext"] +
```
devinsba, line 20: The first of the 2 actual problems is here

Lets compare to the method used for the other spring-boot tests in this project:

```
  @Override
  Map<String, Serializable> expectedExtraServerTags(ServerEndpoint endpoint) {
    ["servlet.path": endpoint.path, "servlet.context": "/$servletContext"] +
    extraServerTags
  }
```

We lose the endpoint information in `servlet.path` **BUT** I this is correct and there shouldn't be a `servlet.path` tag unless the servlet is mounted at a path within the context so the other spring-boot tests should be changed to not expect it.

The spring-boot config for this is:

```
server.servlet-path=/spring-servlet
```

dd-java-agent/instrumentation/spring-webmvc-3.1/spring-webmvc-3.1.gradle
```diff
@@ -51,9 +53,9 @@ dependencies {
 
   testImplementation group: 'org.spockframework', name: 'spock-spring', version: "$versions.spock"
 
-  testImplementation group: 'org.springframework.boot', name: 'spring-boot-starter-test', version: '1.5.17.RELEASE'
-  testImplementation group: 'org.springframework.boot', name: 'spring-boot-starter-web', version: '1.5.17.RELEASE'
-  testImplementation group: 'org.springframework.boot', name: 'spring-boot-starter-security', version: '1.5.17.RELEASE'
+  testImplementation group: 'org.springframework.boot', name: 'spring-boot-starter-test', version: '1.5.22.RELEASE'
+  testImplementation group: 'org.springframework.boot', name: 'spring-boot-starter-web', version: '1.5.22.RELEASE'
+  testImplementation group: 'org.springframework.boot', name: 'spring-boot-starter-security', version: '1.5.22.RELEASE'
```
devinsba, line 58: The versions here didn't matter. I updated since it was easier to get to the documentation for this version

dd-java-agent/instrumentation/spring-webmvc-3.1/src/test/groovy/test/dynamic/DynamicRoutingTest.groovy
```diff
@@ -98,27 +102,41 @@ class DynamicRoutingTest extends HttpServerTest<ConfigurableApplicationContext>
...
+
   @Override
   boolean changesAll404s() {
     true
   }
 
   @Override
   Map<String, Serializable> expectedExtraServerTags(ServerEndpoint endpoint) {
-    ["servlet.path": endpoint.path]
+    ["servlet.path": endpoint.path, "servlet.context": "/"]
```
devinsba, line 123: This is the second issue (happens in multiple test files). Tomcat is tagging a context with a `/` in this configuration, when it doesn't when configured the way it is in its own tests. We know this comes from tomcat because that is the only change that could cause this

dd-java-agent/instrumentation/spring-webmvc-3.1/src/test/groovy/test/boot/SpringBootBasedTest.groovy
```diff
@@ -267,6 +269,30 @@ class SpringBootBasedTest extends HttpServerTest<ConfigurableApplicationContext>
...
+        childOf(trace.span(0))
+        tags {
+          "component" "java-web-servlet-dispatcher"
+          "$Tags.HTTP_ROUTE" "/error"
+          addTags(extraTags)
+          defaultTags()
+        }
+      }
+      trace.span {
+        operationName "spring.handler"
```
eric-ming2, line 286: This looks like a handler span instead of a response span, does this belong in the `responseSpan` method? Could you handle this in the `handlerSpan` method instead?

devinsba: I'd like to put it there but it's actually part of the response handling, its a child of the `servlet.forward` span above

dd-java-agent/instrumentation/spring-webmvc-3.1/src/bootTest/groovy/test/boot/AppConfig.groovy
```diff
@@ -14,21 +14,6 @@ import org.springframework.web.util.UrlPathHelper
...
-    factory.addConnectorCustomizers(
-      new TomcatConnectorCustomizer() {
-        @Override
-        void customize(final Connector connector) {
-          connector.setEnableLookups(true)
-        }
-      })
-
-    return factory
-  }
```
devinsba, line 30: No idea what this was meant to do but it wouldn't work with the way I reused the tests for different app servers

PR #3452, opened 2022-03-18
Support long running spans

amarziali:
# What Does This Do

Add support for long running spans. Allows incomplete spans to be reported to datadog and partial results to be displayed into the live view. 
As well, this PR enables collecting complete traces for spans lasting more than 15 minutes.

The feature is released behind a configuration flag and it's disabled by default.

# How does it work

Pending traces are scanned on a scheduled basis and if a long lasting span is found:
* A copy of this span (same span_id, parent_it, trace_id) is added to the buffer. The span is tagged as partial (metrics[_dd.partial_version] with an incrementing number)
* All finished spans also buffered on the same pending traces are flushed (similar to partial flush scenario)

# Motivation

Today, due to internal datadog architecture, spans lasting more than 15 minutes may result in incomplete tracing. Additionally to this, spans are sent only when all pending ones are closed. Hence, live view won't display those ones.

# Additional Notes

### DIFF
--- a/dd-trace-api/src/main/java/datadog/trace/api/config/TracerConfig.java
+++ b/dd-trace-api/src/main/java/datadog/trace/api/config/TracerConfig.java
@@ -62,9 +62,9 @@ public final class TracerConfig {
   public static final String ENABLE_TRACE_AGENT_V05 = "trace.agent.v0.5.enabled";
   public static final String SAMPLING_MECHANISM_VALIDATION_DISABLED =
       "trace.sampling.mechanism.validation.disabled";
-  public static final String LONG_RUNNING_TRACE_ENABLED = "trace.longrunning.enabled";
+  public static final String LONG_RUNNING_TRACE_ENABLED = "longrunning.enabled";
 
-  public static final String LONG_RUNNING_TRACE_FLUSH_INTERVAL = "trace.longrunning.flush.interval";
+  public static final String LONG_RUNNING_TRACE_FLUSH_INTERVAL = "longrunning.flush.interval";
 
   public static final String CLOCK_SYNC_PERIOD = "trace.clock.sync.period";
 
--- a/dd-trace-api/src/main/java/datadog/trace/api/config/TracerConfig.java
+++ b/dd-trace-api/src/main/java/datadog/trace/api/config/TracerConfig.java
@@ -62,6 +62,9 @@ public final class TracerConfig {
   public static final String ENABLE_TRACE_AGENT_V05 = "trace.agent.v0.5.enabled";
   public static final String SAMPLING_MECHANISM_VALIDATION_DISABLED =
       "trace.sampling.mechanism.validation.disabled";
+  public static final String LONG_RUNNING_TRACE_ENABLED = "trace.longrunning.enabled";
+
+  public static final String LONG_RUNNING_TRACE_FLUSH_INTERVAL = "trace.longrunning.flush.interval";
 
   public static final String CLOCK_SYNC_PERIOD = "trace.clock.sync.period";
 
--- a/dd-trace-api/src/main/java/datadog/trace/api/config/TracerConfig.java
+++ b/dd-trace-api/src/main/java/datadog/trace/api/config/TracerConfig.java
@@ -62,9 +62,9 @@ public final class TracerConfig {
   public static final String ENABLE_TRACE_AGENT_V05 = "trace.agent.v0.5.enabled";
   public static final String SAMPLING_MECHANISM_VALIDATION_DISABLED =
       "trace.sampling.mechanism.validation.disabled";
-  public static final String LONG_RUNNING_TRACE_ENABLED = "longrunning.enabled";
+  public static final String LONG_RUNNING_TRACE_ENABLED = "trace.longrunning.enabled";
 
-  public static final String LONG_RUNNING_TRACE_FLUSH_INTERVAL = "longrunning.flush.interval";
+  public static final String LONG_RUNNING_TRACE_FLUSH_INTERVAL = "trace.longrunning.flush.interval";
 
   public static final String CLOCK_SYNC_PERIOD = "trace.clock.sync.period";
 
--- a/dd-trace-core/src/main/java/datadog/trace/core/CoreSpan.java
+++ b/dd-trace-core/src/main/java/datadog/trace/core/CoreSpan.java
@@ -85,4 +85,6 @@ public interface CoreSpan<T extends CoreSpan<T>> {
   T setFlag(CharSequence name, boolean value);
 
   int samplingPriority();
+
+  Long getPartialVersion();
 }
--- a/dd-trace-core/src/main/java/datadog/trace/core/CoreTracer.java
+++ b/dd-trace-core/src/main/java/datadog/trace/core/CoreTracer.java
@@ -560,7 +560,7 @@ public class CoreTracer implements AgentTracer.TracerAPI {
     return null == mapped ? serviceName : mapped;
   }
 
-  boolean isDisableSamplingMechanismValidation() {
+  boolean   isDisableSamplingMechanismValidation() {
     return disableSamplingMechanismValidation;
   }
 
--- a/dd-trace-core/src/main/java/datadog/trace/core/CoreTracer.java
+++ b/dd-trace-core/src/main/java/datadog/trace/core/CoreTracer.java
@@ -516,7 +516,7 @@ public class CoreTracer implements AgentTracer.TracerAPI {
     this.instrumentationGateway = instrumentationGateway;
 
     if (config.isTraceLongRunningEnabled()) {
-      this.traceKeepAlive = new TraceKeepAlive(config.getTraceLongRunningFlushInterval());
+      this.traceKeepAlive = new TraceKeepAlive(config.getTraceLongRunningFlushInterval() * 1000);
       traceKeepAlive.start();
     } else {
       this.traceKeepAlive = null;
--- a/dd-trace-core/src/main/java/datadog/trace/core/CoreTracer.java
+++ b/dd-trace-core/src/main/java/datadog/trace/core/CoreTracer.java
@@ -1221,9 +1221,6 @@ public class CoreTracer implements AgentTracer.TracerAPI {
     public void run() {
       final CoreTracer tracer = reference.get();
       if (tracer != null) {
-        if (tracer.traceKeepAlive != null) {
-          tracer.traceKeepAlive.stop();
-        }
         tracer.close();
       }
     }
--- a/dd-trace-core/src/main/java/datadog/trace/core/CoreTracer.java
+++ b/dd-trace-core/src/main/java/datadog/trace/core/CoreTracer.java
@@ -162,6 +162,7 @@ public class CoreTracer implements AgentTracer.TracerAPI {
   private final HttpCodec.Extractor extractor;
 
   private final InstrumentationGateway instrumentationGateway;
+  private final TraceKeepAlive traceKeepAlive;
 
   @Override
   public AgentScope.Continuation capture() {
@@ -357,7 +358,6 @@ public class CoreTracer implements AgentTracer.TracerAPI {
       taggedHeaders(config.getRequestHeaderTags());
       partialFlushMinSpans(config.getPartialFlushMinSpans());
       strictTraceWrites(config.isTraceStrictWritesEnabled());
-
       return this;
     }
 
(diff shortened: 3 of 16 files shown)


dd-trace-core/src/main/java/datadog/trace/core/PendingTrace.java
```diff
@@ -61,6 +63,7 @@ PendingTrace create(@Nonnull DDId traceId) {
   private final boolean strictTraceWrites;
 
   private final ConcurrentLinkedDeque<DDSpan> finishedSpans = new ConcurrentLinkedDeque<>();
+  private final ConcurrentLinkedDeque<DDSpan> unfinishedSpans = new ConcurrentLinkedDeque<>();
```
devinsba, line 66: You can move this allocation into the constructor below

devinsba (changes requested):
We need to address the `pendingTraces` map

dd-trace-core/src/main/java/datadog/trace/core/PendingTrace.java
```diff
@@ -153,19 +161,66 @@ void onFinish(final DDSpan span) {
...
+              original.getSpanType(),
+              -1,
+              original.context().getTrace(),
+              original.context().getData(),
+              tracer.isDisableSamplingMechanismValidation());
+      final DDSpan span =
+          DDSpan.create(original.getStartTime() / 1000, contextCopy, original.hasCheckpoints());
+      final long partialVersion = (triggerNanos - span.getStartTime()) / keepAliveIntervalNanos;
+      span.setPartialVersion(partialVersion);
+      span.setMetric(DDTags.DD_PARTIAL_VERSION, partialVersion);
```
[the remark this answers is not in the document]
dougqh: +1

dd-trace-core/src/main/java/datadog/trace/core/CoreSpan.java
```diff
@@ -85,4 +85,6 @@ T setSamplingPriority(
   T setFlag(CharSequence name, boolean value);
 
   int samplingPriority();
+
+  Long getPartialVersion();
```
dougqh, line 89: I don't think I want partially pushed onto the span itself.  It isn't really a data model concept.  It is a transport concept.

dd-trace-core/src/main/java/datadog/trace/core/CoreTracer.java
```diff
@@ -160,6 +160,7 @@ public int compare(final TraceInterceptor o1, final TraceInterceptor o2) {
   private final HttpCodec.Extractor extractor;
 
   private final InstrumentationGateway instrumentationGateway;
+  private final TraceKeepAlive traceKeepAlive;
```
dougqh, line 163: I don't think we need another object for this.  I was thinking it would be handled inside of PendingTrace.

dd-trace-core/src/main/java/datadog/trace/core/TraceKeepAlive.java
```diff
@@ -0,0 +1,72 @@
...
+import java.util.concurrent.ConcurrentMap;
+import java.util.concurrent.TimeUnit;
+import org.slf4j.Logger;
+import org.slf4j.LoggerFactory;
+
+/**
+ * A monitor thread scheduled on a regular basis. It triggers keep-alive spans to be sent for
+ * long-running traces.
+ */
+public class TraceKeepAlive implements AgentTaskScheduler.Task<Boolean> {
```
dougqh, line 16: As noted elsewhere, I don't think we need a separate class for this.  I think this would better be handled inside of PendingTrace.

dd-trace-core/src/main/java/datadog/trace/core/CoreTracer.java
```diff
@@ -539,6 +546,10 @@ public String mapServiceName(String serviceName) {
     return null == mapped ? serviceName : mapped;
   }
 
+  boolean isDisableSamplingMechanismValidation() {
```
dougqh, line 549: Why is this being added?

dd-trace-core/src/main/java/datadog/trace/core/DDSpan.java
```diff
@@ -122,6 +125,14 @@ public boolean isFinished() {
     return durationNano != 0;
   }
 
+  public Long getPartialVersion() {
```
dougqh, line 128: These shouldn't be part of the public span API.  I don't think should be on span at all.

dd-trace-core/src/main/java/datadog/trace/core/DDSpanContext.java
```diff
@@ -531,6 +531,7 @@ public Object unsafeGetTag(final String tag) {
 
   public void processTagsAndBaggage(final MetadataConsumer consumer) {
     synchronized (unsafeTags) {
+      final boolean incomplete = unsafeTags.containsKey(DDTags.DD_PARTIAL_VERSION);
```
dougqh, line 534: We shouldn't be using a tag for partialness.  It isn't really a tag in the data model -- that's just a trick to fit into the existing protocol.

dd-trace-core/src/main/java/datadog/trace/core/DDSpanContext.java
```diff
@@ -616,4 +617,12 @@ private DDSpanContext getTopContext() {
     DDSpan span = trace.getRootSpan();
     return null != span ? span.context() : this;
   }
+
+  String getParentServiceName() {
```
dougqh, line 621: Why are we adding these?  I don't think this is the right place for them.

dd-trace-core/src/main/java/datadog/trace/core/PendingTrace.java
```diff
@@ -105,6 +108,10 @@ private PendingTrace(
     this.traceId = traceId;
     this.pendingTraceBuffer = pendingTraceBuffer;
     this.strictTraceWrites = strictTraceWrites;
+
+    if (tracer.getTraceKeepAlive() != null) {
+      tracer.getTraceKeepAlive().onPendingTraceBegins(this);
```
dougqh, line 113: This is mostly confirming that should just be part of PendingTrace.

dougqh (changes requested):
I've commented on this PR, but I'd really suggest that the PR just be abandoned.

I think we need to back and talk about the protocol and how to make the changes more before jumping into code.

internal-api/src/main/java/datadog/trace/api/Config.java
```diff
@@ -1,57 +1,6 @@
...
-import static datadog.trace.api.ConfigDefaults.DEFAULT_TRACE_AGENT_V05_ENABLED;
-import static datadog.trace.api.ConfigDefaults.DEFAULT_TRACE_ANALYTICS_ENABLED;
-import static datadog.trace.api.ConfigDefaults.DEFAULT_TRACE_ANNOTATIONS;
-import static datadog.trace.api.ConfigDefaults.DEFAULT_TRACE_ENABLED;
-import static datadog.trace.api.ConfigDefaults.DEFAULT_TRACE_EXECUTORS_ALL;
-import static datadog.trace.api.ConfigDefaults.DEFAULT_TRACE_METHODS;
-import static datadog.trace.api.ConfigDefaults.DEFAULT_TRACE_RATE_LIMIT;
-import static datadog.trace.api.ConfigDefaults.DEFAULT_TRACE_REPORT_HOSTNAME;
-import static datadog.trace.api.ConfigDefaults.DEFAULT_TRACE_RESOLVER_ENABLED;
+import static datadog.trace.api.ConfigDefaults.*;
```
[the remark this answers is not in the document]
amarziali: Oh yes. I had to reorganize imports after a rebase. Now should be fixed thanks

dd-trace-api/src/main/java/datadog/trace/api/config/TracerConfig.java
```diff
@@ -62,6 +62,9 @@
   public static final String ENABLE_TRACE_AGENT_V05 = "trace.agent.v0.5.enabled";
   public static final String SAMPLING_MECHANISM_VALIDATION_DISABLED =
       "trace.sampling.mechanism.validation.disabled";
+  public static final String LONG_RUNNING_TRACE_ENABLED = "longrunning.enabled";
+
+  public static final String LONG_RUNNING_TRACE_FLUSH_INTERVAL = "longrunning.flush.interval";
```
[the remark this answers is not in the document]
amarziali: I stripped that prefix to align with what done on go tracer. There is a common convention between tracers on configuration generally speaking?

dd-trace-core/src/main/java/datadog/trace/core/PendingTrace.java
amarziali: should I move also already existing one `finishedSpans` so?

dd-trace-core/src/main/java/datadog/trace/core/DDSpan.java
```diff
@@ -64,7 +64,10 @@ static DDSpan create(
    * combination of millisecond-precision clock and nanosecond-precision offset from start of the
    * trace. See {@link PendingTrace} for details.
    */
-  private final long startTimeNano;
+  private long startTimeNano;
```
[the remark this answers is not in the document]
amarziali: yes. just changed

dd-trace-core/src/main/java/datadog/trace/core/CoreTracer.java
amarziali: Used here: https://github.com/DataDog/dd-trace-java/blob/7b5d30060e0d3bc2218bc230809440a831fd8522/dd-trace-core/src/main/java/datadog/trace/core/PendingTrace.java#L214

I limited to package visibility

dd-trace-core/src/main/java/datadog/trace/core/CoreTracer.java
amarziali: Do you prefer having PendingTrace handling  a scheduled job?

dd-trace-core/src/main/java/datadog/trace/core/DDSpanContext.java
amarziali: Because this information is needed to construct a new SpanContext here: https://github.com/DataDog/dd-trace-java/blob/7b5d30060e0d3bc2218bc230809440a831fd8522/dd-trace-core/src/main/java/datadog/trace/core/PendingTrace.java#L201

I gave only package visibility. Maybe something can be simplified. Please advice

dd-trace-core/src/main/java/datadog/trace/core/PendingTrace.java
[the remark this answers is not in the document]
amarziali: Maybe putting in the metadata like `measured` or `top_level` could be a better option so?

dd-trace-api/src/main/java/datadog/trace/api/config/TracerConfig.java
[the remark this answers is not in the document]
devinsba: Yes, we try to align. The problem is that we support multiple products in our tracer so we try to consistently namespace our configs when they are specific to a product. We don't always do it though so it's probably fine but my preference is for namespacing

dd-trace-core/src/main/java/datadog/trace/core/PendingTrace.java
devinsba: No, that's a required feature. In this case we can avoid the creation of the Deque by only creating inside your if block

dd-trace-api/src/main/java/datadog/trace/api/config/TracerConfig.java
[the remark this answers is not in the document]
amarziali: I added the trace namespace as suggested. Thanks

dd-trace-core/src/main/java/datadog/trace/core/PendingTrace.java
amarziali: Changes are done. Also did another optimization not to populate this list if null (= feature not activated)

dd-trace-core/src/main/java/datadog/trace/core/PendingTrace.java
amarziali: As far I can see, on PendingTrace I would have the possibility to emit keepalives when write is called (e.g. a span of the same trace is finished). But what in case nothing happens and I've just to keep alive an unfinished span? I don't see any other solution than using a scheduled thread. The advantage of externalizing is just use one thread for all the pending traces and separate roles

PR #3453, opened 2022-03-18
Fix issues with features discovery

randomanderson:
# What Does This Do

1) Simplify the features discovery logic by removing needlessly complicated reverse iteration
2) Fix a bug where feature discovery was invoked after every trace request

The agent returns `Datadog-Agent-State` (a hash of the info endpoint) for all endpoints _except_ the info endpoint. Previous to this PR, if the `/info` endpoint exists on the agent, the `state==null` **always**. With trace requests, `state!=null` **always**. As a result, after ever trace request `DDAgentApi.handleAgentChange()` would always call `discover()`.

This PR changes that behavior by probing the discovered trace endpoint once even if the endpoint was found via `/info`

# Motivation

I discovered these issue while adding another discovery endpoint

# Additional Notes

The behavior when dealing Datadog agents that do not return `Datadog-Agent-State` remains unchanged. They will still be feature probed after every trace request.


bantonsson (approved):
Nice find, and great cleanup.

PR #3454, opened 2022-03-21
Skip probing agent trace endpoints when info is available

bantonsson:
# What Does This Do

Takes the result of the Datadog agent `/info` endpoint and avoids probing the endpoints by computing the state hash of the response.

# Motivation

Avoid probing the Datadog agent with incomplete requests that show up as a failure to decode in the agent logs, that might confuse customers/support.

# Additional Notes

I also changed the probing to send valid but empty payloads to the Datadog agent, but that broke the old `test-agent` that's used in some smoke tests, so it was changed back. Replacing the `test-agent` is a bigger task.


randomanderson (approved):
👍

PR #3458, opened 2022-03-23
Fix jakarta instrumentation dependencies

randomanderson:
# What Does This Do
Changes the `jakarta` instrumentation dependency to `compileOnly`. Otherwise, it is added to the final agent jar. Also, adds the api dependency to the `latestDepTest` configuration.

# Motivation
I noticed this issue while debugging another problem

# Additional Notes
We may think of adding custom configuration names like [open-telemetry](https://github.com/open-telemetry/opentelemetry-java-instrumentation/blob/main/conventions/src/main/kotlin/io.opentelemetry.instrumentation.base.gradle.kts) to prevent issues like this. They have `library`, `testLibrary`, and `latestDepTestLibrary` that automatically does the right thing


mcculls (approved):
👍

PR #3469, opened 2022-03-29
Add simple Tracer creating test of published dependencies

bantonsson:
# What Does This Do

Adds a simple test that creates a `Tracer` using the _published_ `dd-trace-ot` dependency.

# Motivation

At least check that the `Tracer` can be created without errors.

# Additional Notes

This would have caught the `0.98.0` missing classes.


test-published-dependencies/ot-pulls-in-api/src/test/java/CreateTracerTest.java
```diff
@@ -0,0 +1,13 @@
...
+
+public class CreateTracerTest {
+
+  @Test
+  void createTracer() {
+    DDTracer tracer = DDTracer.builder().serviceName("TestService").build();
+    GlobalTracer.registerIfAbsent(tracer);
+    datadog.trace.api.GlobalTracer.registerIfAbsent(tracer);
+  }
+}
```
devinsba, line 13: Maybe we should create a trace in this test since the `App` class didn't catch this

test-published-dependencies/gradlew
```diff
@@ -0,0 +1 @@
+../gradlew
```
devinsba, line 1: You'll want to update the circle config to use this symlink instead of the parent executable

devinsba (approved):
LGTM once the circle config is updated

test-published-dependencies/ot-pulls-in-api/src/test/java/CreateTracerTest.java
bantonsson: Sure, I'll add that...

mcculls (approved):

PR #3470, opened 2022-03-29
Refactor startSpan to skip creating a SpanBuilder

eric-ming2:
# What Does This Do
This PR should not change anything semantically; `startSpan` and `buildSpan` should behave the same as before.
# Motivation
Calling `startSpan` creates the span and builds the SpanContext directly, instead of creating a SpanBuilder. This should reduce the memory allocation overhead slightly for each span created.
# Additional Notes
I added tests to confirm that spans created through `startSpan` and `buildSpan` are identical. Currently, `buildSpan` is only used in tests, and has more functionality if you use methods such as `withTag` or `withSpanType`.


dd-trace-core/src/main/java/datadog/trace/core/CoreTracer.java
```diff
@@ -413,12 +413,12 @@ private CoreTracer(
     assert taggedHeaders != null;
 
     this.timeSource = timeSource == null ? SystemTimeSource.INSTANCE : timeSource;
-    this.startTimeNano = this.timeSource.getCurrentTimeNanos();
```
eric-ming2, line 416: Are the unnecessary `this.` supposed to be in the code? My IntelliJ automatically deletes them, and spotlessApply does not bring them back.

dd-trace-core/src/main/java/datadog/trace/core/CoreTracer.java
```diff
@@ -1054,137 +1217,16 @@ public CoreSpanBuilder withTag(final String tag, final Object value) {
      * @return the context
      */
     private DDSpanContext buildSpanContext() {
```
devinsba, line 1219: I would reorder the methods to put the two `buildSpanContext()` methods next to each other

dd-trace-core/src/main/java/datadog/trace/core/CoreTracer.java
devinsba: You might want to turn off "Save Actions" plugin. It will help with performance too

dd-trace-core/src/test/groovy/datadog/trace/core/CoreStartSpanTest.groovy
```diff
@@ -0,0 +1,79 @@
+package datadog.trace.core
+
+import datadog.trace.api.DDId
+import datadog.trace.common.writer.ListWriter
+import datadog.trace.core.test.DDCoreSpecification
+
+class CoreStartSpanTest extends DDCoreSpecification {
```
devinsba, line 7: Maybe rename to `CoreTracerStartSpanTest` so that it's easier to correlate by file name

devinsba (approved):
With these changes are there uses of `AgentTracer.SpanBuilder` remaining? Seems like probably should be close to being able to delete it.

dd-trace-core/src/main/java/datadog/trace/core/CoreTracer.java
eric-ming2: Done, thanks

eric-ming2:
> With these changes are there uses of `AgentTracer.SpanBuilder` remaining? Seems like probably should be close to being able to delete it.

Yes, there are a lot of uses. Whenever a test or instrumentation calls `buildSpan`, `AgentTracer.SpanBuilder` is used. Most of them can be replaced with code that uses `startSpan` instead. However, as mentioned in the Additional Notes, there is some functionality only achievable using an `AgentTracer.SpanBuilder`.

Do I have to go-ahead to start replacing `buildSpan` where it isn't necessary? This involves changes in a lot of files and I don't want to break anything.

devinsba:
I probably wouldn't. I was just wondering. Are there any uses in the actual instrumentations? or is it just tests?

eric-ming2:
> I probably wouldn't. I was just wondering. Are there any uses in the actual instrumentations? or is it just tests?

Good point, it's only used in tests so I shouldn't have to change anything. Thanks :)

eric-ming2:
Because I am unable to create a benchmark that clearly shows a performance benefit from this change, I will not be merging it. I personally think that this change makes the code harder to read because now there are two paths to create a span (not every span is built with a SpanBuilder).

PR #3479, opened 2022-04-01
vertx: pass converted request body / path tokens

cataphract:
# What Does This Do

Adds support in vertx for transmitting to the IG extract path tokensas well as converted request entities (only multipart/urlencoded and json).


cataphract:
I'm moving the urlencoded/multipart decoding to the netty instrumentation.

ValentinZakharov (approved):

dd-java-agent/instrumentation/netty-4.1/src/test/groovy/Netty41ServerTest.groovy
```diff
@@ -80,6 +85,32 @@ class Netty41ServerTest extends HttpServerTest<EventLoopGroup> {
                         content = Unpooled.copiedBuffer(endpoint.bodyForQuery(uri.query), CharsetUtil.UTF_8)
                         response = new DefaultFullHttpResponse(HTTP_1_1, HttpResponseStatus.valueOf(endpoint.status), content)
                         break
+                      case BODY_URLENCODED:
+                        if (msg instanceof FullHttpRequest) {
```
bantonsson, line 89: Is this always true (at least in our tests)?

bantonsson (approved):

dd-java-agent/instrumentation/netty-4.1/src/test/groovy/Netty41ServerTest.groovy
cataphract: It should always be true because I added `HttpObjectAggregator` to the pipeline.

PR #3482, opened 2022-04-06
[CIAPP][Agentless] Payload mapper for `CITestCycle` track.

drodriguezhdez:
As part of the initiative to transition CI Visibility to the agentless mode, the traces generated by the instrumentations need to be serialized in the expected format for the new HTTP public intake.

This PR creates a new `CiTestCycleMapperV1` serialization mapper to transform the trace to the expected format. Additionally, it applies some refactors to the current mappers for the Trace track to reuse all the constants already implemented.

The change in this PR only creates the mapper and has no collateral effects on the current configuration and/or behavior.


dd-trace-core/src/main/java/datadog/trace/common/writer/common/Payload.java
```diff
@@ -1,4 +1,4 @@
-package datadog.trace.common.writer.ddagent;
+package datadog.trace.common.writer.common;
```
devinsba, line 1: I probably would drop the second common here:
```suggestion
package datadog.trace.common.writer;
```

dd-trace-core/src/main/java/datadog/trace/civisibility/writer/ddintake/CiTestCycleMapperV1.java
```diff
@@ -0,0 +1,216 @@
...
+import java.nio.channels.WritableByteChannel;
+import java.nio.charset.StandardCharsets;
+import java.util.Arrays;
+import java.util.List;
+import java.util.Map;
+import okhttp3.RequestBody;
+
+public class CiTestCycleMapperV1 implements RemoteMapper {
+
+  public static final byte[] VERSION = "version".getBytes(StandardCharsets.ISO_8859_1);
```
bantonsson, line 24: Why not `UTF_8`? I know they overlap for characters <= 127, but still.

dd-trace-core/src/main/java/datadog/trace/common/writer/ddagent/TraceMapperV0_4.java
```diff
@@ -18,19 +18,6 @@
 
 public final class TraceMapperV0_4 implements TraceMapper {
 
-  public static final byte[] SERVICE = "service".getBytes(ISO_8859_1);
```
bantonsson, line 21: Wow, why was this not using `UTF_8`?

dd-trace-core/src/test/groovy/datadog/trace/common/writer/DDAgentApiTest.groovy
```diff
@@ -12,7 +12,7 @@ import datadog.trace.api.sampling.SamplingMechanism
 import datadog.trace.common.writer.ddagent.DDAgentApi
 import datadog.communication.ddagent.DDAgentFeaturesDiscovery
 import datadog.trace.common.writer.ddagent.DDAgentResponseListener
-import datadog.trace.common.writer.ddagent.Payload
+
 import datadog.trace.common.writer.ddagent.TraceMapperV0_4
```
bantonsson, line 16: Why the empty line?

dd-trace-core/src/test/groovy/datadog/trace/common/writer/PayloadDispatcherTest.groovy
```diff
@@ -6,7 +6,7 @@ import datadog.trace.api.sampling.PrioritySampling
 import datadog.trace.api.sampling.SamplingMechanism
 import datadog.trace.common.writer.ddagent.DDAgentApi
 import datadog.communication.ddagent.DDAgentFeaturesDiscovery
-import datadog.trace.common.writer.ddagent.Payload
+
 import datadog.trace.common.writer.ddagent.PayloadDispatcher
```
bantonsson, line 10: Here as well?

bantonsson (approved):
LGTM. Only weird thing is `StandardCharsets.ISO_8859_1` which should be `StandardCharsets.UTF_8`, but that was there before.

PR #3489, opened 2022-04-11
Test java 17 in CI

bantonsson:
# What Does This Do

Adds Java 17 as one of the tested versions in CI.

# Motivation

Test on the latest LTS release

# Additional Notes

The RMI instrumentation won't work on Java 17 right now since it inherits across modules in an unsupported way. This needs to be fixed.

### DIFF
--- a/dd-java-agent/instrumentation/elasticsearch/rest-5/rest-5.gradle
+++ b/dd-java-agent/instrumentation/elasticsearch/rest-5/rest-5.gradle
@@ -1,6 +1,9 @@
 // Set properties before any plugins get loaded
 ext {
   minJavaVersionForTests = JavaVersion.VERSION_1_8
+  // TODO Java 17: The embedded elastic search server doesn't work on Java 17
+  //  This can be changed to use test-containers
+  maxJavaVersionForTests = JavaVersion.VERSION_15
 }
 
 muzzle {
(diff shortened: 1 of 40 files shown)


dd-java-agent/instrumentation/elasticsearch/rest-5/rest-5.gradle
```diff
@@ -1,6 +1,7 @@
 // Set properties before any plugins get loaded
 ext {
   minJavaVersionForTests = JavaVersion.VERSION_1_8
+  maxJavaVersionForTests = JavaVersion.VERSION_15
```
devinsba, line 4: I'm guessing elasticsearch server doesn't start with Java 17 too? Might be good to have comments for all these things

devinsba (approved):
It's less changes than I expected. We can get around the cassandra/elasticsearch issues if someone can figure out getting docker working on our circleci runners

dd-java-agent/instrumentation/elasticsearch/rest-5/rest-5.gradle
bantonsson: Yes, you are right. I should go through them and add comments, and a consistent `TODO Java 17 ...` where applicable.

bantonsson:
@devinsba Since our CircleCI runners are `docker` you can't run another `docker` inside that `docker`, as far as I can tell, but you can have other `docker` containers running at the same time (like we do for `memcached`, `rabbitmq`, `mongo`, et.c.) I don't know how much that weighs down the machine though, since they run for the whole duration of the testing.

PR #3493, opened 2022-04-11
Reuse named pipe sockets

randomanderson:
# What Does This Do
Named pipe sockets are stored in a map and reused when requesting the same named pipe.

# Motivation
Named pipes can be opened at most twice (1 client and 1 server). The previous implementation did not take that into account. Different subsystems (tracing, statsd, profiling, etc) opened their own named pipe sockets. This lead to a race where the first subsystem was able to use the named pipe and the rest were not.

# Additional Notes
This is untested because we have an inability to test on windows.

### DIFF
--- a/utils/socket-utils/src/main/java/datadog/common/socket/NamedPipeSocketFactory.java
+++ b/utils/socket-utils/src/main/java/datadog/common/socket/NamedPipeSocketFactory.java
@@ -10,15 +10,8 @@ import javax.net.SocketFactory;
 
 public class NamedPipeSocketFactory extends SocketFactory {
   private static final String NAMED_PIPE_PREFIX = "\\\\.\\pipe\\";
+  private static final Map<File, NamedPipeSocket> fileToSocket = new HashMap<>();
   private final File pipe;
-
-  // Use enum singleton pattern.
-  // This ensures there is at most one INSTANCE per jvm
-  private enum SingletonHolder {
-    INSTANCE;
-
-    final Map<File, NamedPipeSocket> fileToSocket = new HashMap<>();
-  }
 
   public NamedPipeSocketFactory(String pipeName) {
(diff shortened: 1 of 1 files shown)


utils/socket-utils/src/main/java/datadog/common/socket/NamedPipeSocketFactory.java
```diff
@@ -4,12 +4,22 @@
...
+import java.util.Map;
 import javax.net.SocketFactory;
 
 public class NamedPipeSocketFactory extends SocketFactory {
   private static final String NAMED_PIPE_PREFIX = "\\\\.\\pipe\\";
   private final File pipe;
 
+  // Use enum singleton pattern.
+  // This ensures there is at most one INSTANCE per jvm
+  private enum SingletonHolder {
```
bantonsson, line 17: This feels like unnecessary boilerplate for what is essentially
```
private static final Map<File, NamedPipeSocket> fileToSocket = new HashMap<>();
```
and used like
```
synchronized(fileToSocket) {
...
}
```

dougqh: The one benefit of the separate class is that the class initialization lock is basically free.  Once initialized, the check will be optimized away by the JIT.

While I know its idiomatic, I don't like using an enum for this because enums while syntactically concise produce really bloated bytecode.

I also probably wouldn't bother with the holder, since we end up needing a synchronized block anyway.

randomanderson: I might have been overly worried about danger of statics being initialized more than once when a class is loaded from multiple classloaders. I checked and `NamedPipeSocketFactory` is in the `shared` classloader so this should never be an issue. I'll make the changes.

PR #3501, opened 2022-04-14
[AppSec] Skip converting Strongly Encapsulated Internals for WAF

ValentinZakharov:
# What Does This Do
Prevents decomposition of object that are belongs to Strongly Encapsulated Java Internals, in order to avoid throwing [InaccessibleObjectException](https://docs.oracle.com/javase/9/docs/api/java/lang/reflect/InaccessibleObjectException.html) in Java 9+. Represents such objects as strings instead.

# Motivation
Customer reported AppSec false positive issue, when request body contains:
`<error: Unable to make field private final int java.time.LocalDate.year accessible: module java.base does not "opens java.time" to unnamed module @34f15e65>`
and
`<error: unable to make field private final int java.lang.string java.lang.enum.name accessible: module java.base does not "opes java.time" to unnamed module @34f15e65>`
Additionally customer emphasized, they are using [Dropwizard](https://www.dropwizard.io/) running on Java 17.

# Additional Notes
If the request body contains a complex object, AppSec converts it in java standard datatypes with strings for WAF analysis. In order data converted (into map of strings, array of string, etc.) can be properly consumed by WAF.

Decomposition of complex object is done with Java Reflection. Starting from the Java 9, there were introduced Strongly Encapsulated Internals (see: [JEP-260](https://openjdk.java.net/jeps/260), [JEP-396](https://openjdk.java.net/jeps/396), [JEP-403](https://openjdk.java.net/jeps/403)), which made some of standard java classes inaccessible for Reflection. As the result, decomposition throws [InaccessibleObjectException](https://docs.oracle.com/javase/9/docs/api/java/lang/reflect/InaccessibleObjectException.html) for classes like [LocalDate](https://docs.oracle.com/javase/8/docs/api/java/time/LocalDate.html) or [Enum](https://docs.oracle.com/javase/8/docs/api/java/lang/Enum.html) when running on Java 16+.


cataphract:
This doesn't have to do with the java.time or java.lang  specifically right? It's a general JPMS-related problem, no?

ValentinZakharov:
Yes, that's correct, I'm not sure if any classes from another `java.*` packages might be used as POJO. Please correct me If I'm missing something.

bantonsson (approved):

PR #3502, opened 2022-04-14
Fix URL parsing for AWS MySql JDBC Driver

eric-ming2:
Correctly parses JDBC URIs with the prefix `jdbc:mysql:aws://`.

This does not fix the larger issue, so AWS MySQL is still currently unsupported.


dd-java-agent/agent-bootstrap/src/main/java/datadog/trace/bootstrap/instrumentation/jdbc/JDBCConnectionUrlParser.java
```diff
@@ -27,11 +27,13 @@
 public enum JDBCConnectionUrlParser {
   GENERIC_URL_LIKE() {
     @Override
-    DBInfo.Builder doParse(final String jdbcUrl, final DBInfo.Builder builder) {
+    DBInfo.Builder doParse(String jdbcUrl, final DBInfo.Builder builder) {
+      if (jdbcUrl.startsWith("mysql:aws://")) {
+        jdbcUrl = "mysql://" + jdbcUrl.substring(12);
+      }
```
devinsba, line 33: I don't know if this change is necessary, none of the code below uses the protocol portion of the URI object

eric-ming2: In my testing, I found that
`final URI uri = new URI(jdbcUrl);`
uses the protocol portion, and the hostName is not found when "aws" is there. I will double check.

dd-java-agent/agent-bootstrap/src/main/java/datadog/trace/bootstrap/instrumentation/jdbc/JDBCConnectionUrlParser.java
```diff
@@ -175,7 +177,7 @@ DBInfo.Builder doParse(final String jdbcUrl, final DBInfo.Builder builder) {
       }
       final int protoLoc = jdbcUrl.indexOf("://");
       final int typeEndLoc = dbInfo.getType().length();
-      if (protoLoc > typeEndLoc) {
+      if (protoLoc > typeEndLoc && !jdbcUrl.startsWith("mysql:aws://")) {
```
devinsba, line 180: How about:
```suggestion
      if (protoLoc > typeEndLoc && jdbcUrl.substring(typeEndLoc + 1, protoLoc) != "aws") {
```

devinsba: Since we would expect this to fall through to the next if block also, we need to add `.subtype(jdbcUrl.substring(typeEndLoc + 1, protoLoc));` to that line

dd-java-agent/agent-bootstrap/src/main/java/datadog/trace/bootstrap/instrumentation/jdbc/JDBCConnectionUrlParser.java
eric-ming2: Yeah the default URI parser doesn't work when the prefix is 
`mysql:aws`
and I'm not sure why. Just removing "aws" and pretending it's a regular mysql URL feels kind of hacky so I'm open to other suggestions.

dd-java-agent/agent-bootstrap/src/main/java/datadog/trace/bootstrap/instrumentation/jdbc/JDBCConnectionUrlParser.java
eric-ming2: Much cleaner, thanks!

dd-java-agent/agent-bootstrap/src/main/java/datadog/trace/bootstrap/instrumentation/jdbc/JDBCConnectionUrlParser.java
devinsba: Ah, I see it in the javadoc now. So in this case URI isn't the right tool for this job. https://docs.oracle.com/javase/7/docs/api/java/net/URI.html#getScheme()

> The scheme component of a URI, if defined, only contains characters in the alphanum category and in the string "-.+".

devinsba: I think we should trim the schema off at the call site instead, it doesn't need to be included in order to parse correctly

eric-ming2: Done, please let me know what you think!

devinsba (approved):

PR #3505, opened 2022-04-18
[CIAPP][Agentless][Writer] Decouple common buffer+transport logic from DDAgent details

drodriguezhdez:

This PR refactors the code to decouple buffering + common transport logic from specific Datadog Agent implementation details. For that purpose, some common abstract classes and interfaces were created. 

The change in this PR only refactors the code and has no collateral effects on the current configuration and/or behavior. It has no changes related to the agentless mode, just the preparation.

Tested on a spring-boot application.
image

image

### DIFF
--- a/dd-trace-core/src/main/java/datadog/trace/common/writer/PayloadDispatcher.java
+++ b/dd-trace-core/src/main/java/datadog/trace/common/writer/PayloadDispatcher.java
@@ -1,16 +1,11 @@
 package datadog.trace.common.writer;
 
-import datadog.communication.RemoteFeaturesDiscovery;
-import datadog.communication.ddagent.DDAgentFeaturesDiscovery;
 import datadog.communication.monitor.Monitoring;
 import datadog.communication.monitor.Recording;
 import datadog.communication.serialization.ByteBufferConsumer;
 import datadog.communication.serialization.FlushingBuffer;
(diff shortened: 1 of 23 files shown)


dd-trace-core/src/main/java/datadog/trace/common/writer/PayloadDispatcher.java
```diff
@@ -58,39 +56,34 @@ public void onDroppedTrace(int spanCount) {
...
       }
-      String tracesUrl = featuresDiscovery.getTraceEndpoint();
-      if (DDAgentFeaturesDiscovery.V5_ENDPOINT.equalsIgnoreCase(tracesUrl)) {
-        this.traceMapper = new TraceMapperV0_5();
-      } else if (null != tracesUrl) {
-        this.traceMapper = new TraceMapperV0_4();
-      }
-      if (null != traceMapper && null == packer) {
+
+      mapper = mapperDiscovery.getMapper();
```
drodriguezhdez, line 75: Additional layer added to hide mapper selection details in the `PayloadDispatcher` class.

When the DD Agent is used, the mapper discovery class will be `DDAgentMapperDiscovery` where the features discovery instance is used.

bantonsson (approved):

PR #3506, opened 2022-04-18
[Serverless] Auto-instrument AWS Lambda Handler

maxday:
# What Does This Do

Add support for AWS Handler auto-instrumentation

Note : Since the invocation span is created in the extension (to unify logic to detect downstream AWS services across all AWS Lambda supported runtimes), we need to create a `dummy span` with the exact same spanID + traceID. This allows any newly created span inherit from the correct context. 

Screen Shot 2022-04-18 at 9 33 08 AM

# Motivation

Code-free instrumentation for AWS Lambda


dd-java-agent/agent-bootstrap/agent-bootstrap.gradle
```diff
@@ -20,6 +20,12 @@ dependencies {
   api project(':internal-api')
   api project(':internal-api:internal-api-8')
   api project(':dd-java-agent:agent-logging')
+
+  implementation deps.okhttp
+
+  api 'com.fasterxml.jackson.core:jackson-core:2.11.0'
+  api 'com.fasterxml.jackson.core:jackson-databind:2.11.0'
```
maxday, line 27: I am not sure how bas this is for non serverless usage
We definitely need Jackson (or similar) to send a JSON version of the context / user event

randomanderson (commented):
I'll start the ball rolling (this is not a full review, just a few things I noticed):

* `DummyContext` is not needed. Since the headers returned are the standard datadog headers, you can use the normal extraction mechanism.  See `propagate().extract()`

There are problems with the dependencies. Adding classes to the bootstrap means they will be at the top of the classpath hierarchy, with something like jackson, this is sure to cause conflicts.

(normally)
* `LambdaHandler` should be in the instrumentation module and listed as a helper class
* Dependencies should be listed in the instrumentation module
* We use `Moshi` for JSON serialization

The bigger issue is that all classes used by the instrumentation need to be injected into the instrumented classloader. So even just using okhttp and moshi means injecting a lot of classes (all of which have to be listed as helpers).

The correct solution is probably implementing the `LambdaHandler` in `dd-trace-core` and having the instrumentation call that. This way, there are no extra dependencies, you can reuse the Okhttp client

dd-trace-core/src/main/java/datadog/trace/core/CoreTracer.java
```diff
@@ -1210,6 +1212,15 @@ private DDSpanContext buildSpanContext() {
           samplingMechanism = extractedContext.getSamplingMechanism();
           endToEndStartTime = extractedContext.getEndToEndStartTime();
           baggage = extractedContext.getBaggage();
+        } else if (parentContext instanceof DummyLambdaContext) {
+          final DummyLambdaContext dlc = (DummyLambdaContext) parentContext;
+          traceId = dlc.getTraceId();
+          spanId = dlc.getSpanId();
+          parentSpanId = DDId.ZERO;
+          samplingPriority = PrioritySampling.UNSET;
```
hghotra, line 1220: This should also be set in the same way as `TraceId` & `SpanId`

dd-java-agent/agent-bootstrap/agent-bootstrap.gradle
maxday: removed

dd-java-agent/agent-bootstrap/agent-bootstrap.gradle
```diff
@@ -20,6 +20,12 @@ dependencies {
   api project(':internal-api')
   api project(':internal-api:internal-api-8')
   api project(':dd-java-agent:agent-logging')
+
+  implementation deps.okhttp
+
+  api 'com.fasterxml.jackson.core:jackson-core:2.9.10'
+  api 'com.fasterxml.jackson.core:jackson-databind:2.9.10'
```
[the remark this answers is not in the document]
maxday: ✅

dd-trace-core/src/main/java/datadog/trace/core/CoreTracer.java
maxday: ✅

internal-api/src/main/java/datadog/trace/bootstrap/instrumentation/api/DummyLambdaContext.java
```diff
@@ -0,0 +1,53 @@
+package datadog.trace.bootstrap.instrumentation.api;
+
+import datadog.trace.api.DDId;
+import datadog.trace.api.sampling.PrioritySampling;
+import java.util.Collections;
+import java.util.Map;
+
+public class DummyLambdaContext implements AgentSpan.Context {
```
bantonsson, line 8: I think that this context is not needed, and the code could just use the `ExtractedContext` and only fill in the interesting values in it. That would also mean that the changes to `buildSpanContext` can be removed as well.

maxday: you're right! I've renamed it to `LambdaContext` which now extends `ExtractedContext`. This has also drastically simplified the changes to CoreTracer

bantonsson (approved):
LGTM

 There is failure in the check task that needs to be fixed. You can see the problem here https://output.circle-artifacts.com/output/job/31aaaaf5-9915-4b41-8399-e4e2a6bc9817/artifacts/0/reports/dd-trace-core/codenarc/test.html

If you want to run the check locally, then do:
```
./gradlew dd-trace-core:check -PskipTests
```

bantonsson:
Also, please `squash` when merging, or simply `squash` locally and `push -f`. We don't want all these small commits in the history.

bantonsson:
Some more check failures 😉 

This will run `check` for the whole project:

```
./gradlew check -PskipTests
```

hghotra (approved):
👍🏼

devinsba (approved):

Issue #3484, opened 2022-04-08
AppSec: Logs full of warnings "java.lang.IllegalStateException: Raw URI set already"


jokalli2:
Hi,

We have enabled the AppSec functionality. It seems to be working fine, but our service logs are full of warnings. We are using the latest dd-trace-java version and the service is Play Framework 2.8.

The logged error:

[dd.trace 2022-04-08 10:23:53:026 +0300] [application-akka.actor.default-dispatcher-263] WARN datadog.trace.api.gateway.InstrumentationGateway - Callback for EventType{name='server.request.method.uri.raw'} threw.
--
java.lang.IllegalStateException: Raw URI set already
at com.datadog.appsec.gateway.AppSecRequestContext.setRawURI(AppSecRequestContext.java:172)
at com.datadog.appsec.gateway.GatewayBridge$MethodAndRawURICallback.apply(GatewayBridge.java:376)
at com.datadog.appsec.gateway.GatewayBridge$MethodAndRawURICallback.apply(GatewayBridge.java:361)
at datadog.trace.api.gateway.InstrumentationGateway$6.apply(InstrumentationGateway.java:151)
at datadog.trace.api.gateway.InstrumentationGateway$6.apply(InstrumentationGateway.java:145)
at datadog.trace.bootstrap.instrumentation.decorator.HttpServerDecorator.callIGCallbackURI(HttpServerDecorator.java:299)
at datadog.trace.bootstrap.instrumentation.decorator.HttpServerDecorator.onRequest(HttpServerDecorator.java:153)
at datadog.trace.instrumentation.play26.PlayHttpServerDecorator.onRequest(PlayHttpServerDecorator.java:113)
at play.api.mvc.ActionBuilder$anon$9.apply(Action.scala:384)
at play.api.mvc.Action.$anonfun$apply$4(Action.scala:82)
at play.api.libs.streams.StrictAccumulator.$anonfun$mapFuture$4(Accumulator.scala:168)
at scala.util.Try$.apply(Try.scala:210)
at play.api.libs.streams.StrictAccumulator.$anonfun$mapFuture$3(Accumulator.scala:168)
at scala.Function1.$anonfun$andThen$1(Function1.scala:85)
at scala.Function1.$anonfun$andThen$1(Function1.scala:85)
at scala.Function1.$anonfun$andThen$1(Function1.scala:85)
at play.api.libs.streams.StrictAccumulator.run(Accumulator.scala:199)
[...]

PR #3517, opened 2022-04-21
[AppSec] Play/Akka double instrumentation

ValentinZakharov:
# What Does This Do
When setting the same data in the context of the request, the same data will be ignored for further analysis using AppSec WAF.

# Motivation
Issue reported by customer: #3484 

# Additional Notes
Different instrumentations (Play and Akka) populate same data to distinct spans, but all of them consolidates in same AppSecRequestContext, which cause forbidden data replacement. As a workaround we ignores replacement if the values are the same.

### DIFF
--- a/dd-java-agent/appsec/src/main/java/com/datadog/appsec/gateway/AppSecRequestContext.java
+++ b/dd-java-agent/appsec/src/main/java/com/datadog/appsec/gateway/AppSecRequestContext.java
@@ -84,7 +84,9 @@ public class AppSecRequestContext implements DataBundle, Closeable {
         continue;
       }
       Object prev = persistentData.putIfAbsent(address, value);
-      if (prev != null) {
+      if (prev == value) {
+        continue;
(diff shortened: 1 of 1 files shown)


dd-java-agent/appsec/src/main/java/com/datadog/appsec/gateway/AppSecRequestContext.java
```diff
@@ -84,7 +84,7 @@ public void addAll(DataBundle newData) {
         continue;
       }
       Object prev = persistentData.putIfAbsent(address, value);
-      if (prev != null) {
+      if (prev != null && prev != value) {
```
bantonsson, line 87: This code will lead to logging even if we _push_ the same value here. Maybe do a short circuit first?

```
if (prev == value) {
  return;
} else if (prev != null) {
  log.warn(...);
}
...
```

devinsba (approved):

PR #3529, opened 2022-04-28
Remove okhttp dependency from dd-trace-ot

theon:
# What Does This Do

Remove okhttp and okio dependency from dd-trace-ot

# Motivation

We have a lambda that is currently complaining because the JAR file is too big. We depend on only dd-trace-ot and I can see it is pulling in okhttp and okio, but can't see where it is using them.

(draft PR for now as I am shamelessly using this PR to run the tests as struggling to run all the tests locally)

# Additional Notes

### DIFF
--- a/dd-trace-ot/dd-trace-ot.gradle
+++ b/dd-trace-ot/dd-trace-ot.gradle
@@ -42,8 +42,6 @@ dependencies {
   api group: 'io.opentracing.contrib', name: 'opentracing-tracerresolver', version: '0.1.0'
 
   api deps.slf4j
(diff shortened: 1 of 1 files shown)


theon:
Ah dd-trace-ot, bundles a bunch of the other modules like communication. Ignore me!

devinsba:
I think this is a good point from a serverless/lambda point of view. I believe trace reporting in these scenarios is done using logging so the http client really shouldn't be needed (I could be wrong here)

If that is the case you could remove the class files from the jar safely to make the file smaller

PR #3533, opened 2022-05-04
Stop intermittent double Jersey requests

bantonsson:
# What Does This Do

Stops intermittent double Jersey requests by closing the response.

# Motivation

Tests sometimes fail with one extra span, that is a duplicate of the server span.

# Additional Notes


mcculls (approved):
makes sense, thanks

devinsba (approved):

PR #3537, opened 2022-05-06
[CIAPP] Fix span duration for skipped tests

drodriguezhdez:
This PR fixes the span duration for the skipped test spans instrumented by CI Visibility.

**Root cause**: The `finish(duration)` method does expect the end time in microseconds. As the `start time` is based on nanoseconds, the duration was completely wrong for skipped tests.

**Proposed solution**: Adds a new `finishWithDuration` method in the internal DDSpan API to allow us to set a fixed duration.

**Context**: In CI Visibility, the skipped tests need to have a duration of `0 ns`. This issue was not detected before because CI Visibility spans were fixed by the Datadog Agent. Now the CI Visibility traces can be sent directly to a public intake, this bug has been detected.

_**Note:**_ As the tracer has a special treatment for spans with duration == `0 ns`, for the moment let's set the duration to `1 ns` to unblock skipped tests for agentless.

Tested in staging
image

### DIFF
--- a/dd-java-agent/instrumentation/junit-4.10/src/main/java/datadog/trace/instrumentation/junit4/TracingListener.java
+++ b/dd-java-agent/instrumentation/junit-4.10/src/main/java/datadog/trace/instrumentation/junit4/TracingListener.java
@@ -119,7 +119,7 @@ public class TracingListener extends RunListener {
       DECORATE.afterStart(span, version);
       DECORATE.onTestIgnored(span, description, testName, reason);
       DECORATE.beforeFinish(span);
-      span.finish(span.getStartTime());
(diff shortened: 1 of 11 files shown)


drodriguezhdez:
> So parts of the current implementation rely on the fact that a 0 duration means that the Span isn't finished, see here: https://github.com/DataDog/dd-trace-java/pull/3537/files#diff-d0ed96aa9f576a1dbca7a4a49831e10613cbdc9d5ef4523f966ac28783d680ceR151-R153

Right. For the moment, we set the duration to `1 ns` to unblock skipped tests for agentless.

bantonsson (approved):

PR #3539, opened 2022-05-09
[CIAPP][Agentless] Allow setting host URL via configuration

drodriguezhdez:
This PR adds the `dd.civisibility.agentless.url` config property (or environment variable equivalent) to allow configuring the Intake host URL externally. This will be particularly useful for configuring e2e tests on CI Visibility agentless mode.

The change in this PR is a feature used only by CI Visibility. It has no changes that might affect customers in production mode for APM Java.

### DIFF
--- a/internal-api/src/main/java/datadog/trace/api/Config.java
+++ b/internal-api/src/main/java/datadog/trace/api/Config.java
@@ -966,18 +966,17 @@ public class Config {
         configProvider.getBoolean(
             CIVISIBILITY_AGENTLESS_ENABLED, DEFAULT_CIVISIBILITY_AGENTLESS_ENABLED);
 
-    final String ciVisibilityAgentlessUrlStr =
-        configProvider.getString(CIVISIBILITY_AGENTLESS_URL, null);
-    URL parsedCiVisibilityUrl = null;
+    final String ciVisibilityAgentlessUrlStr = configProvider.getString(CIVISIBILITY_AGENTLESS_URL);
+    URI parsedCiVisibilityUri = null;
     if (ciVisibilityAgentlessUrlStr != null && !ciVisibilityAgentlessUrlStr.isEmpty()) {
       try {
-        parsedCiVisibilityUrl = new URL(ciVisibilityAgentlessUrlStr);
-      } catch (MalformedURLException ex) {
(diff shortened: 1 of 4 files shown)


internal-api/src/main/java/datadog/trace/api/Config.java
```diff
@@ -962,6 +966,23 @@ && isJavaVersionAtLeast(8)
         configProvider.getBoolean(
             CIVISIBILITY_AGENTLESS_ENABLED, DEFAULT_CIVISIBILITY_AGENTLESS_ENABLED);
 
+    final String ciVisibilityAgentlessUrlStr =
+        configProvider.getString(CIVISIBILITY_AGENTLESS_URL, null);
+    URL parsedCiVisibilityUrl = null;
+    if (ciVisibilityAgentlessUrlStr != null && !ciVisibilityAgentlessUrlStr.isEmpty()) {
+      try {
+        parsedCiVisibilityUrl = new URL(ciVisibilityAgentlessUrlStr);
```
ValentinZakharov, line 974: I would suggest to use URI conversion additionally (it throws `URISyntaxException`), because plain `new URL(...)` has very limited validation.
```
new URL(...).toURI()
```

internal-api/src/main/java/datadog/trace/api/Config.java
```diff
@@ -962,6 +966,23 @@ && isJavaVersionAtLeast(8)
         configProvider.getBoolean(
             CIVISIBILITY_AGENTLESS_ENABLED, DEFAULT_CIVISIBILITY_AGENTLESS_ENABLED);
 
+    final String ciVisibilityAgentlessUrlStr =
+        configProvider.getString(CIVISIBILITY_AGENTLESS_URL, null);
```
ValentinZakharov, line 970:
```suggestion
        configProvider.getString(CIVISIBILITY_AGENTLESS_URL);
```

internal-api/src/main/java/datadog/trace/api/Config.java
drodriguezhdez: Changed!

ValentinZakharov (approved):

PR #3540, opened 2022-05-09
[CIAPP][Agentless] Add normalization and truncation for CI Visibility Agentless mode

drodriguezhdez:

This PR creates the `DDIntakeTraceInterceptor` to apply normalization to the traces generated before sending them to the public HTTP intake. This normalization is similar to what the Datadog Agent is applying before sending it to the Datadog backend.

The change in this PR is a new feature used only by CI Visibility. It has no changes that might affect customers in production mode for APM Java.


dd-trace-core/src/main/java/datadog/trace/common/writer/ddintake/DDIntakeTraceInterceptor.java
```diff
@@ -0,0 +1,68 @@
...
+
+    if (span.getResourceName() == null || span.getResourceName().length() == 0) {
+      log.debug(
+          "Fixing malformed trace. Resource is empty (reason:resource_empty), setting span.resource={}: {}",
+          span.getOperationName(),
+          span);
+      span.setResourceName(span.getOperationName());
+    }
+
+    final Map<String, Object> tags = span.getTags();
```
bantonsson, line 49: Calling `span.getTags()` will give you an immutable copy of all the tags, which would be throwing exceptions at the `tags.put(...)` calls if the tests covered them.

internal-api/src/main/java/datadog/trace/util/TraceUtils.java
```diff
@@ -0,0 +1,192 @@
...
+    }
+    try {
+      final int code = Integer.parseInt(httpStatusCode);
+      return (code >= 100 && code < 600);
+    } catch (NumberFormatException ex) {
+      return false;
+    }
+  }
+
+  public static String normalizeTag(final String tag) {
```
bantonsson, line 90: It would be great with a doc comment describing the rules for the translations that the code is performing.

drodriguezhdez: Added!

dd-trace-core/src/main/java/datadog/trace/common/writer/ddintake/DDIntakeTraceInterceptor.java
drodriguezhdez: I changed it. Now the `DDSpanContext` is used to modify the tags. Let me know if that should work.

bantonsson (changes requested):
There is no need to do use the `DDSpanContext` in the interceptor.

> Why not just use span.getTag() and span.setTag()? The span.getTags() is costly and unnecessary.

dd-trace-core/src/main/java/datadog/trace/common/writer/ddintake/DDIntakeTraceInterceptor.java
drodriguezhdez: Gotcha. It seems the `MutableSpan` interface does not have "getTag" method. That's why I used the `getTags` method. I changed it to `DDSpan` instead.

bantonsson: 🤦🏼‍♂️ my bad. I just thought that `MutableSpan` would be a usable `interface` but of course not...

bantonsson (approved):

PR #3548, opened 2022-05-11
Fixed race condition

ValentinZakharov:
# What Does This Do
Prevent race condition that occurs if request end event, happens in two parallel threads (because of double instrumentation).

# Motivation
Customer reported about an exception that arise occasionally.

# Additional Notes

### DIFF
--- a/dd-java-agent/appsec/src/main/java/com/datadog/appsec/powerwaf/PowerWAFModule.java
+++ b/dd-java-agent/appsec/src/main/java/com/datadog/appsec/powerwaf/PowerWAFModule.java
@@ -239,12 +239,7 @@ public class PowerWAFModule implements AppSecModule {
     @Override
     public void onEvent(AppSecRequestContext reqCtx, EventType eventType) {
       if (eventType == EventType.REQUEST_END) {
-        Additive additive = reqCtx.getAdditive();
-        if (additive != null) {
-          additive.close();
-        }
-
-        reqCtx.setAdditive(null);
+        reqCtx.closeAdditive();
       }
     }
   }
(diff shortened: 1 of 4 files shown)


dd-java-agent/appsec/src/main/java/com/datadog/appsec/powerwaf/PowerWAFModule.java
```diff
@@ -327,22 +322,9 @@ private Powerwaf.ActionWithData doRunPowerwaf(
...
-          if (wafMetricsEnabled) {
-            metrics = ctxAndAddr.ctx.createMetrics();
-            reqCtx.setWafMetrics(metrics);
-          }
-        } else {
-          metrics = reqCtx.getWafMetrics();
-        }
-      }
+
+      Additive additive = reqCtx.getOrCreateAdditive(ctxAndAddr.ctx, wafMetricsEnabled);
```
bantonsson, line 326: It's not clear to me if there is only a race between multiple threads sending _request end_ or if there could be a _request end_ at the same time as other data becoming available. If the latter is possible, then the `Additive` could be closed while in use since there is no lock held while it's being used.

cataphract: The additive does serialize running and closing: https://github.com/sqreen/AgentJavaNative/blob/master/src/main/java/io/sqreen/powerwaf/Additive.java#L82-L113 and https://github.com/sqreen/AgentJavaNative/blob/master/src/main/java/io/sqreen/powerwaf/Additive.java#L130-L148

bantonsson: Thanks for the pointers to the code in the `Additive`, it makes my concern about closing while in use go away.

Also, from the following stack trace we can see that the failing thread is trying to close the `Additive` from a partial flush, which can most definitely happen while the another thread is either adding data or closing the `Additive` due to a proper request end.

```
Caused by: java.lang.IllegalStateException: This Additive is no longer online
	at io.sqreen.powerwaf.Additive.close(Additive.java:132)
	at com.datadog.appsec.gateway.AppSecRequestContext.close(AppSecRequestContext.java:323)
	at datadog.trace.agent.core.CoreTracer.write(CoreTracer.java:864)
	at datadog.trace.agent.core.PendingTrace.write(PendingTrace.java:283)
	at datadog.trace.agent.core.PendingTrace.partialFlush(PendingTrace.java:242)
	at datadog.trace.agent.core.PendingTrace.decrementRefAndMaybeWrite(PendingTrace.java:230)
	at datadog.trace.agent.core.PendingTrace.onPublish(PendingTrace.java:173)
	at datadog.trace.agent.core.DDSpan.finishAndAddToTrace(DDSpan.java:139)
	at datadog.trace.agent.core.DDSpan.finish(DDSpan.java:150)
```

bantonsson (approved):

PR #3552, opened 2022-05-12
[CIAPP][Agentless] Add retry policy for CI Visibility agentless mode.

drodriguezhdez:

This PR creates a retry policy mechanism for the agentless mode configured in the `DDIntakeApi` and `DDIntakeWriter` classes. The idea is to provide an exponential backoff mechanism to handle temporary backend errors for CI Visibility agentless mode.

This change only affects CI Visibility. The APM production mode is not impacted by this change.


dd-trace-core/src/main/java/datadog/trace/common/writer/RemoteWriter.java
```diff
@@ -100,8 +115,8 @@ public final long getCapacity() {
 
   @Override
   public boolean flush() {
-    if (!closed) { // give up after a second
-      if (traceProcessingWorker.flush(1, TimeUnit.SECONDS)) {
+    if (!closed) {
+      if (traceProcessingWorker.flush(flushTimeout, flushTimeoutUnit)) {
```
drodriguezhdez, line 119: After adding a retry policy for the DDIntakeApi, we need to parameterize the flush timeout to sync it with that. A test detected this. 

This change does not affect the `DDAgentWriter` because that will use the `1 second` flush timeout as it does today.

dd-trace-core/src/main/java/datadog/trace/common/writer/RemoteWriter.java
```diff
@@ -31,24 +31,39 @@ public abstract class RemoteWriter implements Writer {
...
-
   public final HealthMetrics healthMetrics;
 
   protected RemoteWriter(
       final RemoteApi api,
       final TraceProcessingWorker traceProcessingWorker,
       final PayloadDispatcher dispatcher,
       final HealthMetrics healthMetrics,
+      final int flushTimeout,
+      final TimeUnit flushTimeoutUnit,
```
drodriguezhdez, line 46: As we don't have `Duration` (Jdk 1.8) yet. Let me know if you consider we should have a dedicated class for this (encapsulate the value and the unit behind an instance object)

bantonsson: I think it's fine to have them separate. We do that in other places as well.

bantonsson (approved):

PR #3563, opened 2022-05-19
Add debugger config

jpbempel:
# What Does This Do

Add debugger config options

# Motivation

Config options required by upcoming Live Debugger product

# Additional Notes

### DIFF
--- a/dd-trace-api/src/main/java/datadog/trace/api/ConfigDefaults.java
+++ b/dd-trace-api/src/main/java/datadog/trace/api/ConfigDefaults.java
@@ -78,6 +78,17 @@ public final class ConfigDefaults {
   static final boolean DEFAULT_CIVISIBILITY_ENABLED = false;
   static final boolean DEFAULT_CIVISIBILITY_AGENTLESS_ENABLED = false;
 
+  static final boolean DEFAULT_DEBUGGER_ENABLED = false;
+  static final int DEFAULT_DEBUGGER_UPLOAD_TIMEOUT = 30; // seconds
+  static final int DEFAULT_DEBUGGER_UPLOAD_FLUSH_INTERVAL = 0; // ms, 0 = dynamic
+  static final boolean DEFAULT_DEBUGGER_CLASSFILE_DUMP_ENABLED = false;
+  static final int DEFAULT_DEBUGGER_POLL_INTERVAL = 1; // seconds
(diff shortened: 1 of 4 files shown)


internal-api/src/main/java/datadog/trace/api/Config.java
```diff
@@ -457,6 +483,22 @@ public class Config {
...
+  private final boolean debuggerClassFileDumpEnabled;
+  private final int debuggerPollInterval;
+  private final int debuggerDiagnosticsInterval;
+  private final boolean debuggerMetricEnabled;
+  private final String debuggerProbeFileLocation;
+  private final int debuggerBatchSize;
+  private final long debuggerMaxPayloadSize;
+  private final boolean debuggerVerifyByteCode;
+  private final boolean debuggerInstrumentTheWorld;
+  private final String debuggerExcludeFile;
```
devinsba, line 500: Do you want to add these fields to the toString() method too?

dd-trace-api/src/main/java/datadog/trace/api/ConfigDefaults.java
```diff
@@ -78,6 +78,18 @@ public final class ConfigDefaults {
   static final boolean DEFAULT_CIVISIBILITY_ENABLED = false;
   static final boolean DEFAULT_CIVISIBILITY_AGENTLESS_ENABLED = false;
 
+  static final boolean DEFAULT_DEBUGGER_ENABLED = false;
+  static final int DEFAULT_DEBUGGER_UPLOAD_TIMEOUT = 30; // seconds
+  static final int DEFAULT_DEBUGGER_UPLOAD_FLUSH_INTERVAL = 0; // ms, 0 = dynamic
```
devinsba, line 83: Is there a reason to have this one config in milliseconds while all the others are seconds?

devinsba (approved):

internal-api/src/main/java/datadog/trace/api/Config.java
jpbempel: ah true, thanks!

dd-trace-api/src/main/java/datadog/trace/api/ConfigDefaults.java
jpbempel: yes, flushing interval will be no more than 2 seconds here, so having ms precision here makes sense

PR #3564, opened 2022-05-19
Add support for OpenJ9

luhenry:
# What Does This Do

It adds an async-profiler-based controller that would be used for OpenJ9 support.

# Motivation

Support more platforms for Profiling.


dd-java-agent/agent-profiling/profiling-controller-async/src/main/java/com/datadog/profiling/controller/async/AsyncOngoingRecording.java
```diff
@@ -0,0 +1,56 @@
...
+
+  private final OngoingRecording recording;
+  private final Instant started = Instant.now();
+
+  AsyncOngoingRecording(AsyncProfiler asyncProfiler, String recordingName)
+      throws UnsupportedEnvironmentException {
+    log.debug("Creating new recording: {}", recordingName);
+    recording = asyncProfiler.start();
+    if (recording == null) {
+      throw new UnsupportedEnvironmentException("Failed to start auxiliary profiler for Async");
```
jbachorik, line 37:
```suggestion
      throw new UnsupportedEnvironmentException("Failed to start async profiler");
```

dd-java-agent/agent-profiling/profiling-controller/src/main/java/com/datadog/profiling/controller/Controller.java
```diff
@@ -29,5 +29,5 @@ public interface Controller {
    * @return the recording object created.
    */
   @Nonnull
-  OngoingRecording createRecording(@Nonnull String recordingName);
+  OngoingRecording createRecording(@Nonnull String recordingName) throws Exception;
```
jbachorik, line 32: Can this be changed to the specific exception?
Seems like only `UnsupportedEnvironmentException` can be thrown right now, correct?

jbachorik (changes requested):

jbachorik (approved):

PR #3571, opened 2022-05-25
add debugger-scala project

jpbempel:
# What Does This Do
Required to be able to compile Scala code on the fly in unit tests
Cannot easily integrated with target project as Java code is compiled
before Scala code, while we need the opposite. Hence easier to setup
a project we depend on.
Note also, I couldn't convert this code into Java as Scala nested classes are AFAICT not instantiable from Java:
`val run = new global.Run()`

# Motivation
For upcoming Live Debugger product

# Additional Notes


bantonsson (commented):
If this is only for testing, then I think it would be good to have that in the project name.

bantonsson (approved):

PR #3576, opened 2022-05-27
Upgrade jnr-unixsocket to avoid 'use after free' vulnerability

c-fraser:
# What Does This Do

Upgrades the `com.github.jnr:jnr-unixsocket` dependency to the latest version. 

# Motivation

`com.github.jnr:jnr-unixsocket:0.28` depends on `com.github.jnr:jnr-posix:3.0.61` which contains a ["high" security vulnerability](https://security.snyk.io/vuln/SNYK-JAVA-COMGITHUBJNR-1570422). 

The latest `jnr-unixsocket`, version `0.38.17`, [uses com.github.jnr:jnr-posix:3.1.15](https://github.com/jnr/jnr-unixsocket/blob/jnr-unixsocket-0.38.17/pom.xml#L77-L82), which has resolved the vulnerability.

### DIFF
--- a/gradle/dependencies.gradle
+++ b/gradle/dependencies.gradle
@@ -22,7 +22,7 @@ final class CachedData {
     kotlin        : "1.3.72",
     coroutines    : "1.3.0",
     dogstatsd     : "4.0.0",
-    jnr_unixsocket: "0.28",
+    jnr_unixsocket: "0.38.17",
     commons       : "3.2",
     mockito       : '4.4.0',
     moshi         : '1.9.2',


devinsba:
Rerunning to get past flaky tests. I'll merge once I have a green build (other than system-tests)

devinsba:
Ok, so we are going to have to approach this differently. You can see here that the new version requires Java 8 while we have to support 7 at this time

https://app.circleci.com/pipelines/github/DataDog/dd-trace-java/13681/workflows/4bae960f-d1b2-462a-aa9d-73c7496618ef/jobs/300835/tests

We may need to force the correct `jnr-posix` version in all the places the transitive is coming from like `utils/socket-utils/socket-utils.gradle`

bantonsson:
@c-fraser I took a look at the underlying [CVE](https://cve.mitre.org/cgi-bin/cvename.cgi?name=CVE-2014-4043) from 2014, and it is only a vulnerability in glibc prior to `2.20`, so unless you're running on an old OS, this has been fixed at the OS level.

PR #3580, opened 2022-06-02
Add jdbc support for aws-mysql-jdbc driver

devinsba:
# What Does This Do

# Motivation

# Additional Notes


Loganathanc:
i thought so... but it was not working. when i tried only those 2 classes alone.

Loganathanc:
i have tried with your snapshot jar today and its working fine... please push the changes in upcoming release. Thanks

devinsba:
Thanks for the update. Your code was missing two of the classes but your changes made me think it would just work now after the first bug we fixed.

Loganathanc:
when will be part of release? any idea?

Shyam-1990:
@devinsba : Any ETA on adding this change to the release. I have created ticket for the DD support team as well.

devinsba:
Hi, sorry. There should be a release this week.

PR #3582, opened 2022-06-03
Add debugger main project

jpbempel:
# What Does This Do

Implementation of the debugger java agent that handles:
 - Polling & deserialization of debugger configuration from
Remote Configuration
 - generating instrumentation based on probe definitions
 - 2 kinds of probes: snapshot & metric
 - serializing & uploading snapshots coming from probes collecting
data
 - emitting metrics coming from metric probes
 - uploading statuses of probes

# Motivation

Open sourcing Live Debugger product

# Additional Notes
There is no actual additional dependencies per se with Live Debugger and this project.
The only additional dependencies declared is ASM, but it was already packaged in shared part transitively with JNR. Here we are forcing a explicit version (9.2) and package in shared where it was already present.
We need to control the explicit version to be able to parse recent classfiles (JDK17 at least) and may bump it for newer JDK later


dd-java-agent/agent-debugger/src/main/java/com/datadog/debugger/agent/AllowListHelper.java
```diff
@@ -0,0 +1,49 @@
...
+import java.util.HashSet;
+import org.slf4j.Logger;
+import org.slf4j.LoggerFactory;
+
+/** Helper class for handling allowed classes and packages for instrumentation */
+public class AllowListHelper {
+  private static final Logger log = LoggerFactory.getLogger(AllowListHelper.class);
+
+  private final boolean allowAll;
+  private Trie packagePrefixTrie;
```
bantonsson, line 12: From a cursory glance, it feels like both the allow and deny lists could use the `datadog.trace.util.ClassNameTrie`, which is specialized for inclusion/exclusion.

jpbempel: right, updated, thanks!

dd-java-agent/agent-debugger/src/main/java/com/datadog/debugger/poller/PollerHttpClient.java
```diff
@@ -0,0 +1,100 @@
...
+  private final ScheduledExecutorService okHttpExecutorService;
+  private final OkHttpClient client;
+  private final Request request;
+
+  public PollerHttpClient(Request request, Duration requestTimeout) {
+    this.request = request;
+    // This is the same thing OkHttp Dispatcher is doing except thread naming and daemonization
+    okHttpExecutorService =
+        Executors.newSingleThreadScheduledExecutor(
+            new PollerThreadFactory("dd-debugger-poll-http-dispatcher"));
```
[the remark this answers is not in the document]
jpbempel: ok thanks for the info! will do this in a next PR

PR #3591, opened 2022-06-13
Attempt to prevent buffer overflow due to data race

jbachorik:
# What Does This Do

Adds the concept of 'captured' long sequence size to allow concurrent updates to the sequence without causing buffer overflow when persisting that sequence.

# Motivation

Currently, the profiler span context is serialized to a byte buffer which is pre-sized to the sum of sizes of all relevant interval sequences. However, the serialization is done in a subsequent step and this means that there might be few extra intervals added to each sequence in the meantime and writing them out would cause buffer overflow.

# Additional Notes

### DIFF
--- a/dd-java-agent/agent-profiling/profiling-context/src/main/java/com/datadog/profiling/context/LongSequence.java
+++ b/dd-java-agent/agent-profiling/profiling-context/src/main/java/com/datadog/profiling/context/LongSequence.java
@@ -238,12 +238,11 @@ final class LongSequence {
   }
 
   /**
-   * Get the last 'captured' size; if the size has not been captured yet capture and return it now
+   * Get the last 'captured' size
+   *
+   * @return the captured size or {@literal -1} if the size has not been captured yet
    */
   public int getCapturedSize() {
-    if (capturedSize == -1) {
-      capturedSize = size();
-    }
     return capturedSize;
   }
 
--- a/dd-java-agent/agent-profiling/profiling-context/src/main/java/com/datadog/profiling/context/LongSequence.java
+++ b/dd-java-agent/agent-profiling/profiling-context/src/main/java/com/datadog/profiling/context/LongSequence.java
@@ -76,6 +76,7 @@ final class LongSequence {
 
   private AtomicInteger state = new AtomicInteger(0);
   private final int limit;
+  private int capturedSize = -1;
 
   public LongSequence(Allocator allocator) {
     this(allocator, Integer.MAX_VALUE);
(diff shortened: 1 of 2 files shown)


dd-java-agent/agent-profiling/profiling-context/src/main/java/com/datadog/profiling/context/LongSequence.java
```diff
@@ -224,6 +225,28 @@ public int size() {
...
+   *
+   * @return the sequence size
+   */
+  public int captureSize() {
+    capturedSize = size();
+    return capturedSize;
+  }
+
+  /**
+   * Get the last 'captured' size; if the size has not been captured yet capture and return it now
```
bantonsson, line 241: Do we really need to have two different ones? Should there be _capturing_ that overwrites the previous capture?

bantonsson (approved):

dd-java-agent/agent-profiling/profiling-context/src/main/java/com/datadog/profiling/context/LongSequence.java
jbachorik: The idea was to have one method to store the current value and the other method to return that stored value regardless of how many times it is called. I think that would not be possible to achieve with just one method, right?

bantonsson: So I'm a bit confused about the concurrency involved here. I'll add more comments below.

dd-java-agent/agent-profiling/profiling-context/src/main/java/com/datadog/profiling/context/ProfilerTracingContextTracker.java
```diff
@@ -380,11 +380,10 @@ private ByteBuffer encodeIntervals() {
     int encodedDataLimit =
```
bantonsson, line 380: Could it be possible that we have more than one thread in this `encodeIntervals` method at once?

dd-java-agent/agent-profiling/profiling-context/src/main/java/com/datadog/profiling/context/ProfilerTracingContextTracker.java
```diff
@@ -399,6 +398,8 @@ private ByteBuffer encodeIntervals() {
       threadCount++;
       IntervalEncoder.ThreadEncoder threadEncoder = encoder.startThread(threadId);
       LongSequence rawIntervals = threadSequences.get(threadId);
+      int sequenceSize = rawIntervals.getCapturedSize();
```
bantonsson, line 401: If having more than one thread in here is possible, then the `getCapturedSize` here could return a bigger value than was used to compute the `totalSequenceSize`, and wouldn't that lead to a buffer overflow?

jbachorik: There should be only one thread doing the encoding.
The concurrent work we are trying to accommodate here is the spans being activated/deactivated when the trace is being written out.

bantonsson (approved):

luhenry (approved):

PR #3593, opened 2022-06-15
Add Agent.shutdown API

luhenry:
# What Does This Do

It triggers shutdown of ProfilingAgent to support profile upload on shutdown.

# Motivation

This is to be used by other integrations that want to guarantee that data will be flushed and uploaded before the shutdown of the process. It will be used by CI app for example to guarantee we have all profiling data.

# Additional Notes


luhenry:
@danielrs975 that is to be used in CI app to capture profiling data until the end of the process.

bric3 (approved):

PR #2188, opened 2020-12-10
Avoid dns lookup in mongo instrumentations

richardstartin:
`ServerAddress.getSocketAddress()` is implemented as follows, so isn't safe for a tracing library to call

```java
    /**
     * Gets the underlying socket address
     *
     * @return socket address
     */
    public InetSocketAddress getSocketAddress() {
        try {
            return new InetSocketAddress(InetAddress.getByName(host), port);
        } catch (UnknownHostException e) {
            throw new MongoSocketException(e.getMessage(), this, e);
        }
    }
```

PR #3595, opened 2022-06-16
Standardized tags

ValentinZakharov:
# What Does This Do
- Added tags `http.useragent`, `http.client_ip`
- Improved `http.url` - contains query string by default (can be disabled with `http.server.tag.query-string`)

# Motivation

# Additional Notes


dd-java-agent/agent-bootstrap/src/main/java/datadog/trace/bootstrap/Constants.java
```diff
@@ -14,6 +14,7 @@ public final class Constants {
    * datadog.trace.agent.test.SpockRunner#BOOTSTRAP_PACKAGE_PREFIXES_COPY
    */
   public static final String[] BOOTSTRAP_PACKAGE_PREFIXES = {
+    "com.google.re2j",
```
bantonsson, line 17: This should be the _moved_ package name to ensure that it is properly loaded in the bootstrap class loader.

dd-java-agent/agent-bootstrap/src/main/java/datadog/trace/bootstrap/instrumentation/decorator/HttpServerDecorator.java
```diff
@@ -240,27 +243,60 @@ private AgentSpan.Context.Extracted callIGCallbackStart(AgentSpan.Context.Extrac
...
-      igKeyClassifier.done();
-    }
+
+    final TriConsumer<RequestContext<Object>, String, String> reqHeaderCb =
+        cbp.getCallback(EVENTS.requestHeader());
+    final Function<RequestContext<Object>, Flow<Void>> reqHeaderDoneCb =
+        cbp.getCallback(EVENTS.requestHeaderDone());
+
+    HeadersClassifier classifier =
+        new HeadersClassifier() {
```
bantonsson, line 260: Please don't use inline classes when there is so much code involved. It makes it hard to read.

dd-java-agent/agent-bootstrap/src/main/java/datadog/trace/bootstrap/instrumentation/decorator/HttpServerDecorator.java
```diff
@@ -240,27 +243,60 @@ private AgentSpan.Context.Extracted callIGCallbackStart(AgentSpan.Context.Extrac
...
+          public boolean nextHeader(String name, String value) {
+            // Send IG event
+            if (reqHeaderCb != null) {
+              reqHeaderCb.accept(requestContext, name, value);
+            }
+            return true;
+          }
+
+          @Override
+          public void doneHeaders(Map<String, List<String>> headers) {
```
bantonsson, line 271: The whole point of the `AgentPropagation.KeyClassifier` is to not build up any extra `Map` of headers, and not do _expensive_ searches through it, and this does exactly the opposite. Now for every request even if AppSec is off, we will build up an extra `Map` for all the headers.

dd-java-agent/agent-bootstrap/src/main/java/datadog/trace/bootstrap/instrumentation/decorator/HttpServerDecorator.java
```diff
@@ -240,27 +243,60 @@ private AgentSpan.Context.Extracted callIGCallbackStart(AgentSpan.Context.Extrac
...
+            // Resolve User-Agent
+            List<String> userAgent = headers.get("user-agent");
```
bantonsson, line 273: This should be pulled out and stored in the `AgentSpan.Context.Extracted` just like all the other header information.

dd-java-agent/agent-bootstrap/src/main/java/datadog/trace/bootstrap/instrumentation/decorator/HttpServerDecorator.java
```diff
@@ -240,27 +243,60 @@ private AgentSpan.Context.Extracted callIGCallbackStart(AgentSpan.Context.Extrac
...
+            if (userAgent != null && !userAgent.isEmpty()) {
+              span.setTag(Tags.HTTP_USER_AGENT, userAgent.get(0));
+            }
+
+            // Resolve Client IP
+            String clientIpHeader = Config.get().getTraceClientIpHeader();
+            InetAddress inferredAddr = ClientIpResolver.doResolve(clientIpHeader, headers);
```
bantonsson, line 280: Same thing goes for this. All the information needed by the `ClientIpResolver` should be stored in the `AgentSpan.Context.Extracted` and sent in as arguments.

dd-java-agent/agent-bootstrap/src/main/java/datadog/trace/bootstrap/instrumentation/decorator/http/QueryObfuscator.java
```diff
@@ -0,0 +1,36 @@
...
+    if (query.isEmpty()) {
+      return "/";
+    }
+
+    // Use default if regex is undefined
+    if (regex == null) {
+      regex = DEFAULT_OBFUSCATION_PATTERN;
+    }
+
+    Matcher matcher = Pattern.compile(regex).matcher(query);
```
bantonsson, line 30: Here you compile the same `regex` over and over again. There needs to be a shared `Pattern` compiled only once, that you acquire a new `Matcher` from for each invocation.

dd-java-agent/agent-bootstrap/src/main/java/datadog/trace/bootstrap/Constants.java
ValentinZakharov: The shadowJar fixes strings acorss the project automatically. So, this line will be modified during the build.
However, if I change it to `datadog.re2j` and build without relocation `disableShadowRelocate=true` than class will not be loaded to bootstrap classloader. I suppose keep original name will be more safe.
Do you have any dubts about this?

dd-java-agent/agent-bootstrap/src/main/java/datadog/trace/bootstrap/instrumentation/decorator/HttpServerDecorator.java
ValentinZakharov: The problem is `ClientIpResolver` uses strategy which requires to have list of headers read to determine client ip. Since Client ip recognition is based on available headers by priority. That's why I wanted to collect all headers before retrieving client ip.
An alternative option could be recalculate client ip, for each new availaable header based on priority order. In this case, don't have to collect them all of them, but will make more calls of `ClientIpResolver` iteratively.

bantonsson: I'm not saying that you shouldn't collect the information, just that you should store it in the `AgentSpan.Context.Extracted` like all other headers. Don't build a `Map` and don't collect everything.

dd-java-agent/agent-bootstrap/src/main/java/datadog/trace/bootstrap/instrumentation/decorator/http/QueryObfuscator.java
```diff
@@ -0,0 +1,44 @@
...
+
+    // Use default if regex is undefined
+    if (regex == null) {
+      regex = DEFAULT_OBFUSCATION_PATTERN;
+    }
+
+    if (!regex.equals(lastRegex)) {
+      pattern = Pattern.compile(regex);
+      lastRegex = regex;
+    }
```
bantonsson, line 36: This is not thread safe in any way, and you can potentially end up with a `lastRegex` that does not represent `pattern`. If you want to cache a value, you can always use a `FixedSizeCache.ObjectHash<>` with a small size, or you store a `Pair` with the `lastRegex` and `pattern` so you read/write them together.

dd-java-agent/agent-bootstrap/src/main/java/datadog/trace/bootstrap/instrumentation/decorator/HttpServerDecorator.java
```diff
@@ -122,6 +125,14 @@ public AgentSpan onRequest(
       if (forwardedPort != null) {
         span.setTag(Tags.HTTP_FORWARDED_PORT, forwardedPort);
       }
+      String userAgent = context.getUserAgent();
+      if (userAgent != null) {
+        span.setTag(Tags.HTTP_USER_AGENT, userAgent);
+      }
+      InetAddress inferredAddress = ClientIpAddressResolver.doResolve(context);
```
bantonsson, line 132: Could we add an option to disable this code if it is a performance problem?

dd-java-agent/agent-bootstrap/src/main/java/datadog/trace/bootstrap/instrumentation/decorator/HttpServerDecorator.java
```diff
@@ -147,9 +158,16 @@ public AgentSpan onRequest(
           if (config.isHttpServerTagQueryString()) {
             String query =
                 supportsRaw && config.isHttpServerRawQueryString() ? url.rawQuery() : url.query();
+
+            if (query != null && !query.isEmpty()) {
+              query = QueryObfuscator.obfuscate(query, config.getObfuscationQueryRegexp());
```
bantonsson, line 163: Same thing here, could we add an option to disable this code if it is a performance problem?

bantonsson (commented):
This is shaping up nicely. Looking forward to some performance measurements.

dd-trace-core/src/main/java/datadog/trace/core/propagation/ContextInterpreter.java
```diff
@@ -121,6 +203,8 @@ public ContextInterpreter reset() {
     tags = Collections.emptyMap();
     baggage = Collections.emptyMap();
     valid = true;
+    httpHeaders = new HttpHeaders();
```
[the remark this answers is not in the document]
ValentinZakharov: Fixed

dd-java-agent/agent-bootstrap/src/main/java/datadog/trace/bootstrap/instrumentation/decorator/http/QueryObfuscator.java
ValentinZakharov: Fixed

dd-java-agent/agent-bootstrap/agent-bootstrap.gradle
```diff
@@ -23,6 +23,8 @@ dependencies {
   api deps.slf4j
   // ^ Generally a bad idea for libraries, but we're shadowing.
 
+  implementation group: 'com.google.re2j', name: 're2j', version: '1.6'
```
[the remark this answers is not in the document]
ValentinZakharov: Updated to recent version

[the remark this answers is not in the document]
ValentinZakharov: That was old code. removed.

dd-trace-core/src/main/java/datadog/trace/core/tagprocessor/QueryObfuscator.java
```diff
@@ -0,0 +1,81 @@
...
+import org.slf4j.LoggerFactory;
+
+public class QueryObfuscator implements TagsPostProcessor {
+
+  private static final Logger log = LoggerFactory.getLogger(QueryObfuscator.class);
+
+  private static final String DEFAULT_OBFUSCATION_PATTERN =
+      "((?i)(?:p(?:ass)?w(?:or)?d|pass(?:_?phrase)?|secret|(?:api_?|private_?|public_?|access_?|secret_?)key(?:_?id)?|token|consumer_?(?:id|key|secret)|sign(?:ed|ature)?|auth(?:entication|orization)?)(?:(?:\\s|%20)*(?:=|%3D)[^&]+|(?:\"|%22)(?:\\s|%20)*(?::|%3A)(?:\\s|%20)*(?:\"|%22)(?:%2[^2]|%[^2]|[^\"%])+(?:\"|%22))|bearer(?:\\s|%20)+[a-z0-9\\._\\-]|token(?::|%3A)[a-z0-9]{13}|gh[opsu]_[0-9a-zA-Z]{36}|ey[I-L](?:[\\w=-]|%3D)+\\.ey[I-L](?:[\\w=-]|%3D)+(?:\\.(?:[\\w.+\\/=-]|%3D|%2F|%2B)+)?|[\\-]{5}BEGIN(?:[a-z\\s]|%20)+PRIVATE(?:\\s|%20)KEY[\\-]{5}[^\\-]+[\\-]{5}END(?:[a-z\\s]|%20)+PRIVATE(?:\\s|%20)KEY|ssh-rsa(?:\\s|%20)*(?:[a-z0-9\\/\\.+]|%2F|%5C|%2B){100,})";
+
+  private static final DDCache<String, Pattern> cache = DDCaches.newFixedSizeCache(2);
```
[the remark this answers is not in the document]
ValentinZakharov: Removed caching. I got confirmation that we don't need to update regex during the app lifetime.

dd-trace-core/src/main/java/datadog/trace/core/tagprocessor/QueryObfuscator.java
```diff
@@ -0,0 +1,81 @@
...
+      query = Strings.replace(query, matcher.group(), "<redacted>");
+    }
+    return query;
+  }
+
+  @Override
+  public Map<String, Object> processTags(Map<String, Object> unsafeTags) {
+    Object query = unsafeTags.get(DDTags.HTTP_QUERY);
+    if (query instanceof String) {
+      query = obfuscate((String) query, Config.get().getObfuscationQueryRegexp());
```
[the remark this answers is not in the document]
ValentinZakharov: Fixed.

internal-api/src/main/java/datadog/trace/bootstrap/instrumentation/api/TagContext.java
```diff
@@ -52,6 +58,86 @@ public String getForwardedPort() {
     return null;
   }
 
+  @Override
+  public String getForwardedFor() {
+    if (httpHeaders == null) {
```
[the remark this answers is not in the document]
ValentinZakharov: I removed `HttpHeaders.NO_HEADERS`, the idea is to avoid creating `HttpHeaders` if there is no headers.
If any of watched headers will be found during extraction - than we create instance of `HttpHeaders`

dd-java-agent/agent-bootstrap/src/main/java/datadog/trace/bootstrap/instrumentation/decorator/HttpServerDecorator.java
```diff
@@ -176,6 +183,18 @@ public AgentSpan onRequest(
       // TODO: blocking
       callIGCallbackSocketAddress(span, ip, port);
     }
+
+    InetAddress inferredAddress = ClientIpAddressResolver.doResolve(context);
```
mcculls, line 187: Could we put this behind a feature-flag to give us a quick workaround if this causes issues in the field?

(Previously this was only applied when AppSec was enabled, but with this PR it's on for all server tracing)

ValentinZakharov: Done

cataphract (approved):

internal-api/src/main/java/datadog/trace/api/Config.java
```diff
@@ -911,6 +918,21 @@ && isJavaVersionAtLeast(8)
...
+    if (traceClientIpHeader != null) {
+      traceClientIpHeader = traceClientIpHeader.toLowerCase(Locale.ROOT);
+    }
+    this.traceClientIpHeader = traceClientIpHeader;
+
+    this.traceClientIpHeaderDisabled =
+        configProvider.getBoolean(TRACE_CLIENT_IP_HEADER_DISABLED, false);
+
+    this.traceClientIpResolverEnabled =
+        configProvider.getBoolean(TRACE_CLIENT_IP_RESOLVER_ENABLED, true);
```
mcculls, line 934: is it possible to have this default to `false` when app-sec is not enabled, and `true` when app-sec is enabled?

dd-trace-core/src/main/java/datadog/trace/core/propagation/ContextInterpreter.java
```diff
@@ -71,10 +86,19 @@ protected Map<String, String> cleanMapping(Map<String, String> taggedHeaders) {
...
+    if (null != value) {
+      if (FORWARDED_KEY.equalsIgnoreCase(key)) {
+        forwarded = value;
+        hasForwarded = true;
+        return true;
+      }
+      if (FORWARDED_FOR_KEY.equalsIgnoreCase(key)) {
+        if (httpHeaders == null) {
+          httpHeaders = new TagContext.HttpHeaders();
+        }
```
bantonsson, line 98: I think this repetitive `null` check and creation should live in a `private final` helper method.

internal-api/src/main/java/datadog/trace/bootstrap/instrumentation/api/TagContext.java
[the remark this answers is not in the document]
bantonsson: @ValentinZakharov I think that you should add the `HttpHeaders.NO_HEADERS` back and remove all these `null` checks.

internal-api/src/main/java/datadog/trace/api/Config.java
ValentinZakharov: The idea is to collect IP in both cases, if app-sec enabled and disabled.

mcculls: is there a document/RFC explaining why this is needed when app-sec is disabled?

dd-java-agent/agent-bootstrap/src/main/java/datadog/trace/bootstrap/instrumentation/decorator/http/ClientIpAddressResolver.java
```diff
@@ -0,0 +1,419 @@
...
+  public static InetAddress parseIpAddress(String str) {
+    if (str.length() == 0) {
+      return null;
+    }
+    char firstChar = str.charAt(0);
+    if (!(firstChar >= '0' && firstChar <= '9' || firstChar == ':')) {
+      return null; // probably a name instead
+    }
+    try {
+      return InetAddress.getByName(str);
```
mcculls, line 414: note we've seen issues calling `InetAddress.getByName` when creating spans in the past: https://github.com/DataDog/dd-trace-java/pull/2188

ValentinZakharov: Reworked to avoid dns lookup during client ip resolution

bantonsson (approved):

PR #3623, opened 2022-06-29
[alpha] Create tracer injection image for kubernetes

devinsba:
# What Does This Do

This creates a tracer injection image for kubernetes. This will be used by the datadog kubernetes operator to automatically add the tracer to an application pod. The tests here assume that the operator has already done the job of adding this image as an `initializationContainer`

# Additional Notes

- Does not do exhaustive tests of the output
- The test app is not included in this repo yet
- Everything is shell scripts, can be changed to gradle in the future


mcculls (approved):
LGTM

PR #3626, opened 2022-06-30
Properly detect non-OpenJ9 J9 distributions

luhenry:
# What Does This Do

Some distributions of J9 wouldn't be detected as they wouldn't come from the open source Ecplise OpenJ9 distribution.

# Motivation

Fix a customer's case.

# Additional Notes


mcculls (approved):

ValentinZakharov:
I was implementing reading JVM Runtime info for `Telemetry`. But it's not merged yet.
https://github.com/DataDog/dd-trace-java/blob/c98fe7f8668a670e97d4c3db013da78372409f74/internal-api/src/main/java/datadog/trace/api/Platform.java#L116-L124
Does it make sense to use `Platform.java` as a single place for all jvm info?

luhenry:
> Does it make sense to use Platform.java as a single place for all jvm info?

It can make sense. We can do it in a separate PR though.

ValentinZakharov (approved):

PR #3627, opened 2022-06-30
Remote config (disabled by default)

cataphract:
# What Does This Do

Implements remote config and (partially) hooks appsec to it. AppSec is able to reload the WAF configuration from remote config, but activation/deactivation of AppSec is not possible (reserved for upcoming PR).

# Additional Notes

See "[RFC] Integrating with Remote Config in a Tracer" on Google docs.


communication/src/main/java/datadog/communication/ddagent/SharedCommunicationObjects.java
```diff
@@ -31,19 +37,67 @@ public void createRemaining(Config config) {
...
+      String remoteConfigUrl = config.getFinalRemoteConfigUrl();
+      if (remoteConfigUrl != null) {
+        configurationPoller =
+            new ConfigurationPoller(config, getTracerVersion(), remoteConfigUrl, okHttpClient);
+      } else {
+        createRemaining(config);
+        DDAgentFeaturesDiscovery fd = featuresDiscovery(config);
+        String configEndpoint = fd.getConfigEndpoint();
+        if (configEndpoint != null) {
+          remoteConfigUrl = featuresDiscovery.buildUrl(configEndpoint).toString();
```
ValentinZakharov, line 57: Here can be NPE, because `buildUrl` may return `null` if fails.

communication/src/main/java/datadog/communication/ddagent/SharedCommunicationObjects.java
```diff
@@ -31,19 +37,67 @@ public void createRemaining(Config config) {
...
               config.isTraceAgentV05Enabled(),
               config.isTracerMetricsEnabled());
+      featuresDiscovery.discover();
     }
     return featuresDiscovery;
   }
+
+  private static String getTracerVersion() {
+    final StringBuilder sb = new StringBuilder(32);
+    ClassLoader cl = ClassLoader.getSystemClassLoader();
```
ValentinZakharov, line 89: If I'm not mistaken `getSystemClassLoader()` may return `null` if no delegation

communication/src/main/java/datadog/communication/ddagent/SharedCommunicationObjects.java
```diff
@@ -31,19 +37,67 @@ public void createRemaining(Config config) {
...
+    try (final BufferedReader reader =
+        new BufferedReader(
+            new InputStreamReader(
+                cl.getResourceAsStream("dd-java-agent.version"), StandardCharsets.ISO_8859_1))) {
+      for (int c = reader.read(); c != -1; c = reader.read()) {
+        sb.append((char) c);
+      }
```
ValentinZakharov, line 96: Since version is always single line, I assume, we can use `reader.readLine()` instead.

dd-java-agent/agent-bootstrap/src/main/java/datadog/trace/bootstrap/Agent.java
```diff
@@ -69,6 +73,7 @@ private enum AgentFeature {
     STARTUP_LOGS("dd.trace.startup.logs", true),
     PROFILING("dd.profiling.enabled", false),
     APPSEC("dd.appsec.enabled", false),
+    REMOTE_CONFIG("dd.remote_config.enabled", true),
```
ValentinZakharov, line 76: Looks like we are not using `_` (underscore) in option names in Java trace.
Instead everywhere used `-` (dash), see [here](https://github.com/DataDog/dd-trace-java/blob/master/dd-trace-api/src/main/java/datadog/trace/api/config/TracerConfig.java#L44)
Can you confirm that `remote_config` with underscore is agreed across all tracers?

dd-java-agent/agent-bootstrap/src/main/java/datadog/trace/bootstrap/Agent.java
```diff
@@ -400,6 +409,39 @@ public void execute() {
...
+    final StringBuilder sb = new StringBuilder(32);
+    ClassLoader cl = ClassLoader.getSystemClassLoader();
+    try (final BufferedReader reader =
+        new BufferedReader(
+            new InputStreamReader(
+                cl.getResourceAsStream("dd-java-agent.version"), StandardCharsets.ISO_8859_1))) {
+      for (int c = reader.read(); c != -1; c = reader.read()) {
+        sb.append((char) c);
+      }
+    }
```
ValentinZakharov, line 440: Looks like we have duplications of `getAgentVersion` in three places:
- SharedCommunnocationObject
- Agent
- AgentJar

Is there any way to use it from single place?

cataphract: I can get it down to two.

remote-config/src/main/java/datadog/remote_config/tuf/RemoteConfigResponse.java
```diff
@@ -0,0 +1,187 @@
...
+        throw e;
+      } catch (Exception exception) {
+        throw new IntegrityCheckException(
+            "Could not get file contents from remote config, file " + configKey, exception);
+      }
+    }
+    return Optional.empty();
+  }
+
+  private static BigInteger sha256(byte[] bytes) {
```
jpbempel, line 124: why not returning a string?

remote-config/src/main/java/datadog/remote_config/tuf/RemoteConfigResponse.java
```diff
@@ -0,0 +1,187 @@
...
+  }
+
+  public Optional<byte[]> getFileContents(String configKey) {
+    Targets.ConfigTarget configTarget = getTarget(configKey);
+    // tolerate hash not being present for now
+    BigInteger expectedHash = null;
+    if (configTarget != null) {
+      String hashStr = configTarget.hashes.get("sha256");
+      if (hashStr != null) {
+        expectedHash = new BigInteger(hashStr, 16);
```
jpbempel, line 85: What is the added value to reinterpret the hash into a BigInteger?
could keep the resulted String and compare with sha256 as String

jpbempel (commented):
for the package name I suggest just `remoteconfig` instead of `remote_config`

dd-java-agent/agent-bootstrap/src/main/java/datadog/trace/bootstrap/Agent.java
```diff
@@ -70,6 +70,7 @@ private enum AgentFeature {
     PROFILING("dd.profiling.enabled", false),
     APPSEC("dd.appsec.enabled", false),
     IAST("dd.iast.enabled", false),
+    REMOTE_CONFIG("dd.remote_config.enabled", true),
```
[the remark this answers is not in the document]
cataphract: Right, it's unfortunate that there are these two separate mechanisms (`Config` and `AgentFeature`), I presume in order to avoid initializing `Config` too early. I'm correcting the default.

remote-config/src/main/java/datadog/remote_config/tuf/RemoteConfigResponse.java
cataphract: The reason I chose this method was to go around possible differences in the string representation (using a-f or A-F and absence/presence of leading zeros).

remote-config/src/main/java/datadog/remote_config/tuf/RemoteConfigResponse.java
cataphract: See [my other comment](https://github.com/DataDog/dd-trace-java/pull/3627/files/430f8c20fa6c43adff84b90092f55beb29418200#diff-a7eab92a62f6ae291a9219ca5f219e14b9084aadf75b9357d68fed7c4db23bdd).

remote-config/src/main/java/datadog/remote_config/ConfigurationPoller.java
```diff
@@ -0,0 +1,547 @@
...
+      ConfigurationDeserializer<T> deserializer,
+      ConfigurationChangesListener<T> listener) {
+    this.listeners.put(product.name(), new DeserializerAndListener<>(deserializer, listener));
+  }
+
+  public synchronized void removeListener(Product product) {
+    this.listeners.remove(product.name());
+  }
+
+  public synchronized <T> void addFeaturesListener(
```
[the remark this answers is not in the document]
jpbempel: I agree this simplifies the reasoning, however if we stick with a `synchronized` block, at least use an internal dedicated object and a more fine-grained sync block to perform that and including only the relevant part in that block.

communication/src/main/java/datadog/communication/ddagent/SharedCommunicationObjects.java
```diff
@@ -31,18 +34,81 @@ public void createRemaining(Config config) {
...
+        | InstantiationException
+        | IllegalAccessException
+        | InvocationTargetException e) {
+      return null;
+    }
+
+    return configurationPoller;
+  }
+
+  private void maybeSetPoller(Config config)
```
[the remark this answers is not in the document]
jpbempel: yes it's only the impl details of SharedCommunicationObjects. it's more on avoiding to much side effects on that method and having a method returning the created objects, and the call site (in `configurationPoller` method) decide what to do with this created objects (like a method factory if you want).

dd-java-agent/agent-bootstrap/src/main/java/datadog/trace/bootstrap/Agent.java
```diff
@@ -391,6 +395,7 @@ public void execute() {
 
       installDatadogTracer(scoClass, sco);
       maybeStartAppSec(scoClass, sco);
+      maybeStartRemoteConfig(scoClass, sco);
```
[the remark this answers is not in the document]
jpbempel: right, but I still think it's late and some classes are already loaded while we can avoid that because we have locally (at Agent class level) information to avoid the reflection part.

remote-config/src/main/java/datadog/remote_config/tuf/RemoteConfigResponse.java
jpbempel: ok, fine by me

communication/src/main/java/datadog/communication/ddagent/SharedCommunicationObjects.java
```diff
@@ -31,18 +35,78 @@ public void createRemaining(Config config) {
...
+    }
+
+    try {
+      this.configurationPoller = maybeCreatePoller(config);
+    } catch (ClassNotFoundException
+        | NoSuchMethodException
+        | InstantiationException
+        | IllegalAccessException
+        | InvocationTargetException e) {
+      return null;
```
ValentinZakharov, line 55: Do we need any log message for troubleshooting? Or you expect `maybeCreatePoller` will not throw an exceptions?

dd-java-agent/appsec/src/main/java/com/datadog/appsec/config/AppSecConfigServiceImpl.java
```diff
@@ -31,40 +27,44 @@ public class AppSecConfigServiceImpl implements AppSecConfigService {
...
-                Map<String, AppSecConfig> stringObjectMap =
-                    deserializeConfig(Okio.buffer(Okio.source(is)));
-                distributeSubConfigurations(stringObjectMap);
-                this.lastConfig.set(stringObjectMap);
-              } catch (IOException e) {
-                log.error("Error deserializing appsec config", e);
-              }
-            });
+  private void subscribeConfigurationPoller() {
+    this.configurationPoller.addListener(
```
ValentinZakharov, line 47: I think here could be NPE. The `configurationPoller` may be `null` because of `sco.configurationPoller(config);` can return `null` in some cases.

ValentinZakharov: Nevermind, I can see null-checker in caller side.

communication/src/main/java/datadog/communication/ddagent/SharedCommunicationObjects.java
```diff
@@ -31,18 +35,78 @@ public void createRemaining(Config config) {
...
+
+    String remoteConfigUrl = config.getFinalRemoteConfigUrl();
+    if (remoteConfigUrl != null) {
+      return constructor.newInstance(
+          config, TracerVersion.TRACER_VERSION, remoteConfigUrl, okHttpClient);
+    } else {
+      createRemaining(config);
+      DDAgentFeaturesDiscovery fd = featuresDiscovery(config);
+      String configEndpoint = fd.getConfigEndpoint();
+      if (configEndpoint != null) {
```
arbll, line 81: QQ: if someone enables RC in the agent would I have to restart my tracer to have Remote Config ?

arbll: Is this purposeful as a v0 ? The RFC states the following:
```
The <trace-agent>/info endpoint CAN also be used to check if RC is enabled. Regardless, tracers MUST continuously check for new configurations as RC could be enabled in the agent at any point in time.
```

dd-java-agent/agent-bootstrap/src/main/java/datadog/trace/bootstrap/Agent.java
```diff
@@ -431,6 +436,28 @@ public void execute() {
...
+
+    try {
+      Method pollerMethod = scoClass.getMethod("configurationPoller", Config.class);
+      Object poller = pollerMethod.invoke(sco, Config.get());
+      if (poller == null) {
+        log.debug("Remote config is not enabled");
+        return;
+      }
+      Class<?> pollerCls =
+          SHARED_CLASSLOADER.loadClass("datadog.remote_config.ConfigurationPoller");
```
jpbempel, line 452: should be changed as well for `"datadog.remoteconfig.ConfigurationPoller"`

remote-config/src/main/java/datadog/remote_config/ConfigurationPoller.java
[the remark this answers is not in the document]
cataphract: I've limited the code that's covered by the critical section, but I see no point in using a dedicated object rather than `this`, as don't need more than one lock in the class.

dd-java-agent/agent-bootstrap/src/main/java/datadog/trace/bootstrap/Agent.java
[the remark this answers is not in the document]
cataphract: `maybeStartRemoteConfig` now doesn't load the ConfigurationPoller class (and the dependent ones) unless remote config is enabled.

communication/src/main/java/datadog/communication/ddagent/SharedCommunicationObjects.java
cataphract: Hmm, I missed that part and it's going to be annoying to implement that. At this point it's probably a non-issue since remote config needs to be explicitly enabled anyway (and if given an endpoint it will "continuously" try to connect, though with successively larger intervals).

communication/src/main/java/datadog/communication/ddagent/SharedCommunicationObjects.java
cataphract: I've added a logging statement.

ValentinZakharov (approved):

remote-config/src/main/java/datadog/remote_config/ConfigurationPoller.java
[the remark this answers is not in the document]
jpbempel: FTR, Using an internal lock prevent outsider to take also the lock and encapsulate the synchronization logic.
using this expose this lock and some internals of the class, that we may want to keep private for various reasons like changing the lock mechanism

PR #3642, opened 2022-07-07
Glopes/half enable appsec

cataphract:
AppSec is enabled (load & instr. applied) but inactive by default. It can be enabled by remote config features.

This implies that all the appsec related instrumentations are loaded and applied (unless AppSec is explicitly disabled with `-Ddd.appsec.enabled=false`.

AppSec being inactive makes no RequestContext be created. The loaded advice for AppSec skips doing further work on the request if it finds no context associated.


dd-java-agent/instrumentation/spring-webmvc-3.1/src/main/java/datadog/trace/instrumentation/springweb/HttpMessageConverterInstrumentation.java
```diff
@@ -63,30 +62,25 @@ public void adviceTransformations(AdviceTransformation transformation) {
...
+        @Advice.Return final Object obj, @ActiveRequestContext RequestContext reqCtx) {
       if (obj == null) {
         return;
       }
 
-      AgentSpan agentSpan = activeSpan();
-      if (agentSpan == null) {
-        return;
-      }
-
```
bantonsson, line 81: I really like how all the repetitive _defensive_ code is removed by the generated guards!

internal-api/src/main/java/datadog/trace/api/ProductActivationConfig.java
```diff
@@ -0,0 +1,18 @@
+package datadog.trace.api;
+
+public enum ProductActivationConfig {
```
bantonsson, line 3: A little comment about the assumptions would be nice (yes, that can be read in the code as well, but still).

jpbempel (approved):

mcculls:
Could we run some performance tests, comparing master against this branch? (not just request latency, but comparing startup time and heap use)

ValentinZakharov:
I did some measurements that can reflect overhead

- Baseline - pure petstore application
- Tracer - app with running tracer
- Inactive appsec - half enabled but not activated AppSec with enabled Remote config
- AppSec - enabled AppSec with Remote config

**Startup time, from the launch till the first request been processed (seconds):**
```
Baseline:          3.89
Tracer:            5.58 
Inactive AppSec:   5.58 
AppSec:            5.79
```
chart (4)

**Request latency during 2 mins of application running (seconds):**
```
Baseline:         0.001228
Tracer:           0.001465 
Inactive AppSec:  0.001426 
AppSec:           0.001534
```
chart (3)

**Request per second during 2 mins**
```
Baseline:          571.9
Tracer:            480.2 
Inactive AppSec:   484.0 
AppSec:            459.1
```
chart (5)

**Heap usage during 2 mins**
```
Baseline:          208.0
Tracer:            221.8 
Inactive AppSec:   238.2 
AppSec:            240.2
```
chart (6)

**CPU usage during 2 mins**
```
Baseline:          6.3%
Tracer:            6.8%
Inactive AppSec:   6.9% 
AppSec:            7.0%
```
chart (7)

PR #3644, opened 2022-07-07
[alpha] Add a publish step to tracer injection image for kubernetes

devinsba:
# What Does This Do

Adds a publish step for tracer tags that is part of the tracer injection image for kubernetes

# Additional Notes

The jar used in the image is not the exact same file as the one that gets published to maven central. This isn't ideal in my mind so I think this is something we could address


mcculls (approved):
LGTM

PR #3646, opened 2022-07-11
[CIAPP] Add `_dd.ci.env_vars` span tag for every CI Provider

drodriguezhdez:
# What Does This Do

This PR adds the `_dd.ci.env_vars` span tag for every CI Provider supported in CI Visibility for both Pipelines and Tests.
The tag will contain a string in a JSON format with the env vars.

# Motivation

The backend requires these environment variables to correlate tests with pipelines.


nachoBonafonte (approved):

internal-api/src/main/java/datadog/trace/util/Strings.java
```diff
@@ -225,4 +227,25 @@ public static CharSequence truncate(CharSequence input, int limit) {
     }
     return input.subSequence(0, limit);
   }
+
+  public static String toJson(final Map<String, String> map) {
```
liashenko, line 231: The code looks correct, but have you considered using Gson or Jackson?

Also, do we care about the order of attributes?

drodriguezhdez: > The code looks correct, but have you considered using Gson or Jackson?

We try to avoid using 3rd party dependencies if it's possible in the library.

> Also, do we care about the order of attributes?

No, a specific order is not required.

internal-api/src/main/java/datadog/trace/bootstrap/instrumentation/ci/CIProviderInfo.java
```diff
@@ -602,6 +638,9 @@ public String toString() {
           + ", ciWorkspace='"
           + ciWorkspace
           + '\''
+          + ", ciEnvVars='"
+          + ciEnvVars
```
PerfectSlayer, line 642: Can you confirm HashMap as a proper implementation of `toString()` and not a default classname#hash String?

drodriguezhdez: Yes, HashMap has a proper implementation that prints the "{key:value}" form.

PerfectSlayer (approved):

internal-api/src/test/groovy/datadog/trace/bootstrap/instrumentation/ci/CISpecExtractor.groovy
```diff
@@ -21,12 +24,32 @@ class CISpecExtractor {
...
     CISpec(env, tags) {
       this.env = env
       this.tags = tags
     }
+
+    boolean assertTags(Map<String, String> ciTags) {
+      for(String val : JSON_TAGS) {
+        if(!tags[val] && !ciTags[val]) {
+          continue
+        } else if ((tags[val] && !ciTags[val]) || (!tags[val] && ciTags[val])) {
```
liashenko, line 40:
```suggestion
        } else if (!tags[val] || !ciTags[val]) {
```
we can simplify it since we're checking if they're both false above

liashenko (approved):

internal-api/src/test/groovy/datadog/trace/bootstrap/instrumentation/ci/CISpecExtractor.groovy
drodriguezhdez: I think that's not the same. We want to assert that if one has been set, the other one must be set too.

I'd say with your suggestion if one has not been set, it returns false (which it's a valid case because not all the CI providers report envvars).

liashenko: if both are unset, we just continue (#L38), else we check if one is unset and return false.

Isn't it what you're also doing here? checking if only tags or ciTags is unset.

PR #3647, opened 2022-07-11
Util to get stacktrace for reported vulnerabilities

jandro996:
# What Does This Do
Util to do partial walks of stacktraces. Excludes elements that belong to the dd-trace-java project.

# Motivation
In IAST, when we detect a vulnerability, we need to get the class and line where it happened. The relevant class and the line will not be the last in the stack. The relevant class and line will be calculated ignoring internal dd-trace-java classes, usually applying a filter of classes on top of that (e.g standard library).
So the filtering of dd-trace-java elements is built-in, and the API uses streams for further filtering operations.
In most cases, only a few elements need to be walked through (e.g. 2-4), and only in some rare cases the whole stack needs to be examined.

# Additional Notes
This functionality will be used whenever a vulnerability is found, so its performance is critical to vulnerability reporting. To optimize performance, three implementations are included:
- Default implementation: uses Thread.currentThread().getStackTrace() and gets an array with the full stacktrace. This is the most expensive when doing a partial walk, since it needs to create StackTraceElement objects for the whole stacktrace.
- Improve performance for JDK8 with HotSpot ([(JavaLangAccess)](https://github.com/openjdk/jdk/blob/jdk8-b120/jdk/src/share/classes/sun/misc/JavaLangAccess.java)). Relies on JavaLangAccess to create StackTraceElements on-demand.
- Improve performance for JDK9+ ([StackWalker](https://docs.oracle.com/javase/9/docs/api/java/lang/StackWalker.html)). Uses the newer StackWalker API.
 
The implementation is currently in the internal-api-* packages. This was picked arbitrarily to some extent, so other suggestions about locations are welcome. Initially, this API will only be used from the upcoming dd-java-agent/iast module.

# Performance
Here are sample results for the implementations. “Deep” is the number of total elements in the stack. “Limit” is the number of elements retrieved.

**JDK8 with HotSpot implementation performance**

<table>
  <tr>
    <th>Limit</th>
    <th>Deep</th>
    <th>Performance improvement %</th>
  </tr>
  <tr>
    <td>1</td>
    <td>10</td>
    <td>80.46%</td>
  </tr>
<tr>
    <td>3</td>
    <td>10</td>
    <td>75.31%</td>
  </tr>
<tr>
    <td>10</td>
    <td>10</td>
    <td>58.67%</td>
  </tr>
<tr>
    <td>1</td>
    <td>50</td>
    <td>85.42%</td>
  </tr>
<tr>
    <td>3</td>
    <td>50</td>
    <td>82.72%</td>
  </tr>
<tr>
    <td>10</td>
    <td>50</td>
    <td>74.13%</td>
  </tr>
<tr>
    <td>1</td>
    <td>100</td>
    <td>87.05%</td>
  </tr>
<tr>
    <td>3</td>
    <td>100</td>
    <td>85.51%</td>
  </tr>
<tr>
    <td>10</td>
    <td>100</td>
    <td>80.24%</td>
  </tr>
</table>

**JDK9 implementation performance**

<table>
  <tr>
    <th>Limit</th>
    <th>Deep</th>
    <th>Performance improvement %</th>
  </tr>
  <tr>
    <td>1</td>
    <td>10</td>
    <td>71.97%</td>
  </tr>
<tr>
    <td>3</td>
    <td>10</td>
    <td>63.46%</td>
  </tr>
<tr>
    <td>10</td>
    <td>10</td>
    <td>20.49%</td>
  </tr>
<tr>
    <td>1</td>
    <td>50</td>
    <td>82.92%</td>
  </tr>
<tr>
    <td>3</td>
    <td>50</td>
    <td>78.33%</td>
  </tr>
<tr>
    <td>10</td>
    <td>50</td>
    <td>54.04%</td>
  </tr>
<tr>
    <td>1</td>
    <td>100</td>
    <td>88.22%</td>
  </tr>
<tr>
    <td>3</td>
    <td>100</td>
    <td>85.42%</td>
  </tr>
<tr>
    <td>10</td>
    <td>100</td>
    <td>69.66%</td>
  </tr>
</table>

# Future Work
Future versions of this utility may include further performance optimizations:
- Filtering by class name before creating StackTraceElement objects (possible in all implementations except the default).
- Skipping a fixed number of initial frames without inspecting them at all. We will always have a number of elements that are part of dd-trace java, and we can skip them by default. This would save some superfluous allocations of StackTraceElements.

### DIFF
--- a/.github/workflows/codeql-analysis.yml
+++ b/.github/workflows/codeql-analysis.yml
@@ -22,27 +22,27 @@ jobs:
         language: [ 'java' ]
 
     steps:
-    - name: Checkout repository
-      uses: actions/checkout@5a4ac9002d0be2fb38bd78e4b4dbde5606d7042f # 2.3.4
+      - name: Checkout repository
+        uses: actions/checkout@5a4ac9002d0be2fb38bd78e4b4dbde5606d7042f # 2.3.4
 
-    # Initializes the CodeQL tools for scanning.
-    - name: Initialize CodeQL
-      uses: github/codeql-action/init@1a927e9307bc11970b2c679922ebc4d03a5bd980 # 1.0.31
-      with:
-        languages: ${{ matrix.language }}
-        # If you wish to specify custom queries, you can do so here or in a config file.
-        # By default, queries listed here will override any specified in a config file.
-        # Prefix the list here with "+" to use these queries and those in the config file.
-        # queries: ./path/to/local/query, your-org/your-repo/queries@main
+      # Initializes the CodeQL tools for scanning.
+      - name: Initialize CodeQL
+        uses: github/codeql-action/init@1a927e9307bc11970b2c679922ebc4d03a5bd980 # 1.0.31
+        with:
+          languages: ${{ matrix.language }}
+          # If you wish to specify custom queries, you can do so here or in a config file.
+          # By default, queries listed here will override any specified in a config file.
+          # Prefix the list here with "+" to use these queries and those in the config file.
+          # queries: ./path/to/local/query, your-org/your-repo/queries@main
 
-    - name: Build dd-trace-java for creating the CodeQL database
-      run: JAVA_HOME=$JAVA_HOME_8_X64 JAVA_11_HOME=$JAVA_HOME_11_X64 ./gradlew clean :dd-java-agent:shadowJar --build-cache --parallel --stacktrace --no-daemon --max-workers=8
+      - name: Build dd-trace-java for creating the CodeQL database
+        run: JAVA_HOME=$JAVA_HOME_8_X64 JAVA_8_HOME=$JAVA_HOME_8_X64 JAVA_11_HOME=$JAVA_HOME_11_X64 ./gradlew clean :dd-java-agent:shadowJar --build-cache --parallel --stacktrace --no-daemon --max-workers=8
 
-    - name: Perform CodeQL Analysis
-      uses: github/codeql-action/analyze@1a927e9307bc11970b2c679922ebc4d03a5bd980 # 1.0.31
-      
-    - name: upload_artifact
-      uses: actions/upload-artifact@82c141cc518b40d92cc801eee768e7aafc9c2fa2 # 2.3.1
-      with:
-         name: CodeQL Analysis Sarif
-         path: /home/runner/work/dd-trace-java/results/java.sarif    
+      - name: Perform CodeQL Analysis
+        uses: github/codeql-action/analyze@1a927e9307bc11970b2c679922ebc4d03a5bd980 # 1.0.31
+
+      - name: upload_artifact
+        uses: actions/upload-artifact@82c141cc518b40d92cc801eee768e7aafc9c2fa2 # 2.3.1
+        with:
+          name: CodeQL Analysis Sarif
+          path: /home/runner/work/dd-trace-java/results/java.sarif
--- a/gradle.properties
+++ b/gradle.properties
@@ -1,6 +1,5 @@
 org.gradle.parallel=true
 org.gradle.caching=true
-
 org.gradle.jvmargs=-XX:MaxMetaspaceSize=1g
 org.gradle.java.installations.auto-detect=false
 org.gradle.java.installations.auto-download=false
--- a/internal-api/internal-api-8/src/main/java/datadog/trace/util/stacktrace/DefaultStackWalker.java
+++ b/internal-api/internal-api-8/src/main/java/datadog/trace/util/stacktrace/DefaultStackWalker.java
@@ -1,0 +1,20 @@
+package datadog.trace.util.stacktrace;
+
+import java.util.Arrays;
+import java.util.function.Function;
+import java.util.stream.Stream;
+
+public class DefaultStackWalker extends AbstractStackWalker {
+
+  DefaultStackWalker() {}
+
+  @Override
+  public boolean isEnabled() {
+    return true;
+  }
+
+  @Override
+  <T> T doGetStack(final Function<Stream<StackTraceElement>, T> consumer) {
+    return consumer.apply(Arrays.stream(Thread.currentThread().getStackTrace()));
+  }
+}
--- a/internal-api/internal-api-8/src/main/java/datadog/trace/util/stacktrace/StackWalkerFactory.java
+++ b/internal-api/internal-api-8/src/main/java/datadog/trace/util/stacktrace/StackWalkerFactory.java
@@ -1,0 +1,48 @@
+package datadog.trace.util.stacktrace;
+
+import de.thetaphi.forbiddenapis.SuppressForbidden;
+import java.util.Objects;
+import java.util.function.Supplier;
+import java.util.stream.Stream;
+import org.slf4j.Logger;
+import org.slf4j.LoggerFactory;
+
+@SuppressForbidden
+public class StackWalkerFactory {
+
+  private static final Logger LOGGER = LoggerFactory.getLogger(StackWalkerFactory.class);
+
+  public static final StackWalker INSTANCE;
+
+  static {
+    Stream<StackWalker> stream = Stream.of(hotspot(), jdk9()).map(Supplier::get);
+    INSTANCE =
+        stream
+            .filter(Objects::nonNull)
+            .filter(StackWalker::isEnabled)
+            .findFirst()
+            .orElseGet(defaultStackWalker());
+  }
+
+  private static Supplier<StackWalker> defaultStackWalker() {
+    return DefaultStackWalker::new;
+  }
+
+  private static Supplier<StackWalker> hotspot() {
+    return HotSpotStackWalker::new;
+  }
+
+  private static Supplier<StackWalker> jdk9() {
+    return () -> {
+      try {
+        return (StackWalker)
+            Class.forName("datadog.trace.util.stacktrace.JDK9StackWalker")
+                .getDeclaredConstructor()
+                .newInstance();
+      } catch (Throwable e) {
+        LOGGER.warn("JDK9StackWalker not available", e);
+        return null;
+      }
+    };
+  }
+}
--- a/.github/workflows/test-lib-injection.yaml
+++ b/.github/workflows/test-lib-injection.yaml
@@ -23,7 +23,7 @@ jobs:
 
       - name: Build app image
         run: |
-          JAVA_HOME=$JAVA_HOME_8_X64 JAVA_11_HOME=$JAVA_HOME_11_X64 \
+          JAVA_HOME=$JAVA_HOME_8_X64 JAVA_8_HOME=$JAVA_HOME_8_X64 JAVA_11_HOME=$JAVA_HOME_11_X64 \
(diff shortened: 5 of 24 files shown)


smola (changes requested):

internal-api/internal-api-8/src/main/java/datadog/trace/util/stacktrace/DefaultStackWalker.java
```diff
@@ -0,0 +1,34 @@
...
+  @Override
+  public Stream<StackTraceElement> walk() {
+    return doGetStack().filter(e -> isNotDDTraceClass(e.getClassName()));
+  }
+
+  static boolean isNotDDTraceClass(final String className) {
+    return filteredPackages.stream().noneMatch(className::startsWith);
+  }
+
+  protected Stream<StackTraceElement> doGetStack() {
```
manuel-alvarez-alvarez, line 31: Maybe you can drop a comment explaining that this method can be overriden by subclasses (for instance an implementation using JDK9 stack walker)

internal-api/internal-api-8/src/test/java/datadog/trace/util/stacktrace/StackWalkerTest.java
```diff
@@ -0,0 +1,63 @@
...
+  }
+
+  @Test
+  public void filter_DataDog_AppSec_classes_from_StackTraceElements() {
+    // When
+    Stream<StackTraceElement> stackTraceElements = StackWalker.INSTANCE.walk();
+    // Then
+    assertTrue(
+        stackTraceElements.noneMatch(
+            stackTraceElement -> stackTraceElement.toString().startsWith("com.datadog.appsec.")));
```
manuel-alvarez-alvarez, line 61: Since the stack does not contain any `com.datadog.appsec.` classes ... this test does not provide any value. Maybe you can create a mock class in that package, same way as it's done with `Foo`.

internal-api/internal-api-8/src/test/java/com/foo/Foo.java
```diff
@@ -0,0 +1,25 @@
+package com.foo;
```
manuel-alvarez-alvarez, line 1: If bytebuddy is included as a test dependency you can create the classes on the fly giving you more flexibitily for the tests (for instance allowing you to choose the package)

internal-api/internal-api-8/src/main/java/datadog/trace/util/stacktrace/DefaultStackWalker.java
```diff
@@ -0,0 +1,41 @@
...
+import java.util.stream.Collectors;
+import java.util.stream.Stream;
+
+public class DefaultStackWalker implements StackWalker {
+
+  DefaultStackWalker() {}
+
+  protected static Set<String> filteredPackages =
+      Stream.of("datadog.trace.", "com.datadog.appsec.")
+          .collect(Collectors.toCollection(HashSet::new));
```
anderruiz, line 15: I would prefer to create a predicate instead of an Set that then is transformed to an Stream to create a predicate (for each stack line)

internal-api/internal-api-8/src/main/java/datadog/trace/util/stacktrace/DefaultStackWalker.java
anderruiz: I would even say that we should not implement this method here and do it in another class (I know why you did it)

internal-api/internal-api-8/src/main/java/datadog/trace/util/stacktrace/StackWalkerFactory.java
```diff
@@ -0,0 +1,13 @@
...
+
+import java.util.stream.Stream;
+
+public class StackWalkerFactory {
+  static StackWalker getInstance() {
+    // New StackTraceGenerator must be added to the list
+    return Stream.of(new DefaultStackWalker())
+        .filter(StackWalker::isEnabled)
+        .findFirst()
+        .orElse(new DefaultStackWalker());
```
anderruiz, line 11: At this point we know DefaultStackWalker will return true, so, or we do not include in the Stream or we remove .orElse

internal-api/internal-api-8/src/test/java/datadog/trace/util/stacktrace/StackWalkerTest.java
```diff
@@ -0,0 +1,96 @@
...
+  }
+
+  public static class StackTarget implements Supplier<Stream<StackTraceElement>> {
+    @Override
+    public Stream<StackTraceElement> get() {
+      return Stream.empty();
+    }
+  }
+
+  private static Supplier<Stream<StackTraceElement>> getSupplier(final String clazzName)
```
anderruiz, line 79: What is this method doing?

internal-api/internal-api-8/src/main/java/datadog/trace/util/stacktrace/DefaultStackWalker.java
anderruiz:
```suggestion
          protected static Predicate<String> filteredPackages = (cn)->!cn.startsWith("datadog.trace.")&&!cn.startsWith("com.datadog.appsec.)
```

internal-api/internal-api-8/src/main/java/datadog/trace/util/stacktrace/DefaultStackTraceGenerator.java
```diff
@@ -0,0 +1,32 @@
...
+          .collect(Collectors.toCollection(HashSet::new));
+
+  @Override
+  public boolean isEnabled() {
+    return true;
+  }
+
+  @Override
+  public Stream<StackTraceElement> get() {
+    return doGetStack().filter(e -> isNotDDTraceClass(e.getClassName()));
```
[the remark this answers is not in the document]
anderruiz: No, our classes were always excluded

internal-api/internal-api-8/src/test/java/datadog/trace/util/stacktrace/StackWalkerTest.java
manuel-alvarez-alvarez: It dynamically generates mock classes in custom packages to validate stack trace filtering. Another possible and maybe cleaner option would be to separate the testing logic to build the stack from the one to filter it (this way we can create mocks of the full stack giving us more fine grained control)

smola (approved):

manuel-alvarez-alvarez (approved):

smola (approved):

gradle.properties
```diff
@@ -1,8 +1,7 @@
 org.gradle.parallel=true
 org.gradle.caching=true
-
 org.gradle.jvmargs=-XX:MaxMetaspaceSize=1g
 org.gradle.java.installations.auto-detect=false
 org.gradle.java.installations.auto-download=false
 # 8 and 11 is needed to build
-org.gradle.java.installations.fromEnv=JAVA_8_HOME,JAVA_11_HOME
+org.gradle.java.installations.fromEnv=JAVA_HOME,JAVA_8_HOME,JAVA_11_HOME
```
devinsba, line 7: This change shouldn't be necessary. `JAVA_HOME` and `JAVA_8_HOME` should be equivalent

internal-api/internal-api-8/src/main/java/datadog/trace/util/stacktrace/DefaultStackWalker.java
```diff
@@ -0,0 +1,20 @@
...
+  DefaultStackWalker() {}
+
+  @Override
+  public boolean isEnabled() {
+    return true;
+  }
+
+  @Override
+  <T> T doGetStack(final Function<Stream<StackTraceElement>, T> consumer) {
+    return consumer.apply(Arrays.stream(new Throwable().getStackTrace()));
```
devinsba, line 18: This would avoid allocation of a throwable object unnecessarily 
```suggestion
    return consumer.apply(Arrays.stream(Thread.currentThread().getStackTrace()));
```

devinsba (commented):
I think we can avoid `new Throwable()` in most places by using `Thread.currentThread()` to get the stacktrace instead

jandro996:
> 

We changed it in `DefaultStackWalker` but `HotSpotStackWalker` needs `Throwable`  to use `sun.misc.JavaLangAccess` api

jandro996:
> I think we can avoid `new Throwable()` in most places by using `Thread.currentThread()` to get the stacktrace instead

We changed it in `DefaultStackWalker` but `HotSpotStackWalker` needs `Throwable`  to use `sun.misc.JavaLangAccess` api

gradle.properties
jandro996: Yes it seems redundant...  

We need a way to make this config work https://github.com/DataDog/dd-trace-java/blob/e28c1be6b194910664a8064d04f8faf5c9973e7c/internal-api/internal-api-8/internal-api-8.gradle#L17-L21

**CodeQL build** and **Lib injection test** try to build dd-java-agent.jar

https://github.com/DataDog/dd-trace-java/blob/63cc87143a0e3f6d0c0d3ea3b40facf1e3526eda/.github/workflows/codeql-analysis.yml#L38-L39

https://github.com/DataDog/dd-trace-java/blob/63cc87143a0e3f6d0c0d3ea3b40facf1e3526eda/.github/workflows/test-lib-injection.yaml#L44-L45

and our configuration breaks it with this error

```
Where:
Script '/home/runner/work/dd-trace-java/dd-trace-java/gradle/java_no_deps.gradle' line: 265

What went wrong:
A problem occurred evaluating project ':internal-api:internal-api-8'.
No compatible toolchains found for request filter: {languageVersion=8, vendor=any, implementation=vendor-specific} (auto-detect false, auto-download false)
```

At this moment we found 2 possible approaches:

- ~~Edit  `gradle.properties` to add JAVA_HOME allowing valid toolchains search~~

- Edit `codeql-analysis.yml` and `test-lib-injection.yaml` to define `JAVA_8_HOME`

anderruiz (approved):

.github/workflows/codeql-analysis.yml
```diff
@@ -22,27 +22,27 @@ jobs:
...
-
-    - name: Perform CodeQL Analysis
-      uses: github/codeql-action/analyze@1a927e9307bc11970b2c679922ebc4d03a5bd980 # 1.0.31
-      
-    - name: upload_artifact
-      uses: actions/upload-artifact@82c141cc518b40d92cc801eee768e7aafc9c2fa2 # 2.3.1
-      with:
-         name: CodeQL Analysis Sarif
-         path: /home/runner/work/dd-trace-java/results/java.sarif    
+      - name: Checkout repository
```
smola, line 25: Can you avoid changing formatting here, just changing the relevant value?

jandro996: This change is the result of apply spotless automatic code formatting

bantonsson (approved):
I think the performance tables has `Deep` and `Limit` headings mixed up, but apart from that 👍🏻

PR #3649, opened 2022-07-11
Weak hashing detection intrumentation

DDJavierSantos:
# What Does This Do
Instrument the java.security.MessageDigest class so we can detect the use of weak hashing algorightms

# Motivation

To detect vulnerabilities in crypto code when weak algorithms are used. This new functionality will allow us to detect weak hashing vulnerabilites in the IAST module. 

# Additional Notes

Weak cipher algorithm detection is needed to complete the crypto functionality (separate PR).


dd-java-agent/instrumentation/insecurehash/src/main/java/datadog/trace/instrumentation/insecurehash/InsecureHashInstrumentation.java
```diff
@@ -0,0 +1,47 @@
+package datadog.trace.instrumentation.insecurehash;
```
smola, line 1: I think we might want to add this to a `jdk` module? Something like `jdk.insecurehash` (both at Gradle module level and Java package level).

dd-java-agent/instrumentation/insecurehash/src/main/java/datadog/trace/instrumentation/insecurehash/InsecureHashInstrumentation.java
```diff
@@ -0,0 +1,47 @@
...
+    super("insecurehash");
+  }
+
+
+  @Override
+  public String instrumentedType() {
+    return "java.security.MessageDigest";
+  }
+
+  //@Override
```
smola, line 28: Please, remove any commented code from the final PR.

dd-java-agent/instrumentation/insecurehash/src/main/java/datadog/trace/instrumentation/insecurehash/MessageDigestAdvice.java
```diff
@@ -0,0 +1,15 @@
...
+
+import datadog.trace.api.Config;
+import datadog.trace.bootstrap.instrumentation.api.AgentScope;
+import net.bytebuddy.asm.Advice;
+
+public class MessageDigestAdvice {
+
+  @Advice.OnMethodEnter
+  public static void onEnter(@Advice.Argument(value = 0, readOnly = true) String algorithm) {
+    if (Config.get().getInsecureHashingAlgorithms().contains(algorithm)){
```
smola, line 11: We need to define the API to call our vulnerability detection logic from aspects. I bet it won't be in `Config`. I guess we'll have something like `IastEvents.hashAlgorithmFound(algorithm)`.

cc @manuel-alvarez-alvarez

dd-trace-api/src/main/java/datadog/trace/api/ConfigDefaults.java
```diff
@@ -107,5 +112,7 @@ public final class ConfigDefaults {
 
   static final int DEFAULT_RESOLVER_TYPE_POOL_SIZE = 64;
 
+  static final Set<String> DEFAULT_IAST_INSECURE_HASHING_ALGORITHMS = new HashSet<>(Arrays.asList("MD2","RSA-MD4","md4","MD5","SHA")) ;
```
smola, line 115: I think it should be `"SHA-1"` here. Also, what happens with uppercase/lowercase variants?

dd-java-agent/instrumentation/insecurehash/src/main/java/datadog/trace/instrumentation/insecurehash/MessageDigestAdvice.java
DDJavierSantos: But since it will eventually be configurable, it will involve config, right? the code that I added to the Config function?

dd-trace-api/src/main/java/datadog/trace/api/ConfigDefaults.java
DDJavierSantos: this are the ids in java. There are aliases as well. I will add them. Also, we should probably uppercase the argument so we don't need to care about uppercase.

dd-java-agent/instrumentation/insecurehash/src/main/java/datadog/trace/instrumentation/insecurehash/MessageDigestAdvice.java
smola: Amending my previous comment: yes, the list of algorithms might involve config, but we'll want to decouple the instrumentation from the detection logic. So this detection and reporting will eventually be in the IAST module, not embedded into each instrumentation module.

dd-java-agent/instrumentation/insecurehash/src/main/java/datadog/trace/instrumentation/insecurehash/InsecureHashInstrumentation.java
manuel-alvarez-alvarez: I think it might be better to treat the JDK as a library and group all instrumentation targeting JDK classes in a single gradle module (this one can have different submodules for each JDK version or flavour(

dd-java-agent/instrumentation/insecurehash/src/main/java/datadog/trace/instrumentation/insecurehash/InsecureHashInstrumentation.java
```diff
@@ -0,0 +1,50 @@
...
+  public InsecureHashInstrumentation() {
+    super("insecurehash");
+  }
+
+  @Override
+  public String instrumentedType() {
+    return "java.security.MessageDigest";
+  }
+
+  protected boolean defaultEnabled() {
```
manuel-alvarez-alvarez, line 27: Isn't this already implied by extending `Instrumenter.AppSecIast`?

dd-java-agent/instrumentation/insecurehash/src/main/java/datadog/trace/instrumentation/insecurehash/MessageDigestAdvice.java
manuel-alvarez-alvarez: I agree with pretty much everything Santiago said already, doing the separation between instrumentation and IAST logic will be beneficial in the long run.

dd-java-agent/instrumentation/insecurehash/src/main/java/datadog/trace/instrumentation/insecurehash/InsecureHashInstrumentation.java
anderruiz: I agree with both. Keep in mind that most of the instrumentation will end up in that module anyway

dd-java-agent/instrumentation/insecurehash/src/main/java/datadog/trace/instrumentation/insecurehash/InsecureHashInstrumentation.java
```diff
@@ -0,0 +1,50 @@
...
+import net.bytebuddy.asm.Advice;
+import net.bytebuddy.description.method.MethodDescription;
+import net.bytebuddy.matcher.ElementMatchers;
+
+@AutoService(Instrumenter.class)
+public class InsecureHashInstrumentation extends Instrumenter.AppSecIast
+    implements Instrumenter.ForSingleType {
+
+  public InsecureHashInstrumentation() {
+    super("insecurehash");
```
anderruiz, line 19:
```suggestion
    super("weakhash");
```

dd-java-agent/instrumentation/insecurehash/src/main/java/datadog/trace/instrumentation/insecurehash/InsecureHashInstrumentation.java
```diff
@@ -0,0 +1,50 @@
...
+  }
+
+  @Override
+  public void adviceTransformations(AdviceTransformation transformation) {
+    transformation.applyAdvice(
+        NameMatchers.<MethodDescription>named("getInstance")
+            .and(takesArgument(0, String.class))
+            .and(isPublic())
+            .and(ElementMatchers.isStatic()),
+        InsecureHashInstrumentation.class.getName() + "$MessageDigestAdvice");
```
anderruiz, line 38:
```suggestion
        WeakHashInstrumentation.class.getName() + "$MessageDigestAdvice");
```

dd-trace-api/src/main/java/datadog/trace/api/ConfigDefaults.java
```diff
@@ -107,5 +112,8 @@ public final class ConfigDefaults {
 
   static final int DEFAULT_RESOLVER_TYPE_POOL_SIZE = 64;
 
+  static final Set<String> DEFAULT_IAST_INSECURE_HASHING_ALGORITHMS =
+      new HashSet<>(Arrays.asList("MD2", "RSA-MD4", "md4", "MD5", "SHA"));
+
```
anderruiz, line 117: Could we use ClassNameTrie or something similar not to have the list here?

dd-trace-api/src/main/java/datadog/trace/api/config/AppSecConfig.java
```diff
@@ -15,5 +15,8 @@ public final class AppSecConfig {
   public static final String APPSEC_OBFUSCATION_PARAMETER_VALUE_REGEXP =
       "appsec.obfuscation.parameter_value_regexp";
 
+  public static final String APPSEC_IAST_INSECURE_HASHING_ALGORITHMS =
+      "appsec.iast.insecure_hashing_algorigthms";
```
anderruiz, line 19:
```suggestion
      "appsec.iast.insecure_hashing_algorithms";
```

dd-java-agent/instrumentation/insecurehash/src/main/java/datadog/trace/instrumentation/insecurehash/InsecureHashInstrumentation.java
```diff
@@ -0,0 +1,50 @@
...
+  }
+
+  public static class MessageDigestAdvice {
+
+    @Advice.OnMethodEnter
+    public static void onEnter(@Advice.Argument(value = 0, readOnly = true) String algorithm) {
+      if (Config.get().getInsecureHashingAlgorithms().contains(algorithm)) {
```
anderruiz, line 45: Shouldn't we cache Config?

settings.gradle
```diff
@@ -167,6 +167,7 @@ include ':dd-java-agent:instrumentation:hibernate:core-4.3'
 include ':dd-java-agent:instrumentation:http-url-connection'
 include ':dd-java-agent:instrumentation:hystrix-1.4'
 include ':dd-java-agent:instrumentation:ignite-2.0'
+include ':dd-java-agent:instrumentation:insecurehash'
```
anderruiz, line 170:
```suggestion
include ':dd-java-agent:instrumentation:weakhash'
```

dd-java-agent/instrumentation/insecurehash/src/main/java/datadog/trace/instrumentation/insecurehash/InsecureHashInstrumentation.java
anderruiz: There are additional filtering features that we should include here

dd-java-agent/instrumentation/insecurehash/src/main/java/datadog/trace/instrumentation/insecurehash/InsecureHashInstrumentation.java
DDJavierSantos: finally it has been agreed to create a instrumentation module named java-security

dd-java-agent/instrumentation/java-security/java-security.gradle
```diff
@@ -0,0 +1,24 @@
...
+    dirName = 'test'
+  }
+}
+
+tasks.withType(Test).configureEach {
+  jvmArgs += ['-Ddd.iast.enabled=true']
+}
+
+dependencies {
+  testImplementation group: 'org.bouncycastle', name: 'bcprov-jdk18on', version: '1.71'
```
ValentinZakharov, line 23: You are using `Bouncy Castle Provider` for jdk 1.8+, but some of our CircleCI tests are running on jdk 1.7 and as the result `bouncycastle` tests are failing. May be it make sense to use older version - `bcprov-jdk15on`, or prevent to run this test in java7.

dd-java-agent/instrumentation/java-security/src/main/java/datadog/trace/instrumentation/java/security/WeakHashInstrumentation.java
```diff
@@ -0,0 +1,57 @@
...
+
+  @Override
+  public String instrumentedType() {
+    return "java.security.MessageDigest";
+  }
+
+  @Override
+  public boolean isEnabled() {
+    return super.isEnabled();
+  }
```
ValentinZakharov, line 33: I'm sorry, I don't get what was the reason of overriding `isEnabled()` if we call base method?

dd-java-agent/instrumentation/java-security/java-security.gradle
DDJavierSantos: Done

dd-java-agent/instrumentation/java-security/src/main/java/datadog/trace/instrumentation/java/security/WeakHashInstrumentation.java
DDJavierSantos: Done

jbachorik:
This is looking funny - the PR is showing like you are going to re-apply all 100+ commits which are already in master.
Can you, please, recheck and make sure only the real changes are to be reviewed?

dd-java-agent/instrumentation/java-security/src/main/java/datadog/trace/instrumentation/java/security/WeakHashInstrumentation.java
```diff
@@ -0,0 +1,57 @@
...
+            .and(isPublic())
+            .and(ElementMatchers.isStatic()),
+        WeakHashInstrumentation.class.getName() + "$MessageDigestAdvice");
+  }
+
+  public static class MessageDigestAdvice {
+
+    @Advice.OnMethodEnter
+    public static void onEnter(@Advice.Argument(value = 0, readOnly = true) String algorithm) {
+      if (Config.get().getWeakHashingAlgorithms().contains(algorithm.toUpperCase())) {
```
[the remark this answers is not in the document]
smola: Please, just call the callback here directly, without checking anything else. That way we keep the logic in a single place, and we don't need to check for the algorithm twice.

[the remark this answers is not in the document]
team-sourcecodeintegration: Done

bantonsson (approved):

dd-java-agent/iast/src/main/java/com/datadog/iast/IastModuleImpl.java
```diff
@@ -0,0 +1,34 @@
...
+
+  public void onHashingAlgorithm(String algorithm) {
+    if (Config.get().getWeakHashingAlgorithms().contains(algorithm.toUpperCase())) {
+      // get StackTraceElement for the callee of MessageDigest
+      StackTraceElement stackTraceElement =
+          StackWalkerFactory.INSTANCE
+              .walk(
+                  stack ->
+                      stack.filter(s -> !s.getClassName().equals("java.security.MessageDigest")))
+              .findFirst()
```
smola, line 23: You need to do `.findFirst().get()` within the `walk` call. The stream is not valid once it leaves the scope of `walk`.

smola (changes requested):

smola (approved):

PR #3660, opened 2022-07-16
JDK9 stackwalker support for getstacktrace

anderruiz:
# What Does This Do

This PR includes support for JDK9 StackWalker (whenever it is available) to retrieve stacktrace elements excluding those which are part of the agent

# Motivation

JDK9 StackWalker is a faster and the preferred way to retrieve stack traces if it is available

# Additional Notes

Future updates will allow skipping frames, limiting the amount of them and other different options


jandro996:
Rebase with last changes in alejandro.gonzalez/Get_stacktrace_for_reported_vulnerabilities :)

PR #3662, opened 2022-07-18
HotSpotStackWalker utility implementation

jandro996:
# What Does This Do

HotSpotStackWalker utility implementation  that improves DefaultStackWalker implementation for JDK8 with hotspot.

# Motivation

Improve performance obtaining stacktrace

# Additional Notes

This PR continues [Util to get stacktrace for reported vulnerabilities](https://github.com/DataDog/dd-trace-java/pull/3647)

### DIFF
--- a/internal-api/internal-api-8/src/main/java/datadog/trace/util/stacktrace/HotSpotStackWalker.java
+++ b/internal-api/internal-api-8/src/main/java/datadog/trace/util/stacktrace/HotSpotStackWalker.java
@@ -1,0 +1,63 @@
+package datadog.trace.util.stacktrace;
+
+import java.util.Iterator;
+import java.util.NoSuchElementException;
+import java.util.stream.Stream;
+import java.util.stream.StreamSupport;
+
+public class HotSpotStackWalker extends AbstractStackWalker {
+
+  private final sun.misc.JavaLangAccess access = sun.misc.SharedSecrets.getJavaLangAccess();
+
+  @Override
+  public boolean isEnabled() {
+    try {
+      String str = System.getProperty("java.version");
+      if (str.length() > 3
+          && str.startsWith("1.")
+          && Integer.parseInt(Character.toString(str.charAt(2))) >= 7) {
+        if (access != null) {
+          access.getStackTraceElement(new Throwable(), 0);
+          return true;
+        }
+      }
+    } catch (Throwable localThrowable) {
+    }
+    return false;
+  }
+
+  @Override
+  Stream<StackTraceElement> doGetStack() {
+
+    Throwable throwable = new Throwable();
+
+    Iterable<StackTraceElement> iterable =
+        () ->
+            new Iterator<StackTraceElement>() {
+
+              int index = 0;
+
+              final int size = access.getStackTraceDepth(throwable);
+
+              @Override
+              public boolean hasNext() {
+                return index < size;
+              }
+
+              @Override
+              public StackTraceElement next() {
+                if (index >= size) {
+                  throw new NoSuchElementException();
+                }
+                return access.getStackTraceElement(throwable, index++);
+              }
+
+              @Override
+              public void remove() {
+                throw new UnsupportedOperationException();
+              }
+            };
+
+    return StreamSupport.stream(iterable.spliterator(), false);
+  }
+}
--- a/internal-api/internal-api-8/src/main/java/datadog/trace/util/stacktrace/HotSpotStackWalker.java
+++ b/internal-api/internal-api-8/src/main/java/datadog/trace/util/stacktrace/HotSpotStackWalker.java
@@ -1,5 +1,6 @@
 package datadog.trace.util.stacktrace;
 
+import datadog.trace.api.Platform;
 import java.util.Iterator;
 import java.util.NoSuchElementException;
 import java.util.stream.Stream;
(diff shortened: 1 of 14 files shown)


internal-api/internal-api-8/src/main/java/datadog/trace/util/stacktrace/HotSpotStackWalker.java
```diff
@@ -0,0 +1,63 @@
...
+import java.util.stream.StreamSupport;
+
+public class HotSpotStackWalker extends AbstractStackWalker {
+
+  private final sun.misc.JavaLangAccess access = sun.misc.SharedSecrets.getJavaLangAccess();
+
+  @Override
+  public boolean isEnabled() {
+    try {
+      String str = System.getProperty("java.version");
```
jbachorik, line 15: You can use `Platform.isJavaVersion(7)` instead.

internal-api/internal-api-8/src/main/java/datadog/trace/util/stacktrace/HotSpotStackWalker.java
```diff
@@ -0,0 +1,63 @@
+package datadog.trace.util.stacktrace;
+
+import java.util.Iterator;
+import java.util.NoSuchElementException;
+import java.util.stream.Stream;
+import java.util.stream.StreamSupport;
+
+public class HotSpotStackWalker extends AbstractStackWalker {
+
+  private final sun.misc.JavaLangAccess access = sun.misc.SharedSecrets.getJavaLangAccess();
```
jbachorik, line 10: This will blow up for anything newer than JDK 8 - `SharedSecrets` is an internal JDK class and as such is completely off-limits since JDK 15 and on top of that it was moved to a different package since JDK 9 (`jdk.internal.access` - https://github.com/openjdk/jdk11u-dev/blob/75519b1bfb0e5dda9b43ecd75d0c0031205d15a5/src/java.base/share/classes/jdk/internal/misc/SharedSecrets.java#L26)

internal-api/internal-api-8/src/main/java/datadog/trace/util/stacktrace/HotSpotStackWalker.java
```diff
@@ -0,0 +1,63 @@
...
+              public boolean hasNext() {
+                return index < size;
+              }
+
+              @Override
+              public StackTraceElement next() {
+                if (index >= size) {
+                  throw new NoSuchElementException();
+                }
+                return access.getStackTraceElement(throwable, index++);
```
jbachorik, line 52: What is the reason for using `JavaLangAccess` to do this?
When I check the implementation, it turns out that it is just delegating to `Throwable` to get the stack trace elements (https://github.com/openjdk/jdk8u-dev/blob/40a18594ba21d7a345ff82792fc361f7e691e669/jdk/src/share/classes/java/lang/System.java#L1264). Wouldn't then this be 100% equivalent to the `DefaultStackWalker` implementation?

internal-api/internal-api-8/src/main/java/datadog/trace/util/stacktrace/HotSpotStackWalker.java
jandro996: Thanks,  I forgot to check the access init! On the other hand we are wondering how should we tag tests for different JDK versions and vendors... Do you know if we should add something to CI configuration, to gradle, or tests themselves?

smola: > This will blow up for anything newer than JDK 8 

Note that this is expected. Other than the fallback implementation, there will be a Java 9+ implementation, and this one used only for Java 8 when supported.

anderruiz: @jandro996 keep in mind that the use of the internal API should be  try/catched and isEnabled should return false if it was not available, As it is, it is not correct. @jbachorik HotSpotStackWalker is a faster way to retrieve stacks than the traditional one that gathers all the stack, in DD it will only be used for JDK8. For newer JDK9  the preferred java.lang.StackWalker will be used instead (pending PR). Originally in Hdiv HotSpot implementation was used for JDK6-8.

internal-api/internal-api-8/src/main/java/datadog/trace/util/stacktrace/HotSpotStackWalker.java
anderruiz: The difference is that while using Throwable all StackTraceElements are created, with this implementation you can decide which ones needs to be retrieved, allowing to skip several, or to stop iterating.

The options to skip and to stop are not implemented in this PR but this will be required in the future

internal-api/internal-api-8/src/main/java/datadog/trace/util/stacktrace/HotSpotStackWalker.java
jbachorik: The `JavaLangAccess.getStackTraceElement()` is delegating to `Throwable.getStackTraceElement()` (https://github.com/openjdk/jdk8u-dev/blob/40a18594ba21d7a345ff82792fc361f7e691e669/jdk/src/share/classes/java/lang/System.java#L1264) - I wonder how this can be faster than calling `Throwable.getStackTrace()` 🤔

PR #3663, opened 2022-07-18
Bridge between instrumentation and IAST

manuel-alvarez-alvarez:
# What Does This Do
Introduces a bridge class between instrumentations and an IAST module. All hooks in IAST instrumentation will call static methods of this bridge.

# Motivation
Due to the potentially large amount of callbacks the IAST module will receive and performance requirements for certain operations, we have decided to introduce a custom bridge for direct communication between instrumentations and IAST.
This bridge will have static methods for each of the callbacks that the IAST will be installing. Static methods are selected mostly because they need simpler instrumentation, with less stack operations and better performance.
The callbacks will be added to client code by instrumentation modules. The callback implementations will be in the dd-java-agent/iast module.

# Additional Notes
`IASTModule` will be implemented in future pull requests.

### DIFF
--- a/internal-api/src/main/java/datadog/trace/api/iast/InstrumentationBridge.java
+++ b/internal-api/src/main/java/datadog/trace/api/iast/InstrumentationBridge.java
@@ -1,0 +1,58 @@
+package datadog.trace.api.iast;
+
+import javax.annotation.Nonnull;
+import org.slf4j.Logger;
+import org.slf4j.LoggerFactory;
+
+/**
+ * Bridge between instrumentations and {@link IastModule} that contains the business logic relative
+ * to vulnerability detection. The class contains a list of {@code public static} methods that will
+ * be injected into the bytecode via {@code invokestatic} instructions. It's important that all
+ * methods are protected from exception leakage.
+ */
+public abstract class InstrumentationBridge {
+
+  private static final Logger LOG = LoggerFactory.getLogger(InstrumentationBridge.class);
+
+  private static IastModule MODULE;
+
+  private InstrumentationBridge() {}
+
+  public static void registerIastModule(final IastModule module) {
+    MODULE = module;
+  }
+
+  /**
+   * Executed when access to a cryptographic cipher is detected
+   *
+   * <p>{@link javax.crypto.Cipher#getInstance(String)}
+   */
+  public static void onCipherGetInstance(@Nonnull final String algorithm) {
+    try {
+      if (MODULE != null) {
+        MODULE.onCipherAlgorithm(algorithm);
+      }
+    } catch (final Throwable t) {
+      onUnexpectedException("Callback for onCipher threw.", t);
+    }
+  }
+
+  /**
+   * Executed when access to a message digest algorithm is detected
+   *
+   * <p>{@link java.security.MessageDigest#getInstance(String)}
+   */
+  public static void onMessageDigestGetInstance(@Nonnull final String algorithm) {
+    try {
+      if (MODULE != null) {
+        MODULE.onHashingAlgorithm(algorithm);
+      }
+    } catch (final Throwable t) {
+      onUnexpectedException("Callback for onHash threw.", t);
+    }
+  }
+
+  private static void onUnexpectedException(final String message, final Throwable error) {
+    LOG.warn(message, error);
+  }
+}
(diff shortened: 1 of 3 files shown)


internal-api/src/main/java/datadog/trace/api/iast/IASTModule.java
```diff
@@ -0,0 +1,8 @@
+package datadog.trace.api.iast;
+
+public interface IASTModule {
+
+  void onCipher(final String algorithm);
```
anderruiz, line 5: Is final keyword inserted by save actions? seems useless

internal-api/src/main/java/datadog/trace/api/iast/InstrumentationBridge.java
```diff
@@ -0,0 +1,28 @@
...
+import datadog.trace.api.function.Supplier;
+
+public abstract class InstrumentationBridge {
+
+  static Supplier<IASTModule> MODULE = initializeModule();
+
+  private InstrumentationBridge() {}
+
+  public static void onCipher(final String algorithm) {
+    MODULE.get().onCipher(algorithm);
```
anderruiz, line 12: If we want to use this approach at the end, we should consider having the IASTModule as a field in the class instead of using a Supplier (one additional method call per hit)

anderruiz (approved):

internal-api/src/main/java/datadog/trace/api/iast/IASTModule.java
manuel-alvarez-alvarez: Nope, it was me, and yes it's just noise, removed :wink:

internal-api/src/main/java/datadog/trace/api/iast/InstrumentationBridge.java
manuel-alvarez-alvarez: My initial idea was to allow setting an empty instance of the module when the overhead control engine kicks in, but it's just weird to do it in here. Changed it so it's now a field backed instance injected via SPI.

What other approaches do you have in mind for the bridge?

jandro996 (approved):

smola (approved):

ValentinZakharov (approved):

anderruiz (approved):

internal-api/src/main/java/datadog/trace/api/iast/InstrumentationBridge.java
```diff
@@ -0,0 +1,50 @@
...
+   *
+   * <p>{@link javax.crypto.Cipher#getInstance(String)}
+   */
+  public static void onCipherGetInstance(final String algorithm) {
+    try {
+      if (MODULE != null) {
+        MODULE.onCipherAlgorithm(algorithm);
+      }
+    } catch (Throwable t) {
+      LOG.warn("Callback for onCipher threw.", t);
```
dougqh, line 32: Hmm, we really need a facility for where to send unexpected exceptions.  It has been discussed before, but for now, I guess logging is okay.

internal-api/src/main/java/datadog/trace/api/iast/InstrumentationBridge.java
```diff
@@ -0,0 +1,50 @@
...
+ * to vulnerability detection. The class contains a list of {@code public static} methods that will
+ * be injected into the bytecode via {@code invokestatic} instructions. It's important that all
+ * methods are protected from exception leakage.
+ */
+public abstract class InstrumentationBridge {
+
+  private static final Logger LOG = LoggerFactory.getLogger(InstrumentationBridge.class);
+
+  // TODO (malvarez) inject current instance of the module via SPI or other compatible approach
+  static IASTModule MODULE;
```
dougqh, line 17: I think SPI would be overkill.  It also adds start-up overhead of interrogating ClassLoaders which we'd prefer to avoid.

manuel-alvarez-alvarez: Yes I agree, the reason to include the SPI comment is that it was used in all instrumentations. If possible I think build time IoC is much better than at runtime, will it be possible to use something like [Dagger](https://dagger.dev/) in the tracer?

internal-api/src/main/java/datadog/trace/api/iast/InstrumentationBridge.java
manuel-alvarez-alvarez: Yes, one of the reasons of the module is to have a common place to catch the unexpected and when such a facility is built, forward all exceptions.

bantonsson (approved):

PR #3671, opened 2022-07-21
IG support for IAST 

cataphract:
Stores different "request context data" in DDSpanContext for AppSec and IAST and makes the InstrumentationGateway have two separate sets of callbacks.


smola (approved):
This would work for us on the IAST side.

ValentinZakharov:
Proposed implementation looks fine, but I don't really like all that duplications for `APPSEC` and `IAST`. I'd rather prefer some generic approach, because we might need to add more different contexts in future.

cataphract:
> Proposed implementation looks fine, but I don't really like all that duplications for `APPSEC` and `IAST`. I'd rather prefer some generic approach, because we might need to add more different contexts in future.

That's reasonable, but to support an arbitrary number of contexts, the overhead would need to go up because it would imply (I think) extra indirection (e.g. storing the contexts inside maps or at least arrays).

PR #3673, opened 2022-07-22
IAST vulnerability reporting API

smola:
# What Does This Do

Add initial vulnerability reporting API for IAST. This PR includes:
* `iast` module and initial skeleton and initialization of the IAST subsystem.
* IAST callbacks when requests start, and end.
* Request context for IAST data.
* Vulnerability model.
* Reporter to add vulnerabilities to a request, and serialize them at the end of the request.

When IAST is enabled, we’ll add vulnerabilities for each request in a JSON-encoded object, stored in an internal tag (_dd.iast.json). The vulnerability objects are accumulated during the request, and encoded to JSON at the end of the request.

# Motivation

This API will be used whenever a vulnerability is found. It will attach a vulnerability to the current request, and it will encode them in an internal tag in JSON once the request ends.

# Additional Notes

Out of scope of this PR but coming up:
* Deduplicate vulnerabilities before reporting to the backend.
* Limit the number of vulnerabilities per request.

### DIFF
--- a/dd-java-agent/agent-bootstrap/src/main/java/datadog/trace/bootstrap/Agent.java
+++ b/dd-java-agent/agent-bootstrap/src/main/java/datadog/trace/bootstrap/Agent.java
@@ -374,6 +374,7 @@ public class Agent {
 
       installDatadogTracer(scoClass, sco);
       maybeStartAppSec(scoClass, sco);
+      maybeStartIast(scoClass, sco);
       maybeStartRemoteConfig(scoClass, sco);
 
       if (telemetryEnabled) {
@@ -593,6 +594,33 @@ public class Agent {
       appSecInstallerMethod.invoke(null, ss, sco);
     } catch (final Throwable ex) {
       log.warn("Not starting AppSec subsystem: {}", ex.getMessage());
+    }
+  }
+
+  private static void maybeStartIast(Class<?> scoClass, Object o) {
+    if (iastEnabled) {
+      if (isJavaVersionAtLeast(8)) {
+        try {
+          SubscriptionService ss =
+              AgentTracer.get().getSubscriptionService(RequestContextSlot.IAST);
+          startIast(ss, scoClass, o);
+        } catch (Exception e) {
+          log.error("Error starting IAST subsystem", e);
+        }
+      } else {
+        log.warn("IAST subsystem requires Java 8 or later to run");
+      }
+    }
+  }
+
+  private static void startIast(SubscriptionService ss, Class<?> scoClass, Object sco) {
+    try {
+      final Class<?> appSecSysClass = AGENT_CLASSLOADER.loadClass("com.datadog.iast.IastSystem");
+      final Method iastInstallerMethod =
+          appSecSysClass.getMethod("start", SubscriptionService.class);
+      iastInstallerMethod.invoke(null, ss);
+    } catch (final Throwable e) {
+      log.warn("Not starting IAST subsystem", e);
     }
   }
 
--- a/dd-java-agent/iast/iast.gradle
+++ b/dd-java-agent/iast/iast.gradle
@@ -1,0 +1,55 @@
+plugins {
+  id "com.github.johnrengelman.shadow"
+}
+
+apply from: "$rootDir/gradle/java.gradle"
+apply from: "$rootDir/gradle/version.gradle"
+
+dependencies {
+  api deps.slf4j
+  implementation project(':internal-api')
+  implementation project(':internal-api:internal-api-8')
+  implementation project(':internal-api:internal-api-9')
+  implementation group: 'com.squareup.moshi', name: 'moshi', version: versions.moshi
+
+  testImplementation deps.bytebuddy
+  testImplementation project(':utils:test-utils')
+}
+
+sourceCompatibility = JavaVersion.VERSION_1_8
+targetCompatibility = JavaVersion.VERSION_1_8
+
+shadowJar {
+  dependencies deps.excludeShared
+}
+
+jar {
+  archiveClassifier = 'unbundled'
+}
+
+ext {
+  minimumBranchCoverage = 0.6
+  minimumInstructionCoverage = 0.8
+  excludedClassesCoverage = [
+    // Avoid coverage measurement of model getters atm
+    'com.datadog.iast.model.Evidence',
+    'com.datadog.iast.model.Range',
+    'com.datadog.iast.model.Source',
+    'com.datadog.iast.model.Vulnerability',
+    // Small JsonAdapters with unimplemented fromJson
+    'com.datadog.iast.model.json.SourceTypeAdapter',
+  ]
+  excludedClassesBranchCoverage = []
+  excludedClassesInstructionCoverage = []
+  minJavaVersionForTests = JavaVersion.VERSION_1_8
+}
+
+tasks.withType(Test).configureEach {
+  jvmArgs += ['-Ddd.iast.enabled=true']
+}
+def rootDir = project.rootDir
+spotless {
+  java {
+    target 'src/**/*.java'
+  }
+}
--- a/dd-java-agent/iast/src/main/java/com/datadog/iast/IastRequestContext.java
+++ b/dd-java-agent/iast/src/main/java/com/datadog/iast/IastRequestContext.java
@@ -1,0 +1,24 @@
+package com.datadog.iast;
+
+import com.datadog.iast.model.VulnerabilityBatch;
+import java.util.concurrent.atomic.AtomicBoolean;
+
+public class IastRequestContext {
+
+  private final VulnerabilityBatch vulnerabilityBatch;
+
+  private final AtomicBoolean spanDataIsSet;
+
+  public IastRequestContext() {
+    this.vulnerabilityBatch = new VulnerabilityBatch();
+    this.spanDataIsSet = new AtomicBoolean(false);
+  }
+
+  public VulnerabilityBatch getVulnerabilityBatch() {
+    return vulnerabilityBatch;
+  }
+
+  public boolean getAndSetSpanDataIsSet() {
+    return spanDataIsSet.getAndSet(true);
+  }
+}
--- a/dd-java-agent/iast/src/main/java/com/datadog/iast/IastSystem.java
+++ b/dd-java-agent/iast/src/main/java/com/datadog/iast/IastSystem.java
@@ -1,0 +1,31 @@
+package com.datadog.iast;
+
+import datadog.trace.api.Config;
+import datadog.trace.api.function.Supplier;
+import datadog.trace.api.gateway.EventType;
+import datadog.trace.api.gateway.Events;
+import datadog.trace.api.gateway.Flow;
+import datadog.trace.api.gateway.SubscriptionService;
+import org.slf4j.Logger;
+import org.slf4j.LoggerFactory;
+
+public class IastSystem {
+
+  private static final Logger log = LoggerFactory.getLogger(IastSystem.class);
+
+  public static void start(final SubscriptionService ss) {
+    final Config config = Config.get();
+    if (!config.isIastEnabled()) {
+      log.debug("IAST is disabled");
+      return;
+    }
+    log.debug("IAST is starting");
+
+    registerRequestStartedCallback(ss);
+  }
+
+  private static void registerRequestStartedCallback(final SubscriptionService ss) {
+    final EventType<Supplier<Flow<Object>>> event = Events.get().requestStarted();
+    ss.registerCallback(event, () -> new Flow.ResultFlow<>(new IastRequestContext()));
+  }
+}
--- a/dd-java-agent/iast/src/main/java/com/datadog/iast/Reporter.java
+++ b/dd-java-agent/iast/src/main/java/com/datadog/iast/Reporter.java
@@ -1,0 +1,32 @@
+package com.datadog.iast;
+
+import com.datadog.iast.model.Vulnerability;
+import com.datadog.iast.model.VulnerabilityBatch;
+import datadog.trace.api.gateway.RequestContext;
+import datadog.trace.api.gateway.RequestContextSlot;
+import datadog.trace.bootstrap.instrumentation.api.AgentSpan;
+
+/** Reports IAST vulnerabilities. */
+public final class Reporter {
+
+  private Reporter() {}
+
+  public static void report(final AgentSpan span, final Vulnerability vulnerability) {
+    if (span == null) {
+      return;
+    }
+    final RequestContext reqCtx = span.getRequestContext();
+    if (reqCtx == null) {
+      return;
+    }
+    final IastRequestContext ctx = reqCtx.getData(RequestContextSlot.IAST);
+    if (ctx == null) {
+      return;
+    }
+    final VulnerabilityBatch batch = ctx.getVulnerabilityBatch();
+    batch.add(vulnerability);
+    if (!ctx.getAndSetSpanDataIsSet()) {
+      reqCtx.getTraceSegment().setDataTop("iast", batch);
+    }
+  }
+}
--- a/dd-java-agent/iast/src/main/java/com/datadog/iast/model/Evidence.java
+++ b/dd-java-agent/iast/src/main/java/com/datadog/iast/model/Evidence.java
@@ -1,0 +1,28 @@
+package com.datadog.iast.model;
+
+import javax.annotation.Nonnull;
+import javax.annotation.Nullable;
+
+public final class Evidence {
+
+  private final @Nonnull String value;
+
+  private final @Nullable Range[] ranges;
+
+  public Evidence(final String value) {
+    this(value, null);
+  }
+
+  public Evidence(final String value, final Range[] ranges) {
+    this.value = value;
+    this.ranges = ranges;
+  }
+
+  public String getValue() {
+    return value;
+  }
+
+  public Range[] getRanges() {
+    return ranges;
+  }
+}
--- a/dd-java-agent/iast/src/main/java/com/datadog/iast/model/Location.java
+++ b/dd-java-agent/iast/src/main/java/com/datadog/iast/model/Location.java
@@ -1,0 +1,25 @@
+package com.datadog.iast.model;
+
+public final class Location {
+
+  private final String path;
+
+  private final int line;
+
+  private Location(final String path, final int line) {
+    this.path = path;
+    this.line = line;
+  }
+
+  public static Location forStack(final StackTraceElement stack) {
+    return new Location(stack.getClassName(), stack.getLineNumber());
+  }
+
+  public String getPath() {
+    return path;
+  }
+
+  public int getLine() {
+    return line;
+  }
+}
--- a/dd-java-agent/iast/src/main/java/com/datadog/iast/model/Range.java
+++ b/dd-java-agent/iast/src/main/java/com/datadog/iast/model/Range.java
@@ -1,0 +1,29 @@
+package com.datadog.iast.model;
+
+import com.datadog.iast.model.json.SourceIndex;
+import javax.annotation.Nonnegative;
+import javax.annotation.Nonnull;
+
+public final class Range {
+  private final @Nonnegative int start;
+  private final @Nonnegative int length;
+  private final @Nonnull @SourceIndex Source source;
+
+  public Range(final int start, final int length, final Source source) {
+    this.start = start;
+    this.length = length;
+    this.source = source;
+  }
+
+  public int getStart() {
+    return start;
+  }
+
+  public int getLength() {
+    return length;
+  }
+
+  public Source getSource() {
+    return source;
+  }
+}
--- a/dd-java-agent/iast/src/main/java/com/datadog/iast/model/Source.java
+++ b/dd-java-agent/iast/src/main/java/com/datadog/iast/model/Source.java
@@ -1,0 +1,27 @@
+package com.datadog.iast.model;
+
+import com.datadog.iast.model.json.SourceTypeString;
+
+public final class Source {
+  private final @SourceTypeString byte origin;
+  private final String name;
+  private final String value;
+
+  public Source(final byte origin, final String name, final String value) {
+    this.origin = origin;
+    this.name = name;
+    this.value = value;
+  }
+
+  public byte getOrigin() {
+    return origin;
+  }
+
+  public String getName() {
+    return name;
+  }
+
+  public String getValue() {
+    return value;
+  }
+}
--- a/dd-java-agent/iast/src/main/java/com/datadog/iast/model/SourceType.java
+++ b/dd-java-agent/iast/src/main/java/com/datadog/iast/model/SourceType.java
@@ -1,0 +1,24 @@
+package com.datadog.iast.model;
+
+public final class SourceType {
+
+  private SourceType() {}
+
+  public static final byte NONE = 0;
+
+  public static final byte REQUEST_QUERY_PARAMETER = 1;
+  private static final String REQUEST_QUERY_PARAMETER_STRING = "http.url_details.queryString";
+  public static final byte REQUEST_PATH = 2;
+  private static final String REQUEST_PATH_STRING = "http.url_details.path";
+
+  public static String toString(final byte sourceType) {
+    switch (sourceType) {
+      case REQUEST_QUERY_PARAMETER:
+        return REQUEST_QUERY_PARAMETER_STRING;
+      case REQUEST_PATH:
+        return REQUEST_PATH_STRING;
+      default:
+        return null;
+    }
+  }
+}
--- a/dd-java-agent/iast/src/main/java/com/datadog/iast/model/Vulnerability.java
+++ b/dd-java-agent/iast/src/main/java/com/datadog/iast/model/Vulnerability.java
@@ -1,0 +1,32 @@
+package com.datadog.iast.model;
+
+import javax.annotation.Nonnull;
+import javax.annotation.Nullable;
+
+public final class Vulnerability {
+
+  private final @Nonnull VulnerabilityType type;
+
+  private final @Nonnull Location location;
+
+  private final @Nullable Evidence evidence;
+
+  public Vulnerability(
+      final VulnerabilityType type, final Location location, final Evidence evidence) {
+    this.type = type;
+    this.location = location;
+    this.evidence = evidence;
+  }
+
+  public VulnerabilityType getType() {
+    return type;
+  }
+
+  public Location getLocation() {
+    return location;
+  }
+
+  public Evidence getEvidence() {
+    return evidence;
+  }
+}
--- a/dd-java-agent/iast/src/main/java/com/datadog/iast/model/VulnerabilityBatch.java
+++ b/dd-java-agent/iast/src/main/java/com/datadog/iast/model/VulnerabilityBatch.java
@@ -1,0 +1,41 @@
+package com.datadog.iast.model;
+
+import com.datadog.iast.model.json.VulnerabilityEncoding;
+import java.util.ArrayList;
+import java.util.List;
+
+/** Collects vulnerabilities and serializes to JSON lazily on {@link #toString()}. */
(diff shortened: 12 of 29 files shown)


dd-java-agent/agent-bootstrap/src/main/java/datadog/trace/bootstrap/Agent.java
```diff
@@ -617,6 +631,26 @@ private static void startAppSec(InstrumentationGateway gw, Class<?> scoClass, Ob
...
+      return;
+    }
+
+    InstrumentationGateway gw = AgentTracer.get().instrumentationGateway();
+    startIast(gw, scoClass, o);
+  }
+
+  private static void startIast(InstrumentationGateway gw, Class<?> scoClass, Object sco) {
+    try {
+      final Class<?> iastSysClass = APPSEC_CLASSLOADER.loadClass("com.datadog.iast.IastSystem");
```
manuel-alvarez-alvarez, line 645: Shouldn't be `IAST_CLASSLOADER`?

dd-java-agent/iast/src/main/java/com/datadog/iast/IastSystem.java
```diff
@@ -0,0 +1,86 @@
...
+          event,
+          (RequestContext<Object> ctx, IGSpanInfo span) -> {
+            Reporter.finalizeReports(ctx);
+            return Flow.ResultFlow.empty();
+          });
+    } else {
+      ig.registerCallback(
+          event,
+          (RequestContext<Object> ctx, IGSpanInfo span) -> {
+            Reporter.finalizeReports(ctx);
```
manuel-alvarez-alvarez, line 58: We should `try/catch` the reporter it to prevent short circuiting the callback

dd-java-agent/iast/src/main/java/com/datadog/iast/model/Range.java
```diff
@@ -0,0 +1,30 @@
+package com.datadog.iast.model;
+
+import com.datadog.iast.model.json.SourceIndex;
+import javax.annotation.Nonnegative;
+import javax.annotation.Nonnull;
+
+public class Range {
+  private @Nonnegative int start;
+  private @Nonnegative int length;
+  private @Nonnull @SourceIndex Source source;
```
manuel-alvarez-alvarez, line 10: Neat!

smola: This is the beautiful part of it. The not-so-beautiful is in the `JsonAdapter` :wink:

dd-java-agent/iast/src/main/java/com/datadog/iast/model/Vulnerability.java
```diff
@@ -0,0 +1,86 @@
+package com.datadog.iast.model;
+
+import datadog.trace.util.stacktrace.StackWalkerFactory;
+import javax.annotation.Nonnull;
+import javax.annotation.Nullable;
+
+public class Vulnerability {
```
manuel-alvarez-alvarez, line 7: Is Lombok an option?, writing/maintaining getters/sertters/builders/equals/hashcodes make me  :cry:

internal-api/src/main/java/datadog/trace/api/gateway/RequestContext.java
```diff
@@ -9,5 +9,7 @@
 public interface RequestContext<D> {
   D getData();
 
+  Object getIastContext();
```
smola, line 12: You can ignore this parts. It needs a rebase from https://github.com/DataDog/dd-trace-java/pull/3671 once it's ready.

dd-java-agent/iast/src/test/groovy/com/datadog/iast/model/json/VulnerabilityEncodingTest.groovy
```diff
@@ -0,0 +1,152 @@
...
+      .evidence(Evidence.forAlgorithm("MD5"))
+      .location(Location.forStack(new StackTraceElement("foo", "foo", "foo", 1)))
+      .build()
+    ])
+
+    when:
+    def result = VulnerabilityEncoding.toJson(value)
+
+    then:
+    slurper.parseText(result) == slurper.parseText("""{
```
manuel-alvarez-alvarez, line 31: Once `GStrings` came and kick me in the ass (probably nowadays it's fine) but I still prefer `' '` for one liners and `''' '''` for multiline strings.

dd-java-agent/iast/iast.gradle
```diff
@@ -0,0 +1,71 @@
+plugins {
+  id "com.github.johnrengelman.shadow"
+  id "me.champeau.jmh"
```
manuel-alvarez-alvarez, line 3: jmh is just noise for now, not sure if we want to remove it until we have the benchmarks

dd-java-agent/iast/src/test/groovy/com/datadog/iast/ReporterTest.groovy
```diff
@@ -0,0 +1,57 @@
...
+import datadog.trace.api.TraceSegment
+import datadog.trace.api.gateway.RequestContext
+import datadog.trace.bootstrap.instrumentation.api.AgentSpan
+import datadog.trace.test.util.DDSpecification
+
+class ReporterTest extends DDSpecification {
+
+  void 'basic vulnerability reporting'() {
+    given:
+    def traceSegment = Mock(TraceSegment.class)
```
manuel-alvarez-alvarez, line 15: Nowadays I use `final` much more than `def`

dd-java-agent/iast/src/main/java/com/datadog/iast/model/Vulnerability.java
smola: Lombok does not seem to be currently used across the project :shrug:

dd-java-agent/iast/src/main/java/com/datadog/iast/model/EvidenceType.java
```diff
@@ -0,0 +1,6 @@
+package com.datadog.iast.model;
+
+public enum EvidenceType {
```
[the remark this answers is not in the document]
smola: Nope, I just missed the change. I'll change the PR.

dd-java-agent/iast/src/main/java/com/datadog/iast/IastSystem.java
smola: Thanks. It makes sense. I'm going to defer this change until we have some decision around https://github.com/DataDog/dd-trace-java/pull/3671

dd-java-agent/iast/src/main/java/com/datadog/iast/model/json/AdapterFactory.java
```diff
@@ -0,0 +1,128 @@
...
+import javax.annotation.Nullable;
+
+/**
+ * Stateful adapter for {@link VulnerabilityBatch}. It will fill encode vulnerabilityes, collecting
+ * all references to {@link Source}, and replacing them with indexes. Then it will encode a sources
+ * list at the end, in the same order that they were found within the vulnerabilities.
+ */
+class AdapterFactory implements JsonAdapter.Factory {
+
+  private final ThreadLocal<Context> contextThreadLocal = ThreadLocal.withInitial(Context::new);
```
manuel-alvarez-alvarez, line 28: Why not static? it will allow you to declare all inner classes static as well

dd-java-agent/iast/src/test/java/com/datadog/iast/AnyStackRunner.java
```diff
@@ -0,0 +1,29 @@
...
+
+import static net.bytebuddy.matcher.ElementMatchers.named;
+
+import java.lang.reflect.Method;
+import net.bytebuddy.ByteBuddy;
+import net.bytebuddy.implementation.MethodCall;
+
+public class AnyStackRunner {
+
+  public static void callWithinStack(
```
manuel-alvarez-alvarez, line 11: :magic_wand: , I think @jandro996  used a similar approach for stack calculations

dd-java-agent/iast/src/main/java/com/datadog/iast/IastRequestContext.java
```diff
@@ -0,0 +1,23 @@
+package com.datadog.iast;
+
+import com.datadog.iast.model.Vulnerability;
+import java.util.ArrayList;
+import java.util.List;
+
+public class IastRequestContext {
```
dougqh, line 7: I've been discussing some with ASM.  Long term, I don't think we want to have multiple RequestContext objects floating around.

What I'd like to do is push the notion of a Request/JobContext into the core of the tracer itself, but then provide a facility for the different product components to hang structure objects onto it.

Probably something like...
IastData data = Tracer.initiatingJobContext().get(IastData.class);

dd-java-agent/iast/src/main/java/com/datadog/iast/IastRequestContext.java
```diff
@@ -0,0 +1,23 @@
...
+
+  private List<Vulnerability> vulnerabilities;
```
dougqh, line 9: Access to vulnerabilities needs to be thread safe

dd-java-agent/iast/src/main/java/com/datadog/iast/Reporter.java
```diff
@@ -0,0 +1,61 @@
...
+      return;
+    }
+    ctx.addVulnerability(vulnerability);
+  }
+
+  public static void finalizeReports(final RequestContext<Object> requestContext) {
+    if (requestContext == null) {
+      return;
+    }
+    final IastRequestContext ctx = (IastRequestContext) requestContext.getIastContext();
```
dougqh, line 33: Why do we need cast?  Because this is coming from another component?
That seems less than ideal.

dd-java-agent/iast/src/main/java/com/datadog/iast/model/Evidence.java
```diff
@@ -0,0 +1,38 @@
+package com.datadog.iast.model;
+
+import javax.annotation.Nonnull;
+import javax.annotation.Nullable;
+
+public class Evidence {
+  private final @Nonnull EvidenceType type;
```
dougqh, line 7: Very nice, I like use of `@Nonnull`.

dd-java-agent/iast/src/main/java/com/datadog/iast/model/Evidence.java
```diff
@@ -0,0 +1,38 @@
+package com.datadog.iast.model;
+
+import javax.annotation.Nonnull;
+import javax.annotation.Nullable;
+
+public class Evidence {
```
dougqh, line 6: This class looks like an immutable object.  It should probably be final.

dd-java-agent/iast/src/main/java/com/datadog/iast/model/Location.java
```diff
@@ -0,0 +1,25 @@
+package com.datadog.iast.model;
+
+public class Location {
```
dougqh, line 3: final?

dd-java-agent/iast/src/main/java/com/datadog/iast/model/Source.java
```diff
@@ -0,0 +1,27 @@
+package com.datadog.iast.model;
+
+public class Source {
```
dougqh, line 3: final

dd-java-agent/iast/src/main/java/com/datadog/iast/model/SourceType.java
```diff
@@ -0,0 +1,6 @@
+package com.datadog.iast.model;
+
+public enum SourceType {
```
dougqh, line 3: enums are rather bloated.  If we're using this often, I'd prefer we just use a primitive.

dd-java-agent/iast/src/main/java/com/datadog/iast/model/Vulnerability.java
```diff
@@ -0,0 +1,86 @@
...
+
+  public interface BuilderMore {
+    BuilderMore location(Location location);
+
+    BuilderMore computeLocation();
+
+    Vulnerability build();
+  }
+
+  public static final class Builder implements BuilderNeedsType, BuilderNeedsEvidence, BuilderMore {
```
dougqh, line 54: I'm assuming vulnerabilities are relatively rare, so I think the builder is okay.  But in general, I want to avoid builders because they result in needless allocation.

dd-java-agent/iast/src/main/java/com/datadog/iast/model/json/AdapterFactory.java
```diff
@@ -0,0 +1,128 @@
...
+  public class SourceIndexAdapter extends JsonAdapter<Source> {
+
+    @Override
+    public void toJson(JsonWriter writer, @Nullable @SourceIndex Source value) throws IOException {
+      if (value == null) {
+        return;
+      }
+      final Context ctx = contextThreadLocal.get();
+      final int index =
+          ctx.sourceIndexMap.computeIfAbsent(
```
dougqh, line 66: We should be trying to avoid capturing lambdas.  That's going to result in more allocation.

dd-java-agent/iast/src/main/java/com/datadog/iast/model/json/AdapterFactory.java
```diff
@@ -0,0 +1,128 @@
...
+ * Stateful adapter for {@link VulnerabilityBatch}. It will fill encode vulnerabilityes, collecting
+ * all references to {@link Source}, and replacing them with indexes. Then it will encode a sources
+ * list at the end, in the same order that they were found within the vulnerabilities.
+ */
+class AdapterFactory implements JsonAdapter.Factory {
+
+  private final ThreadLocal<Context> contextThreadLocal = ThreadLocal.withInitial(Context::new);
+
+  private static class Context {
+    private final List<Source> sources;
```
dougqh, line 31: Creation of yet another context class worries me.  That's probably the best thing we can do right now given our inflexible core, but we should devise something better.

dd-java-agent/iast/src/main/java/com/datadog/iast/model/json/AdapterFactory.java
```diff
@@ -0,0 +1,128 @@
...
+        }
+      }
+      return null;
+    } else if (VulnerabilityBatch.class.equals(Types.getRawType(type))) {
+      return new VulnerabilityBatchAdapter(moshi);
+    }
+    return null;
+  }
+
+  public class SourceIndexAdapter extends JsonAdapter<Source> {
```
dougqh, line 57: static

dd-java-agent/iast/src/main/java/com/datadog/iast/model/json/AdapterFactory.java
```diff
@@ -0,0 +1,128 @@
...
+    }
+
+    @Nullable
+    @Override
+    public @SourceIndex Source fromJson(JsonReader reader) throws IOException {
+      throw new UnsupportedOperationException("Source deserialization is not supported");
+    }
+  }
+
+  public class VulnerabilityBatchAdapter extends JsonAdapter<VulnerabilityBatch> {
```
dougqh, line 82: static

dd-java-agent/iast/src/main/java/com/datadog/iast/model/json/AdapterFactory.java
```diff
@@ -0,0 +1,128 @@
...
+
+    @Override
+    public void toJson(JsonWriter writer, @Nullable VulnerabilityBatch value) throws IOException {
+      if (value == null) {
+        writer.nullValue();
+        return;
+      }
+
+      try {
+        final Context ctx = contextThreadLocal.get();
```
dougqh, line 102: Nit, logically the resource acquisition should be outside the try.  Functionally, I realize this makes no difference here.

internal-api/src/main/java/datadog/trace/api/gateway/RequestContextHolder.java
```diff
@@ -0,0 +1,34 @@
+package datadog.trace.api.gateway;
+
+import datadog.trace.api.TraceSegment;
+
+public class RequestContextHolder implements RequestContext<Object> {
```
dougqh, line 5: Having this be RequestContext<Object> seems to defeat the point of using generics in the first place.  If we need to relocate RequestContext to a more shareable location, we should do it.

dd-java-agent/agent-bootstrap/src/main/java/datadog/trace/bootstrap/Agent.java
```diff
@@ -617,6 +631,26 @@ private static void startAppSec(InstrumentationGateway gw, Class<?> scoClass, Ob
...
+
+    InstrumentationGateway gw = AgentTracer.get().instrumentationGateway();
+    startIast(gw, scoClass, o);
+  }
+
+  private static void startIast(InstrumentationGateway gw, Class<?> scoClass, Object sco) {
+    try {
+      final Class<?> iastSysClass = IAST_CLASSLOADER.loadClass("com.datadog.iast.IastSystem");
+      final Method iastInstallerMethod =
+          iastSysClass.getMethod("start", InstrumentationGateway.class);
```
dougqh, line 647: I know this isn't new to this code, but the need to use reflection to call start is to me a code smell.

Our sub-systems should implement a common interface and we should call a "start" method in conventional fashion rather than resorting to reflection.

dd-java-agent/iast/src/main/java/com/datadog/iast/IastRequestContext.java
smola: Makes sense. See https://github.com/DataDog/dd-trace-java/pull/3671 for something that might be more aligned with that.

internal-api/src/main/java/datadog/trace/api/gateway/RequestContextHolder.java
smola: Absolutely, the whole RequestContext stuff is pending a change on this PR once https://github.com/DataDog/dd-trace-java/pull/3671 is ready.

dd-java-agent/iast/src/main/java/com/datadog/iast/model/json/AdapterFactory.java
smola: Note that this one just kicks in during serialization. It is created at the beginning of serialization at the end of a request, and reset at the end of such serialization. We start with a list of vulnerabilities, collected during a request, and serialized at the end of the request. Since the same source objects are often referenced in several times (both within the same vulnerability, and across multiple vulnerabilities) we are serializing them as:

```json
{
  "vulnerabilities": {
      /* ... */
      "sourceIndex": 0
      /* ... */
   },
   "sources": [
     /* ... */
   ]
}
```

I have not found a way to do this with Moshi except by:
* Writing custom `JsonAdapter`s for the whole hierarchy.
* Or, writing a custom `JsonAdapter` for the outer-most object and sources, and using a thread-local to share context between them.

I'm happy to change this to a better approach if anyone has any suggestion. I would be particularly interested in codegen approaches to improve performance.

dd-java-agent/iast/src/main/java/com/datadog/iast/IastRequestContext.java
smola: I have changed this to a `ConcurrentLinkedQueue`. However, the way we accumulate vulnerabilities in a request might get changed significantly once we implement vulnerability de-duplication (working on this in a separate PR next week, tentatively).

dd-java-agent/iast/src/main/java/com/datadog/iast/model/json/AdapterFactory.java
smola: Changed to avoid `computeIfAbsent` altogether.

dd-java-agent/iast/src/main/java/com/datadog/iast/IastRequestContext.java
```diff
@@ -0,0 +1,23 @@
...
+public class IastRequestContext {
+
+  private final ConcurrentLinkedQueue<Vulnerability> vulnerabilities;
+
+  public IastRequestContext() {
+    this.vulnerabilities = new ConcurrentLinkedQueue<>();
+  }
+
+  public List<Vulnerability> getVulnerabilities() {
+    return new ArrayList<>(vulnerabilities);
```
anderruiz, line 17: We may not need this defensive copy now, it was a requirement in Hdiv but perhaps not here

dd-java-agent/iast/src/main/java/com/datadog/iast/Reporter.java
```diff
@@ -0,0 +1,61 @@
...
+    }
+    final RequestContext<Object> reqCtx = span.getRequestContext();
+    if (reqCtx == null) {
+      return null;
+    }
+    final Object data = reqCtx.getIastContext();
+    if (!(data instanceof IastRequestContext)) {
+      return null;
+    }
+    return (IastRequestContext) data;
```
anderruiz, line 59: I see too many operations for this critical path. I hope it gets better with future changes.

anderruiz (approved):

internal-api/src/main/java/datadog/trace/api/gateway/RequestContextHolder.java
smola: Removed this class and used the approach implemented at https://github.com/DataDog/dd-trace-java/pull/3671
If we need any follow up refactor to that, we may do it in a separate PR.

dd-java-agent/iast/src/main/java/com/datadog/iast/IastRequestContext.java
smola: It's currently just a copy to conform to a list, since we currently have a queue, and we collect everything once the request ends.

dd-java-agent/iast/src/main/java/com/datadog/iast/IastRequestContext.java
jandro996: @dougqh has made the following comments in #3694 about this issue

https://github.com/DataDog/dd-trace-java/pull/3694/files/60818cad2585aa15daa697921667d27e92ad45b2#r935976690

https://github.com/DataDog/dd-trace-java/pull/3694/files/60818cad2585aa15daa697921667d27e92ad45b2#r935978248

dd-java-agent/iast/src/main/java/com/datadog/iast/Reporter.java
smola: It's one less step now, but it still might be improved in the future.

dd-java-agent/agent-bootstrap/src/main/java/datadog/trace/bootstrap/Agent.java
smola: I have opened a task in our backlog to take care of this, both for our code, and other products' code. We'll try to do it during next sprint.

dd-java-agent/iast/src/main/java/com/datadog/iast/model/Vulnerability.java
smola: I have removed the builder. In this case, it's not that much better than the plain constructor anyway.

dd-java-agent/iast/src/main/java/com/datadog/iast/model/SourceType.java
smola: I have changed the type for this to a `byte`. We'll have a reference to a source type for each "input" of the application (each used header, query parameter, body parameter, etc, counts as an input). Which it shouldn't be that much for average cases, but some applications can manage to use a larger number.

It's still a string for the serialization format.

dd-java-agent/iast/src/main/java/com/datadog/iast/Reporter.java
smola: Yep. The cast is gone, although there's still effectively a cast in the RequestContext.

jandro996 (approved):

dd-java-agent/iast/src/main/java/com/datadog/iast/Reporter.java
```diff
@@ -0,0 +1,32 @@
...
+/** Reports IAST vulnerabilities. */
+public final class Reporter {
+
+  private Reporter() {}
+
+  public static void report(final AgentSpan span, final Vulnerability vulnerability) {
+    if (span == null) {
+      return;
+    }
+    final RequestContext reqCtx = span.getRequestContext();
```
ValentinZakharov, line 18: Direct method to get specific request context from Span can make code more concise (logical).
So at the end we can have something like:
```
IastRequestContext ctx = span.getRequestContext(RequestContextSlot.IAST);
```
But it's more for future improvements.

ValentinZakharov: It's confusing me to see similar names in `RequestContext` and `IastRequestContext`/`AppSecRequestContext`, however they are different things and belongs to a distinct levels of abstraction.

dd-java-agent/iast/src/main/java/com/datadog/iast/model/SourceType.java
```diff
@@ -0,0 +1,24 @@
...
+  private SourceType() {}
+
+  public static final byte NONE = 0;
+
+  public static final byte REQUEST_QUERY_PARAMETER = 1;
+  private static final String REQUEST_QUERY_PARAMETER_STRING = "http.url_details.queryString";
+  public static final byte REQUEST_PATH = 2;
+  private static final String REQUEST_PATH_STRING = "http.url_details.path";
+
+  public static String toString(final byte sourceType) {
```
ValentinZakharov, line 14: Is there specific reason to not use Enum?

dd-java-agent/iast/src/main/java/com/datadog/iast/model/VulnerabilityBatch.java
```diff
@@ -0,0 +1,40 @@
...
+      if (vulnerabilities == null) {
+        this.vulnerabilities = new ArrayList<>();
+      }
+      vulnerabilities.add(v);
+    }
+  }
+
+  public List<Vulnerability> getVulnerabilities() {
+    return vulnerabilities;
+  }
```
ValentinZakharov, line 25: Method provides `vulnerabilities` to an outside caller, i.e. we can't guarantee synchronized access to the list. May be better to use come `concurrent` implementation of list, instead of `ArrayList`?

dd-java-agent/iast/src/main/java/com/datadog/iast/model/SourceType.java
smola: Runtime overhead. I changed this after a previous review: https://github.com/DataDog/dd-trace-java/pull/3673#discussion_r927704796

dd-java-agent/iast/src/main/java/com/datadog/iast/Reporter.java
smola: > Direct method to get specific request context from Span can make code more concise

In principle, yes, but note that we also need to operate on the `RequestContext` here. See below:

```java
reqCtx.getTraceSegment().setDataTop("iast", batch);
```

> It's confusing me to see similar names [...]

I agree. Although in terms of naming, I'm not sure if we can do anything other than dropping `Request` from the later: `IastRequestContext`/`AppSecRequestContext` to `IastContext`/`AppSecContext`.

dd-java-agent/iast/src/main/java/com/datadog/iast/model/VulnerabilityBatch.java
smola: I have to think about this. I didn't really want to expose `vulnerabilities` outside, since it should only be accessed for serialization, which is already synchronized (see the `toString` method below). I could do one of the following:

* Move everything in the `json` subpackage (where the encoders are) to this package, and make `vulnerabilities` only visible within the `model` package.
* and/or, document `getVulnerabilities` as not thread-safe and requiring external synchronization.

PR #3675, opened 2022-07-23
[EXPERIMENTAL] IAST Callback reloader

anderruiz:
# What Does This Do

Includes the ability to reload InstrumentationBridge replacing it with an empty implementation to short-circuit the calls

# Motivation

Being able to short circuit the IAST calls is interesting, both in case of an error but also to do remote activation

# Additional Notes

**This is an experimental branch. Not meant to be merged.**


internal-api/src/main/java/datadog/trace/api/iast/InstrumentationBridgeShortCircuit.java
```diff
@@ -0,0 +1,18 @@
+package datadog.trace.api.iast;
+
+import org.slf4j.Logger;
+import org.slf4j.LoggerFactory;
+
+public abstract class InstrumentationBridgeShortCircuit {
```
manuel-alvarez-alvarez, line 6: Bonus points if we create this class on the fly

dougqh:
I understand the desire to short-circuit the calls, but I'd like to understand why we've chosen this approach to do it.

I see that the callback only takes Strings.  That's good because otherwise I'd be concerned by the cost of constructing parameters of the calling side -- and we need a different mechanism anyway.

However given what we're trying to do here, I think there are a couple possibly simpler options.
1 - We can short-circuit via boolean switch
2 - We can dynamic dispatch to nop implementation

Those both rely on the JVMs speculative optimization, but I'm curious if we've benchmarked our options before settling on bytecode manipulation.  But relying on redefinition has the same downsides.  Redefinition will result in a stop the world pause and surge of recompilation.  Depending on the specifics that's also true for dynamic dispatch, but can easily be avoided.

The simple solution of boolean still leads to deoptimization, but not of the stop the world variety.

Before introducing such mechanisms into the code base, it would nice if the approach was discussed with a larger audience first.

PR #3678, opened 2022-07-26
Add operation name to the endpoint event

mar-kolya:
# What Does This Do

# Motivation

# Additional Notes


dd-trace-core/jfr-openjdk/src/main/java/datadog/trace/core/jfr/openjdk/EndpointEvent.java
```diff
@@ -19,6 +19,9 @@ public class EndpointEvent extends Event implements EndpointTracker {
   @Label("Endpoint")
   private String endpoint = "unknown";
 
+  @Label("Operation")
+  private String operation = "unknown";
```
dougqh, line 23: I don't think we should have magic values in the data model.  I would argue that null is a better signifier of no value than "unknown".

PR #3690, opened 2022-08-01
Improved dependency resolution

ValentinZakharov:
# What Does This Do
This modification improves dependency resolution strategy with extra cases for telemetry

# Motivation

# Additional Notes


telemetry/src/test/groovy/datadog/telemetry/dependency/DependencyServiceImplSpecification.groovy
```diff
@@ -93,6 +93,33 @@ class DependencyServiceImplSpecification extends DDSpecification {
...
+    knownJarCheck(
+      jarName: jar,
+      name: name,
+      version: version)
+
+    where:
+    jar                  |  name        | version
+    'asm-util-9.2.jar'   | 'asm-util'   | '9.2'
+    'agrona-1.7.2.jar'   | 'agrona'     | '1.7.2'
+    'caffeine-2.8.5.jar' | 'caffeine'   | '2.8.5'
```
devinsba, line 120: How does this behave if the version number is not included in the name of the jar file?

devinsba (approved):
Looks like a positive change to me

telemetry/src/test/groovy/datadog/telemetry/dependency/DependencyServiceImplSpecification.groovy
ValentinZakharov: It tries to find version in `pom` in `manifest` or in filename in accordance to priority. In case if nothing found - then we conclude dependency as non-resolvable and calculate hash of the file to resolve with maven repo on backend side later.

PR #3694, opened 2022-08-02
[IAST] Add overhead controller to add caps on analyzed requests and reported vulnerabilities

jandro996:
# What Does This Do

Implement Overhead Control Engine (OCE) for IAST operations. It creates a module that by design ensures the maximum overhead for IAST operations. The OCE will be responsible for assuring performance and short circuiting the logic if it reaches the maximum.

# Motivation

Low overhead is a mandatory requirement for a software that is running in production. The main constraint is that having to introduce many hooks for taint tracking keeping a low overhead is a challenging task.

The OCE is an element that by design ensures that the overhead does not go over a maximum limit. It will measure operations being executed in a request and it will deactivate detection (and therefore reduce the overhead to nearly 0) if a certain threshold is reached.

# Additional Notes

First approach set threshold (configurable) to: 

- Only about 30% of the requests will be analysed
- 2 concurrent request
- 2 vulnerabilities per request
- ~2 vulnerabilities out of request~ (reporting out of request is not supported yet)
 
Future versions will implement a more fine-tuned policy for different operation types. The current PR implements one of the simplest policies.

Depends on https://github.com/DataDog/system-tests/pull/483

### DIFF
--- a/dd-java-agent/iast/src/main/java/com/datadog/iast/IastModuleImpl.java
+++ b/dd-java-agent/iast/src/main/java/com/datadog/iast/IastModuleImpl.java
@@ -4,8 +4,11 @@ import com.datadog.iast.model.Evidence;
 import com.datadog.iast.model.Location;
 import com.datadog.iast.model.Vulnerability;
 import com.datadog.iast.model.VulnerabilityType;
+import com.datadog.iast.overhead.Operations;
+import com.datadog.iast.overhead.OverheadController;
 import datadog.trace.api.Config;
 import datadog.trace.api.iast.IastModule;
+import datadog.trace.bootstrap.instrumentation.api.AgentSpan;
 import datadog.trace.bootstrap.instrumentation.api.AgentTracer;
 import datadog.trace.util.stacktrace.StackWalker;
 import datadog.trace.util.stacktrace.StackWalkerFactory;
@@ -15,11 +18,14 @@ public final class IastModuleImpl implements IastModule {
 
   private final Config config;
   private final Reporter reporter;
+  private final OverheadController overheadController;
   private final StackWalker stackWalker = StackWalkerFactory.INSTANCE;
 
-  public IastModuleImpl(final Config config, final Reporter reporter) {
+  public IastModuleImpl(
+      final Config config, final Reporter reporter, final OverheadController overheadController) {
     this.config = config;
     this.reporter = reporter;
+    this.overheadController = overheadController;
   }
 
   public void onCipherAlgorithm(String algorithm) {
@@ -28,6 +34,10 @@ public final class IastModuleImpl implements IastModule {
     }
     final String algorithmId = algorithm.toUpperCase(Locale.ROOT);
     if (!config.getIastWeakCipherAlgorithms().contains(algorithmId)) {
+      return;
+    }
+    final AgentSpan span = AgentTracer.activeSpan();
+    if (!overheadController.consumeQuota(Operations.REPORT_VULNERABILITY, span)) {
       return;
     }
     // get StackTraceElement for the callee of MessageDigest
@@ -44,7 +54,7 @@ public final class IastModuleImpl implements IastModule {
             VulnerabilityType.WEAK_CIPHER,
             Location.forStack(stackTraceElement),
             new Evidence(algorithm));
-    reporter.report(AgentTracer.activeSpan(), vulnerability);
+    reporter.report(span, vulnerability);
   }
 
   public void onHashingAlgorithm(String algorithm) {
@@ -53,6 +63,10 @@ public final class IastModuleImpl implements IastModule {
     }
     final String algorithmId = algorithm.toUpperCase(Locale.ROOT);
     if (!config.getIastWeakHashAlgorithms().contains(algorithmId)) {
+      return;
+    }
+    final AgentSpan span = AgentTracer.activeSpan();
+    if (!overheadController.consumeQuota(Operations.REPORT_VULNERABILITY, span)) {
       return;
     }
     // get StackTraceElement for the caller of MessageDigest
@@ -69,6 +83,6 @@ public final class IastModuleImpl implements IastModule {
             VulnerabilityType.WEAK_HASH,
             Location.forStack(stackTraceElement),
             new Evidence(algorithm));
-    reporter.report(AgentTracer.activeSpan(), vulnerability);
+    reporter.report(span, vulnerability);
   }
 }
--- a/dd-java-agent/iast/src/main/java/com/datadog/iast/IastRequestContext.java
+++ b/dd-java-agent/iast/src/main/java/com/datadog/iast/IastRequestContext.java
@@ -1,6 +1,7 @@
 package com.datadog.iast;
 
 import com.datadog.iast.model.VulnerabilityBatch;
+import com.datadog.iast.overhead.OverheadContext;
 import java.util.concurrent.atomic.AtomicBoolean;
 
 public class IastRequestContext {
@@ -9,9 +10,12 @@ public class IastRequestContext {
 
   private final AtomicBoolean spanDataIsSet;
 
+  private final OverheadContext overheadContext;
+
   public IastRequestContext() {
     this.vulnerabilityBatch = new VulnerabilityBatch();
     this.spanDataIsSet = new AtomicBoolean(false);
+    this.overheadContext = new OverheadContext();
   }
 
   public VulnerabilityBatch getVulnerabilityBatch() {
@@ -20,5 +24,9 @@ public class IastRequestContext {
 
   public boolean getAndSetSpanDataIsSet() {
     return spanDataIsSet.getAndSet(true);
+  }
+
+  public OverheadContext getOverheadContext() {
+    return overheadContext;
   }
 }
--- a/dd-java-agent/iast/src/main/java/com/datadog/iast/overhead/OverheadContext.java
+++ b/dd-java-agent/iast/src/main/java/com/datadog/iast/overhead/OverheadContext.java
@@ -1,0 +1,28 @@
+package com.datadog.iast.overhead;
+
+import datadog.trace.api.Config;
+import java.util.concurrent.atomic.AtomicInteger;
+
+public class OverheadContext {
+
+  private static final int MAX_OPERATIONS = Config.get().getIastMaxContextOperations();
+
+  private final AtomicInteger availableOperations = new AtomicInteger(MAX_OPERATIONS);
+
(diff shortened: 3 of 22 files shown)


dd-java-agent/iast/src/main/java/com/datadog/iast/IastRequestContext.java
```diff
@@ -9,8 +13,11 @@ public class IastRequestContext {
 
   private final ConcurrentLinkedQueue<Vulnerability> vulnerabilities;
 
+  private final OverheadController overheadController;
+
   public IastRequestContext() {
-    this.vulnerabilities = new ConcurrentLinkedQueue<>();
+    vulnerabilities = new ConcurrentLinkedQueue<>();
```
dougqh, line 19: We're not allowed to have any unbounded data structures.

dougqh: You are probably also better off with a a synchronized list than a concurrent linked queue for performance reasons.  The list needs to be thread safe, but if it is primarily accessed from a single thread a synchronized ArrayList will be better.

dd-java-agent/iast/src/main/java/com/datadog/iast/IastRequestContext.java
```diff
@@ -20,4 +27,19 @@ public List<Vulnerability> getVulnerabilities() {
...
+
+  public OverheadController getOverheadController() {
+    return overheadController;
+  }
+
+  public static IastRequestContext getRequestContext(final AgentSpan span) {
+    if (span == null) {
+      return null;
+    }
+    final RequestContext reqCtx = span.getRequestContext();
```
dougqh, line 39: Yes, this makes sense.  Long term, I'd like to flip this so that every span has an owning request context, but this is fine for now.

dd-java-agent/iast/src/main/java/com/datadog/iast/oce/OverheadController.java
```diff
@@ -0,0 +1,44 @@
...
+import datadog.trace.bootstrap.instrumentation.api.AgentSpan;
+import java.util.concurrent.atomic.AtomicInteger;
+
+public class OverheadController {
+
+  protected static final int MAX_CONCURRENT = 2;
+
+  protected static AtomicInteger concurrentRequestTokens = new AtomicInteger(MAX_CONCURRENT);
+
+  protected boolean requestAnalyzed = false;
```
dougqh, line 13: This needs to be thread safe.

dd-java-agent/iast/src/main/java/com/datadog/iast/oce/OverheadController.java
```diff
@@ -0,0 +1,44 @@
...
+
+  protected AtomicInteger concurrentVulnerabilityTokens = new AtomicInteger(MAX_CONCURRENT);
+
+  public static Quota hasQuota(final LongRunningOperation operation, final AgentSpan span) {
```
dougqh, line 17: This method has an odd name.  It sounds like it would return a boolean but its unconditionally constructing a Quota.

dd-java-agent/iast/src/main/java/com/datadog/iast/oce/OverheadController.java
```diff
@@ -0,0 +1,44 @@
...
+    return new Quota() {
```
dougqh, line 18: How many operations do we expect?  I don't want a lot of spurious object allocation.

dougqh:
I'm seeing a number of thread safety and overhead issues in these pull requests.  I think it would be good to discuss the threading requirements of objects in the tracer.

dd-java-agent/iast/src/main/java/com/datadog/iast/IastRequestContext.java
jandro996: > We're not allowed to have any unbounded data structures. In practice, we don't check bound everything, but we need to understand how big this collection can be.

Thanks for the advice! I think no one in the IAST Java team was aware of this 

This implementation came from another branch PR https://github.com/DataDog/dd-trace-java/pull/3673. I see that there is an open discussion about this issue https://github.com/DataDog/dd-trace-java/pull/3673#discussion_r930940869

@smola, I've added a link to this conversation in you PR

dd-java-agent/iast/src/main/java/com/datadog/iast/oce/OverheadController.java
jandro996: We spect one per request... as you say this is not well designed, we are changing it considering your advice

dd-java-agent/iast/src/main/java/com/datadog/iast/IastRequestContext.java
smola: @dougqh Sure. There's two key pieces that I omitted from https://github.com/DataDog/dd-trace-java/pull/3673, but will come up in the following PRs:

* Limit the vulnerabilities reported per request. While we haven't decided all the details about caps, we'll likely start with something like 2 or 3 vulnerabilities per request. Depending on how comfortable we're with performance overhead, we'll increase it, but it's likely to always remain low (e.g. less than 10 or 100).
* We are adding deduplication of vulnerabilities, so we'll hold at most one vulnerability that is considered a duplicate (e.g. same vulnerability type, in the same class, and same line). So it will probably make sense to change the structure here to accommodate that.

We're also considering limiting the maximum number of requests that we analyze concurrently. In that case, we might afford to do something like having a fixed pool of contexts with initialized structures (e.g. plain arrays).

dd-java-agent/iast/src/main/java/com/datadog/iast/IastModuleImpl.java
```diff
@@ -30,6 +36,10 @@ public void onCipherAlgorithm(String algorithm) {
     if (!config.getIastWeakCipherAlgorithms().contains(algorithmId)) {
```
manuel-alvarez-alvarez, line 36: Conceptually shouldn't this be executed after the overhead controller?

dd-java-agent/iast/src/main/java/com/datadog/iast/overhead/OverheadContext.java
```diff
@@ -0,0 +1,28 @@
...
+
+  private final AtomicInteger availableOperations = new AtomicInteger(MAX_OPERATIONS);
+
+  public int getAvailableQuota() {
+    return availableOperations.get();
+  }
+
+  public boolean consumeQuota(final int delta) {
+    final int availableAfter = availableOperations.addAndGet(-delta);
+    if (availableAfter < 0) {
```
manuel-alvarez-alvarez, line 18: There's a race condition here if `delta > 1`, I don't think it's that important but ...

dd-java-agent/iast/src/main/java/com/datadog/iast/overhead/OverheadController.java
```diff
@@ -0,0 +1,96 @@
...
+      return false;
+    }
+    if (availableRequests.get() <= 0) {
+      return false;
+    }
+    final int availableAfterAcquire = availableRequests.decrementAndGet();
+    if (availableAfterAcquire < 0) {
+      availableRequests.incrementAndGet();
+      return false;
+    }
```
manuel-alvarez-alvarez, line 44: Maybe we should do:

```suggestion
    if (availableRequests.getAndUpdate(value -> value <= 0 ? 0 : value - 1) <= 0) {
      return false;
    }
```

dd-java-agent/iast/src/main/java/com/datadog/iast/overhead/OverheadContext.java
```diff
@@ -0,0 +1,28 @@
...
+      availableOperations.addAndGet(delta);
+      return false;
+    }
```
manuel-alvarez-alvarez, line 21: Will this work?

```suggestion
    final int available = availableOperations.getAndUpdate(value -> value < delta ? value : value - delta);
    if (available < delta) {
      return false;
    }
```

manuel-alvarez-alvarez (changes requested):

dd-java-agent/iast/src/test/groovy/com/datadog/iast/RequestEndedHandlerTest.groovy
```diff
@@ -0,0 +1,57 @@
...
+    def flow = handler.apply(reqCtx, spanInfo)
+
+    then:
+    flow.getAction() == Flow.Action.Noop.INSTANCE
+    flow.getResult() == null
+    1 * reqCtx.getData(RequestContextSlot.IAST) >> iastCtx
+    1 * reqCtx.getTraceSegment() >> traceSegment
+    1 * traceSegment.setTagTop("_dd.iast.enabled", 1)
+    1 * overheadController.releaseRequest()
+    0 * _
```
manuel-alvarez-alvarez, line 35: Nice one! I will copy this for my PRs moving forward :smile:

dd-java-agent/iast/src/main/java/com/datadog/iast/IastModuleImpl.java
smola: I think we shouldn't consume quota for non-weak algorithms? I could check `hasQuota`, but that would require also getting the active span, increasing the overhead for the happy path of non-weak algorithms...

dd-java-agent/iast/src/main/java/com/datadog/iast/overhead/OverheadController.java
smola: I have applied similar logic to all operations now. Performance is similar in the absence of concurrency. I'm sure we'll revisit this again once we do more benchmarks and profiling including concurrency.

dd-java-agent/iast/src/main/java/com/datadog/iast/IastModuleImpl.java
manuel-alvarez-alvarez: To my mind the reasoning regarding the quota in terms of vulnerabilities is the following:

Is there quota to check for vulnerability X?
* Yes: Check for the vulnerability and report it if found
* NO: Continue

In this particular case checking  the vulnerability is  much faster than the quota system, but others might not be. For instance, in a SQL injection you will surely first check for the quota and then continue.

Imagine we have an analyzer system, will you keep the quota check before or after calling the analyzer to check for the weak algorith?

manuel-alvarez-alvarez (approved):

smola:
@dougqh Is there any remaining thread safety concern for the latest code?

PR #3695, opened 2022-08-02
Propagate datastreams pathway in Kafka Streams instrumentation

hokitam:
# What Does This Do
This PR adds datastreams pathway propagation to Kafka Streams instrumentation.

`KafkaStreamTaskInstrumentation` was modified to store a `StreamTaskContext` class instead of only the `AgentScope`, so that we can also store the application ID of the kafka streams application. The application ID is needed to properly tag the pathway context. It is fetched from the input StreamsConfig argument from the StreamTask's constructor.


dd-java-agent/instrumentation/kafka-streams-0.11/src/main/java/datadog/trace/instrumentation/kafka_streams/KafkaStreamTaskInstrumentation.java
```diff
@@ -53,17 +59,41 @@ public String[] helperClassNames() {
...
       packageName + ".StampedRecordContextVisitor",
+      packageName + ".StreamTaskContext",
     };
   }
 
   @Override
   public Map<String, String> contextStore() {
-    return singletonMap(
-        "org.apache.kafka.streams.processor.internals.StreamTask", AgentScope.class.getName());
+    Map<String, String> contextStores = new HashMap<>();
```
dougqh, line 68: Why was this changed from a singleton map?  This is adding overhead (albeit small) unnecessarily.

hokitam: thanks for catching that! I was originally exploring adding 2 context stores here but later removed the second call. Didn't realize there would be an overhead increased. Fixed!

dd-java-agent/instrumentation/kafka-streams-0.11/src/main/java/datadog/trace/instrumentation/kafka_streams/KafkaStreamTaskInstrumentation.java
```diff
@@ -53,17 +59,41 @@ public String[] helperClassNames() {
...
+    contextStores.put(
+        "org.apache.kafka.streams.processor.internals.StreamTask",
+        StreamTaskContext.class.getName());
+
+    return contextStores;
   }
 
   @Override
   public void adviceTransformations(AdviceTransformation transformation) {
+    // StreamsConfig was the 4th input argument to StreamTask's constructor in kafka versions 2.6 to
```
dougqh, line 75: To the best of my knowledge, we typically separate the different versions into separate instrumentations.  However, I'm not the most knowledgeable about this.  I suggested asking the larger Java tracing team for clarification.

dd-java-agent/instrumentation/kafka-streams-0.11/src/main/java/datadog/trace/instrumentation/kafka_streams/KafkaStreamTaskInstrumentation.java
```diff
@@ -191,6 +297,21 @@ public static void start(
           // The queueSpan will be finished after inner span has been activated to ensure that
           // spans are written out together by TraceStructureWriter when running in strict mode
         }
+
+        PathwayContext pathwayContext = propagate().extractBinaryPathwayContext(record, PR_GETTER);
+        span.mergePathwayContext(pathwayContext);
+        List<String> edgeTags = new ArrayList<>();
```
dougqh, line 300: The list allocation could be more efficient.

First, the ArrayList could be sized to the maximum number of tags -- e.g. 3.
But my bigger concern is that it looks we're creating an array on every message and we're doing some costly string concatenation.

Typically for relatively static values we use a cache to avoid the redundant work and allocation -- like topic & group in this case.

dd-java-agent/instrumentation/kafka-streams-0.11/src/main/java/datadog/trace/instrumentation/kafka_streams/KafkaStreamTaskInstrumentation.java
```diff
@@ -157,13 +257,17 @@ public static void start(
         queueSpan.finish();
       }
 
-      InstrumentationContext.get(StreamTask.class, AgentScope.class).put(task, agentScope);
+      if (streamTaskContext == null) {
```
dougqh, line 257: Hmm, it is a bit unfortunate that we need to create a StreamTaskContext rather than hanging the data onto an existing span / scope.  Unfortunately, I guess that's what the API necessitates today.  For now, I guess this is fine.

dd-java-agent/instrumentation/kafka-streams-0.11/src/main/java/datadog/trace/instrumentation/kafka_streams/KafkaStreamTaskInstrumentation.java
```diff
@@ -191,6 +297,21 @@ public static void start(
...
+        edgeTags.add("topic:" + record.topic());
+
+        if (streamTaskContext != null) {
+          String applicationId = streamTaskContext.getApplicationId();
+          if (applicationId != null) {
+            edgeTags.add("group:" + applicationId);
+          }
+        }
+        // Kafka Streams uses the application ID as the consumer group.id.
+        AgentTracer.get().setDataStreamCheckpoint(span, edgeTags);
```
dougqh, line 311: Long term, I think we should more deeply integrate checkpoint support into the context system.  But given that a good context system doesn't exist yet, this is fine as is.

dougqh:
While I think there are some problems because of our poor core APIs, overall, I think this looks good.

I would like to see the performance issues with the tag handling addressed.

dd-java-agent/instrumentation/kafka-streams-0.11/src/main/java/datadog/trace/instrumentation/kafka_streams/KafkaStreamTaskInstrumentation.java
hokitam: Started a thread about this on the Java tracing team's slack channel! I kept it this way for now since we have separate instrumentation for pre-2.7 and post-2.7 in the current instrumentation (i.e. StartSpanAdvice vs StartSpanAdvice27) too, but open to break it out to separate instrumentations if that's better.

dd-java-agent/instrumentation/kafka-streams-0.11/src/main/java/datadog/trace/instrumentation/kafka_streams/KafkaStreamTaskInstrumentation.java
hokitam: I created a new TagCache class that constructs separate caches for topic and group, and also initiated the arraylist with capacity 3 as you suggested. PTAL!

hokitam:
Thank you for the thorough review Doug! I have made changes to the tag handling logic, let me know if that aligns with your suggestions.

dd-java-agent/instrumentation/kafka-streams-0.11/src/main/java/datadog/trace/instrumentation/kafka_streams/KafkaStreamTaskInstrumentation.java
hokitam: acknowledged. ty!

dd-java-agent/instrumentation/kafka-streams-0.11/src/main/java/datadog/trace/instrumentation/kafka_streams/KafkaStreamTaskInstrumentation.java
hokitam: acknowledged. ty!

dd-java-agent/instrumentation/kafka-streams-0.11/src/main/java/datadog/trace/instrumentation/kafka_streams/KafkaStreamTaskInstrumentation.java
hokitam: Discussion with the team suggests that we typically only change instrumentation version if there's a large refactoring change between versions and new instrumentation is needed, otherwise there's no real firm rules. Some instrumentations use advice version and others use muzzle to prevent versions loading from the same time.

dd-java-agent/instrumentation/kafka-streams-0.11/src/main/java/datadog/trace/instrumentation/kafka_streams/KafkaStreamTaskInstrumentation.java
hokitam: Discussion with the team suggests that we typically only change instrumentation version if there's a large refactoring change between versions and new instrumentation is needed, otherwise there's no real firm rules. Some instrumentations use advice version and others use muzzle to prevent versions loading from the same time.

dd-java-agent/instrumentation/kafka-streams-0.11/src/main/java/datadog/trace/instrumentation/kafka_streams/KafkaStreamTaskInstrumentation.java
```diff
@@ -191,6 +300,21 @@ public static void start(
           // The queueSpan will be finished after inner span has been activated to ensure that
           // spans are written out together by TraceStructureWriter when running in strict mode
         }
+
+        PathwayContext pathwayContext = propagate().extractBinaryPathwayContext(record, PR_GETTER);
+        span.mergePathwayContext(pathwayContext);
+        List<String> edgeTags = new ArrayList<>(3);
```
dougqh, line 306: I think long term we'll probably need a lower overhead API.  But for now, I think this is fine especially given that's its behind a configuration flag.

dougqh (approved):

dd-java-agent/instrumentation/kafka-streams-0.11/src/main/java/datadog/trace/instrumentation/kafka_streams/KafkaStreamTaskInstrumentation.java
hokitam: ack, thank you for the review! In terms of overhead is the biggest concern the tags, or something else?

PR #3700, opened 2022-08-04
[alpha] Improve lib-injection testing

devinsba:
# What Does This Do

- Adds the test application to the repo and publishes it to the Github package repo as part of the lib-injection testing workflow. The application is a spring-webflux app written with functional bean definitions and running with Java 11 to minimize startup time. Spring was used because the spring-boot docker image builder uses buildpacks so seemed more robust for testing than just a simple hand made image
- A single entrypoint for all the tasks that can be run
- A test matrix is used for the 4 variants (manually configured pod/datadog operator)+(tcp/uds)

# Additional Notes

Still waiting for the first admission controller release in order to remove all snapshot deps `datadog/cluster-agent-dev:master`. Will do so in a followup PR when that happens


lib-injection/src/test/resources/end-to-end/helm-values.yaml
```diff
@@ -0,0 +1,23 @@
+targetSystem: "linux"
+clusterAgent:
+  enabled: true
+  image:
+    name: ""
+    tag: ahmed-auto-instru
```
ahmed-mez, line 6:
```suggestion
    tag: master
```
The code is merged, you don't have to pull my custom branch.

lib-injection/src/test/resources/end-to-end/helm-values.yaml
```diff
@@ -0,0 +1,23 @@
...
+    repository: datadog/cluster-agent-dev
```
ahmed-mez, line 7: We can use the official repo `datadog/cluster-agent` once we release cluster agent v1.23 (in ~2 months)

lib-injection/src/test/resources/end-to-end/helm-values.yaml
```diff
@@ -0,0 +1,23 @@
...
+  enabled: false
+datadog:
+  clusterName: local
+  tags: []
+  # datadog.kubelet.tlsVerify should be `false` on kind and minikube
+  # to establish communication with the kubelet
+  kubelet:
+    tlsVerify: "false"
+  logs:
+    enabled: true
```
ahmed-mez, line 21: You won't have logs collection because the node agent is disabled.

lib-injection/src/test/resources/end-to-end/helm-values.yaml
devinsba: Changed in the template in `functions.sh`

lib-injection/src/test/resources/end-to-end/helm-values.yaml
devinsba: Removed in the template in `functions.sh`

mcculls (approved):
LGTM

ahmed-mez (commented):
Nit: To avoid any confusion I think it's best to not use the term `operator`.
There is already a project called [datadog-operator](https://github.com/DataDog/datadog-operator) and it has nothing to do with auto instrumentation (at least at the moment). Alternatively, we can use the term `admission-controller` or `datadog-cluster-agent`.

ahmed-mez (approved):
🙇

PR #3701, opened 2022-08-04
Implement blocking servlet/tomcat/jetty

cataphract:
Implements blocking behavior for servlet/tomcat/jetty. Blocking means a fixed error response is send and request handling is skipped.


dd-java-agent/instrumentation/tomcat-5.5/src/main/java/datadog/trace/instrumentation/tomcat/TomcatBlockingHelper.java
```diff
@@ -0,0 +1,59 @@
...
+  private static final Logger log = LoggerFactory.getLogger(TomcatBlockingHelper.class);
+  private static final int STATUS_CODE = 403;
+  private static final String RESPONSE_TEXT = "Access denied (request blocked)";
+  private static final byte[] RESPONSE_BODY = RESPONSE_TEXT.getBytes(StandardCharsets.US_ASCII);
+  private static final MethodHandle GET_OUTPUT_STREAM;
+
+  static {
+    MethodHandle mh = null;
+    try {
+      Method getOutputStream = Response.class.getMethod("getOutputStream");
```
dougqh, line 22: Why do the lookup via reflection to then use a method handle?
Why not just use MethodHandles.lookup directly?

dd-java-agent/instrumentation/tomcat-5.5/src/main/java/datadog/trace/instrumentation/tomcat/TomcatBlockingHelper.java
```diff
@@ -0,0 +1,59 @@
...
+
+  public static void commitBlockingResponse(Response resp) {
+    if (!start(resp, STATUS_CODE)) {
+      return;
+    }
+
+    resp.setHeader("Content-length", Integer.toString(RESPONSE_BODY.length));
+    resp.setHeader("Content-type", "text/plain");
+    try {
+      OutputStream os = (OutputStream) GET_OUTPUT_STREAM.invoke(resp);
```
dougqh, line 38: Why do we need reflection to call Response.getOutputStream?

dd-java-agent/instrumentation/tomcat-5.5/src/main/java/datadog/trace/instrumentation/tomcat/TomcatBlockingHelper.java
cataphract: No reason at all, just inartful on my part.

dd-java-agent/instrumentation/tomcat-5.5/src/main/java/datadog/trace/instrumentation/tomcat/TomcatBlockingHelper.java
cataphract: The return value changes between `OutputStream` and `ServletOutputStream (extends OutputStream)`, so the method reference changes.

dd-java-agent/instrumentation/tomcat-5.5/src/main/java/datadog/trace/instrumentation/tomcat/TomcatBlockingHelper.java
cataphract: Ah, actually the reason is the same I gave in the other comment. When using `lookup` I need to specify the return type, whereas with reflection I don't.

cataphract:
Can only be merged after changes in the WAF/default AppSec configuration. Right now, the system tests cannot pass because the WAF returns a blocking instruction and the system tests expect no blocking.

ValentinZakharov:
I can see usage of `span.markForBlocking()` in case of:
- reading/parsing URL
- reading http headers
- reading socket address

but I can't see such blocking ability for:
- http response (we have few rules that check `server.response.status`)
- parsing request body 

Was it skipped intentionally?

dd-java-agent/instrumentation/jetty-common/src/main/java/datadog/trace/instrumentation/jetty9/HandleVisitor.java
```diff
@@ -0,0 +1,285 @@
...
+ * under a condition.
+ */
+public class HandleVisitor extends MethodVisitor {
+  private static final Logger log = LoggerFactory.getLogger(HandleVisitor.class);
+
+  private boolean lookForStore;
+  private int agentSpanVar = -1;
+  private boolean success;
+  private final String methodName;
+  private boolean foundFirstDispatch;
```
ValentinZakharov, line 27: It seems `foundFirstDispatch` is unused

ValentinZakharov:
I can see usage of `IGSpanInfo.isToBeBlocked()` as a blocking indicator for Servlet/Tomcat/Undertow, but not for Jetty. I don't really understand how is Jetty aware about blocking?

dd-java-agent/instrumentation/servlet-common/src/main/java/datadog/trace/instrumentation/servlet/ServletBlockingHelper.java
```diff
@@ -0,0 +1,45 @@
...
+import java.nio.charset.StandardCharsets;
+import javax.servlet.http.HttpServletResponse;
+import org.slf4j.Logger;
+import org.slf4j.LoggerFactory;
+
+public class ServletBlockingHelper {
+  private static final Logger log = LoggerFactory.getLogger(ServletBlockingHelper.class);
+  private static final int STATUS_CODE = 403;
+  private static final String RESPONSE_TEXT = "Access denied (request blocked)";
+  private static final byte[] RESPONSE_BODY = RESPONSE_TEXT.getBytes(StandardCharsets.US_ASCII);
```
ValentinZakharov, line 14: Blocking info (body, status etc.) are currently duplicated for each blocking instrumentation. Did you consider any general approach to have blocking data in same place?
May be some decorator with defined blocking logic can fit our requirements?

ValentinZakharov (approved):

cataphract:
> I can see usage of `span.markForBlocking()` in case of:
> 
> * reading/parsing URL
> * reading http headers
> * reading socket address
> 
> but I can't see such blocking ability for:
> 
> * http response (we have few rules that check `server.response.status`)
> * parsing request body
> 
> Was it skipped intentionally?

Yes, it was skipped intentionally. IIRC, the plan is to first only block on ip addresses, and then only on users.

cataphract:
> I can see usage of `IGSpanInfo.isToBeBlocked()` as a blocking indicator for Servlet/Tomcat/Undertow, but not for Jetty. I don't really understand how is Jetty aware about blocking?

It is used for Jetty. See https://github.com/DataDog/dd-trace-java/pull/3701/files#diff-8d4f557f90cc211125aed1eaf1c1232fec4a1b9be60e96959e77098f08add2a9R76-R81

dd-java-agent/agent-bootstrap/src/main/resources/datadog/trace/bootstrap/blocking/template.json
```diff
@@ -0,0 +1 @@
+{"errors": [{"title": "Access denied", "detail": "Your request has been blocked and was not executed."}]}
```
nizox, line 1:
```suggestion
{"errors": [{"title": "You've been blocked", "detail": "Sorry, you cannot access this page. Please contact the customer service team. Security provided by Datadog."}]}
```

dd-java-agent/agent-bootstrap/src/main/resources/datadog/trace/bootstrap/blocking/template.html
```diff
@@ -0,0 +1,99 @@
...
+
+    footer p {
+      font-size: 16px
+    }
+  </style>
+</head>
+
+<body>
+<main>
+  <p>Sorry, you cannot access this page. Please contact our custom support team.</p>
```
nizox, line 90:
```suggestion
  <p>Sorry, you cannot access this page. Please contact the customer service team.</p>
```

smola (approved):
We discussed possible ways to simplify this instrumentation using our new call site instrumentation framework. However, it will not be possible. We'll need:

* Supporting inserting a call site advice only in specific classes (we expect to do this at some point).
* Adding some further facilities to call private methods without reflection.

So, it's not happening (yet).

PR #3703, opened 2022-08-05
ygree/graphql-java

ygree:
# What Does This Do

GraphQL-Java instrumentation

image
image

# Motivation

Provide GraphQL-Java support out of the box

# Additional Notes

Feature request: https://datadoghq.atlassian.net/browse/FRAPMS-2235

- [x] add error handling
- [x] replace possibly sensitive query values with `?` except for `null` or boolean values


dd-java-agent/instrumentation/graphql-java-14.0/src/main/java8/datadog/trace/instrumentation/graphqljava/GraphQLQuerySanitizer.java
```diff
@@ -0,0 +1,47 @@
...
+import graphql.language.NullValue;
+import graphql.language.Value;
+import graphql.language.VariableReference;
+import graphql.util.TraversalControl;
+import graphql.util.TraverserContext;
+import graphql.util.TreeTransformerUtil;
+
+public final class GraphQLQuerySanitizer extends NodeVisitorStub {
+  private static final NodeVisitor ReplaceValuesVisitor = new GraphQLQuerySanitizer();
+  private static final AstTransformer astTransformer = new AstTransformer();
```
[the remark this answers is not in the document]
ygree: Ops, excuse my Scala habit for not using all capitals for static finals :) Fixed now.

dd-java-agent/instrumentation/graphql-java-14.0/graphql-java-14.0.gradle
```diff
@@ -0,0 +1,31 @@
+// Set properties before any plugins get loaded
+ext {
+  minJavaVersionForTests = JavaVersion.VERSION_1_8
+}
+
+muzzle {
+  pass {
+    group = "com.graphql-java"
+    module = 'graphql-java'
+    versions = "[14.0,19.0)"
```
bantonsson, line 10: So stopping muzzle at `19.0` does not seem right, so I had to dig into this. There are _weirdly_ versioned artifacts using dates, starting from `2018-...` so the muzzle range should be `'[14.0,2018)'`. You also would need to do some fail matching, so the whole muzzle block would be:
```
muzzle {
  pass {
    group = "com.graphql-java"
    module = 'graphql-java'
    versions = '[14.0,2018)' // exclude dates as version numbers
    // earlier versions are missing some classes and will require separate instrumentation
  }
  fail {
    group = "com.graphql-java"
    module = 'graphql-java'
    versions = '[1.2,14.0)'
  }
}
```

dd-java-agent/instrumentation/graphql-java-14.0/graphql-java-14.0.gradle
```diff
@@ -0,0 +1,31 @@
...
+
+testSets {
+  latestDepTest {
+    dirName = 'test'
+  }
+}
+
+dependencies {
+  main_java8CompileOnly group: 'com.graphql-java', name: 'graphql-java', version: '18.1'
+  testImplementation group: 'com.graphql-java', name: 'graphql-java', version: '18.1'
```
bantonsson, line 26: Why not use `14.0` since that is the lowest version that will work according to the muzzle match?

dd-java-agent/instrumentation/graphql-java-14.0/graphql-java-14.0.gradle
```diff
@@ -0,0 +1,31 @@
...
+
+  testImplementation project(':dd-java-agent:instrumentation:trace-annotation')
+
+  latestDepTestImplementation group: 'com.graphql-java', name: 'graphql-java', version: '+'
```
bantonsson, line 30: In the same vein of weird version numbers, this does not test what you think it does, since it pulls in the highest version number ever that just happens to be `230521-nf-execution`. To get the latest but not anything weird, change `version` to a range like this `'[19+,2018)'`. That pulls in `19.1`, which passes muzzle but fails the `latestDepTest` since the error message seems to have changed.

A simple way to check what is actually used is to run the `dependencies` task like this:
```
./gradlew :dd-java-agent:instrumentation:graphql-java-14.0:dependencies --configuration latestDepTestRuntimeClasspath
```

bantonsson (changes requested):

dd-java-agent/instrumentation/graphql-java-14.0/graphql-java-14.0.gradle
ygree: Good point! I'll fix it accordingly. Thanks for digging into this!

dd-java-agent/instrumentation/graphql-java-14.0/graphql-java-14.0.gradle
ygree: Good point! I'll fix it. Thanks for explanation!

bantonsson (approved):

PR #3706, opened 2022-08-09
[IAST] Call Site Instrumentation support

manuel-alvarez-alvarez:
# What Does This Do
This PR contains the basics to do Call Site Instrumentation in the tracer. It defines a new type of instrumenter `datadog.trace.agent.tooling.bytebuddy.csi.CallSiteInstrumenter` that is able to register instances of `datadog.trace.agent.tooling.csi.CallSiteAdvice` to perform the instrumentation via ASM.

# Motivation
Instrumentations related to IAST often affect core parts of the JDK (String, StringBuilder...) where using callee instrumentation (by default in ByteBuddy) is not good enough for performance. Call site instrumentation focuses on the calls to the instrumented methods enabling the use of inclusion/exclusion lists to fine tune where to instrument.

# Additional Notes
Following this PR comes [[IAST] Call Site Instrumentation plugin #3729](https://github.com/DataDog/dd-trace-java/pull/3729) where advices can be autogenerated by a gradle plugin. It also showcases an example on  how to use CSI in the tracer.

# Benchmark results
The main goal of this results are to validate that call site instrumentation is feasible in terms of startup time of an application:
| Benchmark  | Mode | Cnt | Score | Error | Units |
| ------------- | ------------- |------------- |------------- |------------- |------------- |
|CallSiteBenchmark.callSite   | ss  | 10 | 1572.961 | ± 36.189 | ms/op |
|CallSiteBenchmark.callee     | ss  | 10 | 1551.316 | ± 31.888 | ms/op |
|CallSiteBenchmark.none      |  ss  | 10 | 1337.549 | ± 52.882 | ms/op |

### DIFF
--- a/dd-java-agent/agent-tooling/src/main/java/datadog/trace/agent/tooling/bytebuddy/csi/Advices.java
+++ b/dd-java-agent/agent-tooling/src/main/java/datadog/trace/agent/tooling/bytebuddy/csi/Advices.java
@@ -178,8 +178,8 @@ public class Advices {
     }
 
     /**
-     * This class will try to parse the constant pool of the class in order to discover the required
-     * advices to parse it.
+     * This class will try to parse the constant pool of the class in order to discover if any
+     * configured advices should be applied.
      */
     class ConstantPoolInstrospector implements AdviceIntrospector {
 
--- a/dd-java-agent/agent-tooling/src/main/java/datadog/trace/agent/tooling/bytebuddy/csi/Advices.java
+++ b/dd-java-agent/agent-tooling/src/main/java/datadog/trace/agent/tooling/bytebuddy/csi/Advices.java
@@ -1,0 +1,274 @@
+package datadog.trace.agent.tooling.bytebuddy.csi;
+
+import datadog.trace.agent.tooling.bytebuddy.ClassFileLocators;
+import datadog.trace.agent.tooling.csi.CallSiteAdvice;
+import datadog.trace.agent.tooling.csi.Pointcut;
+import java.util.Arrays;
+import java.util.Collections;
+import java.util.HashMap;
+import java.util.HashSet;
+import java.util.Map;
+import java.util.Set;
+import javax.annotation.Nonnull;
+import net.bytebuddy.description.type.TypeDescription;
+import net.bytebuddy.dynamic.ClassFileLocator;
+import net.bytebuddy.jar.asm.ClassReader;
+import net.bytebuddy.utility.OpenedClassReader;
+import org.slf4j.Logger;
+import org.slf4j.LoggerFactory;
+
+/**
+ * Class holding a collection of {@link CallSiteAdvice} indexed by the declaring type, method name
+ * and method description, it stores the collection of helper classes required by the advices.
+ */
+public class Advices {
+
+  private static final Logger LOG = LoggerFactory.getLogger(Advices.class);
+
+  public static final Advices EMPTY =
+      new Advices(
+          Collections.<String, Map<String, Map<String, CallSiteAdvice>>>emptyMap(),
+          new String[0],
+          AdviceIntrospector.NoOpAdviceInstrospector.INSTANCE) {
+        @Override
+        public boolean isEmpty() {
+          return true;
+        }
+      };
+
+  private final Map<String, Map<String, Map<String, CallSiteAdvice>>> advices;
+
+  private final String[] helpers;
+
+  private final AdviceIntrospector introspector;
+
+  private Advices(
+      final Map<String, Map<String, Map<String, CallSiteAdvice>>> advices,
+      final String[] helpers,
+      final AdviceIntrospector introspector) {
+    this.advices = Collections.unmodifiableMap(advices);
+    this.helpers = helpers;
+    this.introspector = introspector;
+  }
+
+  public static Advices fromCallSites(@Nonnull final CallSiteAdvice... advices) {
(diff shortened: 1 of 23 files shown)


jandro996 (approved):

buildSrc/src/main/java/datadog/trace/plugin/csi/impl/AsmSpecificationBuilder.java
```diff
@@ -0,0 +1,236 @@
+package datadog.trace.plugin.csi.impl;
```
anderruiz, line 1: This class does the parsing of the callbacks to search for @CallSite annotations to find the instrumentation spots

buildSrc/src/main/java/datadog/trace/plugin/csi/AdviceGenerator.java
```diff
@@ -0,0 +1,59 @@
...
+    }
+
+    @Override
+    public boolean isSuccess() {
+      return super.isSuccess() && advices.stream().allMatch(AdviceResult::isSuccess);
+    }
+
+    public List<AdviceResult> getAdvices() {
+      return advices;
+    }
```
ValentinZakharov, line 31: Such uncertainty scares a bit, we are not sure how exactly caller will be using the list (just add? may be get? remove?)

buildSrc/src/main/java/datadog/trace/plugin/csi/impl/AsmSpecificationBuilder.java
manuel-alvarez-alvarez: This PR has been divided in two, one for the instrumenter and another for the gradle plugin, this one now relates to just the instrumenter.

buildSrc/src/main/java/datadog/trace/plugin/csi/AdviceGenerator.java
manuel-alvarez-alvarez: This PR has been divided in two, one for the instrumenter and another for the gradle plugin, this one now relates to just the instrumenter.

smola (approved):

anderruiz (approved):

bantonsson (changes requested):
Great PR!

There is only one thing that I think needs fixing in some way before merging.

I was a bit confused by the benchmarks, and it feels very counter intuitive that there should be any performance difference between `callSite` and `callee` in this case, so I had to poke at them a bit, and the response differs. The `callSite` one gets back `Hello! [Transformed]` and the `callee` one gets back `Hello! [Transformed] [Transformed]`. So we're not comparing the same things.

manuel-alvarez-alvarez:
> Great PR!
> 
> There is only one thing that I think needs fixing in some way before merging.
> 
> I was a bit confused by the benchmarks, and it feels very counter intuitive that there should be any performance difference between `callSite` and `callee` in this case, so I had to poke at them a bit, and the response differs. The `callSite` one gets back `Hello! [Transformed]` and the `callee` one gets back `Hello! [Transformed] [Transformed]`. So we're not comparing the same things.

@bantonsson, first of all thanks for the feedback :smiley:, much appreciated!

My idea with the benchmark is to validate that doing call site instrumentation is feasible in terms of startup time as long as there's a proper exclusion list.

In this particular case I wanted to showcase the instrumentation of `javax.servlet.http.HttpServlet#getParameter` as it's something we need for IAST. It's true that when using callee, due to wrappers/facades/... you end up being executed a few times in the stack of a request (that's why we got the double `[Transformed]` response). I've changed to instrument only the real implementation of the interface (in this particular case it's the apache connector).

Let me know if I should do something else about it.

dd-java-agent/agent-tooling/src/main/java/datadog/trace/agent/tooling/bytebuddy/csi/Advices.java
```diff
@@ -0,0 +1,274 @@
...
+          final @Nonnull Advices advices,
+          final @Nonnull TypeDescription type,
+          final ClassLoader classLoader) {
+        return advices;
+      }
+    }
+
+    /**
+     * This class will try to parse the constant pool of the class in order to discover the required
+     * advices to parse it.
```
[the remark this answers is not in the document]
manuel-alvarez-alvarez: Ops! my bad I don't know what I was ~~parsing~~ thinking when I wrote it, fixed!

bantonsson (approved):

dd-java-agent/agent-tooling/src/main/java/datadog/trace/agent/tooling/bytebuddy/csi/CallSiteInstrumenter.java
```diff
@@ -0,0 +1,57 @@
...
+import datadog.trace.agent.tooling.csi.CallSiteAdvice;
+import java.util.ServiceLoader;
+import javax.annotation.Nonnull;
+
+/**
+ * Instrumented using a {@link CallSiteTransformer} to perform the actual instrumentation. It will
+ * fetch the instance of the {@link CallSiteAdvice} from SPI according to the specified marker
+ * interface.
+ */
+public abstract class CallSiteInstrumenter extends Instrumenter.Default
```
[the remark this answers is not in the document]
manuel-alvarez-alvarez: Exactly, couldn't have explained it better. Our idea is to provide a `CallSiteInstrumenter` that can be used by any other product if so desired.

DDJavierSantos (approved):

PR #3708, opened 2022-08-10
Dependencies resolution distributed in time.

ValentinZakharov:
# What Does This Do
Reworked dependency resolution. Now - distributed in time.
Always computing hash if there is no reliable (built-in) maven info inside the dependency jar-file.

# Motivation
Dependency resolution and hash computation can take a lot of resource depends on jar-file size. In practice this will looks like CPU usage spike at application startup.

# Additional Notes
In current implementation we are resolving one new dependency every second. We assume that 1 second is enough to traverse entries of one jar and compute hash. E.g. in one minute we will recognise 60 dependencies without huge impact to a customer's application.

### DIFF
--- a/telemetry/src/main/java/datadog/telemetry/TelemetrySystem.java
+++ b/telemetry/src/main/java/datadog/telemetry/TelemetrySystem.java
@@ -19,11 +19,13 @@ public class TelemetrySystem {
   private static final Logger log = LoggerFactory.getLogger(TelemetrySystem.class);
 
   private static Thread TELEMETRY_THREAD;
+  private static DependencyService DEPENDENCY_SERVICE;
 
   static DependencyService createDependencyService(Instrumentation instrumentation) {
     if (instrumentation != null) {
       DependencyServiceImpl dependencyService = new DependencyServiceImpl();
       dependencyService.installOn(instrumentation);
+      dependencyService.schedulePeriodicResolution();
       return dependencyService;
     }
     return null;
@@ -33,6 +35,7 @@ public class TelemetrySystem {
       TelemetryService telemetryService,
       OkHttpClient okHttpClient,
       DependencyService dependencyService) {
+    DEPENDENCY_SERVICE = dependencyService;
     TelemetryRunnable telemetryRunnable =
         new TelemetryRunnable(
             okHttpClient,
@@ -55,6 +58,9 @@ public class TelemetrySystem {
   }
 
   public static void stop() {
+    if (DEPENDENCY_SERVICE != null) {
+      DEPENDENCY_SERVICE.stop();
+    }
     if (TELEMETRY_THREAD != null) {
       TELEMETRY_THREAD.interrupt();
       try {
--- a/telemetry/src/main/java/datadog/telemetry/dependency/DependencyResolver.java
+++ b/telemetry/src/main/java/datadog/telemetry/dependency/DependencyResolver.java
@@ -1,0 +1,137 @@
+package datadog.telemetry.dependency;
+
+import java.io.File;
+import java.io.IOException;
+import java.io.InputStream;
+import java.net.JarURLConnection;
+import java.net.URI;
+import java.net.URL;
+import java.nio.file.Files;
+import java.util.jar.Attributes;
+import java.util.jar.JarFile;
+import java.util.jar.Manifest;
+import org.slf4j.Logger;
+import org.slf4j.LoggerFactory;
+
+public class DependencyResolver {
+
+  private static final Logger log = LoggerFactory.getLogger(DependencyResolver.class);
+  private static final String JAR_SUFFIX = ".jar";
+
+  public static Dependency resolve(URI uri) {
+    return identifyLibrary(uri);
+  }
+
+  /**
+   * Identify library from a URI
+   *
+   * @param uri URI to a dependency
+   * @return dependency, or null if unable to qualify jar
+   */
+  // package private for testing
+  static Dependency identifyLibrary(URI uri) {
+    String scheme = uri.getScheme();
+    Dependency dependency = null;
+    try {
+      if ("file".equals(scheme)) {
+        dependency = identifyLibrary(new File(uri));
+      } else if ("jar".equals(scheme)) {
+        dependency = getNestedDependency(uri);
+      }
+    } catch (RuntimeException rte) {
+      log.warn("Failed to determine dependency for uri {}", uri, rte);
+    }
+    // TODO : moving jboss vfs here is probably a idea
+    // it might however require to do somme checks to make sure it's only applied to jboss
+    // and not any application server that also uses vfs:// locations
+
+    return dependency;
+  }
+
+  /**
+   * Identify a library from a .jar file
+   *
+   * @param jar jar dependency
+   * @return detected dependency, {@code null} if unable to get dependency from jar
+   */
+  static Dependency identifyLibrary(File jar) {
+    if (!jar.exists()) {
+      log.warn("unable to find dependency {}", jar);
+      return null;
+    } else if (!jar.getName().endsWith(JAR_SUFFIX)) {
+      log.debug("unsupported file dependency type : {}", jar);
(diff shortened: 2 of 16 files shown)


telemetry/src/main/java/datadog/telemetry/dependency/DependencyResolver.java
```diff
@@ -0,0 +1,137 @@
...
+import org.slf4j.LoggerFactory;
+
+public class DependencyResolver {
+
+  private static final Logger log = LoggerFactory.getLogger(DependencyResolver.class);
+  private static final String JAR_SUFFIX = ".jar";
+
+  public Dependency resolve(URI uri) {
+    return identifyLibrary(uri);
+  }
```
cataphract, line 23: Why can't this all be `static`? Just for testing?

telemetry/src/main/java/datadog/telemetry/dependency/DependencyResolverQueue.java
```diff
@@ -0,0 +1,73 @@
...
+
+import java.net.URI;
+import java.util.HashSet;
+import java.util.Set;
+import java.util.concurrent.BlockingQueue;
+import java.util.concurrent.LinkedBlockingQueue;
+import org.slf4j.Logger;
+import org.slf4j.LoggerFactory;
+
+public class DependencyResolverQueue {
```
cataphract, line 11: I found this class initially confusing. It presents itself as a queue, but it's neither a java `Queue`, nor is it merely a plain queue -- it  actually has expensive processing on its `pollDependency` method. I don't have great suggestions though.

telemetry/src/main/java/datadog/telemetry/dependency/DependencyServiceImpl.java
```diff
@@ -51,52 +37,17 @@ public void installOn(Instrumentation instrumentation) {
 
   @Override
   public Collection<Dependency> determineNewDependencies() {
```
cataphract, line 39: The name of this method no longer reflects what it's doing. Maybe something like `drainDeterminedDependencies`.

telemetry/src/main/java/datadog/telemetry/dependency/DependencyServiceImpl.java
```diff
@@ -1,43 +1,29 @@
...
+  private final DependencyResolverQueue resolverQueue = new DependencyResolverQueue();
 
-  private Set<URI> currentSet; // guarded by this
-  private volatile JbossVirtualFileHelper jbossVirtualFileHelper;
+  private final BlockingQueue<Dependency> newDependencies = new LinkedBlockingQueue<>();
 
   public DependencyServiceImpl() {
-    this.currentSet = new HashSet<>();
+    AgentTaskScheduler.INSTANCE.scheduleAtFixedRate(
+        AgentTaskScheduler.RunnableTask.INSTANCE, this, 0, 1000L, TimeUnit.MILLISECONDS);
```
cataphract, line 26: It's a bad idea to have this sort of side effects on a constructor, first of all for testing.

You can have a separate method:

```java
public void schedulePeriodicResolution() {
  AgentTaskScheduler.INSTANCE.scheduleAtFixedRate(
      AgentTaskScheduler.RunnableTask.INSTANCE,
      this::resolveOneDependency, 0, 1000L, TimeUnit.MILLISECONDS);
}

public void resolveOneDependency() {
  Dependency dep = resolverQueue.pollDependency();
  if (dep != null) {
    log.info("Resolved dep {}", dep.getName());
    newDependencies.add(dep);
  }
}
```

telemetry/src/main/java/datadog/telemetry/dependency/DependencyResolverQueue.java
```diff
@@ -0,0 +1,73 @@
...
+
+  private static final Logger log = LoggerFactory.getLogger(DependencyResolverQueue.class);
+
+  private final DependencyResolver resolver;
+  private final BlockingQueue<URI> newUrlsQueue;
```
cataphract, line 16: You're not interesting in any blocking, so you can use just `ConcurrentLinkedQueue`.

telemetry/src/main/java/datadog/telemetry/dependency/DependencyServiceImpl.java
cataphract: In testing at least, you will want to cancel the scheduling as well. For this, you will need to save the `Scheduled` object.

telemetry/src/main/java/datadog/telemetry/dependency/DependencyResolver.java
ValentinZakharov: Fixed

telemetry/src/main/java/datadog/telemetry/dependency/DependencyResolverQueue.java
ValentinZakharov: Basically this class is the resolver, with the ability to queue multiple dependencies to resolve and future consumption (polling). That's why I was thinking `ResolverQueue` reflects its concept. However, you are right, `Queue` word is confusing.

telemetry/src/main/java/datadog/telemetry/dependency/DependencyServiceImpl.java
ValentinZakharov: Fixed

telemetry/src/main/java/datadog/telemetry/dependency/DependencyServiceImpl.java
ValentinZakharov: Fixed

telemetry/src/main/java/datadog/telemetry/dependency/DependencyResolverQueue.java
ValentinZakharov: Fixed

telemetry/src/test/groovy/datadog/telemetry/dependency/DependencyServiceImplSpecification.groovy
```diff
@@ -253,15 +91,15 @@ class DependencyServiceImplSpecification extends DDSpecification {
     t.transform(null, null, null, null, null)
 
     then:
-    depService.determineNewDependencies().isEmpty()
+    depService.drainDeterminedDependencies().isEmpty()
```
cataphract, line 94: A call to `resolveOneDependency` should be made here and in the other methods that expect no new dependency to be detected, otherwise `drainDeterminedDependencies()` being empty doesn't mean that the input to the transformer was ignored (the uri may have been added to the queue, but just no processed).

telemetry/src/main/java/datadog/telemetry/TelemetrySystem.java
```diff
@@ -24,6 +24,7 @@ static DependencyService createDependencyService(Instrumentation instrumentation
     if (instrumentation != null) {
       DependencyServiceImpl dependencyService = new DependencyServiceImpl();
       dependencyService.installOn(instrumentation);
+      dependencyService.schedulePeriodicResolution();
```
cataphract, line 27: We have tests calling this method (via `startTelemetry`), so to avoid causing side effects for other tests there should be a mechanism to stop the periodic action (and this mechanism should at the very least be called during the tests).

telemetry/src/test/groovy/datadog/telemetry/dependency/DependencyServiceImplSpecification.groovy
ValentinZakharov: Right, I've missed this. Fixed

telemetry/src/main/java/datadog/telemetry/TelemetrySystem.java
ValentinZakharov: Added `stop()` method to interrupt scheduled task

cataphract (approved):

PR #3711, opened 2022-08-10
Add instrumentation for main method

luhenry:
# Motivation

This is for a PoC of CI App an Profiling instrumentation.

### DIFF
--- a/dd-java-agent/instrumentation/main/src/main/java/datadog/trace/instrumentation/main/MainInstrumentation.java
+++ b/dd-java-agent/instrumentation/main/src/main/java/datadog/trace/instrumentation/main/MainInstrumentation.java
@@ -1,0 +1,52 @@
+package datadog.trace.instrumentation.main;
+
+import static datadog.trace.agent.tooling.bytebuddy.matcher.NameMatchers.named;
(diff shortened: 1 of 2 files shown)


dd-java-agent/instrumentation/main/src/main/java/datadog/trace/instrumentation/main/MainInstrumentation.java
```diff
@@ -27,6 +31,11 @@ protected boolean defaultEnabled() {
     return false;
   }
 
+  @Override
+  public ElementMatcher<TypeDescription> hierarchyMatcher() {
+    return extendsClass(named("java.lang.Object"));
```
jbachorik, line 36: This will be really expensive :(

luhenry: This is not meant to be merged in any way. That's only to have some artifacts to test it out on our backend's CI.

luhenry:
Closing as I don't need the PR to trigger CircleCI builds, I can just do it by hand.

PR #3712, opened 2022-08-10
[IAST] Vulnerability hashing and deduplication V0

jandro996:
# What Does This Do

- Add the `hash` property to vulnerability model,  this will be used to determine if two vulnerabilities are equals
- Add logic to `Reporter` that manages duplicated vulnerabilities reporting

# Motivation

Because we discover vulnerabilities by analyzing the requests, and different requests can lead to vulnerabilities with same type in the same file and line, we need mechanisms to not be reporting vulnerabilities persistently referencing same code issue.

# Additional Notes
PR fixing system tests: https://github.com/DataDog/system-tests/pull/508

For this first version:

- We are going to define that whenever a vulnerability has same type and occurs in the same file and line, they will be considered equals regardless of the rest of the parameters. This is the reason to add a hash field composed with type, file and line to the vulnerability model. 

- There will have two working modes:
     - `iast.deduplication.enabled=true` (Default): Vulnerabilities with same hash can be reported one time per run time. It is necessary to maintain a cache with already reported vulnerabilities hashes that can cause performance issues if not limited, for this reasons once 1000 vulnerabilities have been reported the cache will be reset.
    - `iast.deduplication.enabled=false` this working mode is demo oriented and skips deduplication

### DIFF
--- a/dd-java-agent/iast/src/main/java/com/datadog/iast/IastSystem.java
+++ b/dd-java-agent/iast/src/main/java/com/datadog/iast/IastSystem.java
@@ -28,7 +28,7 @@ public class IastSystem {
     }
     log.debug("IAST is starting");
 
-    final Reporter reporter = new Reporter();
+    final Reporter reporter = new Reporter(config);
     final OverheadController overheadController =
         new OverheadController(config, AgentTaskScheduler.INSTANCE);
     final IastModule iastModule = new IastModuleImpl(config, reporter, overheadController);
--- a/dd-java-agent/iast/src/main/java/com/datadog/iast/Reporter.java
+++ b/dd-java-agent/iast/src/main/java/com/datadog/iast/Reporter.java
@@ -2,16 +2,32 @@ package com.datadog.iast;
 
 import com.datadog.iast.model.Vulnerability;
 import com.datadog.iast.model.VulnerabilityBatch;
+import datadog.trace.api.Config;
 import datadog.trace.api.DDTags;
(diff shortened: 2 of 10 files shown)


dd-java-agent/iast/src/main/java/com/datadog/iast/deduplication/DeduplicationControllerFactory.java
```diff
@@ -0,0 +1,60 @@
...
+    return new DeduplicationController() {
+
+      private final Set<Long> hashes = new HashSet<>(Arrays.asList(new Long[1000]));
+
+      @Override
+      public boolean isADuplicate(Vulnerability vulnerability) {
+        boolean isADuplicate = hashes.contains(vulnerability.getHash());
+        if (!isADuplicate) {
+          if (hashes.size() == 1000) {
+            hashes.clear();
```
DDJavierSantos, line 47: Maybe use hashes.remove(hashes.iterator().next()) that will remove the first entry from the hashset

manuel-alvarez-alvarez: We want to start over once the limit has been reached, so removing the first entry is not enough.

bantonsson (approved):

smola:
Depends on: https://github.com/DataDog/system-tests/pull/508

smola (approved):

manuel-alvarez-alvarez (approved):

PR #3717, opened 2022-08-12
Minor improvements for per-span context tracking

jbachorik:
# What Does This Do
This PR reduces the number of 'context changes' we are tracking.
This is achieved by:
- obeying the sampling decision and not tracking spans which are eligible for dropping
- improving integration between the scope manager and context tracking

During addressing the second part I also realised that for the profiling purposes the 'start span', 'end span', 'start thread migration' and 'end thread migration' checkpoints are not really needed as we are interested only when a span 'is doing some work'. We seem to be achieving the same coverage by fixing the integration with the scope manager and just using 'start task' and 'end task' checkpoints.

# Motivation
The main driving force here is the attempt to reduce the amount of generated data while keeping the relevant coverage.

# Additional Notes
The tests asserting the `CPU` and `CPU | END` checkpoints were relaxed as the actual numbers seem to fluctuate wildly, especially on CI servers. The actual event counts are not really that interesting anyway since the correctness of the emitted data is validated by special validators.

### DIFF
--- a/dd-trace-core/jfr-openjdk/src/main/java/datadog/trace/core/jfr/openjdk/JFRCheckpointer.java
+++ b/dd-trace-core/jfr-openjdk/src/main/java/datadog/trace/core/jfr/openjdk/JFRCheckpointer.java
@@ -149,7 +149,7 @@ public class JFRCheckpointer implements Checkpointer, ProfilingListener<Profilin
   }
 
   private void tryEmitCheckpoint(final AgentSpan span, final int flags) {
-    if (flags != 2 && flags != 3) {
+    if (flags != CPU && flags != (CPU | END)) {
       // emit only CPU and CPU | END events
       return;
     }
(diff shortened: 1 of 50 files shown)


dd-trace-core/jfr-openjdk/src/main/java/datadog/trace/core/jfr/openjdk/JFRCheckpointer.java
```diff
@@ -149,6 +149,10 @@ public void onData(ProfilingSnapshot observable) {
   }
 
   private void tryEmitCheckpoint(final AgentSpan span, final int flags) {
+    if (flags != 2 && flags != 3) {
```
luhenry, line 152: Instead of magic variables 2 and 3, why not `flag != CPU && flags != (CPU | END)`?

dd-trace-core/src/main/java/datadog/trace/core/DDSpan.java
```diff
@@ -57,7 +57,10 @@ static DDSpan create(
     final DDSpan span = new DDSpan(timestampMicro, context, emitCheckpoints);
     log.debug("Started span: {}", span);
     context.getTrace().registerSpan(span);
-    span.tracingContextTracker = TracingContextTrackerFactory.instance(span);
+    span.tracingContextTracker =
+        span.eligibleForDropping()
+            ? TracingContextTracker.EMPTY
+            : TracingContextTrackerFactory.instance(span);
```
luhenry, line 63: That will break Wall Time profiler for Endpoint Profiling. Could you push this logic into `PerSpanTrackingContextTrackerFactory` instead?

luhenry (approved):

PR #3719, opened 2022-08-17
Add timestamp to debugger probe diagnostics messages

shatzi:
# What Does This Do

Adds timestamp to probe diagnostics message so the status would appear when it was sent and not when it was received by the backend.

# Motivation

Two bugs:
1. Wrong probe status appear when multiple status messages are sent on the same batch
2. timestamp not updated when probe status been reemitted

# Additional Notes

### DIFF
--- a/dd-java-agent/agent-debugger/src/main/java/com/datadog/debugger/sink/ProbeStatusSink.java
+++ b/dd-java-agent/agent-debugger/src/main/java/com/datadog/debugger/sink/ProbeStatusSink.java
@@ -27,7 +27,7 @@ public class ProbeStatusSink {
 
   private final Builder messageBuilder;
   private final Map<String, TimedMessage> probeStatuses = new ConcurrentHashMap<>();
-  private final ArrayBlockingQueue<ProbeStatus> queue = new ArrayBlockingQueue<>(1000);
+  private final ArrayBlockingQueue<ProbeStatus> queue;
   private final Duration interval;
   private final int batchSize;
   private final boolean isInstrumentTheWorld;
@@ -36,6 +36,7 @@ public class ProbeStatusSink {
     this.messageBuilder = new Builder(config);
     this.interval = Duration.ofSeconds(config.getDebuggerDiagnosticsInterval());
     this.batchSize = config.getDebuggerUploadBatchSize();
+    this.queue = new ArrayBlockingQueue<>(2 * this.batchSize);
     this.isInstrumentTheWorld = config.isDebuggerInstrumentTheWorld();
   }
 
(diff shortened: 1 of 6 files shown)


dd-java-agent/agent-debugger/src/main/java/com/datadog/debugger/sink/ProbeStatusSink.java
```diff
@@ -36,6 +36,7 @@ public class ProbeStatusSink {
     this.messageBuilder = new Builder(config);
     this.interval = Duration.ofSeconds(config.getDebuggerDiagnosticsInterval());
     this.batchSize = config.getDebuggerUploadBatchSize();
+    this.queue = new ArrayBlockingQueue<>(2 * this.batchSize);
```
cimi, line 39: We've changed the capacity of the queue from 1000 to `2 * batchSize` and we've also changed the enqueueing logic. We can have at most 100 probes in an application, what's the current batch size?

I'm trying to understand if we need to keep the old capacity - I saw we're doing checks on `missingCapacity` below, but we only check once if there are items that couldn't be enqueuend when the queue was full.

If the batch size is too small compared to the 100 max probes per service we can drop messages.

cimi (approved):
This implementation passes all our integration tests and is not currently used anywhere so I'm ok with merging as it is.

In the future I think we should simplify this logic - maybe we can get rid of the queue entirely and use the probe id -> probe status map to handle this.

dd-java-agent/agent-debugger/src/main/java/com/datadog/debugger/sink/ProbeStatusSink.java
shatzi: we no longer "drop" messages as delete messages from queue. assume we get 100 probes and each probe generate receive and installed or error = 200 messages. but if the queue is full, we only keep track of the latest status per probe and we will push it on the next batch. 

I have decided to reduce from 1000 to 200 as we no longer "clear" the queue. so pushing probe status at the end would take 10 cycles of flushes those messages. so I thought a buffer of 1 batch would be a good tradeoff.

But maybe increase it back to 3 (or even 10) cycles would be good as we send them every 1 sec. WDYT?

cimi: The batch size is configurable, the default is currently set to 100: https://github.com/DataDog/dd-trace-java/blob/78fff6c24f975c64f72b131edc172841c1ac6843/dd-trace-api/src/main/java/datadog/trace/api/ConfigDefaults.java#L90

I don't think anyone would need to override this (and I don't even know if it's possible now through properties or otherwise).

We can leave it like this now and refactor this part when everything is working.

PR #3720, opened 2022-08-17
Add container-id and tags to remote config requests

shatzi:
# What Does This Do

Adds tags to remote config request (+ container-id sent as an http header)

# Motivation
Allow users to choose which tracers to apply probes based on tags query.
The tags that we are adding propagated to the backend and indexed so the user can define query to choose which tracers should apply debugger probes.

# Additional Notes

Those changes will need to be merged with #3627 as this change adds tags to the RemoteConfigRequest and containerId to PollerHttpClient


dd-java-agent/agent-debugger/src/main/java/com/datadog/debugger/poller/PollerRequestFactory.java
```diff
@@ -16,15 +16,19 @@
 class PollerRequestFactory {
   static final String HEADER_DD_API_KEY = "DD-API-KEY";
   static final String HEADER_DEBUGGER_TRACKING_ID = "X-Datadog-HostId";
```
cimi, line 18: I thought we had removed these? They are no longer used since we're using Remote Config. We can clean this up in a separate PR.

dd-java-agent/agent-debugger/src/main/java/com/datadog/debugger/poller/ConfigurationPoller.java
```diff
@@ -100,7 +106,8 @@ public ConfigurationPoller(
         listener,
         new RatelimitedLogger(log, MINUTES_BETWEEN_ERROR_LOG, TimeUnit.MINUTES),
         "",
-        AgentTaskScheduler.INSTANCE);
+        AgentTaskScheduler.INSTANCE,
+        "");
```
cimi, line 110: Why do we use a blank string here instead of `ContainerInfo.get().getContainerId()`? I think the result of that is cached, so it'd be better to get the real value instead of setting it to a blank string. (also applies below)

I'm not sure where these methods are used - but even if only for tests I think this is fine. The tests should handle `ContainerInfo` gracefully.

dd-java-agent/agent-debugger/src/main/java/com/datadog/debugger/poller/PollerRequestFactory.java
```diff
@@ -16,15 +16,19 @@
...
 
-  static Request newConfigurationRequest(Config config, String url, Moshi moshi) {
+  static Request newConfigurationRequest(
+      Config config, String url, Moshi moshi, String containerId) {
     Request.Builder requestBuilder = new Request.Builder().url(parseUrl(url)).get();
     MediaType applicationJson = MediaType.parse("application/json");
     RequestBody requestBody =
         RequestBody.create(applicationJson, buildRemoteConfigJson(config, moshi));
     requestBuilder.post(requestBody);
+    requestBuilder.addHeader(HEADER_CONTAINER_ID, containerId);
```
cimi, line 31: Should we skip adding the header if the container id string is empty or null?

dd-java-agent/agent-debugger/src/main/java/com/datadog/debugger/poller/PollerRequestFactory.java
```diff
@@ -46,10 +50,11 @@ private static HttpUrl parseUrl(String appProbesUrl) {
   }
 
   private static String buildRemoteConfigJson(Config config, Moshi moshi) {
-    String trackingId = getTrackingId(config);
+    String runtimeId = getTrackingId(config);
```
cimi, line 53: Let's also rename the method to `getRuntimeId`?

communication/src/main/java/datadog/communication/ddagent/TracerVersion.java
```diff
@@ -0,0 +1,23 @@
...
+        new BufferedReader(
+            new InputStreamReader(
+                cl.getResourceAsStream("dd-java-agent.version"), StandardCharsets.ISO_8859_1))) {
+      String line = reader.readLine();
+      return line != null ? line : "0.0.0";
+    } catch (IOException e) {
+      return "0.0.0";
+    }
+  }
+}
```
cimi, line 23: We could inline this where we use it and extract the utility method in a separate PR. I saw we get the version in a couple of other different places.

This may also fix the coverage failure we see in CI.

cimi (approved):
:+1: Thanks! 

I've left a few comments, since this code is not used yet I'm ok with merging as is and addressing the comments in a future PR.

All the RC code in this PR will be removed in favour of the shared client from the PR linked in the description.

dd-java-agent/agent-debugger/src/main/java/com/datadog/debugger/poller/ConfigurationPoller.java
shatzi: Good suggestions

communication/src/main/java/datadog/communication/ddagent/TracerVersion.java
shatzi: PR #3627 was merged.

remote-config/src/main/java/datadog/remoteconfig/SizeCheckedInputStream.java
```diff
@@ -17,25 +17,40 @@ public SizeCheckedInputStream(InputStream in, long maxSize) {
...
   public int read(byte[] b) throws IOException {
-    checkSize(b.length);
-    return in.read(b);
+    if (maxSize == currentSize) {
+      throw new IOException("Reached maximum bytes for this stream: " + maxSize);
+    }
+
+    long maxLength = Math.min(b.length, maxSize - currentSize);
+    checkSize(maxLength);
+    return updateCurrentSize(in.read(b, 0, (int) maxLength));
```
jpbempel, line 35: I am confused how this need to be different from read(byte[], int, int) ?
in fact this method should call directly `read(b, 0, b.length)`

dd-java-agent/agent-installer/src/main/java/datadog/trace/agent/installer/AgentInstaller.java
```diff
@@ -37,6 +37,7 @@ public class AgentInstaller {
   public static void installBytebuddyAgent(Instrumentation inst) {
     if (Config.get().isTraceEnabled()
         || Config.get().isProfilingEnabled()
+        || Config.get().isDebuggerEnabled()
```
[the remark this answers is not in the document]
shatzi: removed

remote-config/src/main/java/datadog/remoteconfig/SizeCheckedInputStream.java
shatzi: you are right. now it just call read(b,0,b.length) and safe checks moved to read with offset call.

PR #3721, opened 2022-08-17
JVM shutdown instrumentation

jbachorik:
# What Does This Do
This adds a custom instrumentation to intercept JVM shutdown before any of the regular shutdown hooks have the chance to run.

# Motivation
Allow the agent to be orderly brought to halt - although this would usually be achieved by custom shutdown hooks it is unfortunately not possible for the profiler because the shutdown hooks are run in no particular order and JFR is registering its own shutdown hook which might be run before our profiler hook, causing the profiler data to disappear.

# Additional Notes
The instrumentation is disabled by default and needs to be manually enabled.


dd-java-agent/agent-bootstrap/src/main/java/datadog/trace/bootstrap/Agent.java
```diff
@@ -700,6 +700,7 @@ public void withTracer(Tracer tracer) {
   private static void shutdownProfilingAgent(final boolean sync) {
     if (AGENT_CLASSLOADER == null) {
       // It wasn't started, so no need to shut it down
+      log.info("Profiler has not been initialized yet");
```
luhenry, line 703: `debug` level instead?

dd-java-agent/agent-builder/src/main/java/datadog/trace/agent/tooling/AgentInstaller.java
```diff
@@ -55,6 +56,11 @@ public static void installBytebuddyAgent(final Instrumentation inst) {
         || Config.get().isIastEnabled()
         || Config.get().isCiVisibilityEnabled()) {
       installBytebuddyAgent(inst, false, new AgentBuilder.Listener[0]);
+      try {
+        inst.retransformClasses(System.class);
+      } catch (UnmodifiableClassException e) {
+        log.warn("", e);
```
luhenry, line 62: Is that a leftover? It should likely have a comment otherwise so we can easily find this log line.

dd-java-agent/agent-profiling/profiling-controller/src/main/java/com/datadog/profiling/controller/ProfilingSystem.java
```diff
@@ -109,7 +109,7 @@ public ProfilingSystem(
       throw new ConfigurationException("Upload period must be positive.");
     }
 
-    // Note: is is important to not keep reference to the threadLocalRandom beyond the constructor
+    // Note: ii is important to not keep reference to the threadLocalRandom beyond the constructor
```
luhenry, line 112:
```suggestion
    // Note: it is important to not keep reference to the threadLocalRandom beyond the constructor
```

dd-java-agent/instrumentation/main/main.gradle
```diff
@@ -0,0 +1,12 @@
+apply from: "$rootDir/gradle/java.gradle"
```
luhenry, line 1: We could rename this instrumentation from `main` to `shutdown`. We are doing a bit more than `main` instrumentation now.

luhenry (approved):

dd-java-agent/agent-bootstrap/src/main/java/datadog/trace/bootstrap/Agent.java
jbachorik: Removed

dd-java-agent/agent-builder/src/main/java/datadog/trace/agent/tooling/AgentInstaller.java
jbachorik: Yeah. Interestingly enough, it is easier to spot the leftovers in the GH diff view than in the IDE :( 
Cleaned it up ...

dd-java-agent/agent-profiling/profiling-controller/src/main/java/com/datadog/profiling/controller/ProfilingSystem.java
jbachorik: Done

dd-java-agent/instrumentation/main/main.gradle
jbachorik: Renamed. It just took a while to commit because macos throttled CPU to ~25% while rebuilding the project structure after the module name change :)

PR #3723, opened 2022-08-17
Include Datadog tags in Print/LogWriter

ygree:
# What Does This Do

Includes [Datadog Tags](https://github.com/DataDog/dd-trace-java/pull/3611) into Printer/Log Writer output

# Additional Notes

### DIFF
--- a/dd-trace-core/src/main/java/datadog/trace/common/writer/DDSpanJsonAdapter.java
+++ b/dd-trace-core/src/main/java/datadog/trace/common/writer/DDSpanJsonAdapter.java
@@ -6,6 +6,8 @@ import com.squareup.moshi.Moshi;
 import com.squareup.moshi.Types;
 import datadog.trace.api.DDId;
 import datadog.trace.core.DDSpan;
+import datadog.trace.core.Metadata;
+import datadog.trace.core.MetadataConsumer;
 import java.io.IOException;
 import java.lang.annotation.Annotation;
 import java.lang.reflect.Type;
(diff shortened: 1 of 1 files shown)


dd-trace-core/src/main/java/datadog/trace/common/writer/DDSpanJsonAdapter.java
```diff
@@ -73,8 +75,9 @@ public void toJson(final com.squareup.moshi.JsonWriter writer, final DDSpan span
     writer.endObject();
     writer.name("meta");
     writer.beginObject();
-    final Map<String, Object> tags = span.getTags();
-    for (final Map.Entry<String, String> entry : span.context().getBaggageItems().entrySet()) {
+    Metadata metadata = spanMetadata(span);
+    final Map<String, Object> tags = metadata.getTags();
```
bantonsson, line 79: This is not thread safe. The tags live in a normal `HashMap` and they are only _locked_ during the `processTagsAndBaggage` call. Furthermore this call will not have the `thread.id`, `thread.name`, `http.status_code`, et.c. that are only added in the `span.getTags()` call.

dd-trace-core/src/main/java/datadog/trace/common/writer/DDSpanJsonAdapter.java
```diff
@@ -90,6 +93,18 @@ public void toJson(final com.squareup.moshi.JsonWriter writer, final DDSpan span
     writer.endObject();
   }
 
+  private Metadata spanMetadata(DDSpan span) {
+    final Metadata[] metadata = new Metadata[1];
+    span.processTagsAndBaggage(
+        new MetadataConsumer() {
+          @Override
+          public void accept(Metadata md) {
+            metadata[0] = md;
```
bantonsson, line 102: The `MetaData` is only valid for the duration of this call. Yes, the code should have comments.

bantonsson (changes requested):

PR #3724, opened 2022-08-18
Fix ProcessSupervisorTest flakiness

smola:
# What Does This Do

Adds missing `assert`s within closures.

# Motivation

Spock does not introduce implicit asserts in closures. This means that
everything in the `conditions.eventually` blocks in these tests is being
ignored unless it throws an exception.

In some cases, this makes the `Process respawns when killed` test case
fail on the `secondProcess != null` assertion.

# Additional Notes


bantonsson (approved):
Really nice catch!

PR #3730, opened 2022-08-19
Send additional StatsPoint for each pathway checkpoint to compute availability metric.

hokitam:
# What Does This Do
This change, modeled after Golang PR https://github.com/DataDog/data-streams-go/pull/12/files, introduces the concept of an origin-based StatsPoint, where the timestamp of the StatsPoint is based on the time at origin instead of the current time. When sending the StatsPoint to Datadog, both the current timestamp StatsPoint (today's default StatsPoint) and the origin-based StatsPoint are sent.

# Motivation
Having a new origin-based StatsPoint will allow us to compute a `received points / sent points` ratio, aligned at the same timestamp for a particular pathway, regardless of when the message was produced or consumed.

# Additional Notes

### DIFF
--- a/dd-trace-core/src/main/java/datadog/trace/core/datastreams/DefaultDataStreamsCheckpointer.java
+++ b/dd-trace-core/src/main/java/datadog/trace/core/datastreams/DefaultDataStreamsCheckpointer.java
@@ -42,7 +42,8 @@ public class DefaultDataStreamsCheckpointer
   private static final StatsPoint POISON_PILL =
       new StatsPoint(Collections.<String>emptyList(), 0, 0, 0, 0, 0);
 
-  private final Map<Long, StatsBucket> timeToBucket = new HashMap<>();
+  private final Map<Long, StatsBucket> tsCurrentBuckets = new HashMap<>();
+  private final Map<Long, StatsBucket> tsOriginBuckets = new HashMap<>();
   private final BlockingQueue<StatsPoint> inbox = new MpscBlockingConsumerArrayQueue<>(1024);
   private final DatastreamsPayloadWriter payloadWriter;
   private final DDAgentFeaturesDiscovery features;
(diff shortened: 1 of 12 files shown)


dd-trace-core/src/main/java/datadog/trace/core/datastreams/DefaultDataStreamsCheckpointer.java
```diff
@@ -203,11 +212,13 @@ private long currentBucket(long timestampNanos) {
     return timestampNanos - (timestampNanos % bucketDurationNanos);
   }
 
-  private void flush(long timestampNanos) {
```
dougqh, line 206: It is probably better for this take an Iterable<Map.Entry<...>> rather than Iterator<Map.Entry<...>>.
Iterable-s can be reused Iterator-s cannot.

dd-trace-core/src/main/java/datadog/trace/core/datastreams/DefaultDataStreamsCheckpointer.java
```diff
@@ -177,18 +178,26 @@ public void run() {
...
+            Long tsCurrentBucket = currentBucket(statsPoint.getTimestampNanos());
 
             // FIXME computeIfAbsent() is not available because Java 7
             // No easy way to have Java 8 in core even though datastreams monitoring is 8+ from
             // DDSketch
-            StatsBucket statsBucket = timeToBucket.get(bucket);
-            if (statsBucket == null) {
-              statsBucket = new StatsBucket(bucket, bucketDurationNanos);
-              timeToBucket.put(bucket, statsBucket);
+            StatsBucket tsCurrentStatsBucket = tsCurrentBuckets.get(tsCurrentBucket);
```
dougqh, line 188: We probably need to revisit this.  It looks right this can grow in an unbounded fashion.  That's strictly forbidden inside tracer code.  We should probably convert this Map into ring buffer or something similar instead.

dd-trace-core/src/main/java/datadog/trace/core/datastreams/DefaultDataStreamsCheckpointer.java
```diff
@@ -203,11 +212,13 @@ private long currentBucket(long timestampNanos) {
...
   }
 
-  private void flush(long timestampNanos) {
+  private void flush(
+      long timestampNanos,
+      Iterator<Map.Entry<Long, StatsBucket>> mapIterator,
+      StatsBucket.BucketType bucketType) {
     long currentBucket = currentBucket(timestampNanos);
 
     List<StatsBucket> includedBuckets = new ArrayList<>();
```
dougqh, line 221: How often are we flushing?
I think it might be better to build a transmission object after each flush rather than copying over.
Although, I know this is what the tracing code does, too.

dd-trace-core/src/main/java/datadog/trace/core/datastreams/DefaultDataStreamsCheckpointer.java
```diff
@@ -220,10 +231,19 @@ private void flush(long timestampNanos) {
...
+      payloadWriter.writePayload(includedBuckets, bucketType);
     }
   }
 
+  private void flush(long timestampNanos) {
+    Iterator<Map.Entry<Long, StatsBucket>> tsCurrentIterator =
+        tsCurrentBuckets.entrySet().iterator();
+    flush(timestampNanos, tsCurrentIterator, StatsBucket.BucketType.TIMESTAMP_CURRENT);
+
+    Iterator<Map.Entry<Long, StatsBucket>> tsOriginIterator = tsOriginBuckets.entrySet().iterator();
```
dougqh, line 243: As mentioned elsewhere, it would be better to pass a Set or Iterable of Map.Entry-s rather than an Iterator.

dd-trace-core/src/main/java/datadog/trace/core/datastreams/StatsBucket.java
```diff
@@ -6,6 +6,21 @@
...
+    }
+
+    public String getValue() {
+      return value;
+    }
+  }
+
   private final long startTimeNanos;
   private final long bucketDurationNanos;
   private final Map<Long, StatsGroup> hashToGroup = new HashMap<>();
```
dougqh, line 26: This collection needs to be size constrained.

PR #3739, opened 2022-08-23
Add support for multiple configurations per product

shatzi:
# What Does This Do

Add support for multiple configurations to be applied per product for ConfigurationPoller.
Listeners now accept configKey (string) to indicate which config to apply or remove (indicated by null configuration)

# Motivation

Live debugger can receive multiple configurations.

# Additional Notes

### DIFF
--- a/remote-config/src/main/java/datadog/remoteconfig/ConfigurationPoller.java
+++ b/remote-config/src/main/java/datadog/remoteconfig/ConfigurationPoller.java
@@ -26,13 +26,11 @@ import java.util.Iterator;
 import java.util.List;
 import java.util.Map;
 import java.util.Optional;
-import java.util.Set;
 import java.util.concurrent.TimeUnit;
 import java.util.concurrent.atomic.AtomicInteger;
 import java.util.function.Consumer;
 import java.util.regex.Matcher;
 import java.util.regex.Pattern;
-import java.util.stream.Collectors;
 import okhttp3.Call;
 import okhttp3.OkHttpClient;
 import okhttp3.Request;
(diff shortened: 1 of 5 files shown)


remote-config/src/main/java/datadog/remoteconfig/ConfigurationPoller.java
```diff
@@ -352,6 +316,30 @@ private boolean processConfigKey(
...
 
+    return dl;
+  }
+
+  private boolean notifyConfigurationUpdated(
+      String configKey, byte[] fileContent, PollingRateHinter pollingRateHinter)
+      throws IOException {
+    return extractDeserializerAndListenerFromKey(configKey)
+        .deserializeAndAccept(configKey, fileContent, pollingRateHinter);
+  }
```
jpbempel, line 327: This is not used

shatzi: Good catch, removed.

remote-config/src/test/groovy/datadog/remoteconfig/ConfigurationPollerSpecification.groovy
```diff
@@ -452,7 +452,7 @@ class ConfigurationPollerSpecification extends DDSpecification {
     }
   }
 
-  void 'deserializer can return null to indicate no config should be applied'() {
+  void 'deserializer can return null, indicate this config should be un-applied'() {
```
cataphract, line 455: This changes the semantics. Returning null from the deserializer can be done when the configuration cannot be deserialized (has some unexpected format). I don't think that in that case the configuration should be "unapplied" (pulled). It would IMO make more sense to keep the old configuration.

cataphract (changes requested):
This is missing tests with the now supported multiple configurations.

jpbempel (approved):

cataphract (approved):

PR #3745, opened 2022-08-23
Fix the async-profiler context tracking

jbachorik:
# What Does This Do
Removes the implicit context activation in the tracker constructor

# Motivation
A concurrent change in the context tracking code made this activation redundant and, in fact, it will cause weird span activation intervals.

# Additional Notes

### DIFF
--- a/dd-java-agent/agent-profiling/profiling-context/src/main/java/com/datadog/profiling/context/AsyncProfilerTracingContextTracker.java
+++ b/dd-java-agent/agent-profiling/profiling-context/src/main/java/com/datadog/profiling/context/AsyncProfilerTracingContextTracker.java
@@ -34,12 +34,7 @@ public final class AsyncProfilerTracingContextTracker implements TracingContextT
   }
 
   private static final ThreadLocal<ArrayDeque<Context>> contextsThreadLocal =
-      new ThreadLocal<ArrayDeque<Context>>() {
-        @Override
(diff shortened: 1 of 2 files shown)


luhenry (approved):

dd-java-agent/agent-profiling/profiling-context/src/main/java/com/datadog/profiling/context/AsyncProfilerTracingContextTracker.java
```diff
@@ -6,8 +6,12 @@
 import datadog.trace.bootstrap.instrumentation.api.AgentSpan;
 import java.nio.ByteBuffer;
 import java.util.ArrayDeque;
+import org.slf4j.Logger;
+import org.slf4j.LoggerFactory;
 
 public final class AsyncProfilerTracingContextTracker implements TracingContextTracker {
+  private static final Logger log =
```
devinsba, line 13: Seems like you don't need this anymore

devinsba (approved):

dd-java-agent/agent-profiling/profiling-context/src/main/java/com/datadog/profiling/context/AsyncProfilerTracingContextTracker.java
jbachorik: Gone

PR #3747, opened 2022-08-23
Switch to `system-tests` for lib injection testing

devinsba:
# What Does This Do

Supersedes #3727 

# Motivation

# Additional Notes


bantonsson (approved):
I actually have no idea how this really works, but it looks just like the _nodejs_ one, so it must be good

Kyle-Verhoog (approved):

PR #3751, opened 2022-08-24
On demand port forwarding for testcontainers on CI

manuel-alvarez-alvarez:
# What Does This Do
Updates the `autoforward.py` utility to create the SSH tunnels when the Docker API asks for the actual container information. It will try to create the forward using the same port or pick a random one if it's in use, if a random one is chosen, then the return JSON will be update accordingly. The SSH forward is not established right  away, it will be delayed until the remote endpoint is actually listening, so the tescontainers waiting strategies are not mislead by the forward.

# Motivation
Testcontainers are causing lots of failures in the CI environment due to waiting strategies failing to check if containers are started successfully.

### DIFF
--- a/.circleci/autoforward.py
+++ b/.circleci/autoforward.py
@@ -1,147 +1,733 @@
 #!/usr/bin/env python3
 # -*- coding: utf-8 -*-
-import dataclasses
-import threading
-import sys
-import signal
-import subprocess
+import argparse
+import errno
 import json
-import re
-import time
 import logging
+import os
+import re
+import requests
+import requests_unixsocket
+import signal
+import socket
+import ssl
+import subprocess
+import sys
+import threading
+import time
+import urllib
+from functools import partial
+from http.server import BaseHTTPRequestHandler, ThreadingHTTPServer
+from pathlib import Path
 
 
-@dataclasses.dataclass(frozen=True)
-class Forward:
-  port: int
+def parse_args():
+  parser = argparse.ArgumentParser(description='Simple TCP proxy for Docker')
+  parser.add_argument('-p', '--port', dest='local_port', type=int, default=8080, help='Local port of the proxy')
+  parser.add_argument('-r', '--remote', dest='remote',
+                      help='Remote IP or host name (http://0.0.0.0:2375, unix:///var/run/docker.sock)')
+  parser.add_argument('-f', '--forward', dest='forward', default='remote-docker',
+                      help='Remote IP or host name to forward connections')
+  parser.add_argument('--secure', dest='secure', action='store_true')
+  parser.set_defaults(secure=False)
+  parser.add_argument('-sk', '--server-key', dest='server_key', help='Key for the local server endpoint')
+  parser.add_argument('-sc', '--server-cert', dest='server_cert', help='Certificate for the local server endpoint')
+  parser.add_argument('-rk', '--remote-key', dest='remote_key', help='Key for the remote endpoint')
+  parser.add_argument('-rc', '--remote-cert', dest='remote_cert', help='Certificate remote endpoint')
+  parser.add_argument('-rca', '--remote-ca', dest='remote_ca', help='CA for the remote endpoint')
+  parser.add_argument('--verbose', dest='verbose', action='store_true')
+  parser.set_defaults(verbose=False)
+  return parser.parse_args()
 
-  def __ne__(self, other):
-    return not self.__eq__(other)
+
+class TcpUpgrade:
+  """
+  Docker remote API defines what is called a connection Hijacking via "Connection: upgrade" and "Upgrade: tcp" headers
+  This class tries to detect a tcp upgrade handshake and keep the sockets open until the remote docker API endpoint
+  decides to close the connection
+  """
+
+  def __init__(self, proxy_server, timeout=2, buffer_size=1024, encoding='utf-8'):
+    self.proxy_server = proxy_server
+    self.args = self.proxy_server.args
+    self.remote = self.proxy_server.remote
+    self.encoding = encoding
+    self.timeout = timeout
+    self.buffer_size = buffer_size
+
+  def try_tcp_upgrade(self, local_socket, request_line, headers, body):
+    """
+    This method executes the HTTP requests and checks for the upgraded headers, if so, it handles the piped TCP stream
+    """
+    remote_socket = self._connect_to_remote()
+    try:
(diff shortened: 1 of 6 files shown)


.circleci/config.yml
```diff
@@ -117,15 +117,12 @@ commands:
       - run:
```
smola, line 117: The above sections about showing remote Docker ports, changing offsets, and restarting Docker would also be obsolote after this PR.

dd-smoke-tests/src/main/groovy/datadog/smoketest/ProcessManager.groovy
```diff
@@ -177,4 +188,20 @@ abstract class ProcessManager extends Specification {
...
     return "01234567890abcdef123456789ABCDEF"
   }
+
+  public <C> C withContainer(final C container) {
+    if (portForwarder == null) {
+      return container
+    }
+    if (container.getClass().getDeclaredMethod('start') != null) {
+      // test containers
+      final spy = Spy(container)
```
smola, line 198: Awesome!

But I wonder if instantiating this Spy here would break some mock verifications. You could try adding:

```
0 * _ # no more interactions
```

at the end of a `then` block to see if it affects it. Otherwise, we'd need to ensure that the Spy doesn't interfere with recorded interactions.

utils/test-utils/src/main/groovy/datadog/trace/test/util/DDSpecification.groovy
```diff
@@ -232,6 +241,22 @@ abstract class DDSpecification extends Specification {
     rebuildConfig()
   }
 
+  public <C> C withContainer(final C container) {
```
smola, line 244: I would add some Javadoc here explaining what it does, and that we'll consider this mandatory for testcontainers on CI.

manuel-alvarez-alvarez: All this ugly code is now gone, using a docker api proxy that is able to create the forwards on demmand

.circleci/config.yml
```diff
@@ -89,42 +89,43 @@ commands:
...
+            echo extendedKeyUsage = clientAuth > extfile-client.cnf
+            openssl x509 -req -days 365 -sha256 -in client.csr -CA ca.pem -CAkey ca-key.pem -CAcreateserial -out cert.pem -extfile extfile-client.cnf
+            rm -v client.csr server.csr extfile.cnf extfile-client.cnf
+            chmod -v 0400 ca-key.pem key.pem server-key.pem
+            chmod -v 0444 ca.pem server-cert.pem cert.pem
+            
+            cd ~/dd-trace-java
+            # The update is temporal until autossh proves to work fine
+            sudo apt update
+            sudo apt install -y autossh
```
bantonsson, line 127: Shouldn't this be installed onto the [`dd-trace-java-docker-build`](https://github.com/DataDog/dd-trace-java-docker-build) image?

manuel-alvarez-alvarez: Yep, but I wasn't sure about how updates to the base image are handled.

My idea was to get this PR approved and before merging it, do the modification to the base image and remove the apt update and install from this PR. Is it OK like this or should I modify the image first?

Thanks for the feedback!

bantonsson: So what you can do is open a PR with the changes in [`dd-trace-java-docker-build`](https://github.com/DataDog/dd-trace-java-docker-build), and then run test of this PR with that image by changing to the image in this PR to `datadog/dd-trace-java-docker-build:{branch_name}`.

dd-java-agent/instrumentation/rabbitmq-amqp-2.7/src/test/groovy/RabbitMQTest.groovy
```diff
@@ -35,9 +35,8 @@ import static datadog.trace.agent.test.utils.TraceUtils.runUnderTrace
 import static datadog.trace.api.config.TraceInstrumentationConfig.RABBIT_PROPAGATION_DISABLED_EXCHANGES
 import static datadog.trace.api.config.TraceInstrumentationConfig.RABBIT_PROPAGATION_DISABLED_QUEUES
 
-// Do not run tests locally on Java7 since testcontainers are not compatible with Java7
-// It is fine to run on CI because CI provides rabbitmq externally, not through testcontainers
-@Requires({ "true" == System.getenv("CI") || jvm.java8Compatible })
+// Do not run tests on Java7 since testcontainers are not compatible with Java7
+@Requires({ jvm.java8Compatible })
```
bantonsson, line 39: It's fine to skip these tests on Java 7 to avoid spinning up the RabbitMQ container on the side

bantonsson (approved):
I can't remember when I read this much Python 😅

.circleci/autoforward.py
```diff
@@ -1,147 +1,734 @@
...
+  parser.add_argument('-rca', '--remote-ca', dest='remote_ca', help='CA for the remote endpoint')
+  parser.add_argument('--verbose', dest='verbose', action='store_true')
+  parser.set_defaults(verbose=False)
+  return parser.parse_args()
+
+
+class TcpUpgrade:
+  """
+  Docker remote API defines what is called a connection Hijacking via "Connection: upgrade" and "Upgrade: tcp" headers
+  This class tries to detect a tcp upgrade handshake and keep the sockets open until the remote decides to close the
```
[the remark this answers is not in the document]
manuel-alvarez-alvarez: Yep, poor wording my side, i'll rewrite it to _remote docker API endpoint_

dd-java-agent/instrumentation/rabbitmq-amqp-2.7/src/test/groovy/RabbitMQTest.groovy
manuel-alvarez-alvarez: I tested it quite a bit and by changing from `GenericContainer` to the more specific `RabbitMQContainer` with its own wait strategies, the tests are way more robust, so no more RabbitMQ container on the side :smile:

.circleci/config.yml
manuel-alvarez-alvarez: I've removed autossh for now, when I get access to the [dd-trace-java-docker-build](https://github.com/DataDog/dd-trace-java-docker-build) I'll create an update PR and revisit it here.

bantonsson: @manuel-alvarez-alvarez You should have write access now

PR #3752, opened 2022-08-24
Allow some tags to not be used for pathway hash computations

hokitam:
# What Does This Do
This change adds the concept of "hashable tags" for pathway propagation, where only a subset of the tags for each checkpoint would be used towards computing the hash for that pathway. In particular:
- the `List<String>` of tags that we have been passing around in `AgentPropagation`, `PathwayContext`, and their implementing classes have been updated to a `LinkedHashMap<String, String>`. A list of tags that previously would have been passed as `List{"type:kafka", "topic:my-topic", ...}`, for example, will now be `Map{"type": "kafka", "topic": "my-topic", ...}`. This allows us to quickly check each tag against a centralized set of "hashable tags" to determine whether the tag should be used to compute the pathway hash, without needing to split the tags back up. We are using a LinkedHashMap so that the tags can remain sorted. (The list of tags currently were not sorted in the Java implementation, but is sorted in the Golang implementation, so using a LinkedHashMap will allow us to also fix that discrepancy). See `AgentPropagation.java`, `AgentTracer.java`, `PathwayContext.java`, `DefaultPathwayContext.java`, and `CoreTracer.java`.
- A new `TagsProcessor` class has been added and replaces the `TagsCache` classes that have been scattered across individual instrumentations that were previously added for string concatenation optimization. The `TagsProcessor` class contains the list of supported tags, along with caches to compute or fetch the final string tags to be added to the pathway propagation checkpoints. See `TagsProcessor.java` and the deleted `TagsCache.java` classes.
- The existing instrumentations that have been enabled for pathway propagation (Kafka, Kafka Streams, gRPC, and RabbitMQ) have all been updated to send tags as `LinkedHashMaps` instead of `Lists`. See `grpc-1.5/...`, `rabbitmq-amqp-2.7/...`, `kafka-clients-0.11/...`, `kafka-streams-0.11/...`.
- This change also adds a new partition tag for the Kafka tags propagation, and also starts sending the `topic` tag for Kafka producers in addition to consumers. See `KafkaProducerInstrumentation.java` and `TracingIterator.java`.

# Additional Notes
Testing: tested with demo rabbitmq, kafka, and grpc apps. Confirmed pathway tags still showing up as expected on relevant metrics.


dd-java-agent/instrumentation/grpc-1.5/src/main/java/datadog/trace/instrumentation/grpc/client/GrpcClientDecorator.java
```diff
@@ -11,16 +12,24 @@
...
+import java.util.LinkedHashMap;
 import java.util.Set;
 
 public class GrpcClientDecorator extends ClientDecorator {
   public static final CharSequence GRPC_CLIENT = UTF8BytesString.create("grpc.client");
   public static final CharSequence COMPONENT_NAME = UTF8BytesString.create("grpc-client");
   public static final CharSequence GRPC_MESSAGE = UTF8BytesString.create("grpc.message");
-  public static final List<String> PRODUCER_PATHWAY_EDGE_TAGS = Arrays.asList("type:internal");
+
+  private static final LinkedHashMap<String, String> createClientPathwaySortedTags() {
```
bantonsson, line 24: Maybe a bit of a nitpick, but the entries in a `LinkedHashMap` are not _sorted_, but _ordered_, and default iteration order is insertion order. I find calling these _sorted_ in the APIs and variables is confusing since there are no sorting guarantees enforced.

Or if the expectation of the APIs, i.e. `setDataStreamCheckpoint`, are that the iteration order of the tags in the `Map` will be sorted, then there should be a comment on the method stating that.

dd-trace-core/src/main/java/datadog/trace/core/datastreams/TagsProcessor.java
```diff
@@ -0,0 +1,80 @@
...
+    result.put(GROUP_TAG, GROUP_TAG_PREFIX);
+    result.put(EXCHANGE_TAG, EXCHANGE_TAG_PREFIX);
+    result.put(HAS_ROUTING_KEY_TAG, HAS_ROUTING_KEY_TAG_PREFIX);
+    return result;
+  }
+
+  public static final String createTag(String tagKey, String tagValue) {
+    if (TAG_TO_CACHE.containsKey(tagKey) && TAG_TO_PREFIX.containsKey(tagKey)) {
+      return TAG_TO_CACHE.get(tagKey).computeIfAbsent(tagValue, TAG_TO_PREFIX.get(tagKey));
+    }
```
bantonsson, line 77: The `containsKey` followed by `get` will give you double hash table lookups. Instead just look up the values and to `null` checks on the variables.

dd-trace-core/src/main/java/datadog/trace/core/datastreams/DefaultPathwayContext.java
```diff
@@ -83,16 +97,23 @@ public void setCheckpoint(List<String> tags, Consumer<StatsPoint> pointConsumer)
...
-      long newHash = generatePathwayHash(edgeTags, hash);
+      for (Map.Entry<String, String> entry : sortedTags.entrySet()) {
+        String tag = TagsProcessor.createTag(entry.getKey(), entry.getValue());
+        if (hashableTagKeys.contains(entry.getKey())) {
+          hashableTags.add(tag);
+        }
+        allTags.add(tag);
+      }
+
+      long newHash = generatePathwayHash(hashableTags, hash);
```
bantonsson, line 109: Building up the `hashableTags` array list just to send it into the hash function feels like double work. One idea could be to have a `PathwayHashBuilder` that wraps the initialization of the `StringBuilder` in `generateNodeHash` that you can feed in the hashable tags into, appending them to the string builder, and then compute the hash in the end.

dd-trace-core/src/main/java/datadog/trace/core/datastreams/TagsProcessor.java
hokitam: thanks for the suggestion! Done.

dd-java-agent/instrumentation/grpc-1.5/src/main/java/datadog/trace/instrumentation/grpc/client/GrpcClientDecorator.java
hokitam: Indeed we want the input tags to be sorted, since it's more expensive to sort them after the fact. Added comments to the interfaces of the 3 methods that takes the sortedTags as parameters!

dd-trace-core/src/main/java/datadog/trace/core/datastreams/DefaultPathwayContext.java
hokitam: Thank you, great idea! Done

bantonsson (approved):

PR #3753, opened 2022-08-25
Allow the 'start-force-first' option on newer JDKs

jbachorik:
# What Does This Do
It turns the `profiling.start-force-first` option from experimental to a regular one.

# Motivation
After https://bugs.openjdk.org/browse/JDK-8233197 fixed we have a good choice of JDKs where this option would be safely supported. Allowing the option with a safety-break when we detect an unsupported JDK seems like the right way to go.

# Additional Notes
The PR also contains some additional cleanups discovered while working on the main feature.

### DIFF
--- a/dd-java-agent/agent-profiling/src/main/java/com/datadog/profiling/agent/ProfilingAgent.java
+++ b/dd-java-agent/agent-profiling/src/main/java/com/datadog/profiling/agent/ProfilingAgent.java
@@ -46,9 +46,15 @@ public class ProfilingAgent {
       final Config config = Config.get();
       final ConfigProvider configProvider = ConfigProvider.getInstance();
 
-      final boolean startForceFirst =
+      boolean startForceFirst =
           configProvider.getBoolean(
               PROFILING_START_FORCE_FIRST, PROFILING_START_FORCE_FIRST_DEFAULT);
+
+      if (!isStartForceFirstSafe()) {
+        log.debug(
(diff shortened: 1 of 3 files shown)


luhenry (approved):

dd-java-agent/agent-profiling/src/main/java/com/datadog/profiling/agent/ProfilingAgent.java
```diff
@@ -119,6 +125,13 @@ public static synchronized void run(final boolean isStartingFirst, ClassLoader a
     }
   }
 
+  private static boolean isStartForceFirstSafe() {
+    return Platform.isJavaVersionAtLeast(14)
+        || (Platform.isJavaVersion(13) && Platform.isJavaVersionAtLeast(13, 0, 4))
+        || (Platform.isJavaVersion(11) && Platform.isJavaVersionAtLeast(11, 0, 8))
+        || (Platform.isJavaVersion(8) && Platform.isJavaVersion(8, 0, 272));
```
bantonsson, line 132: Shouldn't the Java 8 one be `Platform.isJavaVersionAtLeast(8, 0, 272)`, or is this only available in that specific version?

jbachorik: 🤦 Of course :(
Thanks for noticing this! https://github.com/DataDog/dd-trace-java/pull/3758

PR #3754, opened 2022-08-25
Compile for Java 8, and don't test on Java 7

bantonsson:
# What Does This Do
 
Changes source and destination target to Java 8 and removes the Java 7 test job from CircleCI

# Motivation

Version 1.x of dd-trace-java will have Java 8 as a minimum so we can upgrade dependencies

# Additional Notes

Upgraded dependencies and other changes in the build will follow in subsequent PRs

### DIFF
--- a/dd-java-agent/instrumentation/instrumentation.gradle
+++ b/dd-java-agent/instrumentation/instrumentation.gradle
@@ -39,6 +39,7 @@ subprojects { Project subProj ->
 
   subProj.afterEvaluate {
     String jdkCompile = null
+    // TODO:java8 this should be changed and the *_java8 directories folded into the main directory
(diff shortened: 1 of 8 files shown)


devinsba (approved):
:+1:

dd-java-agent/instrumentation/instrumentation.gradle
```diff
@@ -39,6 +39,7 @@ subprojects { Project subProj ->
 
   subProj.afterEvaluate {
     String jdkCompile = null
+    // TODO:java8 this should be changed and the *_java8 directories folded into the main directory
```
PerfectSlayer, line 42: I guess we are not moving the source code  folder yet to keep the `project/java8` branch easier to update?

PerfectSlayer (approved):

dd-java-agent/instrumentation/instrumentation.gradle
bantonsson: yes, that is correct

PR #3757, opened 2022-08-26
Add release update workflow

PerfectSlayer:
# What Does This Do

This action will update the `download-latest-vX` releases body and attachment to the latest version available.

# Motivation

This action takes place in the Java 7 deprecation context.

We will need to create releases with stable tags in order to provide permanent links for download each major version.
Those releases will be based on orphan branch tagged `download-latest-vX` and will contains a copy of the latest attachment available.

On each new release, this action will update this meta release. It can also be trigger manually if needed, proving the major version to update.

# Additional Notes

Most of the logic rely on GitHub script actions and octokit library.
I wonder how to improve readability and maintenance. Feel free to share any idea to improve it.

### DIFF
--- a/.github/workflows/update-latest-release.yml
+++ b/.github/workflows/update-latest-release.yml
@@ -186,14 +186,31 @@ jobs:
             }).catch(error => {
               core.setFailed('Failed to find release: ' + error.message);
             });
+            // Get the current release asset
             const assetFile = 'dd-java-agent.jar';
             const newAssetFile = 'dd-java-agent-${{ steps.check-update.outputs.tag }}.jar';
+            const asset = await github.rest.repos.listReleaseAssets({
+              owner: context.repo.owner,
+              repo: context.repo.repo,
+              release_id: release.id
+            }).then(response => {
+              if (response.status === 200) {
+                for (const asset of response.data) {
+                  if (asset.name === assetFile) {
+                    return asset;
+                  }
+                }
+                return false;
+              } else {
+                core.setFailed('Failed to list release assets.');
+              }
+            });
             // Upload the new version to the release
             const newAsset = await github.rest.repos.uploadReleaseAsset({
               owner: context.repo.owner,
               repo: context.repo.repo,
               release_id: release.id,
-              name: newAssetFile,
+              name: asset ? newAssetFile : assetFile,
               data: fs.readFileSync(assetFile)
             }).then(response => {
               if (response.status === 201) {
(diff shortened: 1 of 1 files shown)


.github/workflows/update-latest-release.yml
```diff
@@ -0,0 +1,250 @@
...
+  workflow_dispatch:
+    inputs:
+      version:
+        description: 'The version to force update'
+        required: true
+        type: 'choice'
+        options:
+        - 'v0'
+        - 'v1'
+        default: 'v1'
```
bantonsson, line 15: I think this will probably be `v0` in the beginning 😉

.github/workflows/update-latest-release.yml
```diff
@@ -0,0 +1,250 @@
...
+                    || (versions[major].minor == minor && versions[major].bugfix < bugfix) // Of if there is an older bugfix version
+                  ) {
+                    versions[major] = { tag, major, minor, bugfix };
+                  }
+                }
+                return versions;
+              }, {});
+              // Find "download latest" releases, indexed by major version
+              const latestReleases = releases.reduce((releases, release) => {
+                if (release.tag_name.match(downloadLatestPattern)) {
```
bantonsson, line 55: Should we use the `downloadLatestPattern` to pull out the `major` version to ensure that it's the correct one? That would also allow the _initial_ release to not have the version number in it, if we add a default `releases[major] = { tag, 0, 0, 0 }`.

PerfectSlayer: That is what I did at first. Then I got two majors, one from the release tag, another from the body content.
I will restore this behavior, compare them, and fail the execution if they differ.

PerfectSlayer: I now get major version from `download-latest-vX` release tag name and compare them to the tag I find its body.
Here is a capture of what happens if versions are mismatching, stoping the workflow execution:
Screenshot 2022-08-26 at 11 45 10

bantonsson: I looked at the updated code, and it feels like it would skip over `download-latest-Vx` release that doesn't have a `vx.y.z` line in it and complain that you should _create_ such a release, which would be confusing since it's obviously there, although with incomplete information in it. Can't we just set a default `releases[releaseMajor] = { tag, 0, 0, 0 }` to just overwrite it with the latest if it doesn't contain a version string?

PerfectSlayer: We can. But it this case, it would mean there is something off with the `download-latest-vX` release.
If it was created without a proper body and attachment, the default value will well trigger the update but will later failed with `Failed to find release asset id.` error when try to remove the original `dd-java-agent.jar`. Hence the error message to create it first.

Would you like me to support the case where the original attachment is missing? (skipping the delete asset call) So I can use the default value and update any empty `download-latest-vX` release?

bantonsson: Yes, I think it would be better to have something that will always try to store the new asset, maybe log after to show that there was something _weird_ with the `download-latest-vX` release.

PerfectSlayer: It is updated.

I keep thinking the logic of this action requires its proper place.
The code does not feel robust and manual testing is more and more difficult as use case increase.
If any other project need it, I would rather create a separate project for the action and write unit tests for.

.github/workflows/update-latest-release.yml
```diff
@@ -54,6 +54,11 @@ jobs:
                 const match = release.tag_name.match(downloadLatestPattern);
                 if (match) {
                   const releaseMajor = parseInt(match[1]);
+                  releases[releaseMajor] = {
+                    tag: 'v' + releaseMajor + '.0.0',
```
bantonsson, line 58: I'm getting confused by these tags. I thought the tag had the same form as the release name `download-latest-vX`.

.github/workflows/update-latest-release.yml
bantonsson: Should we break it out into a separate project right away then?

.github/workflows/update-latest-release.yml
PerfectSlayer: This tag is the tag of the related version (`vX.Y.Z`) contained in the release body, not the tag of the `download-release-vX` itself.

It allows me to compare the version included in the `download-release-vX` body to the one that trigger the build (or the latest available when manually triggered).

I wish I could create proper data structure 😞

PerfectSlayer:
Closed. Replaced by #3784

PR #3759, opened 2022-08-26
Introduce a hard limit for in-flight spans with tracked context

jbachorik:
# What Does This Do
This is adding another config knob for the maximum number of in-flight spans which will be tracked by the per-span context tracking mechanism.
Once the limit is reached the per-span context tracking will be automatically disabled for the application as there is a good chance the application is rapidly generating high number of spans and tracking them would blow up the memory and cost a lot of CPU in terms of GC and maintaining the delay queues. This kind of application is simply not possible to use with the per-span context tracking and therefore turning it off as soon as possible is the best thing we can do.

# Motivation
There are applications out there generating 10s of 1000s of spans per second. This is putting a huge strain on the per-span context tracking because we need to keep extra data attached to each tracked span - both in terms of extra memory and extra CPU cycles. In the end the profiled/traced application can experience overhead of several 100s of percent which is definitely not something we want to see. In such cases it is better to turn off the tracking as soon as we detect a possible problem.

# Additional Notes

### DIFF
--- a/dd-java-agent/agent-profiling/profiling-context/src/main/java/com/datadog/profiling/context/PerSpanTracingContextTrackerFactory.java
+++ b/dd-java-agent/agent-profiling/profiling-context/src/main/java/com/datadog/profiling/context/PerSpanTracingContextTrackerFactory.java
@@ -73,14 +73,15 @@ public final class PerSpanTracingContextTrackerFactory
           int drainedAll = 0;
           int drained = 0;
           int queue = 0;
-          while (queue < target.length
-              && (drained = target[queue++].drainTo(timeouts, capacity)) > 0) {
-            if (log.isDebugEnabled()) {
-              log.debug("Drained {} inactive trackers", drained);
+          while (queue < target.length) {
+            while ((drained = target[queue++].drainTo(timeouts, capacity)) > 0) {
+              if (log.isDebugEnabled()) {
+                log.debug("Drained {} inactive trackers", drained);
+              }
+              timeouts.forEach(TracingContextTracker.DelayedTracker::cleanup);
+              timeouts.clear();
+              drainedAll += drained;
             }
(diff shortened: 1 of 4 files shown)


dd-java-agent/agent-profiling/profiling-context/src/main/java/com/datadog/profiling/context/PerSpanTracingContextTrackerFactory.java
```diff
@@ -109,6 +120,16 @@ private void initializeInactiveTrackerCleanup(long checkPeriodMs) {
             ? Allocators.directAllocator(reservedMemorySize, 32)
             : Allocators.heapAllocator(reservedMemorySize, 32);
     this.inactivityDelay = inactivityDelayNs;
+    int numDelayQueues = Math.min(Math.max(Runtime.getRuntime().availableProcessors() / 2, 2), 8);
```
bantonsson, line 123: The PR description only talks about the in-flight spans, but there is a split of queues over thread id in here as well. I assume that is on purpose.

bantonsson (approved):

dd-java-agent/agent-profiling/profiling-context/src/main/java/com/datadog/profiling/context/PerSpanTracingContextTrackerFactory.java
jbachorik: Ah, sorry. This came up during stabilisation - the reason is that after a certain number of elements the re-heaping of the delay queue becomes quite expensive. Also, the split might help to reduce contention.

bantonsson: I'm all for the split. Just wanted to make sure. 👍🏻

PR #3760, opened 2022-08-26
Implement weak cipher detection

DDJavierSantos:
# What Does This Do
Instrument the javax.crypto.Cipher class so we can detect and report the use of weak cipher algorihtms

# Motivation
This will allow us to detect and report vulnerabilities in crypto code when weak cipher algorithms are used in customer code. 

# Additional Notes
In order for the functionality to be complete, we need the backend to filter and display the vulnerability information we add to the span.


bantonsson (approved):

dd-java-agent/instrumentation/java-security/src/test/groovy/test/UnavailableHashAlgorithmTest.groovy
```diff
@@ -0,0 +1,22 @@
...
+import datadog.trace.agent.test.AgentTestRunner
+
+import java.security.MessageDigest
+import java.security.NoSuchAlgorithmException
+
+import static datadog.trace.agent.test.utils.TraceUtils.runUnderTrace
+
+class UnavailableHashAlgorithmTest extends AgentTestRunner {
+
+  def "test instrumentation"() {
```
smola, line 12: @DDJavierSantos Can you move these different test cases to a single suite (e.g. `WeakCipherInstrumentationTest`)?

smola (changes requested):

smola (approved):

PR #3761, opened 2022-08-26
Extract message and kind from native Java crash

luhenry:
# What Does This Do

This allows to integrate crash tracking to error tracking product by extracting a message and a kind. We also want to extract a stacktrace.

### DIFF
--- a/dd-java-agent/agent-crashtracking/src/main/java/com/datadog/crashtracking/CrashUploader.java
+++ b/dd-java-agent/agent-crashtracking/src/main/java/com/datadog/crashtracking/CrashUploader.java
@@ -20,8 +20,8 @@ import java.io.InputStreamReader;
 import java.io.PrintStream;
 import java.nio.charset.StandardCharsets;
 import java.time.Instant;
-import java.util.Arrays;
 import java.util.ArrayList;
+import java.util.Arrays;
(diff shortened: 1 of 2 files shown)


dd-java-agent/agent-crashtracking/src/main/java/com/datadog/crashtracking/CrashUploader.java
```diff
@@ -142,6 +151,93 @@ void uploadToLogs(@Nonnull List<String> filesContent, @Nonnull PrintStream out)
...
+          Pattern.DOTALL | Pattern.MULTILINE);
+
+  private String extractErrorStackTrace(String fileContent) {
+    // Matcher matcher = errorStackTracePattern.matcher(fileContent);
+    // if (!matcher.find()) {
+    //   System.err.println("No match found for error.stack");
+    //   return null;
+    // }
+    // return matcher.group();
+    return null;
```
jbachorik, line 238: Is this going to be implemented now?
If not, you could just note down a TODO here and remove the commented out lines.

luhenry: Fixed.

jbachorik (approved):

PR #3769, opened 2022-08-30
Fix issue in default stack walker implementation

smola:
# What Does This Do
`Thread.currentThread().getStackTrace()` seems to be the same as `new
Throwable.getStackTrace()` for the general case, but it results in an
additional frame. This might result in a slight performance loss, and
also was the source of a bug where we got an unexpected result.

Refactored stack walker tests to spot this kind of issue.

# Motivation

# Additional Notes

This is a follow up to https://github.com/DataDog/dd-trace-java/pull/3647

### DIFF
--- a/internal-api/internal-api-8/src/main/java/datadog/trace/util/stacktrace/DefaultStackWalker.java
+++ b/internal-api/internal-api-8/src/main/java/datadog/trace/util/stacktrace/DefaultStackWalker.java
@@ -15,6 +15,6 @@ public class DefaultStackWalker extends AbstractStackWalker {
 
   @Override
   <T> T doGetStack(final Function<Stream<StackTraceElement>, T> consumer) {
-    return consumer.apply(Arrays.stream(Thread.currentThread().getStackTrace()));
+    return consumer.apply(Arrays.stream(new Throwable().getStackTrace()));
   }
 }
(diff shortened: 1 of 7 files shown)


internal-api/internal-api-8/src/main/java/datadog/trace/util/stacktrace/DefaultStackWalker.java
```diff
@@ -15,6 +15,6 @@ public boolean isEnabled() {
 
   @Override
   <T> T doGetStack(final Function<Stream<StackTraceElement>, T> consumer) {
-    return consumer.apply(Arrays.stream(Thread.currentThread().getStackTrace()));
+    return consumer.apply(Arrays.stream(new Throwable().getStackTrace()));
```
smola, line 18: For reviewers: this is the actual fix. The rest is just all testing refactor I did to find the issue.

internal-api/internal-api-8/src/test/java/datadog/trace/util/stacktrace/DefaultStackWalkerTest.java
```diff
@@ -1,66 +1,47 @@
...
 
   @Test
   public void defaultStackWalker_must_be_enabled() {
-    assertTrue(defaultStackWalker.isEnabled());
+    assertTrue(stackWalker.isEnabled());
   }
 
   @Test
-  public void walk_retrieves_stackTraceElements() {
+  public void walk_from_non_datadog_class() {
```
smola, line 22: For reviewers: This is the test case that reproduced the issue.

manuel-alvarez-alvarez (approved):

internal-api/internal-api-8/src/main/java/datadog/trace/util/stacktrace/DefaultStackWalker.java
bantonsson: Yup, you would get the additional `Thread.getStackTrace()` frame.

PR #3770, opened 2022-08-30
Avoid warning when using StackWalker on JDK <9

smola:
# What Does This Do
If we cannot instantiate JDK9StackWalker, a warning is printed, which is
fair if we're on JDK 9+, but pointless otherwise. Check the JDK version
for trying to instantiate StackWalker implementations, to avoid
unnecessary noise.

# Motivation

# Additional Notes

cc @robertomonteromiguel this removes the warning you found

### DIFF
--- a/internal-api/internal-api-8/src/main/java/datadog/trace/util/stacktrace/StackWalkerFactory.java
+++ b/internal-api/internal-api-8/src/main/java/datadog/trace/util/stacktrace/StackWalkerFactory.java
@@ -1,5 +1,6 @@
 package datadog.trace.util.stacktrace;
 
+import datadog.trace.api.Platform;
 import de.thetaphi.forbiddenapis.SuppressForbidden;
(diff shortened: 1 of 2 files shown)


internal-api/internal-api-8/src/main/java/datadog/trace/util/stacktrace/StackWalkerFactory.java
```diff
@@ -29,11 +30,17 @@ private static Supplier<StackWalker> defaultStackWalker() {
   }
 
   private static Supplier<StackWalker> hotspot() {
+    if (!Platform.isJavaVersion(8)) {
+      return null;
+    }
```
bantonsson, line 35: Makes sense to check for Java 8 here to avoid creating the instance, and not only in the `isEnabled` method.

bantonsson (approved):

PR #3775, opened 2022-08-31
Handle invalid URL at HTTP client creation

PerfectSlayer:
# What Does This Do

A quick change in HTTP client builder in order to support `null` URL.
Some URL might be null when using `Http.parse` instead of `HttpUrl.get` and break the client creation while testing for URL scheme.

# Motivation

# Additional Notes


jpbempel (approved):
LGTM

bantonsson (approved):

PR #3776, opened 2022-08-31
Dependency detection for multi-pom fat jars

anderruiz:
# What Does This Do

This PR adds support for third party library detection when a fat jar is used but which keeps pom.properties for all dependencies inside the final file. In this particular case it was testes for dropwizard generated jar files, but it should work for any project using maven-shade-plugin in traditional scenarios. In case more than one pom.properties is found inside the jar more than one dependency will be created.

# Motivation

Improve support for third party library detection in fatjar cases. This PR solves part of the cases (when the maven information is still present for those dependencies) but not all of them


ValentinZakharov (approved):

bantonsson (commented):
It would be good to have instructions for how to generate the new test  jar files in the `README`, like for the `spring-boot-app.jar`

anderruiz:
Include some readme instructions about how some test jars were created

bantonsson (approved):

PR #3782, opened 2022-09-01
Add autossh to maintain forwards for testcontainers

manuel-alvarez-alvarez:
# What Does This Do
Use autossh to ensure that SSH forwards for testcontainers are kept running during the builds on CI

# Motivation
If the tunnel breaks for any reason it can cause test to fail

### DIFF
--- a/.circleci/autoforward.py
+++ b/.circleci/autoforward.py
@@ -466,12 +466,13 @@ class PortForwarder:
     """
     Tries to forward via SSH the port that is booked by the local socket
     """
-    if self._is_forwarding() or not self.socket:
+    if self._is_forwarding():
       return
 
     if self._is_remote_port_listening():
-      self.socket.close()
-      self.socket = None
+      if self.socket:
+        self.socket.close()
+        self.socket = None
       try:
         self.process = self._forward_port()
         logging.info('SSH_FORWARD %s -> %s established', self.remote_port, self.local_port)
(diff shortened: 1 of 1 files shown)


.circleci/autoforward.py
```diff
@@ -466,12 +466,13 @@ def forward(self):
...
     """
-    if self._is_forwarding() or not self.socket:
+    if self._is_forwarding():
       return
 
     if self._is_remote_port_listening():
-      self.socket.close()
-      self.socket = None
+      if self.socket:
+        self.socket.close()
```
manuel-alvarez-alvarez, line 474: This might happens if the autossh procces exits during its execution, it's unlikely but it's better to take it into account

bantonsson (approved):

.circleci/config.yml
```diff
@@ -3,7 +3,7 @@ version: 2.1
 defaults: &defaults
   working_directory: ~/dd-trace-java
   docker:
-    - image: &default_container datadog/dd-trace-java-docker-build:latest
+    - image: &default_container datadog/dd-trace-java-docker-build:malvarez_add-autossh
```
smola, line 6: I guess this will be changed before merge, once https://github.com/DataDog/dd-trace-java-docker-build/pull/39 is merged, but approved otherwise.

smola (approved):

.circleci/config.yml
manuel-alvarez-alvarez: Yes, it's changed now :smile:

PR #3807, opened 2022-09-07
Fix weak cipher vulnerability detection

anderruiz:
# What Does This Do

Fix weak cipher vulnerability detection to keep into account algorithm variants (i.e. padding usage)

# Motivation

The current version was only detecting simple algorithms, but not their variants, which is the most traditional way of configuring a cipher. This PR handles that issue


dd-java-agent/iast/src/test/groovy/com/datadog/iast/IastModuleImplCipherTest.groovy
```diff
@@ -27,30 +27,32 @@ class IastModuleImplCipherTest extends IastModuleImplTestBase {
...
-    "PBEWithSHA1AndRC4_128" | _
-    "PBEWithHmacSHA1AndAES_128" | _
+    "RC2"                         | _
+    "Blowfish"                    | _
+    "ARCFOUR"                     | _
+    "DESedeWrap"                  | _
+    "PBEWithSHA1AndRC2_128"       | _
+    "PBEWithSHA1AndRC4_40"        | _
+    "PBEWithSHA1AndRC4_128"       | _
+    "PBEWithHmacSHA1AndAES_128"   | _
```
bantonsson, line 55: I can't see any _negative_ tests here, so how do you know what the matcher matches?

anderruiz: Negative tests were already covered before for in "iast module not blocklisted cipher algorithm", in the end, anything that does not match is not vulnerable. The regex is directly an automatically generated regex from the original values (the ones that were in the Set) applying an | (OR) and collapsing it to make slightly faster

bantonsson: Apparently Wednesday is the new Monday and I couldn't read the whole test file. Ignore my comment.

bantonsson (approved):

dd-trace-api/src/main/java/datadog/trace/api/ConfigDefaults.java
```diff
@@ -81,32 +81,8 @@ public final class ConfigDefaults {
...
-              "PBEWITHHMACSHA1ANDAES_256",
-              "PBEWITHHMACSHA224ANDAES_256",
-              "PBEWITHHMACSHA256ANDAES_256",
-              "PBEWITHHMACSHA384ANDAES_256",
-              "PBEWITHHMACSHA512ANDAES_256",
-              "RC2",
-              "BLOWFISH",
-              "ARCFOUR"));
+  static final String DEFAULT_IAST_WEAK_CIPHER_ALGORITHMS =
+      "^(?:PBEWITH(?:HMACSHA(?:2(?:24ANDAES_(?:128|256)|56ANDAES_(?:128|256))|384ANDAES_(?:128|256)|512ANDAES_(?:128|256)|1ANDAES_(?:128|256))|SHA1AND(?:RC(?:2_(?:128|40)|4_(?:128|40))|DESEDE)|MD5AND(?:TRIPLEDES|DES))|DES(?:EDE(?:WRAP)?)?|BLOWFISH|ARCFOUR|RC2).*$";
```
smola, line 85: @anderruiz If we want to be more strict, we can end the regexp with `/.*$` instead of `.*$` to check only for variants as specified with slashes. But the current regexp is also fine with me.

smola (approved):

internal-api/src/main/java/datadog/trace/api/Config.java
```diff
@@ -1087,10 +1088,9 @@ && isJavaVersionAtLeast(8)
     iastWeakHashAlgorithms =
         tryMakeImmutableSet(
             configProvider.getSet(IAST_WEAK_HASH_ALGORITHMS, DEFAULT_IAST_WEAK_HASH_ALGORITHMS));
-    iastWeakCipherAlgorithms =
-        tryMakeImmutableSet(
-            configProvider.getSet(
-                IAST_WEAK_CIPHER_ALGORITHMS, DEFAULT_IAST_WEAK_CIPHER_ALGORITHMS));
+    String pattern =
+        configProvider.getString(IAST_WEAK_CIPHER_ALGORITHMS, DEFAULT_IAST_WEAK_CIPHER_ALGORITHMS);
+    iastWeakCipherAlgorithms = Pattern.compile(pattern);
```
smola, line 1093: @anderruiz Hm. Here we should guard against crashes if the client specifies an invalid regexp.

smola (changes requested):

internal-api/src/main/java/datadog/trace/api/Config.java
anderruiz: Done, It did not include this time, but perhaps we should integrate Hdiv bounded pattern matching

dd-trace-api/src/main/java/datadog/trace/api/ConfigDefaults.java
anderruiz: Done, it makes sense

smola (approved):

PR #3809, opened 2022-09-07
[WIP] getparameter instrumentation

DDJavierSantos:
# What Does This Do
Use the call site plugin to instrument ServletRequest.getParameter, ServletRequest.getParameterValues and ServletRequest.getParameterNames. The instrumented code marks the parameter values and names as tainted so we can track them through code execution to the point where we can detect vulnerabilities. 

# Motivation
The instrumented code marks the parameter values and names as tainted so we can track them through code execution to the point where we can detect vulnerabilities. The goal is to implement SQL injection vulnerabilities.

# Additional Notes
This PR uses a mock object tainter for testing purposes until the real tainter is merged.


dd-java-agent/instrumentation/servlet/request-2/request-2.gradle
```diff
@@ -27,6 +28,9 @@ testSets {
 dependencies {
   compileOnly group: 'javax.servlet', name: 'servlet-api', version: '2.2'
 
+  implementation group: 'net.bytebuddy', name: 'byte-buddy', version: '1.12.12'
```
manuel-alvarez-alvarez, line 31:
```suggestion
  implementation deps.bytebuddy
```

dd-java-agent/instrumentation/servlet/request-2/request-2.gradle
```diff
@@ -27,6 +28,9 @@ testSets {
...
+  implementation project(':dd-java-agent:iast')
```
manuel-alvarez-alvarez, line 32: The only class from the iast module that is needed is the marker interface `com.datadog.iast.IastAdvice`, I think it will be better to put it in the `internal-api` module,  then there's no need to add any extra dependencies

dd-java-agent/instrumentation/servlet/request-2/src/test/java/com/jsantos/test/TestSuite.java
```diff
@@ -0,0 +1,24 @@
+package com.jsantos.test;
```
manuel-alvarez-alvarez, line 1:
```suggestion
package foo.bar.test;
```

manuel-alvarez-alvarez (changes requested):

PR #3811, opened 2022-09-08
Taint tracking API

smola:
# What Does This Do
Add taint tracking API
    
Includes:
    
- A hash map implementation optimized to store tainted objets. This is a
  weak identity hash map, with fixed capacity, some concurrency
  guarantees (it can lose concurrent writes, but not throw), and
  self-purging.
- Callbacks to taint HTTP parameter names and values (in taint tracking
   jargon: sources), and propagate taints on string concatenation.

# Motivation
This is one of the main building blocks to do taint tracking for IAST:

# Additional Notes

### DIFF
--- a/dd-java-agent/iast/src/jmh/java/com/datadog/iast/taint/TaintedMapEmptyBenchmark.java
+++ b/dd-java-agent/iast/src/jmh/java/com/datadog/iast/taint/TaintedMapEmptyBenchmark.java
@@ -1,0 +1,44 @@
+package com.datadog.iast.taint;
+
+import static java.util.concurrent.TimeUnit.MILLISECONDS;
+import static java.util.concurrent.TimeUnit.NANOSECONDS;
+
+import org.openjdk.jmh.annotations.Benchmark;
+import org.openjdk.jmh.annotations.BenchmarkMode;
+import org.openjdk.jmh.annotations.Fork;
+import org.openjdk.jmh.annotations.Level;
+import org.openjdk.jmh.annotations.Measurement;
+import org.openjdk.jmh.annotations.Mode;
+import org.openjdk.jmh.annotations.OperationsPerInvocation;
+import org.openjdk.jmh.annotations.OutputTimeUnit;
+import org.openjdk.jmh.annotations.Scope;
+import org.openjdk.jmh.annotations.Setup;
+import org.openjdk.jmh.annotations.State;
+import org.openjdk.jmh.annotations.Warmup;
+import org.openjdk.jmh.infra.Blackhole;
+
+@Warmup(iterations = 2, time = 1000, timeUnit = MILLISECONDS)
+@Measurement(iterations = 3, time = 1000, timeUnit = MILLISECONDS)
+@Fork(3)
+@OutputTimeUnit(NANOSECONDS)
+@BenchmarkMode(Mode.AverageTime)
+@State(Scope.Benchmark)
+public class TaintedMapEmptyBenchmark {
+
+  private static final int OP_COUNT = 1024;
+  private TaintedMap map;
+  private final Object anyObject = new Object();
+
+  @Setup(Level.Iteration)
+  public void setup() {
+    map = new DefaultTaintedMap();
+  }
+
+  @Benchmark
+  @OperationsPerInvocation(OP_COUNT)
+  public void getFromEmptyMap(final Blackhole bh) {
+    for (int i = 0; i < OP_COUNT; i++) {
+      bh.consume(map.get(anyObject));
+    }
+  }
+}
--- a/dd-java-agent/iast/src/main/java/com/datadog/iast/IastModuleImpl.java
+++ b/dd-java-agent/iast/src/main/java/com/datadog/iast/IastModuleImpl.java
@@ -2,10 +2,16 @@ package com.datadog.iast;
 
 import com.datadog.iast.model.Evidence;
 import com.datadog.iast.model.Location;
+import com.datadog.iast.model.Range;
+import com.datadog.iast.model.Source;
+import com.datadog.iast.model.SourceType;
 import com.datadog.iast.model.Vulnerability;
 import com.datadog.iast.model.VulnerabilityType;
 import com.datadog.iast.overhead.Operations;
 import com.datadog.iast.overhead.OverheadController;
+import com.datadog.iast.taint.Ranges;
+import com.datadog.iast.taint.TaintedObject;
+import com.datadog.iast.taint.TaintedObjects;
 import datadog.trace.api.Config;
 import datadog.trace.api.iast.IastModule;
 import datadog.trace.bootstrap.instrumentation.api.AgentSpan;
@@ -13,6 +19,7 @@ import datadog.trace.bootstrap.instrumentation.api.AgentTracer;
 import datadog.trace.util.stacktrace.StackWalker;
 import datadog.trace.util.stacktrace.StackWalkerFactory;
 import java.util.Locale;
+import javax.annotation.Nullable;
 
 public final class IastModuleImpl implements IastModule {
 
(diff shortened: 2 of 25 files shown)


dd-java-agent/iast/src/jmh/java/com/datadog/iast/taint/TaintedMapEmptyBenchmark.java
```diff
@@ -0,0 +1,44 @@
...
+  private static final int OP_COUNT = 1024;
+  private TaintedMap map;
+  private final Object anyObject = new Object();
+
+  @Setup(Level.Iteration)
+  public void setup() {
+    map = new DefaultTaintedMap();
+  }
+
+  @Benchmark
```
manuel-alvarez-alvarez, line 37: Will it be possible to have something to compare against as baseline?

dd-java-agent/iast/src/main/java/com/datadog/iast/taint/DefaultTaintedMap.java
```diff
@@ -0,0 +1,257 @@
...
+        return 1;
+      }
+    }
+    // If we reach this point, the entry was already removed or put was lost.
+    return 0;
+  }
+
+  @Override
+  public void clear() {
+    while (referenceQueue.poll() != null) {}
```
manuel-alvarez-alvarez, line 192: Why not throw away the current queue a create a new one?

dd-java-agent/iast/src/main/java/com/datadog/iast/taint/DefaultTaintedMap.java
```diff
@@ -0,0 +1,257 @@
...
+      final TaintedObject entry = (TaintedObject) ref;
+      removedCount += remove(entry);
+    }
+
+    if (estimatedSize.addAndGet(-removedCount) > flatModeThreshold) {
+      isFlat = true;
+    }
+
+    // Reset purging flag.
+    isPurging.set(false);
```
manuel-alvarez-alvarez, line 161: I'll rather `try/finally` this block of code (just in case)

dd-java-agent/iast/src/main/java/com/datadog/iast/taint/DefaultTaintedMap.java
```diff
@@ -0,0 +1,257 @@
...
+import java.util.concurrent.atomic.AtomicInteger;
+import javax.annotation.Nonnull;
+import javax.annotation.Nullable;
+
+/**
+ * Default implementation for {@link TaintedMap}. It is optimized for low concurrency scenarios, but
+ * never fail with high concurrency.
+ *
+ * <p><i>Capacity</i> is fixed, so there is no rehashing. Once it reaches a <i>flat mode
+ * threshold</i>, the table switches to flat mode. In this mode, every new put will be inserted at
```
manuel-alvarez-alvarez, line 18: Why don't switching to flat mode if buckets go over a certain threshold as well (we can do the check while iterating though a bucket)?

dd-java-agent/iast/src/main/java/com/datadog/iast/taint/TaintedMap.java
```diff
@@ -0,0 +1,30 @@
...
+public interface TaintedMap extends Iterable<TaintedObject> {
+
+  void put(@Nonnull TaintedObject to);
+
+  @Nullable
+  TaintedObject get(@Nonnull Object key);
+
+  void clear();
+
+  ReferenceQueue<Object> getReferenceQueue();
```
manuel-alvarez-alvarez, line 29: `TaintedMap` has ownership over a `TaintedObject` (a tainted object belongs to the tainted map used to its creation) and having this getter here looks like it's leaking implementation details to me. 

I would love to have a `TaintedMap` in the `TaintedObject` constructor instead of the reference queue, so we can make this getter go away.

utils/test-utils/src/main/java/datadog/trace/test/util/CircularBuffer.java
```diff
@@ -0,0 +1,47 @@
...
+        list.add(el);
+      }
+      iterIndex = (iterIndex + 1) & lengthMask;
+    }
+    return list.iterator();
+  }
+
+  @SuppressWarnings("unchecked")
+  private T[] createBuffer(final int capacity) {
+    return (T[]) new Object[capacity];
```
manuel-alvarez-alvarez, line 45: With `Array#newInstance` : https://docs.oracle.com/javase/7/docs/api/java/lang/reflect/Array.html#newInstance(java.lang.Class,%20int...) you can even create arrays of primitive types (doesn't matter for this particular example)

dd-java-agent/iast/src/test/groovy/com/datadog/iast/taint/TaintedMapTest.groovy
```diff
@@ -0,0 +1,203 @@
...
+        final tuples = new ArrayList<Tuple2<Object, TaintedObject>>(nObjectsPerThread)
+        for (int i = 1; i <= nObjectsPerThread; i++) {
+          final o = new Object()
+          final to = new TaintedObject(o, [] as Range[], map.getReferenceQueue())
+          tuples.add(new Tuple2<Object, TaintedObject>(o, to))
+          objectHolder.put(o, to)
+        }
+        startLatch.countDown()
+        startLatch.await()
+        tuples.each { map.put(it.get(1)) }
```
manuel-alvarez-alvarez, line 179: :vertical_traffic_light: nice one!

dd-java-agent/iast/src/main/java/com/datadog/iast/IastModuleImpl.java
```diff
@@ -71,4 +78,84 @@ public void onHashingAlgorithm(String algorithm) {
...
+      System.arraycopy(rangesLeft, 0, ranges, 0, rangesLeft.length);
+    }
+    if (rangesRight.length > 0) {
+      final int offset = left.length();
+      Ranges.copyShift(rangesRight, ranges, rangesLeft.length, offset);
+    }
+    taintedObjects.taint(result, ranges);
+  }
+
+  private boolean canBeTainted(final String s) {
```
manuel-alvarez-alvarez, line 158: Since this method will be called quite a bit we should make it static  :smile:

dd-java-agent/iast/src/main/java/com/datadog/iast/IastModuleImpl.java
```diff
@@ -71,4 +78,84 @@ public void onHashingAlgorithm(String algorithm) {
...
+    }
+    final TaintedObjects taintedObjects = ctx.getTaintedObjects();
+    taintedObjects.taintInputString(
+        paramValue, new Source(SourceType.REQUEST_PARAMETER_VALUE, paramName, paramValue));
+  }
+
+  @Override
+  public void onConcat(
+      @Nullable String left, @Nullable String right, final @Nullable String result) {
+    if (!canBeTainted(result)) {
```
manuel-alvarez-alvarez, line 114: I think we should move all this logic to separate analyzers (or whatever the name we wan't to use) otherwise this class is going to get huuuuge :cry:

manuel-alvarez-alvarez (approved):

dd-java-agent/iast/src/main/java/com/datadog/iast/taint/DefaultTaintedMap.java
smola: Right. I was thinking about some concurrency scenario, but it's probably pointless. I'll change it.

dd-java-agent/iast/src/main/java/com/datadog/iast/taint/DefaultTaintedMap.java
smola: @manuel-alvarez-alvarez I don't think we have a production use case for iterating all tainted objects?

dd-java-agent/iast/src/main/java/com/datadog/iast/taint/TaintedMap.java
smola: Makes sense.

utils/test-utils/src/main/java/datadog/trace/test/util/CircularBuffer.java
smola: @manuel-alvarez-alvarez `Array.newInstance` requires passing a `Class` object. Would you prefer adding the class object to the constructor here?

bantonsson (approved):

dd-java-agent/iast/src/main/java/com/datadog/iast/taint/DefaultTaintedMap.java
```diff
@@ -0,0 +1,259 @@
...
+      table[index] = cur.next;
+      return 1;
+    }
+    if (cur == null) {
+      return 0;
+    }
+    for (TaintedObject prev = cur.next; cur != null; prev = cur, cur = cur.next) {
+      if (cur == entry) {
+        prev.next = cur.next;
+        return 1;
```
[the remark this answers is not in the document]
smola: @bantonsson Note that `remove` is private and only called from `purge`, which is guarded against concurrent execution. So there is no concurrent removals (unless I missed something). There can be, however, a removal concurrent with a put.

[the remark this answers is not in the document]
bantonsson: Ok, so what happens then if we miss to remove the `TaintedObject` from the first position since `put` could _win_, i.e. be the last writer, and point to `cur` even though it has been removed, then we would never try to remove it again right?

dd-java-agent/iast/src/main/java/com/datadog/iast/taint/DefaultTaintedMap.java
manuel-alvarez-alvarez: We used to have it back in the day, but it's true that we should try to prevent it.

utils/test-utils/src/main/java/datadog/trace/test/util/CircularBuffer.java
manuel-alvarez-alvarez: Don't worry for this concrete case it does not matter.

dd-java-agent/iast/src/main/java/com/datadog/iast/taint/DefaultTaintedMap.java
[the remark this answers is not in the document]
smola: @bantonsson Oh, you're right. This case is tricky. It will not be removed, except if the tainted map grows enough to go above the `isFlat` threshold. If this happens frequently enough, there might be a slight performance degradation for `get` operations, as well as the memory overhead for these objects. However, I think it should not be a problem as long as the number of lost removes does not get too close to the number of puts, which could prevent going over the `isFlat` threshold, and let chains grow indefinitely.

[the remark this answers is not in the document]
bantonsson: I don't expect this to be a problem either, but it would be nice with a comment to help someone in the future if that behavior shows up in a support ticket.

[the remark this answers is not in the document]
smola: :+1: Done.

dd-java-agent/iast/src/main/java/com/datadog/iast/taint/DefaultTaintedMap.java
```diff
@@ -0,0 +1,272 @@
...
+    // removed again,
+    // (until the map goes to flat mode). When a remove is lost under concurrency, this method will
+    // still return 1, and
+    // it will still be subtracted from the map size estimate. If this is infrequent enough, this
+    // would lead to a
+    // performance degradation of get opertions. If this happens extremely frequently, like number
+    // of lost removals
+    // close to number of puts, it could prevent the map from ever going into flat mode, and its
+    // size might become
+    // effectively unbound.
```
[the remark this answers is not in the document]
smola: Ugh, I'll blame spotless :wink:

PR #3816, opened 2022-09-08
Create ArrayLists of the right size in DebuggerSinkTest to avoid many large objects

bantonsson:
# What Does This Do

Tries to avoid OOM in tests due to excessive allocation of large arrays.

# Motivation

# Additional Notes


jpbempel (changes requested):
I am confused: how presizing ArrayLists avoid OOM?

bantonsson:
Yes, you are right that the allocation and throwing away of the `ArrayList` instances can't be the root cause, and I'll look into that. To be honest, I think it all comes down to fragmentation of an over-allocated heap, and pre-sizing will hopefully reduce the likelihood of that. I'll close the PR if you think it messes with the snapshot API, or no longer tests the _proper_ usage.

PR #3817, opened 2022-09-08
direct allocation profiling

richardstartin:
# What Does This Do

Instruments DirectByteBuffer to emit a JFR event when off heap memory is allocated via JNI, `ByteBuffer.allocateDirect`, or mmap.

# Motivation

Help users understand where direct memory is being allocated

# Additional Notes


internal-api/src/main/java/datadog/trace/api/Platform.java
```diff
@@ -9,6 +9,17 @@ public final class Platform {
   private static final Version JAVA_VERSION = parseJavaVersion(System.getProperty("java.version"));
   private static final JvmRuntime RUNTIME = new JvmRuntime();
 
+  public static boolean hasJfr() {
+    /* Check only for the open-sources JFR implementation.
+     * If it is ever needed to support also the closed sourced JDK 8 version the check should be
+     * enhanced.
+     * Need this custom check because ClassLoaderMatchers.hasClassesNamed() does not support bootstrap class loader yet.
+     * Note: the downside of this is that we load some JFR classes at startup.
+     * Note2: we cannot check that we can load ExceptionSampleEvent because it is not available on the class path yet.
```
jbachorik, line 18: This line should be dropped - the check is not specific to `ExceptionSampleEvent` any more

dd-java-agent/instrumentation/java-directbytebuffer/src/main/java11/datadog/trace/instrumentation/directbytebuffer/AllocateDirectAdvice.java
```diff
@@ -0,0 +1,12 @@
+package datadog.trace.instrumentation.directbytebuffer;
+
+import net.bytebuddy.asm.Advice;
+
+import static datadog.trace.bootstrap.instrumentation.directallocation.DirectAllocationEvent.onDirectAllocation;
+
+public class AllocateDirectAdvice {
+  @Advice.OnMethodExit
```
jbachorik, line 8: Would it make sense to have timed events in an 'around' advice?

richardstartin: I think that would be really useful for JMC users, but I wonder how we would visualise slow allocations, I foresaw a flamegraph weighted by allocation size

jbachorik: Would be nice to have multiple dimensions across which you would weight the graph.
AFAIK, it is not possible in the UI yet but it would really make sense to have that support.

richardstartin: Being able to time the allocation would require a different instrumentation approach - bytebuddy won't allow it on constructors

devinsba (approved):

jbachorik (approved):

PR #3825, opened 2022-09-13
Disable latest-jdk-app in debugger-smoke-tests until jgit issue is fixed

bantonsson:
# What Does This Do

Disables the `latest-jdk-app` part of the `debugger-smoke-tests`.

# Motivation

The `latest-jdk-app` project fails often enough with a hard to diagnose `jgit` failure that it is affecting normal CI builds and runs. Disable the test until someone has the time to look at the build properly and fix it.

# Additional Notes


cimi (approved):
:+1: Thanks!

There's also #3826, please sync with @jpbempel on which one we want to merge 🙂

jpbempel (approved):

PR #3826, opened 2022-09-13
Disable latest-jdk-app

jpbempel:
while finding a fix for JGit issue

# What Does This Do

# Motivation

# Additional Notes


bantonsson:
I think I beat you to it by a few seconds 😉 #3825

jpbempel:
race with #3825

PR #3832, opened 2022-09-13
Remote config improvements

cataphract:
# What Does This Do

Implements specs RCTE1 (signature verification), RCTE2 (report non-protocol errors during configuration) and RCTE3 (report "capabilities").

Additionally (2nd commit): if the remote config cannot be discovered at startup, keep trying to discover at at interval of 5 (first retry) or 60 seconds (subsequent retries). This allows remote config to work if the agent is not immediately available on startup.


dd-java-agent/agent-debugger/src/main/java/com/datadog/debugger/agent/ConfigurationUpdater.java
```diff
@@ -72,30 +72,33 @@ public ConfigurationUpdater(
   // Should be called by only one thread
   // Should return true if configuration is correctly applied/un-applied
   // otherwise false to indicate a probelem
-  public boolean accept(Configuration configuration) {
+  public void accept(Configuration configuration) {
```
jpbempel, line 75: please remove the comment above

remote-config/src/main/java/datadog/remoteconfig/ConfigurationPoller.java
```diff
@@ -203,19 +217,33 @@ synchronized void poll(ConfigurationPoller poller) {
...
     Call call = this.httpClient.newCall(request);
     return call.execute();
   }
 
+  private long calculateCapabilities() {
+    return (this.featureListeners.containsKey("asm") ? CAPABILITY_ASM_ACTIVATION : 0)
+        | (this.listeners.containsKey(Product.ASM_DATA.name()) ? CAPABILITY_ASM_IP_BLOCKING : 0)
+        | (this.listeners.containsKey(Product.ASM_DD.name()) ? CAPABILITY_ASM_DD_RULES : 0);
+  }
+
```
jpbempel, line 246: Sounds weird to refer to ASM specific capabilities in this part of the code.

remote-config/src/main/java/datadog/remoteconfig/HexUtils.java
```diff
@@ -0,0 +1,45 @@
...
+        -1, -1, -1, -1, -1, -1, -1, -1, -1, -1, -1, -1, -1, -1, -1, -1, -1, -1, -1, -1, -1, -1, -1,
+        -1, 10, 11, 12, 13, 14, 15
+      };
+  private static final byte[] HEX =
+      new byte[] {48, 49, 50, 51, 52, 53, 54, 55, 56, 57, 97, 98, 99, 100, 101, 102};
+  private static final char[] hex = "0123456789abcdef".toCharArray();
+
+  private HexUtils() {}
+
+  public static byte[] fromHexString(String input) {
```
jpbempel, line 16: Maybe consider this implementation: https://stackoverflow.com/a/140861

remote-config/src/main/java/datadog/remoteconfig/HexUtils.java
```diff
@@ -0,0 +1,45 @@
...
+    }
+  }
+
+  private static int getDec(int index) {
+    try {
+      return DEC[index - 48];
+    } catch (ArrayIndexOutOfBoundsException e) {
+      return -1;
+    }
+  }
```
jpbempel, line 44: Please don't use Exception for that, and test against the limit of the array!

remote-config/src/main/java/datadog/remoteconfig/HexUtils.java
jpbempel: also your crypto lib used provides already some utils: https://github.com/cryptography-cafe/curve25519-elisabeth/blob/597bf78f175b0c00a72af7413ba1817554ea6cee/src/test/java/cafe/cryptography/curve25519/Utils.java

remote-config/src/main/java/datadog/remoteconfig/JsonCanonicalizer.java
```diff
@@ -0,0 +1,107 @@
+package datadog.remoteconfig;
+
+import java.io.ByteArrayOutputStream;
+import java.nio.charset.StandardCharsets;
+import java.util.List;
+import java.util.Map;
+
+public class JsonCanonicalizer {
```
jpbempel, line 8: is it an original implementation?
I think there are some implementation available as dependency like
https://github.com/erdtman/java-json-canonicalization
https://github.com/cyberphone/json-canonicalization

remote-config/src/main/java/datadog/remoteconfig/tuf/RemoteConfigRequest.java
```diff
@@ -50,6 +52,10 @@ public ClientInfo getClient() {
 
   /** Stores client information for Remote Configuration */
   public static class ClientInfo {
+    public static final long CAPABILITY_ASM_ACTIVATION = 1 << 1;
+    public static final long CAPABILITY_ASM_IP_BLOCKING = 1 << 2;
+    public static final long CAPABILITY_ASM_DD_RULES = 1 << 3;
+
```
jpbempel, line 58: same here, not sure we want to have specifc product capability constants here.

jpbempel (changes requested):
I think we can reuse some dependencies to avoid reimplementing some parts

remote-config/src/main/java/datadog/remoteconfig/ConfigurationPoller.java
cataphract: These are global constants defined in the protocol, so I think that makes sense that the constants be defined in the `remoteconfig` module. However, perhaps it would make more sense to have a way for the clients of the poller to turn on the capabilities, rather than having this logic here.

remote-config/src/main/java/datadog/remoteconfig/HexUtils.java
cataphract: The implementation is specialized to only accept 128 or 256-bit data and pads with zeros on the leftmost bytes to ensure that the output has those exact sizes. I'll add a comment to that effect.

remote-config/src/main/java/datadog/remoteconfig/JsonCanonicalizer.java
cataphract: It is an original implementation, yes.

None of these seem to follow the [spec that's mentioned in the RFC](https://wiki.laptop.org/go/Canonical_JSON). In particular, the spec indicates that the only characters that should be escaped are `"` and `\`, while both libraries (which seem to share code between themselves) escape other characters. There are other differences; for instance both allow floating point numbers.

remote-config/src/main/java/datadog/remoteconfig/tuf/RemoteConfigRequest.java
cataphract: See above. The capability definitions are part of the protocol.

remote-config/src/main/java/datadog/remoteconfig/HexUtils.java
```diff
@@ -7,39 +7,37 @@ public class HexUtils {
...
 
-      for (int i = 0; i < result.length; ++i) {
-        result[i] = (byte) ((getDec(inputChars[2 * i]) << 4) + getDec(inputChars[2 * i + 1]));
-      }
+    int bufferSize = input.length() / 2 > 32 ? 64 : 32;
+    int diff = bufferSize * 2 - input.length();
+    if (diff < 0) {
+      throw new IllegalArgumentException("Hex data is too large");
+    } else if (diff > 0) {
+      input = new String(new char[diff]).replace('\0', '0') + input;
```
jpbempel, line 24: Could also use ` Arrays.fill()`

remote-config/src/main/java/datadog/remoteconfig/HexUtils.java
```diff
@@ -7,39 +7,37 @@ public class HexUtils {
...
 
-      return result;
+    for (int i = 0; i < result.length; ++i) {
+      result[i] = (byte) ((getDec(inputChars[2 * i]) << 4) + getDec(inputChars[2 * i + 1]));
     }
+
+    return result;
   }
 
   private static int getDec(int index) {
```
jpbempel, line 36: Can use `Character.digit(x, 16)`

jpbempel (approved):

remote-config/src/main/java/datadog/remoteconfig/HexUtils.java
cataphract: Done. I've used `Array.fill` and `System.arrayCopy` to avoid creating new strings.

remote-config/src/main/java/datadog/remoteconfig/HexUtils.java
cataphract: I think I prefer to used the current method because it's more efficient and doesn't accept code points outside ASCII as digits.

dd-java-agent/agent-debugger/src/main/java/com/datadog/debugger/agent/ConfigurationUpdater.java
```diff
@@ -70,32 +70,32 @@ public ConfigurationUpdater(
   }
 
   // Should be called by only one thread
-  // Should return true if configuration is correctly applied/un-applied
-  // otherwise false to indicate a probelem
-  public boolean accept(Configuration configuration) {
+  // Should throw a runtime exception if there is a problem. The message of
+  // the exception will be reported in the next request to the conf service
```
bantonsson, line 74: I can't really see where or what would report the exception message. How does this machinery work? Are there any negative tests?

cataphract: The tests are in `ConfigurationPollerSpecification`. See `reports error applying configuration` and `reports error during deserialization`.

bantonsson: Ah, now I see. They are put in `collectedCfgErrors` and added to the state.

ValentinZakharov (approved):

PR #3835, opened 2022-09-13
record storage class for S3 PutObject requests

richardstartin:
# What Does This Do

Record storage class for S3 PutObject requests

# Motivation

Since PutObject requests incur different fees according to storage class, this is an interesting tag to capture.

# Additional Notes

### DIFF
--- a/dd-java-agent/instrumentation/aws-java-sdk-2.2/src/main/java8/datadog/trace/instrumentation/aws/v2/AwsSdkClientDecorator.java
+++ b/dd-java-agent/instrumentation/aws-java-sdk-2.2/src/main/java8/datadog/trace/instrumentation/aws/v2/AwsSdkClientDecorator.java
@@ -28,6 +28,9 @@ public class AwsSdkClientDecorator extends HttpClientDecorator<SdkHttpRequest, S
     request
         .getValueForField("Bucket", String.class)
         .ifPresent(name -> span.setTag("aws.bucket.name", name));
+    request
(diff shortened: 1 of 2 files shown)


dd-java-agent/instrumentation/aws-java-sdk-2.2/src/main/java8/datadog/trace/instrumentation/aws/v2/AwsSdkClientDecorator.java
```diff
@@ -28,6 +28,9 @@ public AgentSpan onSdkRequest(final AgentSpan span, final SdkRequest request) {
     request
         .getValueForField("Bucket", String.class)
         .ifPresent(name -> span.setTag("aws.bucket.name", name));
+    request
+        .getValueForField("StorageClass", String.class)
+        .ifPresent(storageClass -> span.setTag("aws.storage.class", storageClass));
```
richardstartin, line 33: metanote: this is not the most efficient way either to access the properties or to set them on the span. If the instrumentation targeted specific `SdkRequest` types, it could just use getters, and the span would not be captured several times in the lambdas here (i.e. this method allocates a lot insidiously).

bantonsson: Yup, lambdas FTW

bantonsson (approved):

PR #3836, opened 2022-09-14
Fix debugger smoke test flakiness

jpbempel:
# What Does This Do
`TracerDebuggerIntegrationTest` seems to face a race between Tracer instrumentation and Debugger instrumentation:
while a class is loading
(`org.springframework.web.servlet.DispatcherServlet`), Tracer is transforming it while Debugger receiving configuration and add new transformer, but this transformer will not be take into account because class is already transforming and after that will be considered as loaded.
Debugger perform also a retransformation for already loaded classes but as this class is currently transforming with Tracer it is not part of loaded classes.
Workaround is to try to re-instrument the class.

While topping Server app for `ServerDebuggerIntegrationTest`, Http server could be stopped before sending Http response. Catch exception as it is not an issue.

# Motivation
Smoke test flakiness

# Additional Notes

### DIFF
--- a/dd-smoke-tests/debugger-integration-tests/src/test/java/datadog/smoketest/BaseIntegrationTest.java
+++ b/dd-smoke-tests/debugger-integration-tests/src/test/java/datadog/smoketest/BaseIntegrationTest.java
@@ -59,7 +59,9 @@ public abstract class BaseIntegrationTest {
   private HttpUrl snapshotUrl;
   protected Path logFilePath;
   protected Process targetProcess;
-  private volatile Configuration currentConfiguration;
+  private Configuration currentConfiguration;
+  private boolean configProvided;
+  protected final Object configLock = new Object();
 
   @BeforeAll
   static void setupAll() throws Exception {
@@ -157,7 +159,12 @@ public abstract class BaseIntegrationTest {
     if (request.getPath().equals("/info")) {
       return agentInfoResponse;
     }
-    Configuration configuration = getCurrentConfiguration();
+    Configuration configuration;
+    synchronized (configLock) {
+      configuration = getCurrentConfiguration();
+      configProvided = true;
+      configLock.notifyAll();
+    }
     if (configuration == null) {
       configuration = createConfig(Collections.emptyList());
     }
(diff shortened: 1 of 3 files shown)


jbachorik:
Will this interference also happen in customer applications? It sounds like it may, doesn't it?

jpbempel:
> Will this interference also happen in customer applications? It sounds like it may, doesn't it?

oh yes it may happen. I don't see any way to workaround this. happy to be wrong on this!

jbachorik:
😢 
The only thing I can think of that may possibly help would be not registering the transformer and collecting the loaded classes until the config has been all received and applied.

jpbempel:
> 😢 The only thing I can think of that may possibly help would be not registering the transformer and collecting the loaded classes until the config has been all received and applied.

Could be a class that is not yet loaded and will be only when we reach code that triggers the load... in couple of hours...
that's why we are adding the transformer first (in case a load happen just after) and then we processing all loaded classes to re-transform them if already loaded.
but for class in loading... we miss it. and no difference with a class not yet loaded.

jbachorik:
> > 😢 The only thing I can think of that may possibly help would be not registering the transformer and collecting the loaded classes until the config has been all received and applied.
> 
> Could be a class that is not yet loaded and will be only when we reach code that triggers the load... in couple of hours... that's why we are adding the transformer first (in case a load happen just after) and then we processing all loaded classes to re-transform them if already loaded. but for class in loading... we miss it. and no difference with a class not yet loaded.

Right ... the `getAllLoadedClasses()` is basically racing against the class transformation :(
What is the worst that could happen? We might silently miss probes for some classes, right?

jpbempel:
> 
> Right ... the `getAllLoadedClasses()` is basically racing against the class transformation :( What is the worst that could happen? We might silently miss probes for some classes, right?

Right, like what happen in the smoke test. probe instrumentation is not happening. You need to change the configuration to re-apply the probes on the class by retransformation, since the class is now loaded by the time you change the config.

cimi (approved):
:+1: Thanks!

PR #3837, opened 2022-09-14
Bumping test forked max heap to 768MB

jpbempel:
# What Does This Do
Add more heap for tests.
Looking closely to heap dump after test OOM shows that there is just not enough heap to hold all required objects in memory. Special tests like
`com.datadog.debugger.sink.DebuggerSinkTest#splitSnapshotBatch`, `com.datadog.debugger.sink.DebuggerSinkTest#tooLargeSnapshot` or
`com.datadog.debugger.sink.DebuggerSinkTest#tooLargeUTF8Snapshot`
 require more memory than usual, without leaking.

# Motivation

fix OOME

# Additional Notes

### DIFF
--- a/.circleci/config.yml
+++ b/.circleci/config.yml
@@ -231,7 +231,7 @@ jobs:
           name: Run tests
           command: >-
             MAVEN_OPTS="-Xms64M -Xmx512M"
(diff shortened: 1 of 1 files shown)


bantonsson:
Yeah, when I took a quick look at a heap dump of this, it turned out that the jacoco code coverage is holding on to a lot of memory, but I couldn't find a way to configure that aspect of it.

bantonsson (approved):

smola:
On a very superficial look to DebuggerSinkTest, it reminded me to an OOM I had when using a mock within inner loop code. If get a mock to something with a few thousand calls, verification info will get quickly out of control.

smola (approved):

jpbempel:
> On a very superficial look to DebuggerSinkTest, it reminded me to an OOM I had when using a mock within inner loop code. If get a mock to something with a few thousand calls, verification info will get quickly out of control.

From the heap dump, we saw dominators are String/byte[] from the ones generated by the tests.

PR #3848, opened 2022-09-16
Fix race condition between RC and debugger

jpbempel:
# What Does This Do
debugger should start and subscribe to RC before starting to poll
Product (Debugger or AppSec) shouldn't start ConfigurationPoller, but only perform in one place once everybody has subscribed to it

# Motivation
fix race condition spotted by smoke tests

# Additional Notes

### DIFF
--- a/dd-java-agent/agent-bootstrap/src/main/java/datadog/trace/bootstrap/Agent.java
+++ b/dd-java-agent/agent-bootstrap/src/main/java/datadog/trace/bootstrap/Agent.java
@@ -378,10 +378,8 @@ public class Agent {
       installDatadogTracer(scoClass, sco);
       maybeStartAppSec(scoClass, sco);
       maybeStartIast(scoClass, sco);
-      if (debuggerEnabled) {
-        // start debugger before remote config to subscribe to it before starting to poll
-        startDebuggerAgent(instrumentation, scoClass, sco);
-      }
+      // start debugger before remote config to subscribe to it before starting to poll
+      maybeStartDebugger(instrumentation, scoClass, sco);
       maybeStartRemoteConfig(scoClass, sco);
 
       if (telemetryEnabled) {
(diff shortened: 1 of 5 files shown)


dd-java-agent/agent-debugger/src/main/java/com/datadog/debugger/agent/DebuggerAgent.java
```diff
@@ -99,6 +99,8 @@ public static synchronized void run(
       } catch (final IllegalStateException ex) {
         // The JVM is already shutting down.
       }
+    } else {
+      log.warn("No configuration poller available from SharedCommunicationObjects");
     }
```
cimi, line 104: This warning is not useful our users - they won't know about `SharedCommunicationObjects` or what a configuration poller is.

This could be:

> Cannot enable Dynamic Instrumentation because Remote Configuration is not enabled

Should we check `remoteConfigEnabled` and this log in the `maybeStartDebugger` method in the bootstrap class? We could then have this as a debug log here and leave it unchanged.

dd-java-agent/agent-bootstrap/src/main/java/datadog/trace/bootstrap/Agent.java
```diff
@@ -378,11 +378,11 @@ public void execute() {
       installDatadogTracer(scoClass, sco);
       maybeStartAppSec(scoClass, sco);
       maybeStartIast(scoClass, sco);
-      maybeStartRemoteConfig(scoClass, sco);
-
       if (debuggerEnabled) {
+        // start debugger before remote config to subscribe to it before starting to poll
         startDebuggerAgent(instrumentation, scoClass, sco);
       }
```
cimi, line 384: For consistency with the other products, should we make a `maybeStartDebugger` method?

cimi (approved):
:+1: Thanks!

bantonsson (approved):

PR #3849, opened 2022-09-16
troubleshooting smoke tests (Richard, don't review it)

jpbempel:
# What Does This Do

# Motivation

# Additional Notes

### DIFF
--- a/remote-config/src/main/java/datadog/remoteconfig/ConfigurationPoller.java
+++ b/remote-config/src/main/java/datadog/remoteconfig/ConfigurationPoller.java
@@ -176,6 +176,7 @@ public class ConfigurationPoller
   }
 
   public void start() {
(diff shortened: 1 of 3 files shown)


remote-config/src/main/java/datadog/remoteconfig/ConfigurationPoller.java
```diff
@@ -176,6 +176,7 @@ public ConfigurationPoller get() {
   }
 
   public void start() {
+    log.debug("starting ConfigurationPoller", new RuntimeException("ConfigurationPoller.start"));
```
richardstartin, line 179: Suggest to wrap this in a check that debug logging is enabled since this will fill in the stack trace and startup time has always been a very sensitive issue for users

jpbempel: this is a temp debug log for troubleshooting smoke tests, will never be merged

PR #3850, opened 2022-09-16
[IAST] Add manual.keep tag when a vulnerability is reported

smola:
# What Does This Do
Add manual.keep tag when a vulnerability is reported.

# Motivation
Given that this should be relatively infrequent, and is an important event, we want to keep traces when a vulnerability is reported.

# Additional Notes


bantonsson (approved):

dd-java-agent/iast/src/main/java/com/datadog/iast/Reporter.java
```diff
@@ -26,7 +28,9 @@ public void report(final AgentSpan span, final Vulnerability vulnerability) {
     final VulnerabilityBatch batch = ctx.getVulnerabilityBatch();
     batch.add(vulnerability);
     if (!ctx.getAndSetSpanDataIsSet()) {
-      reqCtx.getTraceSegment().setDataTop("iast", batch);
+      final TraceSegment segment = reqCtx.getTraceSegment();
+      segment.setDataTop("iast", batch);
+      segment.setTagTop(DDTags.MANUAL_KEEP, true);
```
manuel-alvarez-alvarez, line 33: At some point we have to be careful regarding when to set the `manual.keep` due to increased costs for customers (for instance, we had the idea to only set the flag when a new vulnerability not seen previously pops up)

dougqh: I think we need to define the semantics here a bit more carefully.

I have two concerns...
1 - There needs to be an indicator of why this is being kept, so that it can show up properly in the ingestion page
2 - We need to decide whether to set this when the trace is already being sampled or not

smola: @dougqh Noted this for the next guild meeting. Is this a concern before merging this PR?

bantonsson: I think that it's fine to merge this as is for now, maybe with a comment in the code. The AppSec code has the same issue.

PR #3858, opened 2022-09-20
Add support conditions for duplicated probes

jpbempel:
# What Does This Do
Keep track of capturing probeIds based on condition evaluations before adding snapshots to the sink

# Motivation
Support for conditions for duplicated probes

# Additional Notes

### DIFF
--- a/dd-java-agent/agent-debugger/debugger-bootstrap/src/main/java/datadog/trace/bootstrap/debugger/Snapshot.java
+++ b/dd-java-agent/agent-debugger/debugger-bootstrap/src/main/java/datadog/trace/bootstrap/debugger/Snapshot.java
@@ -9,9 +9,11 @@ import java.lang.ref.WeakReference;
 import java.util.ArrayList;
 import java.util.Collections;
 import java.util.HashMap;
+import java.util.HashSet;
 import java.util.List;
 import java.util.Map;
 import java.util.Objects;
+import java.util.Set;
 import java.util.UUID;
 import java.util.regex.Pattern;
 import org.slf4j.Logger;
@@ -33,9 +35,9 @@ public class Snapshot {
   private final ProbeDetails probe;
   private final String language;
   private final transient CapturedThread thread;
+  private final transient Set<String> capturingProbeIds = new HashSet<>();
   private String traceId; // trace_id
   private String spanId; // span_id
-
   private transient boolean isCapturing = true;
 
   public Snapshot(java.lang.Thread thread, ProbeDetails probe) {
@@ -150,8 +152,10 @@ public class Snapshot {
 
   public void commit() {
     if (!isCapturing) {
-      // should we also report skipped for all additional probes?
-      DebuggerContext.skipSnapshot(getProbe().getId(), DebuggerContext.SkipCause.CONDITION);
+      DebuggerContext.skipSnapshot(probe.id, DebuggerContext.SkipCause.CONDITION);
+      for (ProbeDetails probeDetails : probe.additionalProbes) {
+        DebuggerContext.skipSnapshot(probeDetails.id, DebuggerContext.SkipCause.CONDITION);
+      }
       return;
     }
     // generates id only when effectively committing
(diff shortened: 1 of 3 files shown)


dd-java-agent/agent-debugger/debugger-bootstrap/src/main/java/datadog/trace/bootstrap/debugger/Snapshot.java
```diff
@@ -164,21 +168,29 @@ public void commit() {
...
+                version,
+                timestamp,
+                duration,
+                stack,
+                captures,
+                new ProbeDetails(additionalProbe.id, probe.location, probe.script, probe.tags),
+                language,
+                thread,
+                traceId,
+                spanId));
```
cimi, line 190: Nit: we could add a `.copy(String newId) -> Snapshot` method to the Snapshot class if we're doing this in more than one place.

dd-java-agent/agent-debugger/debugger-bootstrap/src/main/java/datadog/trace/bootstrap/debugger/Snapshot.java
```diff
@@ -188,19 +200,37 @@ public boolean isCapturing() {
...
+    if (isCapturing) {
+      capturingProbeIds.add(probe.id);
+    }
+    List<ProbeDetails> additionalProbes = probe.additionalProbes;
+    if (!additionalProbes.isEmpty()) {
+      for (ProbeDetails additionalProbe : additionalProbes) {
+        if (executeScript(additionalProbe.getScript(), capture, additionalProbe.getId())) {
+          isCapturing = true; // force capturing the current Snapshot
+          capturingProbeIds.add(additionalProbe.getId());
+        }
```
cimi, line 214: I don't understand this part - why do we need `isCapturing` if we keep `capturingProbeIds`? Isn't it equivalent to checking if the collection has length zero?

dd-java-agent/agent-debugger/src/test/java/com/datadog/debugger/agent/CapturedSnapshotTest.java
```diff
@@ -809,10 +805,58 @@ public void simpleConditionTest() throws IOException, URISyntaxException {
...
+        listener.snapshots.get(0).getCaptures().getEntry(), "arg", String.class.getName(), "5");
+  }
+
+  @Test
+  public void mergedProbesWithAllConditionsFalseTest() throws IOException, URISyntaxException {
+    doMergedProbeConditions(
+        new ProbeCondition(DSL.when(DSL.FALSE), "false"),
+        new ProbeCondition(DSL.when(DSL.FALSE), "false"),
+        0);
+  }
```
cimi, line 818: Nit: we could add a test where both conditions are `true` - we do test that two snapshots are emitted when we have a true condition and a probe without, so I guess that covers most of it.

cimi (approved):
Thanks! 

Left a couple of comments, nothing blocking.

dd-java-agent/agent-debugger/debugger-bootstrap/src/main/java/datadog/trace/bootstrap/debugger/Snapshot.java
jpbempel: done

dd-java-agent/agent-debugger/debugger-bootstrap/src/main/java/datadog/trace/bootstrap/debugger/Snapshot.java
jpbempel: I was reluctant to remove this field as it is used by instrumentation, but you're right we can just rely on the set.

dd-java-agent/agent-debugger/src/test/java/com/datadog/debugger/agent/CapturedSnapshotTest.java
jpbempel: done

cimi (approved):
:+1: Thanks!

PR #3873, opened 2022-09-21
Split CircleCI gradle cache into two parts

bantonsson:
# What Does This Do

Splits the cache on CircleCI into two parts. One large with all the dependencies, that is only saved for `master` and `project/...` branches, and a smaller cache that is saved every time. This leads to new dependencies being downloaded multiple times, but that should hopefully not be a big problem for short lived branches.

This cuts 4-5 minutes off a build in a PR.

# Motivation

Slow builds are slow.

# Additional Notes

### DIFF
--- a/.circleci/config.yml
+++ b/.circleci/config.yml
@@ -10,21 +10,16 @@ defaults: &defaults
 # 2) Cache keys are prefix matched, and the most recently updated cache that matches will be picked
 #
 # There is a weekly job that runs on Monday mornings that builds a new cache from scratch.
-#
-# The cache is also saved using the unique git revision to enable the dependent build steps to get a fully populated
-# cache to work from.
-cache_keys: &cache_keys
+dependency_cache_keys: &dependency_cache_keys
   keys:
-    # Dependent steps will find this cache
-    - dd-trace-java-v4-{{ .Branch }}-{{ checksum "_circle_ci_cache_id" }}-{{ .Revision }}
     # New branch commits will find this cache
-    - dd-trace-java-v4-{{ .Branch }}-{{ checksum "_circle_ci_cache_id" }}-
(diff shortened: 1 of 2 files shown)


.circleci/config.yml
```diff
@@ -178,9 +190,34 @@ jobs:
...
+            matches:
+              pattern: "^(master|project/.+)$"
+              value: << pipeline.git.branch >>
+          steps:
+            - save_cache:
+                key: dd-trace-java-dep-v1-{{ .Branch }}-{{ checksum "_circle_ci_cache_id" }}-{{ .Revision }}
+                <<: *dependency_cache_paths
+
+      # Save a full dependency the first time any other PR branch is built (will be skipped
+      # other runs since the cache name will already exist
```
bantonsson, line 205: This could be removed to cut some time on all builds, at the expense of possibly downloading more dependencies in parallel when dependencies are added/updated.

PerfectSlayer (approved):

.circleci/config.yml
```diff
@@ -178,9 +191,34 @@ jobs:
...
+      - when:
+          condition:
+            not:
+              matches:
+                pattern: "^(master|project/.+)$"
+                value: << pipeline.git.branch >>
+          steps:
+            - save_cache:
+                key: dd-trace-java-dep-v1-{{ .Branch }}-{{ checksum "_circle_ci_cache_id" }}-base
+                <<: *dependency_cache_paths
```
bantonsson, line 216: This could be removed to cut some time on all builds, at the expense of possibly downloading more dependencies in parallel when dependencies are added/updated.

bantonsson: Disabled this by commenting it out.

richardstartin (approved):
does this make it faster? 🚀

bantonsson:
Cuts 4-5 minutes off a PR.

smola (approved):

PR #3876, opened 2022-09-21
JSON Trace Sampling Rules

ygree:
# What Does This Do

Adds support for [JSON Trace Sampling Rules](https://github.com/DataDog/dd-opentracing-cpp/blob/master/doc/sampling.md#sampling-rules)

# Motivation

Align tracer settings for rule based sampling with the other tracers.

We already have the way to specify sampling rules with `trace.sampling.service.rules` and/or `trace.sampling.operation.rules`. The existing mechanism allows to define rules to match either the service name or the operation name but not service and operation simultaneously. The old mechanism supports regex (but doesn't document it) whereas the new mechanism only allows exact match (similar to the Go, C++, Python tracers) for a combination of the service and/or operation name. The new mechanism supports JSON format to define the rules (which is aligned with other tracers and described in [Ingestion Controls with OpenTelemetry](https://docs.google.com/document/d/1HRbi1DrBjL_KGeONrPgH7lblgqSLGlV5Ox1p4RL97xM/edit?pli=1#).

# Additional Notes

N/A


internal-api/src/main/java/datadog/trace/api/Config.java
```diff
@@ -481,6 +482,7 @@ public class Config {
 
   private final Map<String, String> traceSamplingServiceRules;
   private final Map<String, String> traceSamplingOperationRules;
+  private final String traceSamplingJsonRules;
```
bantonsson, line 485: This should be in the `toString` for logging as well.

dd-trace-core/src/main/java/datadog/trace/common/sampling/SamplingRule.java
```diff
@@ -73,4 +73,24 @@ protected CharSequence getRelevantString(final T span) {
...
+      super(sampler);
+      this.serviceName = serviceName;
+      this.operationName = operationName;
+    }
+
+    @Override
+    public boolean matches(T span) {
+      return serviceName == null
+          || serviceName.equals(span.getServiceName()) && operationName == null
+          || operationName.contentEquals(span.getOperationName());
```
bantonsson, line 93: Is it not allowed to have a `name` only rule?

bantonsson (commented):
LGTM, only 2 minor things.

dd-trace-core/src/main/java/datadog/trace/common/sampling/SamplingRule.java
ygree: All possible combinations are allowed: service+name, service only, name only and neither (match all). I just forgot to add parentheses. Thanks!

internal-api/src/main/java/datadog/trace/api/Config.java
```diff
@@ -2931,6 +2938,8 @@ public String toString() {
         + traceSamplingServiceRules
         + ", traceSamplingOperationRules="
         + traceSamplingOperationRules
+        + ", traceSamplingJsonRules="
+        + traceSamplingJsonRules
```
bantonsson, line 2942: It feels like this name is a bit strange, since the setting is `trace.sampling.rules`, and the fact that it's JSON, is just the encoding.

bantonsson (approved):

PR #3878, opened 2022-09-22
Telemetry heartbeat interval option.

ValentinZakharov:
# What Does This Do
Introduced `DD_TELEMETRY_HEARTBEAT_INTERVAL` option that should be used to testing purpose only. The value of the variable should determine the interval in which the client sends heartbeat updates in seconds and must be in the range of 1-3600. (The default value is 60).

# Motivation

# Additional Notes

### DIFF
--- a/telemetry/src/main/java/datadog/telemetry/TelemetryRunnable.java
+++ b/telemetry/src/main/java/datadog/telemetry/TelemetryRunnable.java
@@ -91,12 +91,12 @@ public class TelemetryRunnable implements Runnable {
 
   private void successWait() {
     consecutiveFailures = 0;
-    int waitSeconds = telemetryService.getIntervalSeconds();
+    int waitMs = telemetryService.getIntervalMs();
 
     // Wait between iterations no longer than 10 seconds
-    if (waitSeconds > 10) waitSeconds = 10;
+    if (waitMs > 10000) waitMs = 10000;
 
-    sleeper.sleep(waitSeconds);
+    sleeper.sleep(waitMs);
   }
 
   private void failureWait() {
--- a/telemetry/src/main/java/datadog/telemetry/TelemetryRunnable.java
+++ b/telemetry/src/main/java/datadog/telemetry/TelemetryRunnable.java
@@ -91,8 +91,12 @@ public class TelemetryRunnable implements Runnable {
 
   private void successWait() {
     consecutiveFailures = 0;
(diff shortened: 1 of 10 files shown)


telemetry/src/main/java/datadog/telemetry/TelemetryRunnable.java
```diff
@@ -91,8 +91,12 @@ private boolean mainLoopIteration() throws InterruptedException {
...
   private void successWait() {
     consecutiveFailures = 0;
-    int waitSeconds = 10;
-    sleeper.sleep(waitSeconds * 1000L);
+    int waitSeconds = telemetryService.getIntervalSeconds();
+
+    // Wait between iterations no longer than 10 seconds
+    if (waitSeconds > 10) waitSeconds = 10;
+
+    sleeper.sleep(waitSeconds);
```
smola, line 99: I think `sleep` takes ms, but you're passing seconds here?

telemetry/src/main/java/datadog/telemetry/TelemetryServiceImpl.java
```diff
@@ -21,13 +21,11 @@
 
 public class TelemetryServiceImpl implements TelemetryService {
```
cataphract, line 22: You could add a test in this class.

telemetry/src/main/java/datadog/telemetry/TelemetryRunnable.java
```diff
@@ -91,8 +91,12 @@ private boolean mainLoopIteration() throws InterruptedException {
 
   private void successWait() {
     consecutiveFailures = 0;
-    int waitSeconds = 10;
-    sleeper.sleep(waitSeconds * 1000L);
+    int waitMs = telemetryService.getIntervalMs();
```
cataphract, line 94: `getIntervalMs` is a bit of vague name. Maybe `getHeartbeatInterval`. This could also use an explanation of why this wait time is related to the time between heartbeat interval. IIUC it should be like:

> Heartbeat requests are generated by the call to `prepareRequests` if the heartbeat interval has passed without other requests having been made. This wait time controls the frequency with which `prepareRequests` is called, so the wait time should be low enough that there be an opportunity for heatbeats to be generated after the heartbeat interval has elapsed.

cataphract (approved):

smola (approved):

PR #3894, opened 2022-09-26
Add smoke test for probe error status

jpbempel:
# What Does This Do

# Motivation

Cover all probe statuses by smoke tests

# Additional Notes

### DIFF
--- a/dd-smoke-tests/debugger-integration-tests/src/test/java/datadog/smoketest/ServerDebuggerIntegrationTest.java
+++ b/dd-smoke-tests/debugger-integration-tests/src/test/java/datadog/smoketest/ServerDebuggerIntegrationTest.java
@@ -15,6 +15,7 @@ import datadog.trace.bootstrap.debugger.Snapshot;
 import java.io.EOFException;
 import java.io.IOException;
 import java.util.Collections;
+import java.util.HashMap;
 import java.util.HashSet;
 import java.util.List;
 import java.util.concurrent.TimeUnit;
@@ -26,6 +27,7 @@ import okhttp3.Response;
 import okhttp3.mockwebserver.MockResponse;
 import okhttp3.mockwebserver.MockWebServer;
 import okhttp3.mockwebserver.RecordedRequest;
+import org.junit.Assert;
 import org.junit.jupiter.api.AfterEach;
 import org.junit.jupiter.api.BeforeEach;
 import org.junit.jupiter.api.DisplayName;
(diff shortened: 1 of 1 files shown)


dd-smoke-tests/debugger-integration-tests/src/test/java/datadog/smoketest/ServerDebuggerIntegrationTest.java
```diff
@@ -203,6 +205,30 @@ void testDisableEnableProbesUsingAllowList() throws Exception {
...
+    // statuses could be received out of order
+    HashMap<ProbeStatus.Status, ProbeStatus.Diagnostics> statuses = new HashMap<>();
+    ProbeStatus.Diagnostics diagnostics = retrieveProbeStatusRequest().getDiagnostics();
+    statuses.put(diagnostics.getStatus(), diagnostics);
+    diagnostics = retrieveProbeStatusRequest().getDiagnostics();
+    statuses.put(diagnostics.getStatus(), diagnostics);
+    Assert.assertTrue(statuses.containsKey(ProbeStatus.Status.RECEIVED));
+    Assert.assertEquals(
+        "Cannot find datadog/smoketest/debugger/ServerDebuggerTestApplication::unknownMethodName",
+        statuses.get(ProbeStatus.Status.ERROR).getException().getMessage());
```
cimi, line 229: In the UI we show `NO_TYPE: ...` for both missing class and missing method. Could we have a more specific one for when we found the class but not the method?

Not blocking, fine to do later in another PR.

cimi (approved):

dd-smoke-tests/debugger-integration-tests/src/test/java/datadog/smoketest/ServerDebuggerIntegrationTest.java
jpbempel: We cannot have an error for a class not found because by definition, a class can be loaded at any point in the future, so you cannot ensure that is an incorrect class. So only method name can be returned as an error in Java

cimi: Then should we rename `NO_TYPE` to `NO_METHOD`? I don't know where that string that prefixes the message is coming from.

jpbempel: me neither, and is it true for all languages? probably not for native languages (rust/go)

PR #3895, opened 2022-09-26
Appsec: support parameterization of blocking responses

cataphract:
# What Does This Do
Instead of hardcoding the presumed ids of actions that should elicit blocking a request, the spec is fully implemented. The action returned by the WAF is actually interpreted as an action id and the that id is looked up in the "actions" entry of the WAF configuration. Then the type of the action is looked up, and "block_request" is recognized. The parameters of the action (http response code to send and the content type of the response: one of html, json and out) are also taken into account.

### DIFF
--- a/dd-java-agent/instrumentation/jetty-9/jetty-9.gradle
+++ b/dd-java-agent/instrumentation/jetty-9/jetty-9.gradle
@@ -46,3 +46,7 @@ dependencies {
   latestDepTestImplementation group: 'org.eclipse.jetty', name: 'jetty-continuation', version: '9.+'
   latestDepTestImplementation project(':dd-java-agent:instrumentation:jetty-appsec-9.3')
 }
+latestDepTest {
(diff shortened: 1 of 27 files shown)


dd-java-agent/instrumentation/jetty-9/jetty-9.gradle
```diff
@@ -46,3 +46,7 @@ dependencies {
   latestDepTestImplementation group: 'org.eclipse.jetty', name: 'jetty-continuation', version: '9.+'
   latestDepTestImplementation project(':dd-java-agent:instrumentation:jetty-appsec-9.3')
 }
+latestDepTest {
+  jvmArgs += "-Dnet.bytebuddy.dump=/tmp/abc"
```
bantonsson, line 50: Is this a leftover?

bantonsson (approved):

PR #3903, opened 2022-09-27
Use call site instrumentation for weak hash and weak cipher (detect SHA-1)

DDJavierSantos:
# What Does This Do
Change weak hash and weak cipher instrumentation so we use the call site interface instead of the target instrumentation. 

# Motivation
This is needed so we can instrument the use of SHA-1 hash algorithm since it is used by the jar verifier and we need to ignore the jar verifier.

# Additional Notes
This PR includes the IAST instrumenter, which is a requirement for any other call site interface instrumentations in IAST. It also includes an initial IAST class exclusion trie (based on Hdiv's).


dd-java-agent/instrumentation/iast-instrumenter/iast-instrumenter.gradle
```diff
@@ -0,0 +1,27 @@
...
+  pass {
+    coreJdk()
+  }
+}
+
+apply from: "${rootDir}/gradle/java.gradle"
+apply from: "${rootDir}/gradle/tries.gradle"
+
+apply plugin: 'org.unbroken-dome.test-sets'
+apply plugin: 'call-site-instrumentation'
```
manuel-alvarez-alvarez, line 11: Is it needed here?

dd-java-agent/instrumentation/iast-instrumenter/src/main/java/datadog/trace/instrumentation/iastinstrumenter/IastInstrumenter.java
```diff
@@ -0,0 +1,38 @@
...
+  }
+
+  @Override
+  public boolean matches(final TypeDescription target) {
+    return IastExclusionTrie.apply(target.getName()) != 1;
+  }
+
+  @Override
+  public boolean isEnabled() {
+    return true;
```
manuel-alvarez-alvarez, line 31:
```suggestion
    return Config.get().isIastEnabled();
```

dd-java-agent/instrumentation/java-security/src/main/java/datadog/trace/instrumentation/java/security/WeakCipherInstrumentationCallSite.java
```diff
@@ -0,0 +1,17 @@
+package datadog.trace.instrumentation.java.security;
```
manuel-alvarez-alvarez, line 1: You can use `@CallSite.Before` skipping the return parameter and making the instrumentation faster

dd-java-agent/instrumentation/java-security/src/main/java/datadog/trace/instrumentation/java/security/WeakHashInstrumentationCallSite.java
```diff
@@ -0,0 +1,18 @@
+package datadog.trace.instrumentation.java.security;
```
manuel-alvarez-alvarez, line 1: Same as WeakCipher with before instrumentation

dd-java-agent/instrumentation/java-security/src/test/java/test/foo/TestJavaCalls.java
```diff
@@ -0,0 +1,18 @@
+package test.foo;
```
manuel-alvarez-alvarez, line 1: I always use `TestSuite` for this object, perhaps we could use better names to prevent methods growing in these classes (for instance, `TestSecuritySuite`). And I normally put them under the package `foo.bar` I think we should agree in package and naming.

internal-api/src/test/groovy/datadog/trace/api/iast/InstrumentationBridgeTest.groovy
```diff
@@ -76,4 +76,17 @@ class InstrumentationBridgeTest extends DDSpecification {
     1 * module.onCipherAlgorithm(_) >> { throw new Error('Boom!!!') }
     noExceptionThrown()
   }
+
+  def "bridge calls don't leak exceptions for onParameterName and onParameterValue on null parameters"() {
```
manuel-alvarez-alvarez, line 80: This doesn't belong here right?

internal-api/src/main/java/datadog/trace/api/iast/InstrumentationBridge.java
```diff
@@ -52,6 +52,26 @@ public static void onMessageDigestGetInstance(@Nonnull final String algorithm) {
     }
   }
 
+  public static void onParameterName(final String parameterName) {
```
manuel-alvarez-alvarez, line 55: Same as below, this is for another PR right?

dd-java-agent/instrumentation/java-security/src/test/java/test/foo/TestJavaCalls.java
```diff
@@ -0,0 +1,18 @@
...
+import java.security.MessageDigest;
+import java.security.NoSuchAlgorithmException;
+import javax.crypto.Cipher;
+import javax.crypto.NoSuchPaddingException;
+
+public class TestJavaCalls {
+
+  public Cipher getCipherInstance(String algo)
+      throws NoSuchPaddingException, NoSuchAlgorithmException {
+    return Cipher.getInstance(algo);
```
manuel-alvarez-alvarez, line 12: Something that I do in these kind of test suites is to consume the parameters of the stack and the result, like the following:

```java
  public static String toString(final StringBuilder builder) {
    LOGGER.debug("Before string builder toString {}", builder);
    final String result = builder.toString();
    LOGGER.debug("After string builder toString {}", result);
    return result;
  }
```

The reasoning here is that, if the instrumenter screws the stack, when the call site is the last instruction of the method then the issue might not be detected.

manuel-alvarez-alvarez (changes requested):

manuel-alvarez-alvarez (commented):
The filtering of the stack should be changed:
```java
    StackTraceElement stackTraceElement =
        stackWalker.walk(
            stack ->
                stack
                    .filter(s -> !s.getClassName().equals("java.security.MessageDigest"))
                    .findFirst()
                    .get());
```

Ideally we should could use a trie based solution for all vulnerabilities:

```java
StackTraceElement stackTraceElement =
        stackWalker.walk(IastModuleImpl::findValidPackageForVulnerability);
```

```java
  private static StackTraceElement findValidPackageForVulnerability(
      Stream<StackTraceElement> stream) {
    final StackTraceElement[] first = new StackTraceElement[1];
    return stream
        .filter(
            stack -> {
              if (first[0] == null) {
                first[0] = stack;
              }
              return IastExclusionTrie.apply(stack.getClassName()) != 1;
            })
        .findFirst()
        .orElse(first[0]);
  }
```

manuel-alvarez-alvarez (approved):

smola (approved):

smola (changes requested):
I think we need to remove the previous exception for we had globally:

https://github.com/DataDog/dd-trace-java/blob/0d6702eda5c05925f4040083a36e6a8d384b5847/dd-java-agent/agent-tooling/src/main/resources/datadog/trace/agent/tooling/bytebuddy/matcher/ignored_class_name.trie#L43-L44

dd-java-agent/instrumentation/iast-instrumenter/src/main/resources/datadog/trace/instrumentation/iastinstrumenter/iast_exclusion.trie
```diff
@@ -0,0 +1,192 @@
+# Generates 'IastExclusionsTrie.java'
+
+# This file lists classes that are globally ignored for matching/transformation
+# Use 0 to allow transformation of packages or classes beneath ignored packages
+# Wildcard lines end with '.' or '$' and match any sub-packages/nested-classes
```
bantonsson, line 5: This comment is out of date.

dd-java-agent/instrumentation/iast-instrumenter/src/main/resources/datadog/trace/instrumentation/iastinstrumenter/iast_exclusion.trie
```diff
@@ -0,0 +1,192 @@
...
+1 org.wildfly.*
+1 org.xnio.*
+1 org.yaml.*
+1 reactor.*
+1 springfox.*
+1 tech.jhipster.*
+1 weblogic.*
+1 zipkin2.*
+
+
```
bantonsson, line 187: Double newline

dd-smoke-tests/springboot/src/test/groovy/datadog/smoketest/IastSpringBootSmokeTest.groovy
```diff
@@ -95,4 +100,24 @@ class IastSpringBootSmokeTest extends AbstractServerSmokeTest {
...
+
+  def "weak hash vulnerability is present"() {
+    setup:
+    String url = "http://localhost:${httpPort}/weakhash"
+    def request = new Request.Builder().url(url).get().build()
+
+    when:
+    client.newCall(request).execute()
+
+
```
bantonsson, line 112: Double newline. I wish that there was a better formatter for `spock/gradle`.

dd-smoke-tests/springboot/src/test/groovy/datadog/smoketest/IastSpringBootSmokeTest.groovy
```diff
@@ -22,6 +22,11 @@ class IastSpringBootSmokeTest extends AbstractServerSmokeTest {
     command.add(javaPath())
     command.addAll(defaultJavaProperties)
     command.addAll(["-Ddd.appsec.enabled=true", "-Ddd.iast.enabled=true", "-Ddd.iast-request-sampling=100"])
+    command.addAll([
```
manuel-alvarez-alvarez, line 25: This is actually a duplicate from previous line

jandro996 (approved):

dd-java-agent/agent-iast/src/test/groovy/com/datadog/iast/taint/TaintObjectsLogTest.groovy.orig
```diff
@@ -0,0 +1,33 @@
+package com.datadog.iast.taint
```
smola, line 1: @DDJavierSantos  Can we remove this file? It seems an old test, and not being used (`.orig`).

dd-java-agent/instrumentation/iast-instrumenter/src/main/java/datadog/trace/instrumentation/iastinstrumenter/IastInstrumenter.java
```diff
@@ -0,0 +1,39 @@
...
+  }
+
+  @Override
+  public boolean matches(final TypeDescription target) {
+    return IastExclusionTrie.apply(target.getName()) != 1;
+  }
+
+  @Override
+  public boolean isEnabled() {
+    return Config.get().isIastEnabled();
```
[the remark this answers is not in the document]
smola: That `isEnabled` is gone.
There could be possibly more than one instrumenter related to IAST. But all call site instrumentations for IAST will use this single `IastInstrumenter`. This is currently all we introduce, and we expect most of our instrumentation to land here.

dd-java-agent/instrumentation/iast-instrumenter/src/main/resources/datadog/trace/instrumentation/iastinstrumenter/iast_exclusion.trie
bantonsson: This comment is still out of date and **wrong**. Wildcard lines to not end with `'.'` or `'$'`.

bantonsson (approved):

PR #3905, opened 2022-09-27
Fix release workflow

PerfectSlayer:
# What Does This Do

Download releases will be updated when releases are promoted as non pre-release status.

# Motivation

The current workflow runs when releases are published in pre-release state.
So they are filtered from eligible releases to update `download-latest*` releases.

# Additional Notes


smola (approved):

bantonsson (approved):
🎉

PerfectSlayer:
Will see how it behaves for the next release...
I did run a lot of tests and releases on a fork project but it always seems doubtful until I see it working by myself!

PR #3906, opened 2022-09-28
Remove snapshot & probe URLs config

jpbempel:
# What Does This Do

# Motivation
no longer use URLs as everything is going through datadog agent

# Additional Notes


bantonsson (approved):

ojung (approved):
👍

PR #3909, opened 2022-09-28
Add SQL Injection detection

cataphract:
# What Does This Do
Reports calls to JDBC functions called with tainted strings.


dd-java-agent/iast/iast.gradle
```diff
@@ -1,6 +1,7 @@
 plugins {
   id "com.github.johnrengelman.shadow"
   id 'me.champeau.jmh'
+  id 'java-test-fixtures'
```
manuel-alvarez-alvarez, line 4: Didn't know about this little guy, nice!

dd-java-agent/iast/src/main/java/com/datadog/iast/IastModuleImpl.java
```diff
@@ -93,6 +96,59 @@ public void onHashingAlgorithm(String algorithm) {
...
+
+    Vulnerability vulnerability =
+        new Vulnerability(
+            VulnerabilityType.SQL_INJECTION,
+            Location.forStack(stackTraceElement),
+            new Evidence(queryString, taintedObject.getRanges()));
+    reporter.report(span, vulnerability);
+  }
+
+  private static StackTraceElement findFirstFrameForSecondPackage(
```
manuel-alvarez-alvarez, line 126: We need utilities to do proper filtering (probably based on tries). At runtime you never know how many steps you have to traverse the stack until you reach customer code. I've checked for weak cipher and hashing and it's not yet done, so I guess it's ok for now.

cataphract: yeah my opinion is that this is a fool's errand; we should just send the full stacktrace. Even if we manage to find the first frame that belongs to the application, it may not be informative (e.g. if all the application's queries go through some abstraction).

smola: I think that, at some point, we might add the full stacktrace to reported vulnerabilities. However, that would not solve the problem here. We use a location (in Java, class+line) to deduplicate vulnerabilities, and that runs in a performance-sensitive path. We'd like to make this path even faster, which would increase our ability to cover more vulnerabilities within reasonable overheads. Going for the full stacktrace from the beginning would just run counter our performance requirements. There's some more aspects to it we can discuss (e.g. at product level).

dd-java-agent/instrumentation/jdbc/jdbc.gradle
```diff
@@ -89,4 +89,5 @@ tasks.named("latestDepJava11Test").configure {
 
 tasks.withType(Test).configureEach {
   usesService(testcontainersLimit)
+  jvmArgs += ['-Dnet.bytebuddy.dump=/tmp/aa']
```
manuel-alvarez-alvarez, line 92: Remember to remove this before the final merge :smile:

dd-java-agent/instrumentation/jdbc/src/main/java/datadog/trace/instrumentation/jdbc/IastStatementCallSite.java
```diff
@@ -0,0 +1,29 @@
...
+    @CallSite.Before("boolean java.sql.Statement.execute(java.lang.String, int)"),
+    @CallSite.Before("boolean java.sql.Statement.execute(java.lang.String, int[])"),
+    @CallSite.Before("boolean java.sql.Statement.execute(java.lang.String, java.lang.String[])"),
+    @CallSite.Before("long java.sql.Statement.executeLargeUpdate(java.lang.String)"),
+    @CallSite.Before("long java.sql.Statement.executeLargeUpdate(java.lang.String, int)"),
+    @CallSite.Before("long java.sql.Statement.executeLargeUpdate(java.lang.String, int[])"),
+    @CallSite.Before(
+        "long java.sql.Statement.executeLargeUpdate(java.lang.String, java.lang.String[])"),
+  })
+  public static void beforeExecute(@CallSite.Argument String sql) {
```
manuel-alvarez-alvarez, line 26: I'll rather explicitly set the argument index to keep it consistent with `IastConnectionCallSite`

dd-java-agent/instrumentation/jdbc/src/test/groovy/foo/bar/IastJdbcCallSites.groovy
```diff
@@ -0,0 +1,100 @@
...
+
+import groovy.transform.CompileStatic
+
+import java.sql.CallableStatement
+import java.sql.Connection
+import java.sql.PreparedStatement
+import java.sql.ResultSet
+import java.sql.Statement
+
+@CompileStatic
```
manuel-alvarez-alvarez, line 11: Nice!

dd-java-agent/agent-tooling/src/main/java/datadog/trace/agent/tooling/bytebuddy/csi/CallSiteUtils.java
```diff
@@ -21,22 +31,45 @@ public abstract class CallSiteUtils {
     BOX_HANDLERS[Type.FLOAT] = new JdkBoxingHandler("java/lang/Float", "F", "floatValue");
     BOX_HANDLERS[Type.LONG] = new JdkBoxingHandler("java/lang/Long", "J", "longValue");
     BOX_HANDLERS[Type.DOUBLE] = new JdkBoxingHandler("java/lang/Double", "D", "doubleValue");
+
+    STACK_MANIP_TABLE.put("12|1", new int[] {SWAP, DUP_X1});
```
manuel-alvarez-alvarez, line 35: Maybe not for this PR or we will never merge it, but how much time do you think it will take to move all the stack duplication logic to compile time (moving it to the plugin)?.

I'm worried about startup time (as always) and now with the table I think we should pre-compute all the instructions while generating the call sites.

manuel-alvarez-alvarez (approved):

dd-java-agent/agent-iast/src/main/java/com/datadog/iast/IastModuleImpl.java
```diff
@@ -93,6 +86,48 @@ public void onHashingAlgorithm(String algorithm) {
...
+    if (taintedObject == null) {
+      return;
+    }
+
+    if (!overheadController.consumeQuota(Operations.REPORT_VULNERABILITY, span)) {
+      return;
+    }
+
+    StackTraceElement stackTraceElement =
+        stackWalker.walk(IastModuleImpl::findValidPackageForVulnerability);
```
anderruiz, line 106: As it is difficult to guess it directly for the PR, silly question, does this approach cover cases similar to Spring JDBCTemplate? We need to instrument it (Real SQL call is there) while it needs to be ignored in the stack calculation

dd-java-agent/agent-tooling/src/main/java/datadog/trace/agent/tooling/bytebuddy/csi/CallSiteUtils.java
```diff
@@ -215,21 +301,32 @@ private static void pushArray(
...
+    for (int idx : indices) {
+      final Type type = methodArgumentTypes[idx];
+      loadNthArgFromArray(mv, type, idx);
+    }
+  }
+
+  private static void loadNthArgFromArray(MethodVisitor mv, Type type, int argIdx) {
+    final int stackObjectSize = type.getSize();
+    // 1. duplicate the array
+    mv.visitInsn(Opcodes.DUP);
```
anderruiz, line 319: Shouldn't this be in a static import for coherence?

dd-java-agent/agent-iast/src/main/java/com/datadog/iast/IastModuleImpl.java
smola: I think we need an ad hoc review of Spring, both adding exclusions/inclusions, and tests. It's the same issue with Servlet's getParameter.
But it's something I would rather do in a separate PR for Spring support.

buildSrc/call-site-instrumentation-plugin/build.gradle.kts
```diff
@@ -92,3 +92,8 @@ tasks.build {
 tasks.test {
   useJUnitPlatform()
 }
+
+project.afterEvaluate {
```
smola, line 96: cc @manuel-alvarez-alvarez 
If we do this, we should probably add the check step to CI? Since, after this change, I guess the tests won't run at all?
Can we get this to a separate PR?

manuel-alvarez-alvarez: It's not only the tests but spotless as well, it's true that developing with it under `buildSrc` it's a bit of a hassle

smola: @cataphract Can we move this to a separate PR? We should also make sure that at least spotless runs.

smola: I have removed these, rebased, and adapted to API changes from master. However, I still get failures on `FreemarkerAdviceGeneratorTest`.

PR #3910, opened 2022-09-28
Nayeem kamal/greenplum jdbc

nayeem-kamal:
# What Does This Do
Adds the option to use hierarchy matching for jdbc if the driver in use is not supported by name based matching

# Motivation
Greenplum JDBC obfuscates class names and cannot be consistently matched with name-based matching

# Additional Notes

### DIFF
--- a/dd-java-agent/instrumentation/jdbc/src/main/java/datadog/trace/instrumentation/jdbc/ConnectionHierarchyInstrumentation.java
+++ b/dd-java-agent/instrumentation/jdbc/src/main/java/datadog/trace/instrumentation/jdbc/ConnectionHierarchyInstrumentation.java
@@ -1,4 +1,3 @@
-
 package datadog.trace.instrumentation.jdbc;
 
 import static datadog.trace.agent.tooling.bytebuddy.matcher.HierarchyMatchers.implementsInterface;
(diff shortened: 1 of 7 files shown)


dd-java-agent/instrumentation/jdbc/src/main/java/datadog/trace/instrumentation/jdbc/ConnectionHierarchyInstrumentation.java
```diff
@@ -0,0 +1,33 @@
...
+import datadog.trace.agent.tooling.Instrumenter;
+import datadog.trace.api.Config;
+import net.bytebuddy.description.type.TypeDescription;
+import net.bytebuddy.matcher.ElementMatcher;
+
+@AutoService(Instrumenter.class)
+public class ConnectionHierarchyInstrumentation extends AbstractConnectionInstrumentation
+    implements Instrumenter.ForTypeHierarchy {
+  public ConnectionHierarchyInstrumentation() {
+    super("jdbc", "jdbcMatcher");
```
devinsba, line 16: Since the class name for this is stable I don't think you should build this as a generic thing. I would make the driver name the second param

devinsba:
If you change the instrumentation classes to be specific to this driver then you can get rid of all the changes to config and use the built in instrumentation enable disable functionality

nayeem-kamal:
Closing this PR because I am having issues with rebasing with master

PR #3915, opened 2022-09-29
String taint tracking - substring and subSequence methods

jandro996:
# What Does This Do

Adds all the instrumentation needed to perform taint tacking in substring operations for String objects

# Motivation

IAST requires to track all modifications that happen to strings in the code, this PR implements the substring and subSequence operations for String class

# Performance

| Benchmark | Mode | Cnt |  Score |  Error | Units |
| ----------- | ------ | --- | ------- | ----- |------|
|StringSubsequenceBenchmark.baseline    |          ss | 15000 |  29.736| ±  2.788|  ns/op
|StringSubsequenceBenchmark.iastDisabled    |      ss  |15000  | 30.538 |±  3.634| ns/op
|StringSubsequenceBenchmark.taintedLoseRange  |    ss|  15000  | 67.920 |±  7.945|  ns/op
|StringSubsequenceBenchmark.taintedModifyRange  |  ss | 15000 | 256.353| ± 15.960|  ns/op


dd-java-agent/instrumentation/java-lang/src/test/groovy/datadog/trace/instrumentation/java/lang/StringCallSiteTest.groovy
```diff
@@ -27,4 +27,46 @@ class StringCallSiteTest extends AgentTestRunner {
...
+    setup:
+    final iastModule = Mock(IastModule)
+    InstrumentationBridge.registerIastModule(iastModule)
+
+    when:
+    final result = TestSuite.stringSubSequence('012345', 1, 5)
+
+    then:
+    result == '1234'
+    1 * iastModule.onStringSubSequence('012345', 1, 5, '1234')
```
manuel-alvarez-alvarez, line 69: You can try with other types of `CharSequence` like `StringBuilder` or `StringBuffer`

jandro996: We are instrumenting String.subsequence so there is no need to test with StringBuilder or StringBuffer :)

dd-java-agent/agent-iast/src/main/java/com/datadog/iast/taint/Ranges.java
```diff
@@ -25,4 +25,64 @@ public static void copyShift(
...
+      if (newLength > 0) {
+        newRanges[newRangeIndex] = createIfDifferent(rangeSelf, newStart, newLength);
+        newRangeIndex++;
+      }
+    }
+
+    return newRanges;
+  }
+
+  private static int getSubstringSkippedRanges(int offset, int length, Range[] ranges) {
```
manuel-alvarez-alvarez, line 68: To me it would make more sense to return an array with two elements that will be the start and end indexes of the included ranges:

```java
private static int[] getIncludedRangesInterval(int offset, int length, Range[] ranges) {
  // start: index of the first included range
  // end: index of the first non included range or length of ranges
  int start = -1, end = -1;
  // compute ranges
  return new int[] { start, end };
}
```

Then method to create the ranges is far easier.

dd-java-agent/instrumentation/java-lang/src/main/java/datadog/trace/instrumentation/java/lang/StringCallSite.java
```diff
@@ -18,4 +18,34 @@ public static String afterConcat(
...
+  public static String afterSubstring(
+      @CallSite.This final String self,
+      @CallSite.Argument final int beginIndex,
+      @CallSite.Return final String result) {
+    InstrumentationBridge.onStringSubSequence(
+        self, beginIndex, self != null ? self.length() : 0, result);
+    return result;
+  }
+
+  @CallSite.After("java.lang.String java.lang.String.substring(int, int)")
```
manuel-alvarez-alvarez, line 32: Can't you join the two last call sites in an array? the signatures look compatible

dd-java-agent/agent-iast/src/main/java/com/datadog/iast/taint/Ranges.java
```diff
@@ -25,4 +25,64 @@ public static void copyShift(
...
+      final int newEnd = newStart + newLength;
+      if (newStart < 0) {
+        newLength = newLength + newStart;
+        newStart = 0;
+      }
+      if (newEnd > length) {
+        newLength = length - newStart;
+      }
+      if (newLength > 0) {
+        newRanges[newRangeIndex] = createIfDifferent(rangeSelf, newStart, newLength);
```
manuel-alvarez-alvarez, line 60: This is far simpler, if `offset == 0` then they are equal, otherwise you have to offset it.

dd-java-agent/instrumentation/java-lang/src/main/java/datadog/trace/instrumentation/java/lang/StringCallSite.java
manuel-alvarez-alvarez: We've checked and we cannot ATM, the code to copy parameters in the stack will fail.

dd-java-agent/agent-iast/src/main/java/com/datadog/iast/propagation/StringModuleImpl.java
```diff
@@ -165,6 +166,32 @@ public void onStringConcatFactory(
     taintedObjects.taint(result, targetRanges);
   }
 
+  @Override
+  @SuppressFBWarnings("ES_COMPARING_PARAMETER_STRING_WITH_EQ")
+  public void onStringSubSequence(
+      @Nullable String self, int beginIndex, int endIndex, @Nullable CharSequence result) {
```
smola, line 172: `self` can be `@NotNull` if this is an after aspect.

dd-java-agent/agent-iast/src/main/java/com/datadog/iast/taint/Ranges.java
```diff
@@ -83,6 +81,66 @@ public static <E> RangesProvider<E> rangesProviderFor(
...
+        if (newLength > 0) {
+          newRanges[newRangeIndex] = new Range(newStart, newLength, range.getSource());
+        }
+      }
+    }
+
+    return newRanges;
+  }
+
+  public static int[] getIncludedRangesInterval(
```
smola, line 123: Maybe this code could be included directly at `forSubstring`, to avoid the array creation here.

dd-java-agent/agent-iast/src/test/groovy/com/datadog/iast/propagation/StringModuleTest.groovy
```diff
@@ -439,6 +439,95 @@ class StringModuleTest extends IastModuleImplTestBase {
     "==>123<==" | "==>456<==" | "==>123<====>456<=="
   }
 
+  void 'onStringSubSequence null ,empty or string not change after subsequence (#self, #beginIndex, #endIndex)'(final String self, final int beginIndex, final int endIndex) {
```
smola, line 442:
```suggestion
  void 'onStringSubSequence null, empty or string not changed after subsequence (#self, #beginIndex, #endIndex)'(final String self, final int beginIndex, final int endIndex) {
```

dd-java-agent/agent-iast/src/test/groovy/com/datadog/iast/propagation/StringModuleTest.groovy
```diff
@@ -439,6 +439,95 @@ class StringModuleTest extends IastModuleImplTestBase {
...
+    given:
+    final result = self?.substring(beginIndex, endIndex)
+    when:
```
smola, line 445: I don't think we have style rules about it, but most ouf our Spock tests have a space before each block:

```
when:
foo

then:
bar

File: dd-java-agent/agent-iast/src/test/groovy/com/datadog/iast/propagation/StringModuleTest.groovy [hunk 9]
```diff
@@ -439,6 +439,95 @@ class StringModuleTest extends IastModuleImplTestBase {
...
+    def to = ctx.getTaintedObjects().get(result)
+    if (shouldBeTainted) {
+      assert to != null
+      assert to.get() == result
+      assert taintFormat(to.get() as String, to.getRanges()) == expected
+    } else {
+      assert to == null
+    }
+    where:
+    self                      | beginIndex | endIndex | expected
```
smola, line 504: I miss some cases where `self` is fully tainted with a single range.

dd-java-agent/agent-iast/src/main/java/com/datadog/iast/propagation/StringModuleImpl.java
```diff
@@ -165,6 +166,32 @@ public void onStringConcatFactory(
...
+    if (!canBeTainted(self) || !canBeTainted(result) || self == result) {
```
manuel-alvarez-alvarez, line 173: I'll rather do:
```java
if (self == result || !canBeTainted(result)) {
    return;
}
```

Given that result is a subset of result it should be enough

dd-java-agent/agent-iast/src/main/java/com/datadog/iast/taint/Ranges.java
manuel-alvarez-alvarez: I think that the idea in mind is to reuse the method for trims as well, probably we can use a long and do some bit shifting ops to prevent the array creation.

dd-java-agent/agent-iast/src/test/groovy/com/datadog/iast/propagation/StringModuleTest.groovy
jandro996: Thanks!!! there was a bug when self is fully tainted with a single range

dd-java-agent/agent-iast/src/main/java/com/datadog/iast/taint/Ranges.java
jandro996: As @manuel-alvarez-alvarez this is implemented to be compatible with https://github.com/DataDog/dd-trace-java/pull/3951/files/409749c2f20c7a6e0632c628cd930cdb318ed516#diff-4d93e9e8d535d005f9c6484971aabf453ee4b7f3ed09f6e4a78522e0dfb04b63

smola (approved):

manuel-alvarez-alvarez (approved):

PR #3917, opened 2022-09-29
[smoke tests] New basic test for java module system

robertomonteromiguel:
# What Does This Do

New test to check java module system.

# Motivation

We need to detect java module system failures

# Additional Notes
 
This test is adopted from jdk regression tests. 
Check: http://hg.openjdk.java.net/jdk9/jdk9/jdk/file/f8b19df2115a/test/java/lang/ModuleTests/BasicModuleTest.java


robertomonteromiguel:
- I think the smoke tests didn't fail. Is the bug fixed?
- Findbug is failing. I am using forgiven API for my test. I don't know if this check is necessary for smoke tests ¿?  
- Other checks are failing, but I think there are not related with the new test
Findbug failure:
`Caused by: de.thetaphi.forbiddenapis.ForbiddenApiException: Check for forbidden API calls failed while scanning class 'datadog.smoketest.moduleapp.ModuleApplication' (ModuleApplication.java): java.lang.ClassNotFoundException: java.lang.Module (while looking up details about referenced class 'java.lang.Module')
`

dd-smoke-tests/java9-modules/src/main/java11/datadog/smoketest/moduleapp/ModuleApplication.java
```diff
@@ -3,5 +3,24 @@
 public class ModuleApplication {
   public static void main(final String[] args) throws InterruptedException {
     Thread.sleep(600);
+    ModuleApplication.testBaeModuleRead();
+  }
+  public static void testBaeModuleRead() {
```
robertomonteromiguel, line 8: typo name. I will change it

PR #3934, opened 2022-09-30
Use a version of gradle 7.5.1 patched for cgroup memory

bantonsson:
# What Does This Do

Changes the build to use a patched version of gradle 7.5.1 that understands CircleCI cgroup memory.

# Motivation

The CircleCI build step is balancing on the edge of 100% memory and sometimes just fails. This is because the gradle memory aware runners get the full machine memory information and not the container related one. This patched version (which is a stop gap until fixes are contributed upstream) reads container memory information from the cgroup filesystem and tops out the build step on ~75% memory.

# Additional Notes


PerfectSlayer (approved):

richardstartin (approved):

smola (approved):

gradle/wrapper/gradle-wrapper.properties
```diff
@@ -1,7 +1,9 @@
 distributionBase=GRADLE_USER_HOME
 distributionPath=wrapper/dists
-distributionSha256Sum=db9c8211ed63f61f60292c69e80d89196f9eb36665e369e7f00ac4cc841c2219
+#distributionSha256Sum=db9c8211ed63f61f60292c69e80d89196f9eb36665e369e7f00ac4cc841c2219
 # Please note that the version specific cache directory in .circleci/config.yml needs to match this version
-distributionUrl=https\://services.gradle.org/distributions/gradle-7.5.1-all.zip
+#distributionUrl=https\://services.gradle.org/distributions/gradle-7.5.1-all.zip
```
ValentinZakharov, line 5: Is that intentionally kept old record to be able rollback?

dd-smoke-tests/quarkus/quarkus.gradle
```diff
@@ -13,7 +13,7 @@ def appBuildDir = "$buildDir/application"
 // define the task that builds the quarkus project
 tasks.register('quarkusBuild', Exec) {
   workingDir "$appDir"
-  environment += ["GRADLE_OPTS": "-Xmx512M"]
+  environment += ["GRADLE_OPTS": "-Dorg.gradle.jvmargs='-Xmx512M -Xms64M'"]
```
jpbempel, line 16: Just curious: what is the point to set the initial size here?
why not, if set, using the same as Xmx?

bantonsson: Very good point. It's a copy pasta thing. It too can be removed.

PR #3938, opened 2022-10-03
StringConcatFactory taint tracking

manuel-alvarez-alvarez:
# What Does This Do
Add taint tracking support for `+` concatenation in JDK >=9 through `StringConcatFactory`

# Motivation
JDKs with support for invoke dynamic use `StringConcatFactory` in order to do String concatenation through the `+` operator, we need to support it in order to do traint tracking.


jandro996 (approved):

dd-java-agent/agent-iast/src/main/java/com/datadog/iast/IastModuleImpl.java
```diff
@@ -210,6 +212,54 @@ public void onStringBuilderToString(
...
+    }
+    if (rangeCount == 0) {
+      return;
+    }
+
+    final Range[] targetRanges = new Range[rangeCount];
+    int offset = 0, rangeIndex = 0;
+    for (int item : recipeOffsets) {
+      if (item < 0) {
+        offset += (-item);
```
smola, line 249: Interesting way to write `offset -= item`?
Or is it to make it clear that we're adding a positive number (after flipping the sign)?

dd-java-agent/instrumentation/java-lang/java-lang-9/src/main/java/datadog/trace/instrumentation/java/lang/invoke/StringConcatFactoryCallSite.java
```diff
@@ -0,0 +1,180 @@
...
+
+  /**
+   * Preprocess the recipe and create an array of offsets where for each offset:
+   *
+   * <ul>
+   *   <li><code>offset < 0</code> length of the current recipe chunk with sign changed
+   *   <li><code>offset >= 0</code> index of the argument
+   * </ul>
+   */
+  private static int[] preprocessRecipe(final String recipe, final Object[] constants) {
```
smola, line 78: I think we considered before moving this to build time? If it is possible, we might want to optimize that in the future (not in this PR).

smola (approved):

dd-java-agent/instrumentation/java-lang/java-lang-9/src/main/java/datadog/trace/instrumentation/java/lang/invoke/StringConcatFactoryCallSite.java
manuel-alvarez-alvarez: What do you mean by build time?, this processing happens when an invoke dynamic is found (instrumentation time), and not when doing the invoke dynamic as before (run time). Is that what you meant?

smola: Discussed offline. I meant actual build time, but it's not really possible.

PR #3942, opened 2022-10-03
authenticate github requests to avoid api throttling

richardstartin:
# What Does This Do

Authenticates github requests using a new PAT installed in circleci

# Motivation

Avoid API throttling

# Additional Notes


randomanderson (approved):

mcculls (approved):

bantonsson:
It would be nice if the diff blob had a descriptive _name_. Also I don't think the diff code works as intended, since the output of the command is:
```
fatal: ambiguous argument '"master"': unknown revision or path not in the working tree.
Use '--' to separate paths from revisions, like this:
'git <command> [<revision>...] -- [<file>...]'
```

PR #3946, opened 2022-10-04
Improve tracer stats behavior

PerfectSlayer:
# What Does This Do

This PR improve the stats behavior by:
- Adding missing `synthetic` to the aggregation key,
- Support the `_dd.measured` custom tag to mark span as measured.
  Measured spans will take part of stats like top level span.

# Motivation

Those changes improves the results of [Java tracer system tests](https://github.com/DataDog/system-tests/pull/504).

# Additional Notes

Synthetic span detection is based on `origin` value comparison.
The comparison I use calls `CharSequence.toString()`. I fear there might be unnecessary allocation behind it but I didn't think I can rely on `CharSequence.equals` according to the Javadoc:

> The result of comparing two objects that implement CharSequence is therefore, in general, undefined.

Let me know if there is a better solution for this comparison.

### DIFF
--- a/dd-trace-api/src/main/java/datadog/trace/api/DDTags.java
+++ b/dd-trace-api/src/main/java/datadog/trace/api/DDTags.java
@@ -40,4 +40,5 @@ public class DDTags {
   public static final String ORIGIN_KEY = "_dd.origin";
   public static final String LIBRARY_VERSION_TAG_KEY = "library_version";
   public static final String CI_ENV_VARS = "_dd.ci.env_vars";
+  public static final String MEASURED = "_dd.measured";
 }
--- a/dd-trace-api/src/main/java/datadog/trace/api/DDTags.java
+++ b/dd-trace-api/src/main/java/datadog/trace/api/DDTags.java
@@ -16,14 +16,14 @@ public class DDTags {
 
   public static final String ERROR_MSG = "error.msg"; // string representing the error message
   public static final String ERROR_TYPE = "error.type"; // string representing the type of the error
-  public static final String ERROR_STACK = "error.stack"; // human readable version of the stack
+  public static final String ERROR_STACK = "error.stack"; // human-readable version of the stack
 
   public static final String ANALYTICS_SAMPLE_RATE = "_dd1.sr.eausr";
   @Deprecated public static final String EVENT_SAMPLE_RATE = ANALYTICS_SAMPLE_RATE;
 
   /** Manually force tracer to be keep the trace */
   public static final String MANUAL_KEEP = "manual.keep";
-  /** Manually force tracer to be drop the trace */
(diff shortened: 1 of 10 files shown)


dd-trace-core/src/main/java/datadog/trace/common/metrics/MetricKey.java
```diff
@@ -11,27 +11,31 @@ public final class MetricKey {
...
-    // the constants are 31^4, 31^3, 31^2, 31^1, 31^0
+    // the constants are 31^5, 31^4, 31^3, 31^2, 31^1, 31^0
     this.hash =
         923521 * this.resource.hashCode()
             + 29791 * this.service.hashCode()
             + 961 * this.operationName.hashCode()
             + 31 * this.type.hashCode()
-            + httpStatusCode;
+            + httpStatusCode
+            + (this.synthetics ? 28629151 : 0);
```
bantonsson, line 38: I know it doesn't matter, but it feels weird do have the largest constant at the end.

dd-trace-api/src/main/java/datadog/trace/api/DDTags.java
```diff
@@ -16,14 +16,14 @@ public class DDTags {
...
-  public static final String ERROR_STACK = "error.stack"; // human readable version of the stack
+  public static final String ERROR_STACK = "error.stack"; // human-readable version of the stack
 
   public static final String ANALYTICS_SAMPLE_RATE = "_dd1.sr.eausr";
   @Deprecated public static final String EVENT_SAMPLE_RATE = ANALYTICS_SAMPLE_RATE;
 
   /** Manually force tracer to be keep the trace */
   public static final String MANUAL_KEEP = "manual.keep";
-  /** Manually force tracer to be drop the trace */
+  /** Manually force tracer to be dropped the trace */
```
bantonsson, line 26: Please change both this sentence and the one above `MANUAL_KEEP` to something along the lines of:
`Manually force the tracer to drop/keep the trace`

bantonsson (approved):

dd-trace-core/src/main/java/datadog/trace/common/metrics/MetricKey.java
PerfectSlayer: Actually, I have a quick question about this part: why don't we use `Objects.hash()`?

richardstartin: there's a comment above it explaining why

PerfectSlayer: Oh sorry… though night and a 2.5h commute this morning 😩 
 I'll make sure to think twice before asking today.
Thanks for talking time answering me.

PR #3947, opened 2022-10-04
Collapse build and check on CircleCI

bantonsson:
# What Does This Do

Collapses `build` and `check` to not run tests if there are any code violations.

# Motivation

# Additional Notes

### DIFF
--- a/.circleci/config.yml
+++ b/.circleci/config.yml
@@ -202,6 +202,9 @@ jobs:
       - restore_cache:
           <<: *dependency_cache_keys
 
+      - restore_cache:
+          <<: *build_cache_keys
+
       - run:
           name: Build Project
           command: >-
--- a/.circleci/config.yml
+++ b/.circleci/config.yml
@@ -209,10 +209,10 @@ jobs:
           name: Build Project
           command: >-
             MAVEN_OPTS="-Xms64M -Xmx256M"
-            GRADLE_OPTS="-Dorg.gradle.jvmargs='-Xmx1900M -Xms64M -XX:ErrorFile=/tmp/hs_err_pid%p.log -XX:+HeapDumpOnOutOfMemoryError -XX:HeapDumpPath=/tmp'"
-            ./gradlew clean compile shadowJar
+            GRADLE_OPTS="-Dorg.gradle.jvmargs='-Xmx2048M -Xms2048M -XX:ErrorFile=/tmp/hs_err_pid%p.log -XX:+HeapDumpOnOutOfMemoryError -XX:HeapDumpPath=/tmp'"
+            ./gradlew clean compile shadowJar check -PskipTests
             << pipeline.parameters.gradle_flags >>
-            --max-workers=4
+            --max-workers=8
             --rerun-tasks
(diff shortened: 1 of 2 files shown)


richardstartin (approved):
bonus points if we can speed it up, but this will reduce total build time as check is always one of the slowest fan out tasks and can't be skipped.

bantonsson:
This makes the `Build Project` step of the `build` task go from ~4.5 minutes to ~8.5 minutes. I don't think that's acceptable. The only way to speed this up seems to be enabling the config cache and upgrading a number of plugins, which will have to wait until java-8 is the minimum.

richardstartin:
> This makes the `Build Project` step of the `build` task go from ~4.5 minutes to ~8.5 minutes. I don't think that's acceptable. The only way to speed this up seems to be enabling the config cache and upgrading a number of plugins, which will have to wait until java-8 is the minimum.

I'm not sure that's the right metric to focus on. The build _job_ increases from 9.5 to 13 minutes because the gradle build itself takes 4 minutes longer, but completely eliminates a 12 min+ check job. So you go from 9.5 + 12 to 13.5. If you need to run slow jobs anyway, you won't care because they mostly mask the 12 mins check latency, but if you can skip those anyway the build is faster because of this change.

bantonsson:
Ok, fine. But the current approach adds some unnecessary time for reports and other things that get duplicated since they are now saved in the `workspace`. I'll look into it a bit more.

randomanderson:
@bantonsson @richardstartin I was curious so I looked at the [build scan](https://scans.gradle.com/s/bii4arh3iokkw/performance/configuration):
- The slowest plugin is one that I don't think anyone is using (`com.dorongold.task-tree`)
- The slow executions are dominated by the smoke test builds (maybe there's room for improvement) and `submodulesUpdate` for jmxfetch (should this just be versioned instead of invoking `git` during the build?)
- ~24s are wasted writing to a build cache that's never used.

.circleci/config.yml
```diff
@@ -202,12 +202,15 @@ jobs:
...
+      - restore_cache:
+          <<: *build_cache_keys
+
       - run:
           name: Build Project
           command: >-
             MAVEN_OPTS="-Xms64M -Xmx256M"
-            GRADLE_OPTS="-Dorg.gradle.jvmargs='-Xmx1900M -Xms64M -XX:ErrorFile=/tmp/hs_err_pid%p.log -XX:+HeapDumpOnOutOfMemoryError -XX:HeapDumpPath=/tmp'"
-            ./gradlew clean compile shadowJar
+            GRADLE_OPTS="-Dorg.gradle.jvmargs='-Xmx2048M -Xms2048M -XX:ErrorFile=/tmp/hs_err_pid%p.log -XX:+HeapDumpOnOutOfMemoryError -XX:HeapDumpPath=/tmp'"
```
richardstartin, line 212: this runs on a 16 core box, is it worth adding more workers now it's also doing check?

bantonsson: I should definitely try it.

PR #3966, opened 2022-10-05
profile tests

richardstartin:
# What Does This Do

Allows setting `profile: true` on a non-IBM8/non-ORACLE8 test job to collect a JFR

# Motivation
Diagnose test reliability/test problems
# Additional Notes

### DIFF
--- a/.circleci/config.yml
+++ b/.circleci/config.yml
@@ -355,6 +355,9 @@ jobs:
       maxWorkers:
         type: integer
         default: 2
+      profile:
+        type: boolean
+        default: false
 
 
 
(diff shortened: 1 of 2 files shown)


.circleci/config.yml
```diff
@@ -344,8 +347,13 @@ jobs:
       - run:
           name: Run tests
           command: >-
+            if [[ << parameters.profile >> ]] && [[ << parameters.testJvm >> != "IBM8" ]] && [[ << parameters.testJvm >> != "ORACLE8" ]]; 
+            then
+              PROFILER_COMMAND="-XX:+FlightRecorder -XX:StartFlightRecording=settings=profile,filename=/tmp/<< parameters.stage >>-<< parameters.testJvm >>.jfr,dumponexit=true"
```
jpbempel, line 352: nit: `-XX:+FlightRecorder` is useless, JFR is enabled with `StartFlightRecording`
see https://bugs.openjdk.org/browse/JDK-8225312
It was necessary for earlier Oracle JDK8 < 8u40

jpbempel (approved):

.circleci/config.yml
richardstartin: good catch!

PR #3967, opened 2022-10-05
Fix IP blocking (added default AppSec action)

ValentinZakharov:
# What Does This Do
Added missing default `block` action for the AppSec.

# Motivation
AppSec wasn't blocking malicious requests in case if no actions defined in rules

# Additional Notes

### DIFF
--- a/dd-java-agent/appsec/src/main/java/com/datadog/appsec/powerwaf/PowerWAFModule.java
+++ b/dd-java-agent/appsec/src/main/java/com/datadog/appsec/powerwaf/PowerWAFModule.java
@@ -71,7 +71,7 @@ public class PowerWAFModule implements AppSecModule {
               new HashMap<String, Object>() {
                 {
                   put("status_code", 403);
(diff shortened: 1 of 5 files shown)


dd-java-agent/appsec/src/main/java/com/datadog/appsec/powerwaf/PowerWAFModule.java
```diff
@@ -63,6 +63,19 @@ public class PowerWAFModule implements AppSecModule {
...
+          "block",
+          new ActionInfo(
+              "block_request",
+              new HashMap<String, Object>() {
+                {
+                  put("status_code", 403);
+                  put("type", "auto");
+                  put("grpc_status_code", 10);
+                }
+              }));
```
cataphract, line 77: Don't do this, as it will create an extra class. Initialize instead on a static block or create a private static method.

smola (approved):

cataphract (approved):

PR #3970, opened 2022-10-05
Add missing size for map values in snapshots

jpbempel:
# What Does This Do

# Motivation

# Additional Notes


jbachorik (approved):
Looking good

PR #3971, opened 2022-10-05
allocate fewer large strings in openliberty test

richardstartin:
# What Does This Do

Makes the openliberty smoke test lighter weight by not allocating huge strings (which need to be converted to UTF-8) over and over again

# Motivation

Reduce resource requirement for smoke tests so they can run faster => faster, more reliable CI

This is from a profile of just the test side of the openliberty smoke test:

Screenshot 2022-10-05 at 21 02 08

# Additional Notes

### DIFF
--- a/dd-smoke-tests/springboot-openliberty/src/test/groovy/datadog/smoketest/SpringBootOpenLibertySmokeTest.groovy
+++ b/dd-smoke-tests/springboot-openliberty/src/test/groovy/datadog/smoketest/SpringBootOpenLibertySmokeTest.groovy
@@ -68,12 +68,9 @@ class SpringBootOpenLibertySmokeTest extends AbstractServerSmokeTest {
   def "Test concurrent high load requests to Spring Boot running Open Liberty"() {
     def url = "http://localhost:${httpPort}/connect/0"
     def formBody = new FormBody.Builder()
-    def value = new StringBuilder(10_001 * 10)
-    for (int i = 0; i < 10000; i++) {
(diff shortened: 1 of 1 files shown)


mcculls (approved):

dd-smoke-tests/springboot-openliberty/src/test/groovy/datadog/smoketest/SpringBootOpenLibertySmokeTest.groovy
```diff
@@ -68,12 +68,9 @@ class SpringBootOpenLibertySmokeTest extends AbstractServerSmokeTest {
   def "Test concurrent high load requests to Spring Boot running Open Liberty"() {
     def url = "http://localhost:${httpPort}/connect/0"
     def formBody = new FormBody.Builder()
-    def value = new StringBuilder(10_001 * 10)
-    for (int i = 0; i < 10000; i++) {
-      value.append("@@@@@@@@@@") // 10 chars
-    }
+    def value = "not too big!"
```
bantonsson, line 71: So this is quite different from the original test that has really long `Strings`. I took a quick look at the okhttp code and it seems that one of the issues is that `@` is in the set of characters that need encoding, which leads to 2 extra copies of the `String` in an encoded form as well. And of course calling `value.toString()` in the loop is just asking for extra copies.

Anyway, good find, and 100 form fields is still pretty big.

bantonsson (approved):

PR #3974, opened 2022-10-06
enable async-profiler by default on x86/x64

richardstartin:
# What Does This Do

Enables async-profiler on any JVM running on x86/x64 or J9

# Motivation

# Additional Notes

Depends on #3992 because this needs to be tested on more J9 and OpenJ9 distributions

### DIFF
--- a/dd-java-agent/agent-profiling/profiling-controller/src/main/java/com/datadog/profiling/controller/ProfilingSystem.java
+++ b/dd-java-agent/agent-profiling/profiling-controller/src/main/java/com/datadog/profiling/controller/ProfilingSystem.java
@@ -149,6 +149,10 @@ public final class ProfilingSystem {
           uploadPeriod.toMillis(),
           TimeUnit.MILLISECONDS);
       started = true;
+    } catch (UnsupportedEnvironmentException unsupported) {
+      log.warn(
+          "Datadog Profiling was enabled on an unsupported JVM, will not profile application. See {} for more details about supported JVMs.",
+          "https://docs.datadoghq.com/profiler/enabling/java/?tab=commandarguments#requirements");
     } catch (Throwable t) {
       if (t instanceof RuntimeException) {
         // Possibly a wrapped exception related to Oracle JDK 8 JFR MX beans
--- a/dd-smoke-tests/src/main/groovy/datadog/smoketest/AbstractSmokeTest.groovy
+++ b/dd-smoke-tests/src/main/groovy/datadog/smoketest/AbstractSmokeTest.groovy
@@ -91,7 +91,6 @@ abstract class AbstractSmokeTest extends ProcessManager {
     "-Ddd.profiling.start-delay=${PROFILING_START_DELAY_SECONDS}",
     "-Ddd.profiling.upload.period=${PROFILING_RECORDING_UPLOAD_PERIOD_SECONDS}",
     "-Ddd.profiling.url=${getProfilingUrl()}",
-    "-Ddd.profiling.async.enabled=true",
     "-Ddd.profiling.async.wall.enabled=true",
     "-Ddd.profiling.async.alloc.enabled=" + !isIBM,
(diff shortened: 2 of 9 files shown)


dd-trace-api/src/main/java/datadog/trace/api/config/ProfilingConfig.java
```diff
@@ -55,7 +55,7 @@ public final class ProfilingConfig {
   public static final String PROFILING_AUXILIARY_TYPE_DEFAULT = "none";
 
   public static final String PROFILING_ASYNC_ENABLED = "profiling.async.enabled";
-  public static final boolean PROFILING_ASYNC_ENABLED_DEFAULT = false;
+  public static final boolean PROFILING_ASYNC_ENABLED_DEFAULT = System.getProperty("os.arch").contains("x86");;
```
jbachorik, line 58: Would it be possible to use the newly added `Platform.isX86()` method instead?

dd-smoke-tests/src/main/groovy/datadog/smoketest/AbstractSmokeTest.groovy
```diff
@@ -88,7 +89,7 @@ abstract class AbstractSmokeTest extends ProcessManager {
     "-Ddd.profiling.start-delay=${PROFILING_START_DELAY_SECONDS}",
     "-Ddd.profiling.upload.period=${PROFILING_RECORDING_UPLOAD_PERIOD_SECONDS}",
     "-Ddd.profiling.url=${getProfilingUrl()}",
-    "-Ddd.profiling.async.enabled=false",
+    "-Ddd.profiling.async.enabled=${Platform.isX86()}",
```
jbachorik, line 92: IIAM, we can remove this altogether as the default value will be the same (eg. enabled on x86 only)?

jbachorik (commented):
Just a few clarification questions ...

dd-smoke-tests/src/main/groovy/datadog/smoketest/AbstractSmokeTest.groovy
richardstartin: I intend to set up tests on aarch64 soon

internal-api/src/main/java/datadog/trace/api/Platform.java
```diff
@@ -230,6 +230,10 @@ public static boolean isJ9() {
         && System.getProperty("java.vm.name").contains("J9");
   }
 
+  public static boolean isX86() {
+    return System.getProperty("os.arch").contains("x86");
```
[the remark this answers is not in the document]
richardstartin: Can it NPE though?

dd-smoke-tests/src/main/groovy/datadog/smoketest/AbstractSmokeTest.groovy
richardstartin: oh yes good point 🤣

jbachorik (approved):

jbachorik (approved):

dd-java-agent/agent-profiling/profiling-controller/src/main/java/com/datadog/profiling/controller/ProfilingSystem.java
```diff
@@ -149,6 +149,10 @@ private void startProfilingRecording() {
           uploadPeriod.toMillis(),
           TimeUnit.MILLISECONDS);
       started = true;
+    } catch (UnsupportedEnvironmentException unsupported) {
+      log.warn(
+          "Datadog Profiling was enabled on an unsupported JVM, will not profile application. See {} for more details about supported JVMs.",
+          "https://docs.datadoghq.com/profiler/enabling/java/?tab=commandarguments#requirements");
```
jbachorik, line 155: Nice touch!

jbachorik (approved):

PR #3982, opened 2022-10-06
Simplify Sampler and related interface by moving type parameter from the class to method declaration

ygree:
# What Does This Do

Moves the span type parameter from the class declaration to the method declaration.

# Motivation

It will simplify class by removing unnecessary type parameter from it.

# Additional Notes

Please let me know if there are any objections for this approach.
This change is based on #4031.

### DIFF
--- a/dd-trace-core/src/main/java/datadog/trace/common/sampling/Sampler.java
+++ b/dd-trace-core/src/main/java/datadog/trace/common/sampling/Sampler.java
@@ -14,7 +14,7 @@ import org.slf4j.Logger;
 import org.slf4j.LoggerFactory;
 
 /** Main interface to sample a collection of traces. */
-public interface Sampler<T extends CoreSpan<T>> {
+public interface Sampler {
 
   /**
    * Sample a collection of traces based on the parent span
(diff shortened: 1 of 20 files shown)


dd-trace-core/src/main/java/datadog/trace/common/sampling/Sampler.java
```diff
@@ -55,40 +55,38 @@ public static <T extends CoreSpan<T>> Sampler<T> forConfig(final Config config)
...
         }
       } else {
-        sampler = new AllSampler<>();
+        sampler = new AllSampler();
       }
       return sampler;
     }
 
-    public static <T extends CoreSpan<T>> Sampler<T> forConfig(final Properties config) {
+    public static <T extends CoreSpan<T>> Sampler forConfig(final Properties config) {
```
PerfectSlayer, line 89: `T` is no more used here

dd-trace-core/src/main/java/datadog/trace/common/sampling/Sampler.java
```diff
@@ -14,21 +14,21 @@
...
    */
-  boolean sample(T span);
+  <T extends CoreSpan<T>> boolean sample(T span);
 
   final class Builder {
     private static final Logger log = LoggerFactory.getLogger(Builder.class);
 
-    public static <T extends CoreSpan<T>> Sampler<T> forConfig(final Config config) {
-      Sampler<T> sampler;
+    public static <T extends CoreSpan<T>> Sampler forConfig(final Config config) {
```
PerfectSlayer, line 30: `T` is no more used here

dd-trace-core/src/main/java/datadog/trace/common/sampling/SingleSpanSampler.java
```diff
@@ -11,14 +11,14 @@
...
+public interface SingleSpanSampler {
 
-  boolean setSamplingPriority(T span);
+  <T extends CoreSpan<T>> boolean setSamplingPriority(T span);
 
   final class Builder {
     private static final Logger log = LoggerFactory.getLogger(Builder.class);
 
-    public static <T extends CoreSpan<T>> SingleSpanSampler<T> forConfig(Config config) {
+    public static <T extends CoreSpan<T>> SingleSpanSampler forConfig(Config config) {
```
PerfectSlayer, line 21: `T` is no more used here

PerfectSlayer (changes requested):
I think some generics might be removed.

PerfectSlayer (approved):

PR #3987, opened 2022-10-07
Shade all internal dependencies in dd-trace-ot

bantonsson:
# What Does This Do

Shades and bundles all internal dependencies in `dd-trace-ot` only leaving these public dependencies:
```
\--- com.datadoghq:dd-trace-ot
     +--- com.datadoghq:dd-trace-api
     +--- io.opentracing:opentracing-api
     +--- io.opentracing:opentracing-noop
     +--- io.opentracing:opentracing-util
     +--- io.opentracing.contrib:opentracing-tracerresolver
     \--- org.slf4j:slf4j-api
```

# Motivation

Allow the internals of the tracer to evolve without forcing end users to upgrade or switch dependencies.

# Additional Notes


dd-trace-ot/dd-trace-ot.gradle
```diff
@@ -33,7 +33,10 @@ dependencies {
   compileOnly deps.autoserviceAnnotation
 
   api project(':dd-trace-api')
-  api project(':dd-trace-core')
+  implementation (project(':dd-trace-core')) {
+    // why all communication pulls in remote config is beyond me...
+    exclude(group: 'com.datadoghq', module: 'remote-config')
```
PerfectSlayer, line 38: I had the same remark. Can we ask to rework the `remote-config` module to make `communication` to not depend of it?

gradle/dependencies.gradle
```diff
@@ -25,8 +25,8 @@ final class CachedData {
     kotlin        : "1.3.72",
     coroutines    : "1.3.0",
     dogstatsd     : "4.0.0",
-    jnr_unixsocket: "0.28",
-    jnr_ffi       : '2.1.12', // dependency of jnr_unixsocket: update in tandem
+    jnr_unixsocket: "0.38.17",
+    jnr_ffi       : '2.2.12', // dependency of jnr_unixsocket: update in tandem
```
PerfectSlayer, line 29: Is it related related to shading?

PerfectSlayer (approved):

dd-trace-ot/dd-trace-ot.gradle
bantonsson: I think that can be done in `master` but I didn't want to wait for it.

gradle/dependencies.gradle
bantonsson: Nope, it's security related. I should separated it out into a separate PR.

test-published-dependencies/ot-is-shaded/build.gradle
```diff
@@ -0,0 +1,36 @@
...
+}
+
+def currentJavaHomePath = getJavaHomePath(System.getProperty("java.home"))
+def jarFile = configurations.jarfile.filter { it.name.startsWith('dd-trace-ot')}.singleFile
+
+tasks.register('checkJarContents', Exec) {
+  commandLine('./check_jar_contents',
+    '--command', "${currentJavaHomePath}/bin/jar",
+    '--file', "${jarFile}"
+    )
```
PerfectSlayer, line 16: Can the check be implemented in Java using ZIP file system provider?

Or at least define input of the task to no break gradle cache?

bantonsson: To not _break_ the gradle build cache, the tasks would need to produce some `output` as well, which I don't know what that would be.

I'm not going to put any effort into making this cacheable right now since these are generally only run during CI, and take on the order of 5 seconds, including launching the gradle daemon, which is less than the `publishMavenLocal` task that needs to be run before running this.

PR #3990, opened 2022-10-07
Add some info log for debugger

jpbempel:
readjust level for some logs

# What Does This Do

# Motivation

# Additional Notes

### DIFF
--- a/dd-java-agent/agent-debugger/src/main/java/com/datadog/debugger/agent/ConfigurationUpdater.java
+++ b/dd-java-agent/agent-debugger/src/main/java/com/datadog/debugger/agent/ConfigurationUpdater.java
@@ -106,7 +106,7 @@ public class ConfigurationUpdater implements DebuggerContext.ProbeResolver {
     currentConfiguration = newConfiguration;
 
     if (changes.hasProbeRelatedChanges()) {
-      log.info("apply new probe configuration, changes: {}", changes);
+      log.info("Applying new probe configuration, changes: {}", changes);
       handleProbesChanges(changes);
     }
     if (changes.hasRateLimitRelatedChanged()) {
@@ -176,7 +176,7 @@ public class ConfigurationUpdater implements DebuggerContext.ProbeResolver {
 
     // ensures that we have at least re-transformed 1 class
     if (changedClasses.size() > 0) {
-      log.debug("Re-transformation done.");
+      log.debug("Re-transformation done");
     }
   }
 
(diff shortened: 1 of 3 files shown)


dd-java-agent/agent-debugger/src/main/java/com/datadog/debugger/agent/DebuggerAgent.java
```diff
@@ -35,14 +35,12 @@ public class DebuggerAgent {
...
       Instrumentation instrumentation, SharedCommunicationObjects sco) {
-
     Config config = Config.get();
-
     if (!config.isDebuggerEnabled()) {
       log.info("Debugger agent disabled");
       return;
     }
-
+    log.info("Starting debugger...");
```
ojung, line 43: Nit: Are we considering logs that end up in the user application user-facing? Because then this should be DI I suppose.

ojung (approved):

dd-java-agent/agent-debugger/src/main/java/com/datadog/debugger/agent/DebuggerAgent.java
jpbempel: logger name contains debugger because of the package/classes, anyway...

dd-java-agent/agent-debugger/src/main/java/com/datadog/debugger/agent/ConfigurationUpdater.java
```diff
@@ -106,7 +106,7 @@ private void applyNewConfiguration(Configuration newConfiguration) {
     currentConfiguration = newConfiguration;
 
     if (changes.hasProbeRelatedChanges()) {
-      log.debug("apply new probe configuration, changes: {}", changes);
+      log.info("apply new probe configuration, changes: {}", changes);
```
cimi, line 109: s/apply/Applying

dd-java-agent/agent-debugger/src/main/java/com/datadog/debugger/agent/ConfigurationUpdater.java
```diff
@@ -176,7 +176,7 @@ private void handleProbesChanges(ConfigurationComparer changes) {
 
     // ensures that we have at least re-transformed 1 class
     if (changedClasses.size() > 0) {
-      log.info("Re-transformation done.");
+      log.debug("Re-transformation done.");
```
cimi, line 179: Nit: not sure what the convention is in this library, but usually log lines don't end with punctuation - we could remove the trailing dot.

dd-java-agent/agent-debugger/src/main/java/com/datadog/debugger/agent/DebuggerAgent.java
cimi: I agree with Oskar, it's better to say `Dynamic Instrumentation` instead of `debugger`, it's more likely the user knows what it is. Also agree that the logger name will have `Debugger` in it but usually people look at the message.

Regardless, we should remove the `...` (related to the comment above about punctuation).

PR #4002, opened 2022-10-10
jetty blocking: fix calling of old method

cataphract:
Jetty replaced Response.complete() with Response.closeOutput() some time ago. See https://github.com/eclipse/jetty.project/commit/bacff75b311


smola (changes requested):
Can we add a unit test to this module?

cataphract:
> Can we add a unit test to this module?

Done.

smola (approved):

PR #4007, opened 2022-10-11
Refactor createPredicate

jpbempel:
# What Does This Do
raise correct exception when syntax is incorrect instead of having a Json parsing exception

# Motivation
Raise correct error message when syntax in expression language is incorrect

# Additional Notes


ojung (approved):
👍

PR #4011, opened 2022-10-11
Replace summary formater by a builder

jpbempel:
# What Does This Do
build the summary message as context are added to the snapshot to use values before the freeze.

# Motivation
avoid deserializing Json to build summary

# Additional Notes


dd-java-agent/agent-debugger/debugger-bootstrap/src/main/java/datadog/trace/bootstrap/debugger/SnapshotSummaryBuilder.java
```diff
@@ -0,0 +1,152 @@
...
+
+  public void addExit(Snapshot.CapturedContext exit) {
+    if (exit == null) {
+      return;
+    }
+    if (exit.getLocals() == null) {
+      return;
+    }
+    locals = formatCapturedValues(exit.getLocals());
+    CapturedValue capturedReturnValue = exit.getLocals().get("@return");
```
cimi, line 45: Should we remove the `@return` from the locals if we print it separately?

That's something that bothered me about the current summary, we also repeat that.

dd-java-agent/agent-debugger/debugger-bootstrap/src/main/java/datadog/trace/bootstrap/debugger/Snapshot.java
```diff
@@ -37,6 +37,7 @@ public class Snapshot {
   private final transient Set<String> capturingProbeIds = new HashSet<>();
   private String traceId; // trace_id
   private String spanId; // span_id
+  private final transient SnapshotSummaryBuilder summaryBuilder;
```
cimi, line 40: Out of curiosity, since this class is not `Serializable` the transient doesn't do anything?

dd-java-agent/agent-debugger/debugger-bootstrap/src/main/java/datadog/trace/bootstrap/debugger/SnapshotSummaryBuilder.java
```diff
@@ -0,0 +1,152 @@
...
+    if (method == null) {
+      method = formatMethod(probe.getLocation());
+    }
+    sb.append(method);
+    sb.append("(");
+    sb.append(arguments);
+    sb.append(")");
+    if (returnValue != null) {
+      sb.append(": ").append(returnValue);
+    }
```
cimi, line 78: Nit: we could skip initialising arguments to the empty string and do a null check here before appending (so it's consistent with return and locals).

cimi (commented):
:+1: Thanks, looks good!

Only one comment about `@return`, the rest are nits.

I didn't check the utility methods in the builder, I assume you copied them over from the deleted class.

dd-java-agent/agent-debugger/debugger-bootstrap/src/main/java/datadog/trace/bootstrap/debugger/Snapshot.java
jpbempel: it's to avoid (De)Serialization as Json by Moshi

dd-java-agent/agent-debugger/debugger-bootstrap/src/main/java/datadog/trace/bootstrap/debugger/SnapshotSummaryBuilder.java
jpbempel: done

cimi (approved):
:+1: Thank you!

PR #4018, opened 2022-10-12
Collect in snapshot dynamic types for references

jpbempel:
# What Does This Do
Instead of collecting declared (lexical) type for argument, fields or local variable, we collect dynamic type that provides more useful information.

# Motivation
Knowing my argument is in fact a String instead of the Object as declared in the source code.

# Additional Notes

### DIFF
--- a/dd-java-agent/agent-debugger/src/main/java/com/datadog/debugger/instrumentation/MethodProbeInstrumentor.java
+++ b/dd-java-agent/agent-debugger/src/main/java/com/datadog/debugger/instrumentation/MethodProbeInstrumentor.java
@@ -283,12 +283,12 @@ public final class MethodProbeInstrumentor extends Instrumentor {
       // bail out if no args
       return;
     }
-    insnList.add(new InsnNode(Opcodes.DUP));
-    // stack: [capturedcontext, capturedcontext]
-    ldc(insnList, argTypes.length);
-    // stack: [capturedcontext, capturedcontext, int]
-    insnList.add(new TypeInsnNode(Opcodes.ANEWARRAY, CAPTURED_VALUE.getInternalName()));
-    // stack: [capturedcontext, capture, array]
+    insnList.add(new InsnNode(Opcodes.DUP)); // stack: [capturedcontext, capturedcontext]
+    ldc(insnList, argTypes.length); // stack: [capturedcontext, capturedcontext, int]
+    insnList.add(
+        new TypeInsnNode(
+            Opcodes.ANEWARRAY,
+            CAPTURED_VALUE.getInternalName())); // stack: [capturedcontext, capture, array]
     int counter = 0;
     int slot = isStatic ? 0 : 1;
     for (Type argType : argTypes) {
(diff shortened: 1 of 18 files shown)


dd-java-agent/agent-debugger/src/main/java/com/datadog/debugger/instrumentation/Types.java
```diff
@@ -344,4 +344,19 @@ public static String objectOrArrayDesc(String javaType) {
     }
     return buf.toString();
   }
+
+  public static boolean isPrimitive(String type) {
```
shatzi, line 348: why we don't treat string as primitive here as well? do we need to check dynamic type for strings?

dd-java-agent/agent-debugger/src/main/java/com/datadog/debugger/instrumentation/MethodProbeInstrumentor.java
```diff
@@ -297,39 +297,62 @@ private void collectArguments(InsnList insnList) {
...
+      ldc(insnList, currentArgName);
+      // stack: [capturedcontext, capturedcontext, array, array, int, string]
+      InsnList loadArg = new InsnList();
+      loadArg.add(new VarInsnNode(argType.getOpcode(Opcodes.ILOAD), slot));
+      collectDynamicType(argType.getClassName(), insnList, loadArg);
+      insnList.add(new VarInsnNode(argType.getOpcode(Opcodes.ILOAD), slot));
+      // stack: [capturedcontext, capturedcontext, array, array, int, string, type_name, arg]
+      tryBox(argType, insnList);
+      // stack: [capturedcontext, capturedcontext, array, array, int, type_name, object]
+      addCapturedValueOf(insnList, probe.getCapture());
```
shatzi, line 313: Nit: looks like a refactor candidate -> addCaptureValueToArray(counter++,currentArgName, loadArg) as this chunk of code appear multiple times with minor tweaks.

dd-java-agent/agent-debugger/src/main/java/com/datadog/debugger/instrumentation/MethodProbeInstrumentor.java
```diff
@@ -297,39 +297,62 @@ private void collectArguments(InsnList insnList) {
...
         insnList,
         CAPTURE_CONTEXT_TYPE,
         "addArguments",
         Type.VOID_TYPE,
-        Types.asArray(CAPTURED_VALUE, 1)); // // stack: [capturedcontext]
+        Types.asArray(CAPTURED_VALUE, 1));
+    // stack: [capturedcontext]
+  }
+
+  private void collectDynamicType(String declaredType, InsnList insnList, InsnList loadingObjNode) {
```
shatzi, line 328: while I find this very cool instrumentation. why does captureValueOf(object) doesn't do if (obj != null) obj.getClass().getType() : declaredType for every capturedValueOf.

Also, we might need to do so anyhow to give dynamic type information for deeper values. or that is already been done by serialization ?

shatzi (commented):
Very cool change. 

Maybe we can simplify by refactor captureValueOf instead of the instrumentation -> easier to reason about it and maintain.

dd-java-agent/agent-debugger/src/main/java/com/datadog/debugger/instrumentation/Types.java
jpbempel: String is not a primitive in JVM parlance, only those listed types.
String is class like (almost) any others and could be stored in an Object variable or field. So we need to get the dynamic type (hence String) from this variable while for primitives we don't need.

dd-java-agent/agent-debugger/src/main/java/com/datadog/debugger/instrumentation/MethodProbeInstrumentor.java
jpbempel: good point 👍 , I need to revisit that

shatzi (approved):
Looks awesome.

PR #4021, opened 2022-10-12
Refactor IastModule prototype

cataphract:
This is an attempt at increasing cohesion in the IAST instrumentations. The current strategy to do taint propagation goes like this:

1. A class with call site advice is created on the dd-java-agent/instrumentation/xxx module. It looks like this:
```java
@CallSite(spi = IastAdvice.class)
public class StringCallSite {
  @CallSite.After("java.lang.String java.lang.String.concat(java.lang.String)")
  public static String afterConcat(
      @CallSite.This final String self,
      @CallSite.Argument final String param,
      @CallSite.Return final String result) {
    InstrumentationBridge.onStringConcat(self, param, result);
    return result;
  }
}
  ```
 (A call to this static method is injected into the call site, i.e. the advice is not inlined).
2. There is a call on a method on the `InstrumentationBridge` class. This is a boot cl class, defined in the internal-api module. The method in question looks like like this:
```java
public static void onStringConcat(final String self, final String param, final String result) {
  try {
    if (MODULE != null) {
      MODULE.onStringConcat(self, param, result);
    }
  } catch (final Throwable t) {
    onUnexpectedException("Callback for onStringConcat threw.", t);
  }
}
```
`MODULE` here being a static field of type `IastModule` (another boot class).
3. Add `onStringConcat` to the `IastModule` interface (also defined in internal-api).
4. Now implement `onStringConcat` on the implementation of `IastModule`, named — unsurprisingly — `IastModuleImpl`. This class lives in the dd-java-agent:agent-iast module.
5. Add tests on all three modified modules.

The motivation here is to improve modularity. All the code changes should be limited to the dd-java-agent:instrumentation:xxx module in question.

The basic idea of this solution is that instead of going to the `InstrumentationBridge` boot class, the advice methods (`StringCallSite#afterConcat` in the example above) obtain a reference to the tainting method (implemented `IastModuleImpl#onStringConcat` in the example) through an invokedynamic call. This sidesteps any visibility issues resulting from the tainting method being loaded in the datadog classloader.

The bootstrap method of the invokedynamic call is implemented in a bootstrap classloader class named `CallSiteHelperRegistry`. When IAST starts, it registers in this class all of the tainting methods (called "call site helpers" for a lack of better naming). It looks for them by looking at methods annotated with `@CallSiteHelper` inside classes implementing the marker interface `CallSiteHelperContainer`.

Step one from the status quo description is unchanged; `StringCallSite` in the example above would look the same. However, its invokestatic call needs to be changed to an appropriate invokedynamic call. This is done with a compile time instrumentation implemented in `CsiAdviceTransformationPlugin`.


dd-java-agent/agent-tooling/src/main/java/datadog/trace/agent/tooling/bytebuddy/csi/CsiAdviceTransformationPlugin.java
```diff
@@ -0,0 +1,173 @@
...
+
+      Handle bootstrapHandle =
+          new Handle(
+              Opcodes.H_INVOKESTATIC,
+              "datadog/trace/api/iast/CallSiteHelperRegistry",
+              "bootstrap",
+              "(Ljava/lang/invoke/MethodHandles$Lookup;Ljava/lang/String;Ljava/lang/invoke/MethodType;Ljava/lang/String;)Ljava/lang/invoke/CallSite;",
+              false);
+      String helperKey = owner.replace('/', '.') + '\0' + name;
+      super.visitInvokeDynamicInsn(name, descriptor, bootstrapHandle, helperKey);
```
manuel-alvarez-alvarez, line 147: Can you do a benchmark to compare the overhead for invoke static vs dynamic (ideally with different JDKs)? call sites tend to be massive and every ns counts :smile:

dd-java-agent/agent-tooling/src/main/java/datadog/trace/agent/tooling/bytebuddy/csi/CsiAdviceTransformationPlugin.java
```diff
@@ -0,0 +1,173 @@
...
+import net.bytebuddy.implementation.Implementation;
+import net.bytebuddy.jar.asm.ClassVisitor;
+import net.bytebuddy.jar.asm.Handle;
+import net.bytebuddy.jar.asm.MethodVisitor;
+import net.bytebuddy.jar.asm.Opcodes;
+import net.bytebuddy.matcher.ElementMatchers;
+import net.bytebuddy.pool.TypePool;
+import net.bytebuddy.utility.OpenedClassReader;
+
+public class CsiAdviceTransformationPlugin extends Plugin.ForElementMatcher {
```
manuel-alvarez-alvarez, line 27: I like this bytebuddy plugins, but don't hesitate to change the gradle plugin directly to generate the proper java sources.

dd-java-agent/agent-iast/src/main/java/com/datadog/iast/IastSystem.java
```diff
@@ -29,14 +37,25 @@ public static void start(final SubscriptionService ss) {
...
+    initializeCallSiteHelperRegistry();
     final IastModule iastModule = new IastModuleImpl(config, reporter, overheadController);
     InstrumentationBridge.registerIastModule(iastModule);
     registerRequestStartedCallback(ss, overheadController);
     registerRequestEndedCallback(ss, overheadController);
   }
 
+  private static void initializeCallSiteHelperRegistry() {
+    CallSiteHelperRegistry.reset();
+    ServiceLoader<CallSiteHelperContainer> callSiteHelperContainers =
```
manuel-alvarez-alvarez, line 52: Call sites are registered via SPI as well (which I don't like much) if we lean towards registering the helpers as well, it might be better to invest some time in doing some compile time injection when building the shadow jar

internal-api/src/main/java/datadog/trace/api/iast/CallSiteHelperRegistry.java
```diff
@@ -0,0 +1,230 @@
...
+              0,
+              MethodHandles.publicLookup()
+                  .findVirtual(Throwable.class, "getClass", MethodType.methodType(Class.class)));
+
+    } catch (NoSuchMethodException | IllegalAccessException e) {
+      throw new UndeclaredThrowableException(e);
+    }
+  }
+
+  private static MethodHandle makeExceptionSafe(
```
manuel-alvarez-alvarez, line 140: I like this approach for ensuring safety of the call sites, but can't we achieve the same via bytebuddy plugin and remove the overhead penalty at startup?

dd-java-agent/agent-tooling/src/main/java/datadog/trace/agent/tooling/bytebuddy/csi/CsiAdviceTransformationPlugin.java
cataphract: I will, but for the comparison to be apples to apples we need to compare against invokestatic + getfield + invokeinterface

dd-java-agent/agent-iast/src/main/java/com/datadog/iast/IastSystem.java
cataphract: Yeah, my main problem with it is that we need to instantiate the service classes, which actually just contain static methods.

dd-java-agent/agent-tooling/src/main/java/datadog/trace/agent/tooling/bytebuddy/csi/CsiAdviceTransformationPlugin.java
manuel-alvarez-alvarez: Yep, sure it makes no sense otherwise.

 Something that worries me a bit as well is startup time, initial resolution of the invoke dynamic call sites might have a big impact in an application with lots of targets for our instrumentation.

internal-api/src/main/java/datadog/trace/api/iast/CallSiteHelperRegistry.java
```diff
@@ -0,0 +1,230 @@
...
+import java.lang.reflect.Array;
+import java.lang.reflect.Method;
+import java.lang.reflect.Modifier;
+import java.lang.reflect.UndeclaredThrowableException;
+import java.util.HashMap;
+import java.util.Map;
+import org.slf4j.Logger;
+import org.slf4j.LoggerFactory;
+
+public class CallSiteHelperRegistry {
```
manuel-alvarez-alvarez, line 17: Perhaps `CallSiteHelperFactory` is a better name to be aligned with JDK's invoke dynamics

dd-java-agent/agent-tooling/src/main/java/datadog/trace/agent/tooling/bytebuddy/csi/CsiAdviceTransformationPlugin.java
cataphract: I only tested on my laptop, which is unpredictable due to throttling, but invokedynamic seems to have better performance. https://gist.github.com/cataphract/5db5c9f41cf9f067401d73adf1bad0a8#file-results-txt

cataphract: I've pushed the benchmark, in case you want to run it with `gradle :internal-api:jmh -Pinclude='.*HelperInvocationBenchmark'`

dd-java-agent/agent-iast/src/main/java/com/datadog/iast/IastSystem.java
cataphract: I've changed this in the last commit to read the file with services directly. This avoids instantiating the containers. AFAIC this eliminates my main concern, though ofc there's still space for improvement.

dd-java-agent/agent-tooling/src/main/java/datadog/trace/agent/tooling/bytebuddy/csi/CsiAdviceTransformationPlugin.java
cataphract: I've renamed the stuff to expand the scope beyond csi, so I think it makes more sense to have it where it is now then.

internal-api/src/main/java/datadog/trace/api/iast/CallSiteHelperRegistry.java
cataphract: It could be done of course, it's just more trouble. The problem is that I'm not seeing how to use the bytebuddy annotation advice api to do it, since we the support for around advice involves a fallback method handle being passed to the registry. So probably I'd need to do it with asm visitors or (even worse) byte buddy's ravioli api.

I'm inclined to leave this as is for now, but can be persuaded otherwise.

internal-api/src/main/java/datadog/trace/api/iast/CallSiteHelperRegistry.java
cataphract: I think the current name is more appropriate insofar as the class acts primary as registry (map) for method handles under a certain "helper key". You call `registerHelper` to populate the map and `bootstrap` to fetch method handles from the map. The fact that it also transforms the helper methods to some extent (to make them exception-safe) is a bit of an incidental feature.

dougqh: While I understand the motivation behind using invokedynamic, it is going create challenges if you want to support this on Graal native-image in the future.  In general, I think we should be moving away from dynamicism -- not towards it.

cataphract: OK, I'll change more of the instrumentation to compile time.

cataphract: I've changed the instrumentation of the helpers to be done at compile time. Ultimately, in circumstances where we don't care about our advice being reachable by the classloader of the class of the call site instrumentation, the invokedynamic substitution doesn't need to be made.

dd-java-agent/agent-tooling/src/main/java/datadog/trace/agent/tooling/bytebuddy/InvokeDynamicHelpersPlugin.java
```diff
@@ -0,0 +1,521 @@
...
+    }
+
+    private int numArguments() {
+      return this.argTypes.length;
+    }
+
+    private int numLocalArgSlots() {
+      int slots = 0;
+      for (Type type : this.argTypes) {
+        if (type.equals(Type.LONG_TYPE) || type.equals(Type.LONG_TYPE)) {
```
cataphract, line 481: note to self
```suggestion
        if (type.equals(Type.LONG_TYPE) || type.equals(Type.DOUBLE_TYPE)) {
```

dd-java-agent/instrumentation/java-lang/src/main/java/datadog/trace/instrumentation/java/lang/StringHelperContainer.java
```diff
@@ -0,0 +1,262 @@
...
+public class StringHelperContainer implements InvokeDynamicHelperContainer {
+
+  private static final Object[] EMPTY = new Object[0];
+
+  private static final Pattern FORMAT_PATTERN =
+      Pattern.compile("%(\\d+\\$)?([-#+ 0,(\\<]*)?(\\d+)?(\\.\\d+)?([tT])?([a-zA-Z%])");
+
+  /** @see String#format(String, Object...) */
+  @InvokeDynamicHelper(
+      fallbackMethod =
```
manuel-alvarez-alvarez, line 35: Can we _inherit_ the fallback method from the call site definition?

manuel-alvarez-alvarez:
When dealing with sinks (for instance  `WeakHashInstrumentationCallSite`) we need access to the `OverheadControler` and the `Reporter` in order to check if we should report a vulnerability and actually do it. Since `InvokeDynamicHelperContainer` are created via SPI there is no ease solution to wire dependencies, in a quick and dirty example I prepared I passed them along the `IastRequestContext` but don't like it much:

```java
  @InvokeDynamicHelper
  public static void onHashingAlgorithm(@Nullable final String algorithm) {
    if (algorithm == null) {
      return;
    }
    final IastRequestContext context = IastRequestContext.get();
    if (context == null) {
      return;
    }
    final String algorithmId = algorithm.toUpperCase(Locale.ROOT);
    if (!Config.get().getIastWeakHashAlgorithms().contains(algorithmId)) {
      return;
    }
   
    final AgentSpan span = AgentTracer.activeSpan();
    // [*] Wiring of the overhead controller 
    if (!context.getOverheadController().consumeQuota(Operations.REPORT_VULNERABILITY, span)) {
      return;
    }
    StackTraceElement stackTraceElement = currentStack();

    Vulnerability vulnerability =
        new Vulnerability(
            VulnerabilityType.WEAK_HASH,
            Location.forSpanAndStack(span.getSpanId(), stackTraceElement),
            new Evidence(algorithm));
    
    // [*] Wiring of the reporter
    context.getReporter().report(span, vulnerability);
  }
```
Do you have another way of wiring dependencies in mind?

manuel-alvarez-alvarez:
Can you have a look on how to apply the new helpers to the `StringConcatFactoryCallSite`?  the bytebbudy plugin currently fails due to JDK9 APIs and I'm not sure the current approach can be applied easily to invoke dynamic call sites.

dd-java-agent/agent-iast/src/main/java/com/datadog/iast/IastSystem.java
```diff
@@ -49,4 +74,29 @@ private static void registerRequestEndedCallback(
...
+      r.lines()
+          .map(
+              line -> {
+                try {
+                  return (Class<? extends T>) cl.loadClass(line);
+                } catch (ClassNotFoundException e) {
+                  throw new UndeclaredThrowableException(e);
+                }
+              })
+          .forEachOrdered(consumer);
```
manuel-alvarez-alvarez, line 97: If the `map` or `forEachOrdered` lambdas throw and exception the iteration over the `InvokeDynamicHelperContainer` leaving the registry only partially initialized.

dd-java-agent/agent-tooling/src/main/java/datadog/trace/agent/tooling/bytebuddy/InvokeDynamicHelpersPlugin.java
```diff
@@ -0,0 +1,521 @@
...
+   *   } catch (Throwable e) {
+   *     if (e instanceof RealCallThrowable) {
+   *       throw StackUtils.filterDatadog(e.getCause());
+   *     }
+   *     LOG$.warn("Helper XXX has thrown", e);
+   *     return fallbackMethod(args...);
+   *   }
+   * </pre>
+   *
+   * or, if there is no fallback method:
```
manuel-alvarez-alvarez, line 272: We should validate that only `Around` advices can have a fallbackMethod and all `Around` advices should declare the fallback method. 

If for any reason something fails and we don't have a fallback method then we have to break the customer app, returning some default values that might make their application behave unexpectedly and this is something we cannot afford.

dd-java-agent/agent-tooling/src/main/java/datadog/trace/agent/tooling/bytebuddy/InvokeDynamicHelpersPlugin.java
```diff
@@ -0,0 +1,521 @@
...
+        issueDefaultReturn();
+      }
+      super.visitLabel(endFunctionLabel);
+
+      for (LocalVariableVisitation lvVisitation : this.lvVisitations) {
+        lvVisitation.visit(mv);
+      }
+
+      int numLocalArgsSlots = numLocalArgSlots();
+      int ourMaxStack = Math.max(2, numLocalArgsSlots);
```
manuel-alvarez-alvarez, line 380: This is wrong, our max stack is actually 3 instead of two:
```
LOG_INSTANCE
LDC
THROWABLE
```

internal-api/src/main/java/datadog/trace/api/iast/InvokeDynamicHelperRegistry.java
```diff
@@ -0,0 +1,76 @@
...
+      } catch (IllegalAccessException | NoSuchMethodException e) {
+        throw new UndeclaredThrowableException(e);
+      }
+    }
+  }
+
+  private static void registerMethod(MethodHandles.Lookup lookup, Method method)
+      throws IllegalAccessException, NoSuchMethodException {
+    if (!Modifier.isStatic(method.getModifiers())) {
+      throw new IllegalStateException("The helper method " + method + " should be static");
```
manuel-alvarez-alvarez, line 35: It's better to validate this at compile time rather than at runtime, although I don't think it's easy.

internal-api/src/main/java/datadog/trace/api/iast/InvokeDynamicHelperRegistry.java
```diff
@@ -0,0 +1,76 @@
...
+    }
+
+    MethodHandle helperMH = lookup.unreflect(method);
+    Class<?> helperClass = method.getDeclaringClass();
+    String helperMethodName = method.getName();
+
+    String helperKey = calculateHelperKey(helperClass, helperMethodName);
+    if (HELPERS.containsKey(helperKey)) {
+      throw new IllegalStateException(
```
manuel-alvarez-alvarez, line 44: Same as previous comment.

dd-java-agent/instrumentation/java-lang/src/main/java/datadog/trace/instrumentation/java/lang/StringHelperContainer.java
```diff
@@ -0,0 +1,262 @@
...
+          "java/lang/String.format(Ljava/util/Locale;Ljava/lang/String;[Ljava/lang/Object;)Ljava/lang/String;")
+  public static String onStringFormat(
+      @Nullable Locale l, @Nonnull String fmt, @Nullable Object[] args) {
+    if (fmt == null) {
```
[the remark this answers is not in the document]
cataphract: Well there are two advantages:

* You can invalidate the call site is iast is enabled/disabled at runtime.
* You don't need to inject the helpers on every classloader that uses them.

cataphract:
> When dealing with sinks (for instance `WeakHashInstrumentationCallSite`) we need access to the `OverheadControler` and the `Reporter` in order to check if we should report a vulnerability and actually do it. Since `InvokeDynamicHelperContainer` are created via SPI there is no ease solution to wire dependencies, in a quick and dirty example I prepared I passed them along the `IastRequestContext` but don't like it much:
> 
> ```java
>   @InvokeDynamicHelper
>   public static void onHashingAlgorithm(@Nullable final String algorithm) {
>     if (algorithm == null) {
>       return;
>     }
>     final IastRequestContext context = IastRequestContext.get();
>     if (context == null) {
>       return;
>     }
>     final String algorithmId = algorithm.toUpperCase(Locale.ROOT);
>     if (!Config.get().getIastWeakHashAlgorithms().contains(algorithmId)) {
>       return;
>     }
>    
>     final AgentSpan span = AgentTracer.activeSpan();
>     // [*] Wiring of the overhead controller 
>     if (!context.getOverheadController().consumeQuota(Operations.REPORT_VULNERABILITY, span)) {
>       return;
>     }
>     StackTraceElement stackTraceElement = currentStack();
> 
>     Vulnerability vulnerability =
>         new Vulnerability(
>             VulnerabilityType.WEAK_HASH,
>             Location.forSpanAndStack(span.getSpanId(), stackTraceElement),
>             new Evidence(algorithm));
>     
>     // [*] Wiring of the reporter
>     context.getReporter().report(span, vulnerability);
>   }
> ```
> 
> Do you have another way of wiring dependencies in mind?

We could do the injection when the helpers are registered (either by writing a field or doing the injection in the arguments of a static method by returning a MethodHandle where this argument is prepopulated).

smola:
Closing, since we do not expect to work on this short term. We may revisit this approach in the future though.

PR #4025, opened 2022-10-13
Sanitize env version in remote config request

shatzi:
# What Does This Do

When sending request to remote config we ensure service name, env, version are tokenized and sanitized 

# Motivation

env or version need to be tokenized based on DD-rules to ensure both tracer and datadog backend use the same values to match predicates. 

# Additional Notes

Also removed duplicated TagsHelper implementation.


mellon85 (commented):
LGTM

dd-trace-core/src/main/java/datadog/trace/core/StatusLogger.java
```diff
@@ -101,6 +101,10 @@ public void toJson(JsonWriter writer, Config config) throws IOException {
     writer.value(config.isLogsInjectionEnabled());
     writer.name("profiling_enabled");
     writer.value(config.isProfilingEnabled());
+    writer.name("remote_config_enabled");
+    writer.value(config.isRemoteConfigEnabled());
+    writer.name("debugger_enabled");
+    writer.value(config.isDebuggerEnabled());
```
jpbempel, line 107: Thanks!

jpbempel (approved):

PR #4026, opened 2022-10-14
Disable `jdk.ClassLoaderStatistics` unless running on JDK17 or later

richardstartin:
# What Does This Do

Simplifies the event checks in `OpenJdkController` and disables `jdk.ClassLoaderStatistics` on older JVMs

# Motivation

Prior to JDK15, `jdk.ClassLoaderStatistics` was expensive, but it was fixed in [JDK-8244733](https://bugs.openjdk.org/browse/JDK-8244733). The first LTS to support this event efficiently was JDK17, so disable the event before JDK17.

# Additional Notes

### DIFF
--- a/dd-java-agent/agent-profiling/profiling-controller-openjdk/src/main/java/com/datadog/profiling/controller/openjdk/OpenJdkController.java
+++ b/dd-java-agent/agent-profiling/profiling-controller-openjdk/src/main/java/com/datadog/profiling/controller/openjdk/OpenJdkController.java
@@ -16,12 +16,14 @@
 package com.datadog.profiling.controller.openjdk;
 
 import static com.datadog.profiling.controller.ProfilingSupport.*;
+import static datadog.trace.api.Platform.isJavaVersionAtLeast;
 
 import com.datadog.profiling.controller.ConfigurationException;
 import com.datadog.profiling.controller.Controller;
 import com.datadog.profiling.controller.UnsupportedEnvironmentException;
 import com.datadog.profiling.controller.jfr.JfpUtils;
 import com.datadog.profiling.controller.openjdk.events.AvailableProcessorCoresEvent;
+import datadog.trace.api.Platform;
 import datadog.trace.api.config.ProfilingConfig;
 import datadog.trace.bootstrap.config.provider.ConfigProvider;
 import de.thetaphi.forbiddenapis.SuppressForbidden;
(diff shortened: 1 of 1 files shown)


dd-java-agent/agent-profiling/profiling-controller-openjdk/src/main/java/com/datadog/profiling/controller/openjdk/OpenJdkController.java
```diff
@@ -67,30 +68,35 @@ public OpenJdkController(final ConfigProvider configProvider)
...
+        && isEventEnabled(recordingSettings, "jdk.NativeMethodSample")) {
       log.debug(
           "Inexpensive native profiling is not supported for this JDK. "
               + "Disabling NativeMethodSample JFR event.");
       recordingSettings.put("jdk.NativeMethodSample#enabled", "false");
     }
 
+    if (!isJavaVersionAtLeast(17)
+        && isEventEnabled(recordingSettings, "jdk.ClassLoaderStatistics")) {
+      recordingSettings.put("jdk.ClassLoadingStatistics#enabled", "true");
```
jbachorik: ?
```suggestion
      recordingSettings.put("jdk.ClassLoadingStatistics#enabled", "false");
```

richardstartin: sorry, typo, I've just DRYed this out a bit

jbachorik (approved):

ghostdogpr:
Spent the last few days debugging an issue that led me exactly to this problem, what a nice timing! 🙏

dd-java-agent/agent-profiling/profiling-controller-openjdk/src/main/java/com/datadog/profiling/controller/openjdk/OpenJdkController.java
```diff
@@ -163,46 +157,17 @@ public OpenJdkOngoingRecording createRecording(final String recordingName)
...
-  // jdk.ObjectAllocationOutsideTLAB
-
-  boolean isObjectAllocationOutsideTLABEnabledInRecordingSettings(
-      Map<String, String> recordingSettings) {
-    return Boolean.parseBoolean(recordingSettings.get("jdk.ObjectAllocationOutsideTLAB#enabled"));
+  private static void disableEvent(Map<String, String> recordingSettings, String event) {
+    String wasEnabled = recordingSettings.put(event + "#enabled", "false");
+    if (Boolean.parseBoolean(wasEnabled)) {
+      log.debug(
+          "Disabling JFR event {} because it is expensive on the current version of the JVM ({}).",
```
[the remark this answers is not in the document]
jbachorik: Not sure. Considering that this will almost never be visible, it would be better to have this information in the insights/analysis - and I think we have this and it is actually showing up with a nice exclamation mark and yet no customer has ever noticed that before filing an escalation 🤷

PR #4031, opened 2022-10-15
Single Span Sampling

ygree:
# What Does This Do

**NOTE: there is another PR on top of this #3982 that is a refactoring for Sampler and related interface signatures that moves type parameter from the class declaration into the methods.**

Implementation of the single span sampling mechanism described in [this RFC](https://github.com/DataDog/architecture/tree/master/rfcs/apm/integrations/single-span-ingestion-control#recommended-solution-span-sampling-rules).

It introduces SpanProcessingWorker running on its own thread. It is responsible for sampling single spans. It receives unsampled traces from PrioritizationStrategy after that it applies matching rules with regard to the associated rate limiter if any. And then all sampled and unsampled (unless dropping policy is active) spans are sent to TraceProcessingWorker for serialization and transporting to the Agent. 

There are two new Health Metrics:
- `queue.partial.traces` - number of enqueued traces with some unsampled spans dropped
- `queue.dropped.spans` - total number of dropped spans

---

Next diagram shows how this new parts (in the rectangle) integrate into the existing functionality:

```mermaid
graph TD

PS[PrioritizationStrategy]
pQ[primaryQueue]
sQ[secondaryQueue]
TPW[TraceProcessingWorker]
subgraph  
  tQ[traceQueue]
  SPW{SpanProcessingWorker}
  dp{DroppingPolicy}
  or{Is Single Span Sampling enabled?}
end

PS--sampled traces-->pQ
PS--unsampled traces-->or
or--yes-->tQ
or--no-->sQ

pQ--consumed by-->TPW
sQ--consumed by-->TPW

tQ--->SPW
SPW--sampled spans-->pQ
SPW--unsampled spans-->dp
dp--non active-->sQ
```

---

Next diagram shows how new and old parts are constructed:

```mermaid
graph TD

CTB[CoreTracerBuilder]
CT[CoreTracer]
WF[WriterFactory]
TPW[TraceProcessingWorker]
PS[PrioritizationStrategy]
RW[RemoteWriter]
AW[DDAgentWriter]
IW[DDIntakeWriter]
pQ[primaryQueue]
sQ[secondaryQueue]
DP[DroppingPolicy]

subgraph  
  SPW[SpanProcessingWorker]
  SSS[SingleSpanSampler]
end

CTB--creates-->CT
CTB--creates from config-->SSS

CT--creates or uses-->WF

WF--creates-->AW
WF--creates-->IW

RW--creates-->TPW
RW--creates-->DP

AW--extends-->RW
IW--extends-->RW

SPW--uses-->SSS

TPW--creates-->PS
TPW--creates-->SPW

PS--uses-->DP

TPW--creates-->pQ
TPW--creates-->sQ

SPW--uses-->pQ
SPW--uses-->sQ
SPW--uses-->DP

```

# Motivation

This is a replacement for existing Analytics events that have been deprecated for a long time.

# Additional Notes

- [x] add missing tests
- [x] implement necessary health metrics (`partialTraces`, `droppedSpans`)
- [x] glob matching optimization for cases when only glob pattern presents - `*` or no glob pattern `*` or `?` present - `exact match`
- [x] Add a test case in `TraceProcessingWorkerTest` for when `droppingPolicy` is inactive
- [x] Add test cases in `SpanSamplingWorkerTest` when `secondaryQueue` is full or `droppingPolicy` is active


dd-trace-core/src/main/java/datadog/trace/common/sampling/DeterministicSampler.java
```diff
@@ -1,12 +1,39 @@
...
 
 /**
  * This implements the deterministic sampling algorithm used by the Datadog Agent as well as the
  * tracers for other languages
  */
-public class DeterministicSampler<T extends CoreSpan<T>> implements RateSampler<T> {
+public abstract class DeterministicSampler<T extends CoreSpan<T>> implements RateSampler<T> {
+
+  /** Uses trace-id as a sampling id */
+  public static final class TraceSampler<T extends CoreSpan<T>> extends DeterministicSampler<T> {
```
ygree, line 13: Is it Okay to change existing core classes? Can we break customers' code with such changes?

dd-trace-core/src/main/java/datadog/trace/common/sampling/DeterministicSampler.java
```diff
@@ -1,12 +1,39 @@
...
 
+import datadog.trace.api.DDId;
 import datadog.trace.core.CoreSpan;
 
 /**
  * This implements the deterministic sampling algorithm used by the Datadog Agent as well as the
  * tracers for other languages
  */
-public class DeterministicSampler<T extends CoreSpan<T>> implements RateSampler<T> {
+public abstract class DeterministicSampler<T extends CoreSpan<T>> implements RateSampler<T> {
```
ygree, line 10: Ideally, I'd rather move the type parameter along with its constraint into the method declaration like so #3982

dd-trace-core/src/main/java/datadog/trace/common/writer/ddagent/Prioritization.java
```diff
@@ -62,48 +71,82 @@ protected void blockingOffer(final Queue<Object> queue, final Object event) {
...
-        return primary.offer(trace);
+        return primary.offer(trace)
+            ? PublishResult.ENQUEUED_FOR_SERIALIZATION
+            : PublishResult.DROPPED_BUFFER_OVERFLOW;
       }
       switch (priority) {
         case SAMPLER_DROP:
         case USER_DROP:
-          return !droppingPolicy.active() && secondary.offer(trace);
+          if (spanSampling != null) {
```
ygree, line 134: Should we respect droppingPolicy when Single Span Sampling is enabled?

dd-trace-core/src/main/java/datadog/trace/common/sampling/DeterministicSampler.java
dougqh: I'm not sure.  Do we expose a configuration that let's someone set this?

dd-trace-core/src/main/java/datadog/trace/common/writer/ddagent/Prioritization.java
dougqh: I'm curious to hear what the other tracers did, but I think we have 3 cases...
- send dropped traces - always enqueue
- skip dropped traces - no single span rules - never enqueue
- skip dropped traces w/ single span rules - enqueue, but process differently - go through the chunk and only send the single span selected ones

dd-trace-core/src/main/java/datadog/trace/core/util/GlobPattern.java
```diff
@@ -0,0 +1,56 @@
+package datadog.trace.core.util;
+
+import java.util.regex.Pattern;
+
+public final class GlobPattern {
+
+  public static String globToRegex(String globPattern) {
```
dougqh, line 7: Maybe we should make this private, so we encourage using the compiled form instead.
I don't feel strongly about this.

ygree: Good point! I'll make it private.

dd-trace-core/src/main/java/datadog/trace/common/writer/ddagent/Prioritization.java
ygree: Yeah, I've already reworked this part to send unsampled spans unless the dropping policy is active. Found this inconsistency with system-tests.

dd-trace-core/src/main/java/datadog/trace/common/sampling/RuleBasedSampler.java
```diff
@@ -16,6 +16,7 @@
 import org.slf4j.Logger;
 import org.slf4j.LoggerFactory;
 
+// TODO maybe rename it to TraceSampler?
```
PerfectSlayer, line 19: Agreed :+1: If we could prefix what is related to trace sampling with `Trace`, is would be easier to distinguish what is trace from span sampling.

dd-trace-core/src/main/java/datadog/trace/common/sampling/SingleSpanSampler.java
```diff
@@ -0,0 +1,93 @@
...
+      boolean spanSamplingRulesDefined = spanSamplingRules != null && !spanSamplingRules.isEmpty();
+      boolean spanSamplingRulesFileDefined =
+          spanSamplingRulesFile != null && !spanSamplingRulesFile.isEmpty();
+
+      if (spanSamplingRulesDefined && spanSamplingRulesFileDefined) {
+        log.warn(
+            "Both {} and {} defined. {} will be ignored.",
+            SPAN_SAMPLING_RULES,
+            SPAN_SAMPLING_RULES_FILE,
+            SPAN_SAMPLING_RULES);
```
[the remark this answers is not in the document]
ygree: Good catch! Thanks!

dd-trace-core/src/main/java/datadog/trace/common/sampling/RuleBasedSampler.java
ygree: Okay. I've pushed this renaming to #3982

PerfectSlayer (approved):

dd-trace-core/src/main/java/datadog/trace/common/sampling/RuleBasedSampler.java
PerfectSlayer: Thanks!

PR #4038, opened 2022-10-17
Change the effect of the trace.client-ip-header.disabled option

cataphract:
This option now:
* Disables collection of the forwarding/client ip address headers, plus that of the User-Agent header.
* Disables the calculation of the value for http.client_ip, and prevents that tag from being reported. The exception being if a custom IP address is configured through trace.client-ip-header. In that case, that header will be used to report http.client_ip.
* This distinguishes the new meaning of the option from the previous one, which just disabled custom IP address header collection/processing (even if trace.client-ip-header was given).


smola (changes requested):
Looks good, but it currently fails system tests:

```
./tests/test_standard_tags.py::Test_StandardTagsClientIp
```

```
runner           | FAILED tests/test_standard_tags.py::Test_StandardTagsClientIp::test_client_ip - 2 fails
runner           | FAILED tests/test_standard_tags.py::Test_StandardTagsClientIp::test_client_ip_with_appsec_event - 1 fails
```

See https://github.com/DataDog/system-tests/pull/582

smola:
It seems we have to update some test helpers:
```
	at datadog.trace.agent.test.base.HttpServerTest.test forwarded request(HttpServerTest.groovy:454)
Caused by: java.lang.AssertionError: Tag "http.client_ip": "null" != "1.2.3.4". Expression: (value == expected). Values: value = null, expected = 1.2.3.4
	at datadog.trace.agent.test.asserts.TagsAssert.tag(TagsAssert.groovy:103)
	at datadog.trace.agent.test.asserts.TagsAssert.methodMissing(TagsAssert.groovy:116)
	at datadog.trace.agent.test.base.HttpServerTest.serverSpan_closure5$_closure32(HttpServerTest.groovy:1190)
```

bantonsson (changes requested):
This is a breaking change to a behavior that has been there since April 2021, and there is no way to get back the tags without enabling AppSec.

Maybe we could switch the default of the `trace.client-ip-header` option and think about when that would be overridden by enabling AppSec instead.

dd-java-agent/testing/src/main/groovy/datadog/trace/agent/test/base/WithHttpServer.groovy
```diff
@@ -63,7 +64,7 @@ abstract class WithHttpServer<SERVER> extends AgentTestRunner {
     println "$server started at: $address"
   }
 
-  def cleanupSpec() {
+  void cleanupSpec() {
```
smola, line 67: Ideally we should keep these superfluous changes out of PRs.

cataphract:
> This is a breaking change to a behavior that has been there since April 2021, and there is no way to get back the tags without enabling AppSec.
> 
> Maybe we could switch the default of the `trace.client-ip-header` option and think about when that would be overridden by enabling AppSec instead.

I've added a new option (`trace.client-ip-without-appsec`) to restore the previous behavior.

dd-java-agent/testing/src/main/groovy/datadog/trace/agent/test/base/WithHttpServer.groovy
bantonsson: 💯

internal-api/src/main/java/datadog/trace/api/Config.java
```diff
@@ -1104,6 +1103,9 @@ && isJavaVersionAtLeast(8)
     }
     telemetryHeartbeatInterval = telemetryInterval;
 
+    this.clientIpWithoutAppSec =
+        configProvider.getBoolean(CLIENT_IP_ENABLED, DEFAULT_CLIENT_IP_ENABLED);
+
```
bantonsson, line 1108: Config name and setting name should be similar

dd-java-agent/agent-bootstrap/src/main/java/datadog/trace/bootstrap/instrumentation/decorator/HttpServerDecorator.java
```diff
@@ -108,27 +111,32 @@ public AgentSpan onRequest(
       final REQUEST request,
       final AgentSpan.Context.Extracted context) {
     Config config = Config.get();
+    boolean clientIpResolverEnabled =
+        config.isClientIpWithoutAppSec()
+            || config.isTraceClientIpResolverEnabled() && ActiveSubsystems.APPSEC_ACTIVE;
```
bantonsson, line 116: I think that this conflates two separate things:

1.  Collecting the headers
2. Running the `clientIpResolver` which is more costly

There is still no way to only get the headers like before.

cataphract: It's true enough that those two things are being controlled together, but the default behavior before was to both collect the headers and run the resolver. The purpose of the setting `isClientIpWithoutAppSec` (now renamed) is to restore the previous default behavior. (The cost of the resolver can also be kept in check by setting a custom ip header.)

bantonsson: Yes, you are correct. I was under the impression that the resolver was not run.

bantonsson (approved):

smola (approved):

bantonsson:
Please tag this as a braking change and how to enable the old behavior in the PR description to make writing release notes easy.

Issue #3812, opened 2022-09-08
Kafka client traces missing when using spliterator


LoicTosser:
Hello,

Our Java application uses the java Kafka client library to consume Kafka events.
To listen to events, we use the code below:
```
var records = kafkaConsumer.poll(Duration.ofSeconds(1));
var recordsByPartition = StreamSupport.stream(records.spliterator(), false)
                                                      .collect(groupingBy(ConsumerRecord::partition));
```
In Datadog, the Kafka consumer traces are missing.

I saw that only the `KafkaConsumers.iterator` method is instrumented, not the `spliterator()` one. Still, I don't know if it's the root cause of this issue because the `spliterator()` calls the `iterator()` method.

When using spring-kafka, the Kafka client traces are correctly taken into account in Datadog.

Our app:
Spring-boot 2.7.3
Java 17
Kafka-client 2.8.0
dd-java-agent 0.107.1

PR #4039, opened 2022-10-17
Instrumentation for Spliterator in kafka-clients-0.11

kr-igor:
# What Does This Do
This PR address the gap in coverage for Kafka instrumentation. 
See [this issue](https://github.com/DataDog/dd-trace-java/issues/3812) for details.


hokitam (approved):

devinsba (commented):
I'm not 100% sure but I think the files you moved need to stay in the `java` folder instead of `java8` because they need to be compiled to java 7 bytecode support. I think theres a trick to make the `java8` code link against `java`

kr-igor:
@devinsba thanks for reviewing! I was comparing to other instrumentations and could not find any relevant examples. What would be the best way to validate that this PR brakes Java 7 support? Should tests catch such issues?
cc: @mcculls

PR #4042, opened 2022-10-18
Delay feature discovery in configuration poller

smola:
# What Does This Do

# Motivation

# Additional Notes

### DIFF
--- a/communication/src/main/java/datadog/communication/ddagent/SharedCommunicationObjects.java
+++ b/communication/src/main/java/datadog/communication/ddagent/SharedCommunicationObjects.java
@@ -85,14 +85,7 @@ public class SharedCommunicationObjects {
       configUrlSupplier = new FixedConfigUrlSupplier(remoteConfigUrl);
     } else {
       createRemaining(config);
-      DDAgentFeaturesDiscovery fd = featuresDiscovery(config);
-      String configEndpoint = fd.getConfigEndpoint();
-      if (configEndpoint != null) {
-        remoteConfigUrl = featuresDiscovery.buildUrl(configEndpoint).toString();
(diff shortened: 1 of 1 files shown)


communication/src/main/java/datadog/communication/ddagent/SharedCommunicationObjects.java
```diff
@@ -139,35 +132,43 @@ public String get() {
...
       if (elapsed > retryInterval) {
-        this.featuresDiscovery.discover();
-        retryInterval = 60000;
+
+        if (sco.featuresDiscovery == null) {
+          final DDAgentFeaturesDiscovery fd = sco.featuresDiscovery(config);
+        } else {
+          sco.featuresDiscovery.discover();
+          retryInterval = 60000;
+        }
```
jpbempel, line 161: why not just call:
```
sco.featuresDiscovery.discover();
```
?

smola: I'm not sure I follow. As opposed to? `sco.featuresDiscovery(config)` calls feature discovery internally, I don't repeat that for the first call. It's also a dirty way to distinguish the first request in the base case (which gets a retry interval of 5s).

smola: Discussed offline. It would be better to separate initialization and discover calls. I'm looking into it. However, ensuring that it's always called once requires changing the logic in quite a few places.

jpbempel (approved):
Ok for now, but `featureDiscovery()` needs to be refactored to separate instantiation and discovery in itself

bantonsson (approved):

PR #4044, opened 2022-10-18
Upgrade async-profiler, simplify tracing context filtering

richardstartin:
# What Does This Do

Reuses the same config flag for tracing context enablement, and retires the old flag. No longer sends a command to async-profiler to enable context explicitly since the native side will soon be passive leaving dd-trace-java in control of whether context updates are made. 

# Motivation

# Additional Notes

### DIFF
--- a/dd-java-agent/agent-profiling/profiling-async/src/main/java/com/datadog/profiling/async/AsyncProfiler.java
+++ b/dd-java-agent/agent-profiling/profiling-async/src/main/java/com/datadog/profiling/async/AsyncProfiler.java
@@ -137,6 +137,11 @@ public final class AsyncProfiler {
             os,
             t.getMessage());
       }
+      throw new UnsupportedEnvironmentException(
+          String.format(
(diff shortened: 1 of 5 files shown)


dd-java-agent/agent-profiling/profiling-async/src/main/java/com/datadog/profiling/async/AsyncProfiler.java
```diff
@@ -140,9 +140,8 @@ private static one.profiler.AsyncProfiler inferFromOsAndArch()
     }
     throw new UnsupportedEnvironmentException(
         String.format(
-            "Unable to instantiate async profiler for the detected environment: arch={}, os={}",
-            arch,
-            os));
+            "Unable to instantiate async profiler for the detected environment: arch=%s, os=%s",
```
richardstartin, line 143: random bug fix in apparently dead code (in tests at least)

jbachorik (approved):

PR #4045, opened 2022-10-18
Lazy remote-config Feature Discovery

ValentinZakharov:
# What Does This Do
Remote config Feature Discovery moved to RetryDiscoveryConfigUrlSupplier to make it lazy

# Motivation

# Additional Notes


communication/src/main/java/datadog/communication/ddagent/SharedCommunicationObjects.java
```diff
@@ -138,37 +130,51 @@ public String get() {
...
+      if (configEndpoint != null) {
+        configUrl = fd.buildUrl(configEndpoint).toString();
         return configUrl;
       }
+
       long now = System.currentTimeMillis();
       long elapsed = now - lastTry;
       if (elapsed > retryInterval) {
-        this.featuresDiscovery.discover();
+        fd.discover();
```
smola, line 164: I think that, on the first call, if the agent is not available, this will lead to 2 `discover` calls. One within `featuresDiscovery` and the other here.

PR #4049, opened 2022-10-18
hook context installation and thread exit so profiler can be notified of threads with tracing context

richardstartin:
# What Does This Do

Adds hooks for notifying when:
* the first scope on a thread is created
* threads exit

which allows notification of async-profiler when threads have context attached and when these threads are exited, which allows for filtering of profiled threads.

# Motivation

# Additional Notes


dd-java-agent/agent-profiling/profiling-async/src/main/java/com/datadog/profiling/async/AsyncProfiler.java
```diff
@@ -145,6 +145,18 @@ private static one.profiler.AsyncProfiler inferFromOsAndArch()
             os));
   }
 
+  void addCurrentThread() {
+    if (asyncProfiler != null) {
+      asyncProfiler.addThread(Thread.currentThread());
```
bantonsson, line 150: If we were unable to apply the `Thread` instrumentation so we don't call `removeCurrentThread`, would we run the risk of an unbounded set of threads being added?

richardstartin: Fortunately that's handled on the other side - async-profiler knows when threads exit itself, and since it only cares about the native thread id it only needs 8KiB for a bitset. I am considering whether the Thread instrumentation is actually necessary and whether this callback should only be emitted if the scope stack ThreadLocal is ever removed (perhaps in future code changes).

PR #4058, opened 2022-10-20
Ensure buildSrc can be tested with JDKs > 11

manuel-alvarez-alvarez:
# What Does This Do
Ensure the call site instrumentation plugin is compiled for JDK8 and the same for the plugin tests.

# Motivation
Gradle compiles the code and runs all the tests in the `buildSrc` folder, so any issue completely blocks the build process, in our case a wrong JDK was causing the problem due to the tests failing.

# Additional Notes

### DIFF
--- a/buildSrc/src/test/groovy/InstrumentPluginTest.groovy
+++ b/buildSrc/src/test/groovy/InstrumentPluginTest.groovy
@@ -14,6 +14,9 @@ class InstrumentPluginTest extends Specification {
       id 'java'
       id 'net.bytebuddy.byte-buddy-gradle-plugin'
     }
+    
+    sourceCompatibility = JavaVersion.VERSION_1_8
+    targetCompatibility = JavaVersion.VERSION_1_8
 
     repositories {
       mavenCentral()
--- a/buildSrc/call-site-instrumentation-plugin/build.gradle.kts
+++ b/buildSrc/call-site-instrumentation-plugin/build.gradle.kts
@@ -7,6 +7,11 @@ plugins {
   id("com.github.johnrengelman.shadow") version "7.1.2"
 }
 
+java {
+  sourceCompatibility = JavaVersion.VERSION_1_8
+  targetCompatibility = JavaVersion.VERSION_1_8
+}
+
 spotless {
   java {
     toggleOffOn()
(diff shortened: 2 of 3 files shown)


buildSrc/src/test/groovy/InstrumentPluginTest.groovy
```diff
@@ -14,6 +14,9 @@ class InstrumentPluginTest extends Specification {
       id 'java'
       id 'net.bytebuddy.byte-buddy-gradle-plugin'
     }
+
```
manuel-alvarez-alvarez, line 17: It did fail with a JDK 18:

```
Condition failed with Exception:

new ClassReader(new FileInputStream(classFile)).accept(new ClassVisitor(OpenedClassReader.ASM_API) { @Override FieldVisitor visitField(int access, String fieldName, String descriptor, String signature, Object value) { if ('__TEST__FIELD__' == fieldName) { foundInsertedField = true } return null } }, OpenedClassReader.ASM_API)
|
<java.io.FileInputStream@22954fd4 fd=java.lang.IllegalAccessException: class org.codehaus.groovy.runtime.DefaultGroovyMethods cannot access a member of class java.io.FileInputStream (in module java.base) with modifiers "private final" path=java.lang.IllegalAccessException: class org.codehaus.groovy.runtime.DefaultGroovyMethods cannot access a member of class java.io.FileInputStream (in module java.base) with modifiers "private final" channel=java.lang.IllegalAccessException: class org.codehaus.groovy.runtime.DefaultGroovyMethods cannot access a member of class java.io.FileInputStream (in module java.base) with modifiers "private volatile" closeLock=java.lang.IllegalAccessException: class org.codehaus.groovy.runtime.DefaultGroovyMethods cannot access a member of class java.io.FileInputStream (in module java.base) with modifiers "private final" closed=java.lang.IllegalAccessException: class org.codehaus.groovy.runtime.DefaultGroovyMethods cannot access a member of class java.io.FileInputStream (in module java.base) with modifiers "private volatile">

	at InstrumentPluginTest.test instrument plugin(InstrumentPluginTest.groovy:106)
Caused by: java.lang.IllegalArgumentException: Unsupported class file major version 62
	at org.objectweb.asm.ClassReader.<init>(ClassReader.java:196)
	at org.objectweb.asm.ClassReader.<init>(ClassReader.java:177)
	at org.objectweb.asm.ClassReader.<init>(ClassReader.java:163)
	at org.objectweb.asm.ClassReader.<init>(ClassReader.java:284)
	... 1 more
```

richardstartin (approved):
👍🏻

bantonsson (approved):

PR #4062, opened 2022-10-20
Use async-profiler snapshot api

jbachorik:
# What Does This Do
Adjusts the async-profiler integration to use the lightweight call to dump JFR data.

# Motivation
Reduce the overhead related to stopping and starting async-profiler engine just to obtain the JFR data.

# Additional Notes


richardstartin (approved):
👍🏻

jbachorik:
I am getting weird crashes in tests. Sending back to draft until I identify and fix the cause.

jbachorik:
On closer look async-profiler is making a lot of assumptions all over the code about only one recording being active between start and stop, where stop will tear down all async-profiler infra (timers, signal handlers etc.) to make sure data are not written to recording when that has been released.

All of this does not work well with my attempt to copy the currently written data and create a fresh new recording in order to populate all the necessary JFR metadata. A proper implementation would need much larger changes in the library.

PR #4063, opened 2022-10-20
register ContextThreadListener earlier to capture all tracing interactions

richardstartin:
# What Does This Do

By the time the rest of the profiler gets installed, there might have been a lot of tracing, so we need to install this listener earlier. This imposes constraints on the listener, which its current implementation meets (it doesn't matter when `AsyncProfiler` is created) but care needs to be taken not to introduce a dependency on JFR here in the future.

# Motivation

# Additional Notes

### DIFF
--- a/dd-java-agent/agent-bootstrap/src/main/java/datadog/trace/bootstrap/Agent.java
+++ b/dd-java-agent/agent-bootstrap/src/main/java/datadog/trace/bootstrap/Agent.java
@@ -15,6 +15,7 @@ import static datadog.trace.util.Strings.toEnvVar;
 
 import datadog.trace.api.Checkpointer;
 import datadog.trace.api.Config;
+import datadog.trace.api.GlobalTracer;
 import datadog.trace.api.StatsDClientManager;
 import datadog.trace.api.Tracer;
 import datadog.trace.api.WithGlobalTracer;
@@ -377,6 +378,7 @@ public class Agent {
       }
 
       installDatadogTracer(scoClass, sco);
+      installContextThreadListener();
       maybeStartAppSec(scoClass, sco);
       maybeStartIast(scoClass, sco);
       // start debugger before remote config to subscribe to it before starting to poll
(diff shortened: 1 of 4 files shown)


dd-java-agent/agent-bootstrap/src/main/java/datadog/trace/bootstrap/Agent.java
```diff
@@ -674,6 +676,29 @@ public void withTracer(Tracer tracer) {
...
+      try {
+        Tracer tracer = GlobalTracer.get();
+        if (tracer instanceof AgentTracer.TracerAPI) {
+          ContextThreadListener listener =
+              (ContextThreadListener)
+                  AGENT_CLASSLOADER
+                      .loadClass("com.datadog.profiling.async.ContextThreadFilter")
+                      .getDeclaredConstructor()
+                      .newInstance();
+          ((AgentTracer.TracerAPI) tracer).addThreadContextListener(listener);
```
bantonsson, line 694: Won't this always try to load the `AsyncProfiler` even if it isn't enabled?

dd-java-agent/agent-profiling/profiling-async/src/main/java/com/datadog/profiling/async/ContextThreadFilter.java
```diff
@@ -2,6 +2,10 @@
 
 import datadog.trace.bootstrap.instrumentation.api.ContextThreadListener;
 
+/**
+ * This class must be installed early to be able to see all scope initialisations, which means it
+ * must not be modified to depend on JFR, so that it can be installed before tracing starts.
+ */
```
bantonsson, line 8: Would be good with some warnings in the `AsyncProfiler` code as well, since this code depends on it being created.

bantonsson (commented):
Do the smoke tests exercise this startup change? I can't remember what's enabled.

dd-java-agent/agent-bootstrap/src/main/java/datadog/trace/bootstrap/Agent.java
richardstartin: Yes, but our `AsyncProfiler` wrapper is "environment-safe" and will just not initialise the `AsyncProfiler` interface if it's not enabled or can't be enabled, so these callbacks will essentially go nowhere. Initialising `AsyncProfiler` before the rest of the profiler doesn't have any impact on the rest of the bootstrapping process.

But I could always add `isAsyncProfilerEnabled` to `Config`...

richardstartin: I've got a follow up to avoid any interactions with async-profiler when it isn't enabled, but the listener will be installed regardless

dd-java-agent/agent-profiling/profiling-async/src/main/java/com/datadog/profiling/async/ContextThreadFilter.java
richardstartin: good point, I've addressed this

dd-java-agent/agent-bootstrap/src/main/java/datadog/trace/bootstrap/Agent.java
bantonsson: If the `AsyncProfiler` is _environment-safe_, then it's fine, but I can't see the check of `PROFILING_ASYNC_ENABLED` anywhere in the `AsyncProfiler`.

richardstartin: please look at the second commit

bantonsson: Stop fixing things while I'm typing 😉

bantonsson (approved):

PR #4065, opened 2022-10-20
remove Thread exit instrumentation

richardstartin:
# What Does This Do

Removes the thread exit instrumentation added in #4049 - we don't need it because we have the JVMTI `ThreadEnd` callback to remove threads from profile filters.

# Motivation

# Additional Notes


bantonsson (approved):
🎉

PR #4067, opened 2022-10-21
Enable remote config by default

smola:
# What Does This Do

Enable remote config by default.

Also:

* Make sure that smoke tests for remote activation run on gradle test task.
* Tweak tests to avoid explicit remote config enablement.

# Motivation

In 0.113.0 we changed the default of the telemetry setting in Config to true, but that's not enough, as it needs to be updated in `Agent` too.

# Additional Notes


smola:
This seems to have a negative interaction with `datadog.smoketest.SpringBootOpenLibertySmokeTest`.

jpbempel (approved):

PR #4071, opened 2022-10-21
WIP/POC - Add log injection to logback pattern automatically

devinsba:
# What Does This Do

# Motivation

# Additional Notes


dd-java-agent/instrumentation/logback-1/src/main/java/datadog/trace/instrumentation/logback/LogbackPatternLayoutInstrumentation.java
```diff
@@ -0,0 +1,51 @@
...
+    public static void onEnter(@Advice.Argument(value = 0, readOnly = false) String pattern) {
+      String patternAddition = "- %X{dd.trace_id} %X{dd.span_id} -";
+      boolean addTraceId = !pattern.contains("%X{dd.trace_id}");
+      if (addTraceId) {
+        int messageIndex = pattern.lastIndexOf(" ");
+        if (messageIndex == -1) {
+          messageIndex = pattern.lastIndexOf("\t");
+        }
+        pattern =
+            pattern.substring(0, messageIndex) + patternAddition + pattern.substring(messageIndex);
```
amarziali, line 47: This makes a lots of assumptions about how the pattern may look like. It's true that the message it's usually at the end but, if we want to be sure that the correlation works we also have to make sure that the ingestion and the grok parsing happens. In my opinion, if we want to do this kind of instrumentation, we can also directly force to use json encoding or change the pattern to a deterministic one. It will be a "simple but deterministic default". If a user wants more it will do manually as today. What do you think about?

dd-java-agent/instrumentation/logback-1/src/main/java/datadog/trace/instrumentation/logback/LogbackPatternLayoutInstrumentation.java
```diff
@@ -0,0 +1,51 @@
...
+            .and(named("setPattern"))
+            .and(takesArguments(1))
+            .and(takesArgument(0, String.class)),
+        LogbackPatternLayoutInstrumentation.class.getName() + "$SetPatternAdvice");
+  }
+
+  public static class SetPatternAdvice {
+    @Advice.OnMethodEnter
+    public static void onEnter(@Advice.Argument(value = 0, readOnly = false) String pattern) {
+      String patternAddition = "- %X{dd.trace_id} %X{dd.span_id} -";
```
amarziali, line 39: It may add a double `-` (as highlighted in the tests also) and this may break some grok patterns.

dd-java-agent/instrumentation/logback-1/src/main/java/datadog/trace/instrumentation/logback/LogbackPatternLayoutInstrumentation.java
```diff
@@ -0,0 +1,51 @@
...
+@AutoService(Instrumenter.class)
+public class LogbackPatternLayoutInstrumentation extends Instrumenter.Tracing
+    implements Instrumenter.ForSingleType {
+  public LogbackPatternLayoutInstrumentation() {
+    super("logback");
+  }
+
+  @Override
+  public String instrumentedType() {
+    return "ch.qos.logback.core.pattern.PatternLayoutBase";
```
amarziali, line 22: In an optic of zero configuration logging, should we restrict it only to console appenders? I think this point of interception has a broader audience (like file appenders) and using files requires more manipulations from an user point of view

dd-java-agent/instrumentation/logback-1/src/main/java/datadog/trace/instrumentation/logback/LogbackPatternLayoutInstrumentation.java
devinsba: What do you think about moving the trace ID and span ID right after the first segment, which is usually the date?

amarziali: out of the box supported grok expressions are:

```
java_default (%{_date_ms}|%{_duration})\s+\[%{_thread_name}\]\s+%{_status}\s+%{_logger_name}\s*(%{_context}\s*)?- (?>%{word:dd.trace_id} %{word:dd.span_id} - )?%{data:message}((\n|\t)%{data:error.stack})?

java_slf4j %{_date_slf4j}\s+\[%{_thread_name}\]\s+%{_status}\s+%{_logger_name}\s*(%{_context}\s*)?- (?>%{word:dd.trace_id} %{word:dd.span_id} - )?%{data:message}((\n|\t)%{data:error.stack})?

java_log4j %{_date} %{_status}\s+%{_logger_name}:%{_line}\s+- (?>%{word:dd.trace_id} %{word:dd.span_id} - )?%{data:message}((\n|\t)%{data:error.stack})?

java_fallback (%{_date}|%{_date_ms}) %{_status}\s+(?>%{word:dd.trace_id} %{word:dd.span_id} )?%{data:message}((\n|\t)%{data:error.stack})?
```

So I think it won't work. My personal advice, if you enable this instrumentation, is to change completely the pattern to a static one. So that you have the guarantee that information will be ingested and understood. You may give the ability to provide a custom one via an env variable if more customization is required.

devinsba: I think I agree with this in part. What I am thinking now is, if they application tries to use any default logging pattern then we modify it. Otherwise we view that as the user configuring their logger so it makes them configure it further

PR #4075, opened 2022-10-23
Fix fallback for empty dd.appsec.enabled/DD_APPSEC_ENABLED

smola:
# What Does This Do

Default is `inactive`, but on empty string it was falling back to `false`.

# Motivation

# Additional Notes

* Spotted while testing system-test changes https://github.com/DataDog/system-tests/pull/565 where `DD_APPSEC_ENABLED` was expected to set the default, not `false`.
* We may want to fix other inconsistencies for empty values in system properties and environment variables everywhere.

### DIFF
--- a/dd-java-agent/agent-bootstrap/src/main/java/datadog/trace/bootstrap/Agent.java
+++ b/dd-java-agent/agent-bootstrap/src/main/java/datadog/trace/bootstrap/Agent.java
@@ -894,10 +894,9 @@ public class Agent {
   private static boolean isAppSecFullyDisabled() {
     // must be kept in sync with logic from Config!
     final String featureEnabledSysprop = AgentFeature.APPSEC.systemProp;
-    String settingValue = System.getProperty(featureEnabledSysprop);
-    settingValue = settingValue != null && settingValue.isEmpty() ? null : settingValue;
+    String settingValue = getNullIfEmpty(System.getProperty(featureEnabledSysprop));
     if (settingValue == null) {
-      settingValue = ddGetEnv(featureEnabledSysprop);
+      settingValue = getNullIfEmpty(ddGetEnv(featureEnabledSysprop));
       settingValue = settingValue != null && settingValue.isEmpty() ? null : settingValue;
     }
 
@@ -906,6 +905,13 @@ public class Agent {
         || settingValue.equalsIgnoreCase("true")
         || settingValue.equalsIgnoreCase("1")
         || settingValue.equalsIgnoreCase("inactive"));
+  }
+
+  private static String getNullIfEmpty(final String value) {
+    if (value != null && value.isEmpty()) {
+      return null;
+    }
+    return value;
   }
 
   /** @return configured JMX start delay in seconds */
--- a/dd-java-agent/agent-bootstrap/src/main/java/datadog/trace/bootstrap/Agent.java
+++ b/dd-java-agent/agent-bootstrap/src/main/java/datadog/trace/bootstrap/Agent.java
@@ -908,7 +908,7 @@ public class Agent {
   }
 
   private static String getNullIfEmpty(final String value) {
-    if (value != null && value.isEmpty()) {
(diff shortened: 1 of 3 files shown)


internal-api/src/test/groovy/datadog/trace/api/ConfigTest.groovy
```diff
@@ -2035,6 +2036,68 @@ class ConfigTest extends DDSpecification {
...
+    null       | "false"    | ProductActivationConfig.FULLY_DISABLED
+    null       | "0"        | ProductActivationConfig.FULLY_DISABLED
+    null       | "invalid"  | ProductActivationConfig.FULLY_DISABLED
+    null       | "true"     | ProductActivationConfig.FULLY_ENABLED
+    null       | "1"        | ProductActivationConfig.FULLY_ENABLED
+    ""         | null       | ProductActivationConfig.ENABLED_INACTIVE
+    ""         | ""         | ProductActivationConfig.ENABLED_INACTIVE
+    ""         | "inactive" | ProductActivationConfig.ENABLED_INACTIVE
+    ""         | "false"    | ProductActivationConfig.ENABLED_INACTIVE // FIXME: Should be FULLY_DISABLED?
+    ""         | "0"        | ProductActivationConfig.ENABLED_INACTIVE // FIXME: Should be FULLY_DISABLED?
```
bantonsson, line 2069: I think the behavior for both `sys` and `env` should be the same, i.e. "" is treated as no setting.

smola: Agreed. Done.

dd-java-agent/agent-bootstrap/src/main/java/datadog/trace/bootstrap/Agent.java
```diff
@@ -895,8 +895,10 @@ private static boolean isAppSecFullyDisabled() {
     // must be kept in sync with logic from Config!
     final String featureEnabledSysprop = AgentFeature.APPSEC.systemProp;
     String settingValue = System.getProperty(featureEnabledSysprop);
+    settingValue = settingValue != null && settingValue.isEmpty() ? null : settingValue;
```
jandro996: this can be reused :)

dd-java-agent/agent-bootstrap/src/main/java/datadog/trace/bootstrap/Agent.java
```diff
@@ -906,6 +907,13 @@ private static boolean isAppSecFullyDisabled() {
         || settingValue.equalsIgnoreCase("inactive"));
   }
 
+  private static String getNullIfEmpty(final String value) {
+    if (value != null && value.isEmpty()) {
```
anderruiz, line 911: For me it would be easier in this way

`if(value==null || value.isEmpty())`

internal-api/src/main/java/datadog/trace/api/Config.java
```diff
@@ -1098,7 +1098,17 @@ && isJavaVersionAtLeast(8)
     }
     telemetryHeartbeatInterval = telemetryInterval;
 
-    String appSecEnabled = configProvider.getString(APPSEC_ENABLED, DEFAULT_APPSEC_ENABLED);
+    // ConfigProvider.getString currently doesn't fallback to default for empty strings. So we have
+    // special handling here until we have a general solution for empty string value fallback.
+    String appSecEnabled = configProvider.getString(APPSEC_ENABLED);
+    if (appSecEnabled == null || appSecEnabled.isEmpty()) {
```
anderruiz, line 1104: getNullIfEmpty()==null?

anderruiz (approved):

bantonsson (approved):

internal-api/src/main/java/datadog/trace/api/Config.java
smola: Not available from this package :s

PR #4077, opened 2022-10-23
Ensure that remote config activation test runs on CI

smola:
# What Does This Do

Ensure that remote config activation test runs on CI. It was currently tied to the `check` phase, and not `test`.

# Motivation

# Additional Notes


bantonsson (approved):

cataphract (approved):
Fine for now I guess, but I don't like this sort of solution we use. The problem it has is that when we try to run a single test with the `test` target it always runs all the other test-type targets. We should create a new a testAll or whatever target and make all targets of type test depend on it, and then call it from CI.

PR #4080, opened 2022-10-24
Ignore dependency duplicates while collecting with Telemetry

ValentinZakharov:
# What Does This Do
Added the set of hashes to keep all collected dependencies, to ignore duplicates. Assumed the set of hashes will exist during the whole application life time, because we don't know when exactly new dependency will be loaded.

# Motivation
The same dependency could be loaded twice or more by different class loaders - we can have duplications, that we need to ignore. 

# Additional Notes

### DIFF
--- a/telemetry/src/main/java/datadog/telemetry/dependency/Dependency.java
+++ b/telemetry/src/main/java/datadog/telemetry/dependency/Dependency.java
@@ -13,6 +13,7 @@ import java.util.ArrayList;
 import java.util.Collections;
 import java.util.Enumeration;
 import java.util.List;
+import java.util.Objects;
 import java.util.Properties;
 import java.util.jar.Attributes;
 import java.util.jar.JarEntry;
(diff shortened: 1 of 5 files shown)


telemetry/src/main/java/datadog/telemetry/dependency/Dependency.java
```diff
@@ -269,4 +270,10 @@ private static String parseGroupId(String bundleSymbolicName, String bundleName)
 
     return null;
   }
+
+  // HashCode uses to determine/suppress duplicates when collecting Dependencies.
```
smola, line 274:
```suggestion
  // HashCode used to suppress duplicates when collecting dependencies.
```

telemetry/src/main/java/datadog/telemetry/dependency/Dependency.java
```diff
@@ -269,4 +270,10 @@ private static String parseGroupId(String bundleSymbolicName, String bundleName)
...
+  @Override
+  public int hashCode() {
+    return Objects.hash(name, version);
```
smola, line 277: As discussed offline, deduplication based on name, version is not enough, since we must report the same dep in different sources.

Please, hold this off until I confirm with backend team.

cataphract: Hash codes are not unique. This should be combined with an equals method and a set.

PR #4085, opened 2022-10-24
add span context to exception samples

richardstartin:
# What Does This Do

This adds the active span and local root span ID to exception sample events so they can be used to explain span latency.

# Motivation

# Additional Notes


jbachorik (approved):
:shipit:

PR #4087, opened 2022-10-25
Do not log exceptions when remote config returns 404

smola:
# What Does This Do

If the agent has remote config enabled, the library discovers the endpoint, and then the agent is restarted with remote config disabled, we want to keep the logs quiet. We keep a debug-level message, but no warning or exception.

# Motivation

# Additional Notes


bantonsson (approved):
An extension of this would be to rediscover agent capabilities and disable/enable remote config.

smola:
@bantonsson I agree. I'll check with other people working on remote config. So far, in dotnet and node we're not triggering again discovery of the remote config endpoint once it's been discovered. But it's probably a good idea to retrigger it once we got a 404.

PR #4088, opened 2022-10-25
Fix duplicate GraphQL instrumentation bug

krosen040:
# What Does This Do
When `GraphQLInstrumentation#install` is called with a `ChainedInstrumentation` object the resulting `Instrumentation` contains duplicates (the individual Instrumentation objects contained in the `ChainedInstrumentation` as well as the `ChainedInstrumentation` itself). This fixes that.

# Motivation

# Additional Notes


ygree (approved):
Good catch! Thanks for the fix!

ygree:
Closing this PR because these changes have been cherry-picked into another PR #4118 to be able to run entire test suit (system-tests fail with `DD_API_KEY is missing in env, please add it.`) that is not possible to run from the forked repository branch.

PR #4090, opened 2022-10-25
Create spans for vulnerabilities outside of requests

manuel-alvarez-alvarez:
# What Does This Do
Creates a new span for vulnerabilities where there is no current context

# Motivation
Vulnerabilities are not only found during the execution of a request, they can be present at boot time, background processes ... this PR tries to enable the creation of custom spans when there is no requests.

# Additional Notes

### DIFF
--- a/dd-java-agent/agent-iast/src/main/java/com/datadog/iast/Reporter.java
+++ b/dd-java-agent/agent-iast/src/main/java/com/datadog/iast/Reporter.java
@@ -7,13 +7,22 @@ import datadog.trace.api.DDTags;
 import datadog.trace.api.TraceSegment;
 import datadog.trace.api.gateway.RequestContext;
 import datadog.trace.api.gateway.RequestContextSlot;
+import datadog.trace.bootstrap.instrumentation.api.AgentScope;
 import datadog.trace.bootstrap.instrumentation.api.AgentSpan;
+import datadog.trace.bootstrap.instrumentation.api.AgentTracer;
+import datadog.trace.bootstrap.instrumentation.api.InternalSpanTypes;
+import datadog.trace.bootstrap.instrumentation.api.ScopeSource;
+import datadog.trace.bootstrap.instrumentation.api.TagContext;
 import java.util.Set;
 import java.util.concurrent.ConcurrentHashMap;
 import java.util.function.Predicate;
+import javax.annotation.Nonnull;
+import javax.annotation.Nullable;
 
 /** Reports IAST vulnerabilities. */
 public class Reporter {
+
+  private static final String VULNERABILITY_SPAN_NAME = "vulnerability";
 
   private final Predicate<Vulnerability> duplicated;
 
@@ -29,10 +38,22 @@ public class Reporter {
     this.duplicated = duplicate;
   }
 
-  public void report(final AgentSpan span, final Vulnerability vulnerability) {
+  public void report(@Nullable final AgentSpan span, @Nonnull final Vulnerability vulnerability) {
     if (span == null) {
-      return;
+      final AgentSpan newSpan = startNewSpan();
+      try (final AgentScope autoClosed = tracer().activateSpan(newSpan, ScopeSource.MANUAL)) {
+        vulnerability.getLocation().updateSpan(newSpan.getSpanId());
+        reportVulnerability(newSpan, vulnerability);
+      } finally {
+        newSpan.finish();
+      }
+    } else {
+      reportVulnerability(span, vulnerability);
     }
+  }
+
+  private void reportVulnerability(
+      @Nonnull final AgentSpan span, @Nonnull final Vulnerability vulnerability) {
     final RequestContext reqCtx = span.getRequestContext();
     if (reqCtx == null) {
       return;
@@ -54,6 +75,18 @@ public class Reporter {
       // are kept.
       segment.setTagTop(DDTags.MANUAL_KEEP, true);
     }
+  }
+
+  private AgentSpan startNewSpan() {
+    final AgentSpan.Context tagContext =
+        new TagContext().withRequestContextDataIast(new IastRequestContext());
+    return tracer()
+        .startSpan(VULNERABILITY_SPAN_NAME, tagContext)
+        .setSpanType(InternalSpanTypes.VULNERABILITY);
+  }
+
+  protected AgentTracer.TracerAPI tracer() {
+    return AgentTracer.get();
   }
 
   /**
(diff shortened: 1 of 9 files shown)


dd-smoke-tests/springboot/src/test/groovy/datadog/smoketest/IastSpringBootSmokeTest.groovy
```diff
@@ -21,7 +27,11 @@ class IastSpringBootSmokeTest extends AbstractServerSmokeTest {
     List<String> command = new ArrayList<>()
     command.add(javaPath())
     command.addAll(defaultJavaProperties)
-    command.addAll(["-Ddd.appsec.enabled=true", "-Ddd.iast.enabled=true", "-Ddd.iast-request-sampling=100"])
+    command.addAll([
+      "-Ddd.appsec.enabled=true",
```
ValentinZakharov, line 31: I don't really understand why for IAST smoke test, you need enabled AppSec? 🤔

manuel-alvarez-alvarez:
> I can see, you build IAST smoke test in tracer's SpringBoot app, wouldn't be better to use distinct IAST smoke-test app?

This is inherited from the weak hash smoke tests, what  do you think @smola @DDJavierSantos ?

dd-java-agent/agent-iast/src/main/java/com/datadog/iast/IastAgentSpan.java
```diff
@@ -0,0 +1,74 @@
...
+    return AgentTracer.activeSpan();
+  }
+
+  @SuppressWarnings("resource")
+  static AgentSpan activeOrNewSpan() {
+    final AgentSpan span = activeSpan();
+    if (span != null) {
+      return span;
+    }
+    final AgentScope[] scope = new AgentScope[1];
```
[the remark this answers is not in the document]
manuel-alvarez-alvarez: Lambda can only address `final` variables from the outer closure, that's why we use an array and only address the initial element. Do you know a better way to do it (it's true that it's a bit ugly :cry:)?

dd-java-agent/agent-iast/src/main/java/com/datadog/iast/IastAgentSpan.java
```diff
@@ -0,0 +1,74 @@
...
+    return (AgentSpan)
+        Proxy.newProxyInstance(
```
smola, line 29: We have no internal API for this?

manuel-alvarez-alvarez: Nope, the problem is that I should do delegation rather than a proxy, but the API is so big that it made it very verbose. 

Should I change it @smola?

dd-java-agent/agent-iast/src/main/java/com/datadog/iast/model/Location.java
```diff
@@ -6,7 +6,7 @@ public final class Location {
 
   private final int line;
 
-  private final Long spanId;
+  private Long spanId;
```
smola, line 9: :cry:

smola (approved):

dd-java-agent/agent-iast/src/main/java/com/datadog/iast/model/Location.java
manuel-alvarez-alvarez: We could delay the creation of the vulnerability until we finally have the span, let me try

smola:
@manuel-alvarez-alvarez At the moment, I don't see any value in created separate modules for IAST smoke tests (and definitely not in this PR).

dd-java-agent/agent-iast/src/main/java/com/datadog/iast/model/Location.java
smola: @manuel-alvarez-alvarez I'm not really worried about this. It's ok to do this. If we can avoid it, fine. But it doesn't need to block this PR.

manuel-alvarez-alvarez: Ok, let's keep it like this then, but I dislike it too :smile:

dd-java-agent/agent-iast/src/main/java/com/datadog/iast/Reporter.java
```diff
@@ -7,14 +7,23 @@
...
 import java.util.Set;
 import java.util.concurrent.ConcurrentHashMap;
 import java.util.function.Predicate;
+import javax.annotation.Nonnull;
+import javax.annotation.Nullable;
 
 /** Reports IAST vulnerabilities. */
 public class Reporter {
 
+  private static final String NEW_SPAN_TYPE = "iast.vulnerability";
```
manuel-alvarez-alvarez, line 25: This has to be common for all languages

dd-java-agent/agent-iast/src/main/java/com/datadog/iast/Reporter.java
```diff
@@ -29,10 +38,22 @@ public Reporter(final Config config) {
     this.duplicated = duplicate;
   }
 
-  public void report(final AgentSpan span, final Vulnerability vulnerability) {
+  public void report(@Nullable final AgentSpan span, @Nonnull final Vulnerability vulnerability) {
     if (span == null) {
-      return;
+      final AgentSpan newSpan = startNewSpan();
+      try (final AgentScope ignored = tracer().activateSpan(newSpan, ScopeSource.MANUAL)) {
```
bantonsson, line 44: Using the word `ignored` feels wrong when it's more like `autoClosed`. I immediately started thinking that the scope wasn't closed.

bantonsson (approved):

dd-java-agent/agent-iast/src/main/java/com/datadog/iast/Reporter.java
manuel-alvarez-alvarez: TBH it was intelliJ who suggested to change and use this name, let's name it `autoClosed` :smile:

dd-java-agent/agent-iast/src/main/java/com/datadog/iast/Reporter.java
manuel-alvarez-alvarez: @uurien ack me if you see this comment :smile:

dd-trace-api/src/main/java/datadog/trace/api/DDSpanTypes.java
```diff
@@ -27,4 +27,6 @@ public class DDSpanTypes {
   public static final String GRAPHQL = "graphql";
 
   public static final String TEST = "test";
+
+  public static final String IAST = "iast";
```
manuel-alvarez-alvarez, line 31: This has to be common as well

smola (approved):

PR #881, opened 2019-06-13
Add `dd.trace.db.client.split-by-instance` Config

tylerbenson:
Additional integration testing to follow in separate PR.

PR #4094, opened 2022-10-25
Add `dd.trace.db.client.split-by-host` Config

shunyy:
# What Does This Do
When enabled, sets the db client's span's service.name to the hostname (if available).

# Motivation
Currently, only `dd.trace.db.client.split-by-instance` is supported, but this cannot be used to identify instances with the same instance name (= schema name when using postgres)  but are different services (writer instance/reader instance). For example, the service name of spans from postgres on `example.local/schema1` and `example-replica.local/schema1` will both be `schema1` and cannot be distinguished. I would like to be able to distinguish them by using host as the `service.name` of the span.

ref: https://github.com/DataDog/dd-trace-java/pull/881

# Additional Notes
- With this PR, if both `dd.trace.db.client.split-by-instance=true` and `dd.trace.db.client.split-by-host=true` are specified,  the `service.name` that is resolved in `dd.trace.db.client.split-by-host` will take precedence. Since this may create confusion, I would appreciate it if you have a better way to do this.


marcus-crane:
Ah thanks! This is a feature that would be very useful for the Java tracer and I hope it gets merged in 🙂

shunyy:
Hi.

Will this merge request be seen after some time? I don't know how to request a review, so if anyone knows how to do this, I would appreciate it!

shunyy:
@jpbempel 
Hi.

Sorry if I'm contacting the wrong person, but no one is responding, so I'm going to reach out to you, who seems to work on this repository recently.

Is it possible to be reviewed or get an advice about this please?

marcus-crane:
@richardstartin @bantonsson @piochelepiotr Hi there, would you be able to queue up someone review this PR? We've been waiting for this feature to land but it hasn't had anyone across it. It's a very useful feature that is present in other tracers for customers using database shards 😢

mcculls (approved):
Thanks for the contribution!

PR #4096, opened 2022-10-25
async profiler scope listener

richardstartin:
# What Does This Do

* Extends the `ExtendedScopeListener` interface to accept a local root span ID
* Replaces the async profiler context tracker with a scope listener which calls into `AsyncProfiler`'s context API

# Motivation

Since profiling now only tracks context when scopes are activated and doesn't use any of the custom hooks made available for checkpoints, this simplifies the integration and coalesces two scope stacks.

# Additional Notes


bantonsson (approved):
Nice

dd-java-agent/agent-profiling/profiling-context/src/main/java/com/datadog/profiling/context/AsyncProfilerScopeListener.java
```diff
@@ -0,0 +1,25 @@
...
+public class AsyncProfilerScopeListener implements ExtendedScopeListener {
+
+  private static final AsyncProfiler ASYNC_PROFILER = AsyncProfiler.getInstance();
+
+  @Override
+  public void afterScopeActivated() {}
+
+  @Override
+  public void afterScopeActivated(DDId traceId, DDId localRootSpanId, DDId spanId) {
+    ASYNC_PROFILER.setContext(localRootSpanId.toLong(), spanId.toLong());
```
jbachorik, line 16: We could check `ASYNC_PROFILER.isAvailable()` (the value can be actually cached in a static final field as well) to avoid unnecessary `toLong()` conversions when async profiler is not available.

jbachorik (approved):

dd-java-agent/agent-profiling/profiling-context/src/main/java/com/datadog/profiling/context/AsyncProfilerScopeListener.java
richardstartin: done

PR #4097, opened 2022-10-25
remove unused checkpoints

richardstartin:
# What Does This Do

Removes cruft from past incarnations of tracer/profiler integration to simplify the codebase

# Motivation

# Additional Notes


jbachorik (approved):
I checked that all that was removed was supposed to be removed.
I didn't check that all that was supposed to be removed was actually removed.

PR #4100, opened 2022-10-26
Try to fix flaky tainted map test

manuel-alvarez-alvarez:
# What Does This Do
Tries to fix the flaky tests regarding the tainted map

# Motivation

# Additional Notes

### DIFF
--- a/dd-java-agent/agent-iast/src/test/groovy/com/datadog/iast/taint/TaintedMapTest.groovy
+++ b/dd-java-agent/agent-iast/src/test/groovy/com/datadog/iast/taint/TaintedMapTest.groovy
@@ -3,12 +3,16 @@ package com.datadog.iast.taint
 import com.datadog.iast.model.Range
 import datadog.trace.test.util.CircularBuffer
 import datadog.trace.test.util.DDSpecification
-import datadog.trace.test.util.GCUtils
 
 import java.util.concurrent.ArrayBlockingQueue
+import java.util.concurrent.Callable
 import java.util.concurrent.ConcurrentHashMap
 import java.util.concurrent.CountDownLatch
 import java.util.concurrent.Executors
+import java.util.concurrent.TimeUnit
+
+import static datadog.trace.test.util.GCUtils.awaitGC
+import static java.util.concurrent.TimeUnit.SECONDS
 
 class TaintedMapTest extends DDSpecification {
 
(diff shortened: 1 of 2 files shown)


dd-java-agent/agent-iast/src/test/groovy/com/datadog/iast/taint/TaintedMapTest.groovy
```diff
@@ -281,34 +284,40 @@ class TaintedMapTest extends DDSpecification {
...
+          final o = new FinalizableObject(finalizer)
           final to = new TaintedObject(o, [] as Range[], map.getReferenceQueue())
           tuples.add(new Tuple2<Object, TaintedObject>(o, to))
         }
         startLatch.countDown()
         startLatch.await()
         for (int i = 1; i <= nObjectsPerThread; i++) {
           if (i % nBeforeWaitGC == 0) {
-            GCUtils.awaitGC()
+            tryAwaitGC(toCollect)
```
smola, line 300: I wonder if instead of this we should just call `awaitGC` for every `TaintedObject` which is not in the object holders? I think `awaitGC` admits a parameter to pass a weakref that should be collected.

manuel-alvarez-alvarez: Yep, I was doing some testing, I was a bit scared of waiting for each ref to be GCed individually but it works. It's done like this in the latest version :smile:

dd-java-agent/agent-iast/src/test/groovy/com/datadog/iast/taint/TaintedMapTest.groovy
```diff
@@ -70,19 +72,20 @@ class TaintedMapTest extends DDSpecification {
       assert map.get(o) == to
     }
     (1..iters).each {
-      (1..nObjectsPerIter).each {
+      final refs = (1..nObjectsPerIter).collect {
```
bantonsson, line 75: Is `final` necessary?

dd-java-agent/agent-iast/src/test/groovy/com/datadog/iast/taint/TaintedMapTest.groovy
```diff
@@ -281,6 +284,7 @@ class TaintedMapTest extends DDSpecification {
     def futures = (1..nThreads).collect { thread ->
       executorService.submit({
         ->
+        final unreferenced = new ArrayList<TaintedObject>()
```
bantonsson, line 287: Using `final` here as well, and in multiple places below.

bantonsson (approved):

dd-java-agent/agent-iast/src/test/groovy/com/datadog/iast/taint/TaintedMapTest.groovy
manuel-alvarez-alvarez: In Groovy I tend to favor `final` over `def` when references are not expected to mutate, I know it might cause issues with the magic behind spock if used in the `given` or `setup` labels, but it's not the case here.

Am I missing something why it's better not to use `final`?

bantonsson: Not used to `groovy` and not used in a lot of places in our code.

ValentinZakharov (approved):

jandro996 (approved):

PR #4107, opened 2022-10-27
larger smoke tests image

richardstartin:
# What Does This Do

The smoke tests are currently CPU bound despite having a lot of RAM headroom, and circleci doesn't offer a way to trade RAM for vCPUs https://circleci.com/docs/configuration-reference/#docker-execution-environment so the only option is to double the size of the image for now.

# Motivation

# Additional Notes

### DIFF
--- a/.circleci/config.yml
+++ b/.circleci/config.yml
@@ -446,6 +446,10 @@ jobs:
     <<: *tests
     resource_class: xlarge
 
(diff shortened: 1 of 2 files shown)


smola (approved):
Passed everything at the first attempt and in 22 minutes? Nice.

PR #4108, opened 2022-10-27
Prevent segmentation fault in IBM J9 11

manuel-alvarez-alvarez:
# What Does This Do
Coputes the `StackTraceElement` manually from a `StackFrame`

# Motivation
The method `StackFrame::toStackTraceElement()` was causing segmentation faults in IBM J9 11

# Additional Notes


richardstartin:
I don't think OpenJDK users should pay for a J9 workaround - have you considered using `Platform.isJ9()` and introducing a J9-only code path?

internal-api/internal-api-9/src/main/java/datadog/trace/util/stacktrace/JDK9StackWalker.java
```diff
@@ -25,7 +25,13 @@ public boolean isEnabled() {
 
   @Override
   <T> T doGetStack(final Function<Stream<StackTraceElement>, T> consumer) {
-    return walker.walk(
-        stack -> consumer.apply(stack.map(java.lang.StackWalker.StackFrame::toStackTraceElement)));
+    return walker.walk(stack -> consumer.apply(stack.map(JDK9StackWalker::map)));
+  }
+
+  private static StackTraceElement map(final java.lang.StackWalker.StackFrame frame) {
+    // IBM J9 11 was segfaulting when calling StackFrame::toStackTraceElement()
```
smola, line 32: Please, specify that this is with old patch versions of OpenJ9 11, but recent patch versions are ok.

manuel-alvarez-alvarez: Yep, done!

manuel-alvarez-alvarez:
> I don't think OpenJDK users should pay for a J9 workaround - have you considered using `Platform.isJ9()` and introducing a J9-only code path?

Done!, thanks for the input, I was a bit lazy not doing it like this in the first place :smile:

internal-api/internal-api-9/src/main/java/datadog/trace/util/stacktrace/JDK9StackWalker.java
```diff
@@ -8,11 +9,15 @@
 public class JDK9StackWalker extends AbstractStackWalker {
 
   static java.lang.StackWalker walker;
+  static Function<java.lang.StackWalker.StackFrame, StackTraceElement> mapper;
```
richardstartin, line 12: if you make it `static final` it will constant fold

manuel-alvarez-alvarez: Yep done. 

At first I was modifying the values in the tests, but since we test it in J9 as well should be fine. 

Thanks for the input!

bantonsson (approved):

richardstartin (approved):

smola (approved):

PR #4116, opened 2022-10-28
User blocking support in the WAF

cataphract:
Add support for the WAF (part of Application Security Management) to block requests based on user. This currently relies on a new public API required to use this functionality.


dd-java-agent/appsec/src/main/java/com/datadog/appsec/event/data/KnownAddresses.java
```diff
@@ -93,6 +93,8 @@ public interface KnownAddresses {
 
   Address<Object> GRPC_SERVER_REQUEST_MESSAGE = new Address<>("grpc.server.request.message");
 
+  Address<String> USER_ID = new Address<>("user.id");
```
smola, line 96: `usr.id` ?

smola (approved):
Looks good to me, also already validated in manual testing.

PR #4118, opened 2022-10-28
Fix duplicate GraphQL instrumentation bug

ygree:
# What Does This Do

When GraphQLInstrumentation#install is called with a ChainedInstrumentation object the resulting Instrumentation contains duplicates (the individual Instrumentation objects contained within the ChainedInstrumentation as well as the ChainedInstrumentation itself). This fixes that.

# Motivation

# Additional Notes

Open this PR from the original repository branch to be able run the entire test suit.
Original fix #4088 fails to run system-tests with `DD_API_KEY is missing in env, please add it.` because the PR is open from a fork repository.


ygree:
Merging without a review as the original identical PR #4088 has already been reviewed

PR #4123, opened 2022-10-31
Less verbose error on missing statsd_port

smola:
# What Does This Do
Found during testing. If a feature discovery call got no `statsd_port` in the response, we logged a long NullPointerException.

This PR makes the handling of missing port gentler on logging.

# Motivation

Exception with enabled telemetry and disabled tracer:

# Additional Notes

### DIFF
--- a/communication/src/main/java/datadog/communication/ddagent/DDAgentFeaturesDiscovery.java
+++ b/communication/src/main/java/datadog/communication/ddagent/DDAgentFeaturesDiscovery.java
@@ -228,7 +228,12 @@ public class DDAgentFeaturesDiscovery implements DroppingPolicy {
   private static void discoverStatsDPort(final Map<String, Object> info) {
     try {
       Map<String, ?> config = (Map<String, ?>) info.get("config");
-      int statsdPort = ((Number) config.get("statsd_port")).intValue();
+      final Object statsdPortObj = config.get("statsd_port");
+      if (statsdPortObj == null) {
+        log.debug("statsd_port missing from trace agent /info response");
+        return;
+      }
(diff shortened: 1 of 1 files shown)


jandro996 (approved):

communication/src/main/java/datadog/communication/ddagent/DDAgentFeaturesDiscovery.java
```diff
@@ -228,7 +228,12 @@ private boolean processInfoResponse(String response) {
...
-      int statsdPort = ((Number) config.get("statsd_port")).intValue();
+      final Object statsdPortObj = config.get("statsd_port");
+      if (statsdPortObj == null) {
+        log.debug("statsd_port missing from trace agent /info response");
+        return;
+      }
+      int statsdPort = ((Number) statsdPortObj).intValue();
       DDAgentStatsDClientManager.setDefaultStatsDPort(statsdPort);
     } catch (Throwable ignore) {
       log.debug("statsd_port missing from trace agent /info response", ignore);
```
manuel-alvarez-alvarez, line 239: This is now more an `error` or `warn` message right now, as it's coming from a misconfigured port right?

manuel-alvarez-alvarez (approved):

communication/src/main/java/datadog/communication/ddagent/DDAgentFeaturesDiscovery.java
smola: Yep. It might come from an empty string or malformed input. However, I'm not sure about the range of things we can expect as input here, so I would rather not move from debug to warn/error yet. But we could monitor logs and see if there are still stacktraces, and then decide.

Issue #4051, opened 2022-10-18
GraphQL Instrumentation - Support for async data fetchers returning CompletableFuture


jrogers:
graphql-java supports returning a `CompletableFuture` from a data fetcher to allow for fetchers to be asynchronous. The existing code doesn't handle this case well as it assumes it's a synchronous call - https://github.com/DataDog/dd-trace-java/blob/c0e281a6397388436da82222d38e03c4032fdcf6/dd-java-agent/instrumentation/graphql-java-14.0/src/main/java8/datadog/trace/instrumentation/graphqljava/InstrumentedDataFetcher.java#L43

As a result the spans are shown as almost instantaneous when async, and errors that occur are not tagged on the span of the field fetch (although they do get flagged on the execution span later).

One possible workaround might be checking if the returned value is a `CompletionStage` and add a whenComplete callback in that scenario that sets error on failure and finishes the span.

PR #4132, opened 2022-11-01
Add GraphQL async-fetcher test

ygree:
# What Does This Do

This reproduces the issue when an async fetcher is in use. The span is not marked as errored.

# Motivation

# Additional Notes

Ref: #4051


ygree:
Closing this in favor of #4896

PR #4135, opened 2022-11-02
Add DD_TAGS tags into snapshot/probe status tags

jpbempel:
# What Does This Do
using the tracer global tags that includes DD_TAGS we can add custom tags to the snapshot or probe statuses.

# Motivation
This is useful for Source Code Integration which requires tags for commit sha and repository

# Additional Notes

### DIFF
--- a/dd-java-agent/agent-debugger/src/main/java/com/datadog/debugger/sink/DebuggerSink.java
+++ b/dd-java-agent/agent-debugger/src/main/java/com/datadog/debugger/sink/DebuggerSink.java
@@ -12,6 +12,7 @@ import datadog.trace.core.DDTraceCoreInfo;
 import datadog.trace.util.AgentTaskScheduler;
 import java.util.List;
 import java.util.concurrent.TimeUnit;
+import java.util.stream.Collectors;
 import org.slf4j.Logger;
 import org.slf4j.LoggerFactory;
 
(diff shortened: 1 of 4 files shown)


dd-java-agent/agent-debugger/src/main/java/com/datadog/debugger/sink/DebuggerSink.java
```diff
@@ -81,17 +82,29 @@ public DebuggerSink(
...
-    this.snapshotSink = snapshotSink;
-    this.uploadFlushInterval = config.getDebuggerUploadFlushInterval();
+    if (config.getGlobalTags().isEmpty()) {
+      return debuggerTags;
+    }
+    String globalTags =
+        config.getGlobalTags().entrySet().stream()
+            .map(e -> e.getKey() + ":" + e.getValue())
+            .collect(Collectors.joining(","));
+    return debuggerTags + "," + globalTags;
```
shatzi, line 107: what happen if DD_TAGS contain our default tags, are those overwritten? it is valid to send multiple tags with same key different values?

dd-java-agent/agent-debugger/src/main/java/com/datadog/debugger/sink/DebuggerSink.java
```diff
@@ -81,17 +82,29 @@ public DebuggerSink(
...
+    tags = buildTags(config);
+    this.debuggerMetrics = debuggerMetrics;
+    this.probeStatusSink = probeStatusSink;
+    this.snapshotSink = snapshotSink;
+    this.uploadFlushInterval = config.getDebuggerUploadFlushInterval();
+  }
+
+  private static String buildTags(Config config) {
+    String debuggerTags =
         JsonSnapshotSerializer.IntakeRequest.concatTags(
```
shatzi, line 94: maybe we should extract this function to TagsUtils?

dd-java-agent/agent-debugger/src/main/java/com/datadog/debugger/sink/DebuggerSink.java
```diff
@@ -81,17 +82,29 @@ public DebuggerSink(
...
     this.batchUploader = batchUploader;
-    tags =
+    tags = buildTags(config);
+    this.debuggerMetrics = debuggerMetrics;
+    this.probeStatusSink = probeStatusSink;
+    this.snapshotSink = snapshotSink;
+    this.uploadFlushInterval = config.getDebuggerUploadFlushInterval();
+  }
+
+  private static String buildTags(Config config) {
```
shatzi, line 92: maybe rename to getDefaultTagsMergedWithGlobalTags()

shatzi (approved):
LGTM, some nit comments

dd-java-agent/agent-debugger/src/main/java/com/datadog/debugger/sink/DebuggerSink.java
jpbempel: not overwritten, it's a list not a map. so you will have all the tags, even if duplicates

dd-java-agent/agent-debugger/src/main/java/com/datadog/debugger/sink/DebuggerSink.java
jpbempel: done

dd-java-agent/agent-debugger/src/main/java/com/datadog/debugger/sink/DebuggerSink.java
jpbempel: done

PR #4139, opened 2022-11-02
add option to exclude rabbitmq routingKey from resource name

amarziali:
# What Does This Do

Add an option to exclude rabbitMQ routing key from the publisher's resource name. This can be used to reduce the cardinality of generated spans' resources.

The feature is driven by the system property  `dd.trace.rabbit.include.routingkey.in.resource` (or env `DD_TRACE_RABBIT_INCLUDE_ROUTINGKEY_IN_RESOURCE`) which is enabled by default (it's the current behaviour).

Setting it to false will exclude the routingKey to be part of the span resource name

# Motivation

# Additional Notes

Being a breaking change, the option is disabled by default

### DIFF
--- a/dd-java-agent/instrumentation/rabbitmq-amqp-2.7/src/main/java/datadog/trace/instrumentation/rabbitmq/amqp/RabbitDecorator.java
+++ b/dd-java-agent/instrumentation/rabbitmq-amqp-2.7/src/main/java/datadog/trace/instrumentation/rabbitmq/amqp/RabbitDecorator.java
@@ -95,7 +95,7 @@ public class RabbitDecorator extends MessagingClientDecorator {
         routingKey == null || routingKey.isEmpty()
             ? "<all>"
             : routingKey.startsWith("amq.gen-") ? "<generated>" : routingKey;
-    span.setResourceName("basic.publish " + exchangeName + " -> " + routing);
+    span.setResourceName(buildResourceName("basic.publish", exchangeName, routing));
     span.setTag(AMQP_EXCHANGE, exchange);
     span.setTag(AMQP_ROUTING_KEY, routingKey);
   }
@@ -147,6 +147,15 @@ public class RabbitDecorator extends MessagingClientDecorator {
       return "<generated>";
     }
     return queueName;
+  }
+
+  private String buildResourceName(
+      final String opName, final String exchangeName, final String routingKey) {
+    final String prefix = opName + " " + exchangeName;
+    if (Config.get().isRabbitRoutingKeyExcludedFromResourceName()) {
+      return prefix;
+    }
+    return prefix + " -> " + routingKey;
   }
 
   public TracedDelegatingConsumer wrapConsumer(String queue, Consumer consumer) {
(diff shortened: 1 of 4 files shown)


dd-trace-api/src/main/java/datadog/trace/api/config/TraceInstrumentationConfig.java
```diff
@@ -63,6 +63,9 @@ public final class TraceInstrumentationConfig {
   public static final String RABBIT_PROPAGATION_DISABLED_EXCHANGES =
       "rabbit.propagation.disabled.exchanges";
 
+  public static final String RABBIT_EXCLUDE_ROUTINGKEY_FROM_RESOURCE =
+      "rabbit.exclude.routingkey.from.resource";
```
bantonsson, line 67: This is just a nitpick, but to me it feels better if the setting is worded the other way around, like `rabbit.include.routingkey.in.resource`, with a default of `true`.

dd-java-agent/instrumentation/rabbitmq-amqp-2.7/src/main/java/datadog/trace/instrumentation/rabbitmq/amqp/RabbitDecorator.java
```diff
@@ -149,6 +149,15 @@ private String normalizeQueueName(String queueName) {
     return queueName;
   }
 
+  private String buildResourceName(
+      final String opName, final String exchangeName, final String routingKey) {
+    final String prefix = opName + " " + exchangeName;
+    if (Config.get().isRabbitRoutingKeyExcludedFromResourceName()) {
+      return prefix;
+    }
+    return prefix + " -> " + routingKey;
```
bantonsson, line 158: Please use a `StringBuilder` in this method that can be presized to the max including the routing key.

bantonsson (commented):
Sorry for not getting to this earlier.

dd-trace-api/src/main/java/datadog/trace/api/config/TraceInstrumentationConfig.java
amarziali: Yep I changed accordingly. I agree this is more straightforward

dd-java-agent/instrumentation/rabbitmq-amqp-2.7/src/main/java/datadog/trace/instrumentation/rabbitmq/amqp/RabbitDecorator.java
amarziali: All right I imagined the compiler was already doing some optimization. I allocated a `StringBuilder` with a size like the worst case as suggested

sdwa0:
The system property used in the code is `dd.rabbit.include.routingkey.in.resource` (or env `DD_RABBIT_INCLUDE_ROUTINGKEY_IN_RESOURCE`), which is different than what is mentioned in the description above (there is no `trace.` after `dd.`).

This isn't a big issue, but I just thought mentioning this here in case someone tried the property mentioned in the description and didn't get the desired result.

amarziali:
> 

Thanks for having catched it out. I edited the description accordingly

PR #4140, opened 2022-11-02
"Automagically" create dd_crash_uploader script for crash tracking.

jbachorik:
# What Does This Do
When `-XX:OnError=<action>` JVM arg is specified and `<action>` is a script file named `dd_crash_uploader.sh` or `dd_crash_uploader.bat` (for windows), regardless of in which directory that script should be located, the action script will be 'automagically' generated at the agent startup.
In case of a failure to generate the script (non-accessible location, already existing etc.) this step will be just skipped without affecting the rest of the agent startup.

# Motivation
Although we are able to pick up the crash log file and upload it to our intake, the configuration itself is pretty involved. One needs to create a script to redirect the on-crash action to our `AgentCLI` action, set it up to pick up the crash log file (which can be located in some other place using `-XX:ErrorLogFile` JVM argument) and then wire it back using the `-XX:OnError` JVM argument.
All of this requires a lot of typing or copy-pasting when errors can be introduced. Therefore, having an automated script setup based only on the `-XX:OnError` (and optionally `-XX:ErrorLogFile`) JVM arguments sounds as a good way to improve onboarding, at least until we have full auto-injection support.

# Additional Notes

### DIFF
--- a/dd-java-agent/agent-bootstrap/src/main/java/datadog/trace/bootstrap/Agent.java
+++ b/dd-java-agent/agent-bootstrap/src/main/java/datadog/trace/bootstrap/Agent.java
@@ -16,6 +16,7 @@ import static datadog.trace.util.Strings.toEnvVar;
 import datadog.trace.api.Checkpointer;
 import datadog.trace.api.Config;
 import datadog.trace.api.GlobalTracer;
+import datadog.trace.api.Platform;
 import datadog.trace.api.StatsDClientManager;
 import datadog.trace.api.Tracer;
 import datadog.trace.api.WithGlobalTracer;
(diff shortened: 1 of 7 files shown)


dd-java-agent/agent-bootstrap/src/main/java/datadog/trace/bootstrap/Agent.java
```diff
@@ -646,6 +648,15 @@ private static void startTelemetry(Instrumentation inst, Class<?> scoClass, Obje
     }
   }
 
+  private static void initializeCrashUploader() {
+    try {
+      Class<?> clz = AGENT_CLASSLOADER.loadClass("com.datadog.crashtracking.ScriptInitializer");
```
richardstartin, line 653: shall we check it's not J9 before doing this?

richardstartin (approved):

richardstartin (approved):

PR #3559, opened 2022-05-17
Telemetry; support for reporting dependencies

cataphract:
# What Does This Do

Sends telemetry data to the agent. Right now, only host information and dependencies are reported.
Dependencies are collected by installing a transformer that collects URLs from unique protection domains. The telemetry thread wakes up every 10 seconds to try and determine the name, version and possibly hash for the dependency associated with the URL.

PR #4141, opened 2022-11-02
Drop jnr-jffi dependency from telemetry

smola:
# What Does This Do
Drops jnr-jffi usage from Telemetry.

# Motivation
* Avoids libffi.so extraction by telemetry.
* Avoids class loading race conditions when Telemetry is lazy loaded, and tries to load ASM itself concurrently to other instrumentation code. See https://github.com/DataDog/dd-trace-java/pull/4124.
* Fixes support for Telemetry in non-x86 architectures.

# Additional Notes


richardstartin (approved):

gradle/dependencies.gradle
```diff
@@ -25,7 +25,6 @@ final class CachedData {
     coroutines    : "1.3.0",
     dogstatsd     : "4.1.0",
     jnr_unixsocket: "0.28",
-    jnr_ffi       : '2.1.12', // dependency of jnr_unixsocket: update in tandem
```
bantonsson, line 28: This needs to remain, as per the comment.

smola: jnr-unixsocket-0.28 has a transitive dep on jnr-jffi-2.1.12. I think this was added when a dep on jnr-jffi was added to Telemetry, to ensure we pulled the same version in different modules. See https://github.com/DataDog/dd-trace-java/pull/3559. So I guess it can be done now?

cataphract:
Nothing against in principle but:

1) I don't understand what this has to do with ASM and
2) This seems to silently drop the support for Mac OS? Did you run the tests there?

smola:
@cataphract Nothing to do with ASM, other than tat derived issues affect dependency collection (which is on our roof). Regarding macOS, you're right, I think some fields would be broken by this. Would you mind running tests in macOS? There will be a few fields affected.

smola:
Tests now pass on macOS. See `HostInfoTest` for the difference in output for the new implementation.

PR #4142, opened 2022-11-02
Loop protection for PathwayContext

kr-igor:
# What Does This Do
This PR adds a simple loop protection to the PathwayContext, aiming to detect and resolve situations when two identical operations become logically nested. 

A simple example:

You have a Google PubSub listener, which receives a message and then pushes it to two separate gRPC destinations, sequentially. All 3 operations will share the same context, and each time you're calling setCheckpoint, your PathwayContext will get modified. As the result, the second gRPC outgoing call will consider the first one its parent, creating a loop.


piochelepiotr (approved):

richardstartin:
How does this mechanism guarantee that the map will not grow until the application hits an OOME?

kr-igor:
This map lives as long as the context, values are added to the map at most once per checkpoint, typically one key-value pair (+ HashMap overhead to store these values) per span.

richardstartin (approved):
Nice, looks good to me!

PR #4147, opened 2022-11-03
ConfigCollector: change internal TreeSet to HashSet

smola:
# What Does This Do

* Make `CONFIG_FILTER_LIST` private, since it's not used anywhere else.
* Change it from `TreeSet` to `HashSet` to lower overhead.

# Motivation

# Additional Notes

Follow up to https://github.com/DataDog/dd-trace-java/pull/4144#discussion_r1013038483


ValentinZakharov:
@jpbempel If I recall correctly, it was done to represent `ConfigCollector` as integral object when pass data telemetry Otherwise we will need to expose composited Map or do any other workaround.
see here: 
https://github.com/DataDog/dd-trace-java/blob/master/telemetry/src/main/java/datadog/telemetry/TelemetryRunnable.java#L52
I would be happy to consider better solution.

PR #4157, opened 2022-11-04
AppSec default to false on OpenJ9 and non supported archs

smola:
# What Does This Do

Some OpenJ9-based JVMs (e.g. old IBM SDK 8.0) have issues with AppSec, so we're disabling it by default. This will prevent issues when enabling remote config by default, and it will also disable 1-click activation of AppSec for OpenJ9.

Also disable on architectures other than x86 and amd64, since PowerWAF is not supported anywhere else yet. So trying to enable it later is pointless.

# Motivation

# Additional Notes

### DIFF
--- a/dd-java-agent/agent-bootstrap/src/main/java/datadog/trace/bootstrap/Agent.java
+++ b/dd-java-agent/agent-bootstrap/src/main/java/datadog/trace/bootstrap/Agent.java
@@ -624,14 +624,15 @@ public class Agent {
   private static boolean isSupportedAppSecArch() {
     final String arch = System.getProperty("os.arch");
     if (Platform.isWindows()) {
-      return "amd64".equals(arch) || "x86_64".equals(arch) || "x86".equals(arch);
+      // TODO: Windows bindings need to be built for x86
+      return "amd64".equals(arch) || "x86_64".equals(arch);
     } else if (Platform.isMac()) {
       return "amd64".equals(arch) || "x86_64".equals(arch) || "aarch64".equals(arch);
     } else if (Platform.isLinux()) {
       return "amd64".equals(arch) || "x86_64".equals(arch) || "aarch64".equals(arch);
     }
-    // Still return true in other if unexpected cases, and we'll handle loading errors during AppSec
-    // startup.
+    // Still return true in other if unexpected cases (e.g. SunOS), and we'll handle loading errors
+    // during AppSec startup.
     return true;
   }
 
--- a/dd-java-agent/agent-bootstrap/src/main/java/datadog/trace/bootstrap/Agent.java
+++ b/dd-java-agent/agent-bootstrap/src/main/java/datadog/trace/bootstrap/Agent.java
@@ -158,7 +158,7 @@ public class Agent {
       setSystemPropertyDefault(AgentFeature.APPSEC.getSystemProp(), "false");
     } else if (!isSupportedAppSecArch()) {
       log.debug(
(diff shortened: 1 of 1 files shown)


jandro996 (approved):

dd-java-agent/agent-bootstrap/src/main/java/datadog/trace/bootstrap/Agent.java
```diff
@@ -610,6 +621,20 @@ private static void startAppSec(SubscriptionService ss, Class<?> scoClass, Objec
     }
   }
 
+  private static boolean isSupportedAppSecArch() {
+    final String arch = System.getProperty("os.arch");
+    if (Platform.isWindows()) {
+      return "amd64".equals(arch) || "x86_64".equals(arch) || "x86".equals(arch);
```
cataphract, line 627: x86 is not supported on windows.

cataphract (changes requested):

ValentinZakharov:
Does it make sense to move support checks inside the AppSec? Currently we already have a checker inside the AppSec, wouldn't be better to keep everything in the same place?

dd-java-agent/agent-bootstrap/src/main/java/datadog/trace/bootstrap/Agent.java
smola: @cataphract I think it is? https://github.com/DataDog/libddwaf/releases/tag/1.5.1 
am I missing something?

smola:
@ValentinZakharov Ideally, it would be nice to do the same checks there. But this sets the default as early as possible, in a way that it has effect even in `isFeatureEnabled(APPSEC)` (way before AppSec initialization is attempted). The reason to do this early is to ensure that everything (isFeatureEnabled, Config, ConfigProvider) and everywhere will see the right configuration value.

dd-java-agent/agent-bootstrap/src/main/java/datadog/trace/bootstrap/Agent.java
ValentinZakharov: I guess, @cataphract meant it's not implemented in Java. see [here](https://github.com/sqreen/AgentJavaNative/blob/master/src/main/java/io/sqreen/powerwaf/NativeLibLoader.java#L136-L164)

smola: Oh, I see. Thanks. I will update it.

ValentinZakharov:
@smola, TBH, I'm doubt about this approach. I suppose, if customer is trying to enabled AppSec on `J9` it should see `warning` or `error` in logs, about unsupported platform. However current code displays message in `debug` log level, which usually is not enabled.
Or if the customer, will try to enabled AppSec with 1-click activation - it will be really hard to investegate problem, due the same reason.

In addition, I think, the right way would be to change remote-config `CAPABILITIES` to inform backend, that this specific service is unable to run AppSec by removing the flag. In this case, customer will not see `enabled ASM` button in dashboard (but it's more for future work, I guess).

ValentinZakharov (approved):

smola:
For customers trying 1-click activation, we'll not announce ASM capabilities (since AppSec won't be initialize, so the capability is not added in the first place, right?), so they will not have the option in the first place.

Customers trying to explicitly enable AppSec in an unsupported platform won't be affected by the new default. AppSec will try to load, and it will print a warning if the load fails.

cataphract (approved):

PR #4158, opened 2022-11-04
OpenLiberty: Disable flaky test on OpenJ9 and extend test coverage

DDJavierSantos:
# What Does This Do

- Detect if we are on IBM's J9 JVM and skip the test in that case so we avoid a bug that makes the test flaky.
- Pass the agent's path and parameters using the JAVA_TOOL_OPTIONS environment variable so child processes get instrumented too. Otherwise the client application would not be instrumented since openliberty creates a new process for the client application.
- Add a test that checks for a vulnerability on the client application to make sure it is instrumented.

# Motivation
Fix the test so it is not Flaky. Add a test that tests the real situation where we instrument the client app.

# Additional Notes


dd-smoke-tests/springboot-openliberty/src/test/groovy/datadog/smoketest/SpringBootOpenLibertySmokeVulnerabilityTest.groovy
```diff
@@ -0,0 +1,83 @@
...
+import okhttp3.Request
+import spock.lang.Requires
+import spock.lang.Shared
+import spock.util.concurrent.PollingConditions
+
+import java.util.stream.Collectors
+
+// This test currently fails on IBM JVMs
+@Requires({ !Platform.isJ9() })
+class SpringBootOpenLibertySmokeVulnerabilityTest extends AbstractServerSmokeTest {
```
smola, line 16: Could you rename this to `IastSpringBootOpenLibertySmokeTest`? It'll be more homogenous with the others. Other than that, looks good.

DDJavierSantos: There is already a test with that name. I have kept it and added this extra test.

smola (approved):

PR #4163, opened 2022-11-04
Minor improvement of Dependency collection

ValentinZakharov:
# What Does This Do
Reuse MessageDigest, instead of creating new instance each time

# Motivation

# Additional Notes

### DIFF
--- a/telemetry/src/main/java/datadog/telemetry/dependency/Dependency.java
+++ b/telemetry/src/main/java/datadog/telemetry/dependency/Dependency.java
@@ -4,7 +4,6 @@ import datadog.trace.util.Strings;
 import java.io.File;
 import java.io.IOException;
 import java.io.InputStream;
-import java.lang.reflect.UndeclaredThrowableException;
 import java.math.BigInteger;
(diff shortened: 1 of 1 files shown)


telemetry/src/main/java/datadog/telemetry/dependency/Dependency.java
```diff
@@ -32,6 +32,18 @@ public final class Dependency {
...
 
+  private static final MessageDigest md;
+
+  static {
+    try {
+      md = MessageDigest.getInstance("SHA-1");
+    } catch (NoSuchAlgorithmException e) {
+      log.error("Unable to create cipher", e);
+      // should not happen
+      throw new UndeclaredThrowableException(e);
```
smola, line 43: What happens if this does happen? Where will this exception be caught?

ValentinZakharov: Reworked - now code, code skip hashing, in case of problem and write exception message in the log.

smola (approved):

PR #4167, opened 2022-11-07
Add flags to disable specific IAST vulnerability detections

DDJavierSantos:
…figuration. Defaulted to true.

# What Does This Do
Add configuration flags for IAST weak cipher and IAST weak hash. 

# Motivation
To allow independent configuration of IAST detection by vulnerability type.

# Additional Notes
The features are defaulted to true, so if IAST is enabled, they are enabled.

### DIFF
--- a/dd-java-agent/instrumentation/java-security/src/main/java/datadog/trace/instrumentation/java/security/WeakCipherInstrumentationCallSite.java
+++ b/dd-java-agent/instrumentation/java-security/src/main/java/datadog/trace/instrumentation/java/security/WeakCipherInstrumentationCallSite.java
@@ -1,12 +1,18 @@
 package datadog.trace.instrumentation.java.security;
 
+import static datadog.trace.api.ConfigDefaults.DEFAULT_IAST_WEAK_CIPHER_ENABLED;
+
 import datadog.trace.agent.tooling.csi.CallSite;
+import datadog.trace.api.config.IastConfig;
 import datadog.trace.api.iast.IastAdvice;
 import datadog.trace.api.iast.InstrumentationBridge;
 import datadog.trace.api.iast.sink.WeakCipherModule;
 import java.security.Provider;
 
-@CallSite(spi = IastAdvice.class)
+@CallSite(
+    spi = IastAdvice.class,
+    featureFlag = IastConfig.IAST_WEAK_CIPHER_ENABLED,
+    featureFlagDefault = DEFAULT_IAST_WEAK_CIPHER_ENABLED)
 public class WeakCipherInstrumentationCallSite {
 
   @CallSite.Before("javax.crypto.Cipher javax.crypto.Cipher.getInstance(java.lang.String)")
--- a/dd-java-agent/instrumentation/java-security/src/test/groovy/test/DisabledWeakCipherForkedTest.groovy
+++ b/dd-java-agent/instrumentation/java-security/src/test/groovy/test/DisabledWeakCipherForkedTest.groovy
@@ -1,0 +1,28 @@
+package test
+
+import datadog.trace.agent.test.AgentTestRunner
+import datadog.trace.api.iast.sink.WeakCipherModule
+import datadog.trace.api.iast.InstrumentationBridge
+import foo.bar.TestSuite
+import datadog.trace.api.Config
+
+class DisabledWeakCipherForkedTest extends AgentTestRunner {
+  @Override
+  protected void configurePreAgent() {
+    injectSysConfig("dd.iast.enabled", "true")
+    injectSysConfig("iast.weak-cipher.enabled", "false")
+  }
+
+  def "test weak cipher instrumentation"() {
+    setup:
+    WeakCipherModule module = Mock(WeakCipherModule)
+    InstrumentationBridge.registerIastModule(module)
+
+    when:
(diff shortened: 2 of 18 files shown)


dd-java-agent/instrumentation/java-security/src/test/groovy/test/DisabledWeakCipherForkedTest.groovy
```diff
@@ -0,0 +1,26 @@
...
+import datadog.trace.agent.test.AgentTestRunner
+import datadog.trace.api.iast.IastModule
+import datadog.trace.api.iast.InstrumentationBridge
+
+import javax.crypto.Cipher
+
+class DisabledWeakCipherForkedTest extends AgentTestRunner {
+  @Override
+  protected void configurePreAgent() {
+    injectSysConfig("iast.weak-cipher.detection.enabled", "false")
```
smola, line 12: Maybe just `dd.iast.weak-cipher.enabled`?
cc @anderruiz

DDJavierSantos: false is testing that we disable the flag, there are another test where we test it enabled

dd-java-agent/instrumentation/java-security/src/main/java/datadog/trace/instrumentation/java/security/WeakCipherInstrumentationCallSite.java
```diff
@@ -1,12 +1,18 @@
 package datadog.trace.instrumentation.java.security;
```
jandro996, line 1: All the old sources, propagation and sinks PRs must be rebased and refactored due to https://github.com/DataDog/dd-trace-java/pull/4300

jandro996 (changes requested):

dd-java-agent/instrumentation/java-security/src/main/java/datadog/trace/instrumentation/java/security/WeakCipherInstrumentationCallSite.java
DDJavierSantos: Done

dd-java-agent/instrumentation/java-security/src/test/groovy/test/DisabledWeakCipherForkedTest.groovy
DDJavierSantos: I have changed the flag name to dd.iast.weak-cipher.enabled

internal-api/src/main/java/datadog/trace/api/Config.java
```diff
@@ -588,6 +592,8 @@ static class HostNameHolder {
   private final Pattern iastWeakCipherAlgorithms;
 
   private final boolean iastDeduplicationEnabled;
+  private final boolean iastWeakHashEnabled;
+  private final boolean iastWeakCipherEnabled;
```
bantonsson, line 596: I assume that these variables can be removed now?

DDJavierSantos: We are interested in the flags being reported in Config.toString(). We think it can help us troubleshoot in the future. I have added this variables to the output produced by Config.toString().

bantonsson: I understand that, but going forward we would need some more generic mechanism than adding config flags by hand here for every iast integration, and also, you don't need to add them here since they are only used when printing the config. Another thing is that they would still be printed as _enabled_ even though `iast` itself is not _enabled_.

bantonsson (approved):

internal-api/src/main/java/datadog/trace/api/Config.java
dougqh: Yeah, I really think Config should be more declarative & self-describing.  If we ever do truly overhaul Config, then I'd definitely like it to be able to print out the flags, values, and even defaults without requiring code changes every time.

I'm thinking something akin to how it is done inside the JVM, but alas, not today.

dd-trace-api/src/main/java/datadog/trace/api/config/IastConfig.java
```diff
@@ -11,6 +11,8 @@ public final class IastConfig {
   public static final String IAST_VULNERABILITIES_PER_REQUEST = "iast.vulnerabilities-per-request";
   public static final String IAST_REQUEST_SAMPLING = "iast.request-sampling";
   public static final String IAST_DEDUPLICATION_ENABLED = "iast.deduplication.enabled";
+  public static final String IAST_WEAK_HASH_ENABLED = "iast.weak-hash.enabled";
+  public static final String IAST_WEAK_CIPHER_ENABLED = "iast.weak-cipher.enabled";
```
smola, line 15: What about the other supported vulnerabilities?

jandro996 (commented):
As @smola said [here](https://github.com/DataDog/dd-trace-java/pull/4167/files#r1082271362) we are not covering  SQL_INJECTION, COMMAND_INJECTION, PATH_TRAVERSAL and LDAP_INJECTION:

- Modify each callsite implementation and tests
- Set default values in ConfigDefaults.java

Maybe it would be better to implement each sink in a different PR

PR #4173, opened 2022-11-07
Eager initialization of Moshi in Telemetry

smola:
# What Does This Do
Eager initialization of Moshi can avoid issues with a SecurityManager. If it still fails, it's better that we fail early and avoid starting the telemetry subsystem.

# Motivation

# Additional Notes

### DIFF
--- a/telemetry/src/main/java/datadog/telemetry/TelemetrySystem.java
+++ b/telemetry/src/main/java/datadog/telemetry/TelemetrySystem.java
@@ -1,6 +1,13 @@
 package datadog.telemetry;
 
+import com.squareup.moshi.JsonAdapter;
+import com.squareup.moshi.Moshi;
 import datadog.communication.ddagent.SharedCommunicationObjects;
+import datadog.telemetry.api.AppDependenciesLoaded;
+import datadog.telemetry.api.AppIntegrationsChange;
+import datadog.telemetry.api.AppStarted;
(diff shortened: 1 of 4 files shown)


telemetry/src/main/java/datadog/telemetry/TelemetrySystem.java
```diff
@@ -49,10 +56,24 @@ static Thread createTelemetryRunnable(
...
+    // If it fails for any other reason, we better fail early and avoid starting telemetry.
+    final Moshi moshi =
+        new Moshi.Builder().add(new PolymorphicAdapterFactory(Payload.class)).build();
+    // Ensure that each Payload adapter is initialized.
+    moshi.adapter(AppStarted.class);
+    // FIXME: GenerateMetrics serializer needs to be fixed.
+    // moshi.adapter(GenerateMetrics.class);
+    moshi.adapter(AppDependenciesLoaded.class);
+    moshi.adapter(AppIntegrationsChange.class);
+    final JsonAdapter<Telemetry> jsonAdapter = moshi.adapter(Telemetry.class);
```
ValentinZakharov, line 71: I'm sorry, didn't get what was the reason of moving `Moshi` initialisation outside the `RequestBuilder`?
For me, `TelemetryService` is more about high level of abstraction which is unrelated to details of implementation (e.g. json serialiser)

PR #4177, opened 2022-11-07
Fix blocking on openliberty

cataphract:
The servlet instrumentation is not enough to implement blocking. The path is different on the first request; only on subsequent requests does the servlet instrumentation have an opportunity to run.


smola (approved):
This should have a unit test, but if it works with the smoke and system tests, it's ok for now.

PR #4178, opened 2022-11-07
Exclude tracer deps in telemetry

anderruiz:
# What Does This Do

It does not report tracer internal dependencies in telemetry

# Motivation

Currently internal dependencies were reported as if they were from the application

### DIFF
--- a/telemetry/src/main/java/datadog/telemetry/dependency/LocationsCollectingTransformer.java
+++ b/telemetry/src/main/java/datadog/telemetry/dependency/LocationsCollectingTransformer.java
@@ -13,6 +13,17 @@ import org.slf4j.LoggerFactory;
 class LocationsCollectingTransformer implements ClassFileTransformer {
   private static final Logger log = LoggerFactory.getLogger(LocationsCollectingTransformer.class);
 
+  private static URL agentURL = null;
+
+  static {
+    try {
+      agentURL =
+          LocationsCollectingTransformer.class.getProtectionDomain().getCodeSource().getLocation();
+    } catch (Throwable ex) {
+      log.error("Error getting agent URL");
+    }
+  }
+
   private final DependencyServiceImpl dependencyService;
   private final Set<ProtectionDomain> seenDomains =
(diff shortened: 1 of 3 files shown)


smola (approved):

smola:
@anderruiz I added a commit to fix a spotbugs error. Please, check.

telemetry/src/main/java/datadog/telemetry/dependency/LocationsCollectingTransformer.java
```diff
@@ -13,6 +13,18 @@
 class LocationsCollectingTransformer implements ClassFileTransformer {
   private static final Logger log = LoggerFactory.getLogger(LocationsCollectingTransformer.class);
 
+  private static String agentUrl = null;
```
ValentinZakharov, line 16: May be define `static final` would be better?

telemetry/src/main/java/datadog/telemetry/dependency/LocationsCollectingTransformer.java
```diff
@@ -13,6 +13,18 @@
...
+
+  static {
+    try {
+      final URL agentUrlObj =
+          LocationsCollectingTransformer.class.getProtectionDomain().getCodeSource().getLocation();
+      agentUrl = agentUrlObj == null ? null : agentUrlObj.toString();
+    } catch (Throwable ex) {
+      log.error("Error getting agent URL");
```
ValentinZakharov, line 24: I'd rather use `warning`, because for me `error` is more about the case when we can't proceed, but in this case, we still will be able to collect deps. What do you think?

telemetry/src/main/java/datadog/telemetry/dependency/LocationsCollectingTransformer.java
anderruiz: if I set it to final then I need to set it to null in the Throwable. Anyway this part is gone in the latest commit

smola (approved):

PR #4184, opened 2022-11-08
plugins update

am312:
Update spotless, axion-release, replace unsupported jcraft jsch.  Should fix issues around RSA keys caused by jcraft's jsch used in jgit.

### DIFF
--- a/dd-trace-java.gradle
+++ b/dd-trace-java.gradle
@@ -1,8 +1,11 @@
 buildscript {
-  configurations.classpath {
-    resolutionStrategy {
-      // both axion-release and spotless use jgit, use the axion-release version
-      force 'org.eclipse.jgit:org.eclipse.jgit:5.12.0.202106070339-r'
+  dependencies {
+    classpath "pl.allegro.tech.build:axion-release-plugin:1.14.2"
+  }
+
+  configurations.all {
+    resolutionStrategy.dependencySubstitution {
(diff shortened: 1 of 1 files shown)


dd-trace-java.gradle
```diff
@@ -1,21 +1,24 @@
...
+      substitute module("com.jcraft:jsch") using module("com.github.mwiede:jsch:0.2.4") because "jcraft jsch has been unmaintained for years."
     }
   }
 }
 
 plugins {
   id 'com.github.ben-manes.versions' version '0.27.0'
 
-  id "com.diffplug.spotless" version "5.17.1"
+  id "com.diffplug.spotless" version "6.11.0"
```
bantonsson, line 16: This will fail on some kotlin things unfortunately. I looked at this before, and can share my stashed changes in a branch and you can cherry-pick them or you can create a PR on top of my changes. I'll push them.

bantonsson (commented):
Nice. I'll commit and push my stuff.

dd-trace-java.gradle
bantonsson: My changes are in #4185

bantonsson: I think they need to be updated in sync, since `spotless` also pulls in a version of `jgit`, and it has historically clashed with `axion-release` (hence the forcing of versions).

bantonsson:
@am312 You should be able to do your overrides on top of `master` now.

PR #4186, opened 2022-11-08
Adjust snapshot structure for this object

jpbempel:
# What Does This Do
Consider this as a regular object argument. it should have type and
 field attributes

# Motivation
This argument is a special argument, but it should also handled as a regular object value for simplifying snapshot processing

# Additional Notes


ojung (approved):
👍

PR #4195, opened 2022-11-09
simplify async-profiler config flags

richardstartin:
# What Does This Do

* Disable wallclock profiler by default, which we are less confident in than the CPU profiler. This allows us to GA the CPU profiler without risking users inadvertently using the wallclock profiler.
* When the wallclock profiler is enabled, we will always collapse samples and will always filter threads, so these options are removed. If users disabled these features very large profiles full of irrelevant samples would be produced.
* Enable tracing context by default, this means that whenever async-profiler is used, the profile will contain labeled events which already drastically improves code-hotspots.

# Motivation

# Additional Notes


jbachorik (approved):
LGTM!

PR #4206, opened 2022-11-10
Fix profiler event labeling

richardstartin:
# What Does This Do

Currently the `spanId` and `rootSpanId` fields are transposed, this fixes it and adds a check to detect this.

# Motivation

# Additional Notes


dd-java-agent/agent-profiling/profiling-context/src/main/java/com/datadog/profiling/context/AsyncProfilerScopeListener.java
```diff
@@ -14,7 +14,7 @@ public void afterScopeActivated() {}
   @Override
   public void afterScopeActivated(DDId traceId, DDId localRootSpanId, DDId spanId) {
     if (ASYNC_PROFILER.isAvailable()) {
-      ASYNC_PROFILER.setContext(localRootSpanId.toLong(), spanId.toLong());
+      ASYNC_PROFILER.setContext(spanId.toLong(), localRootSpanId.toLong());
```
jbachorik, line 17: Argh, in these cases you wish Java had named arguments :(

richardstartin: let's just hope the test isn't too flaky :(

jbachorik (approved):

PR #4208, opened 2022-11-10
Improve error message when using oracle jdk 8 w/o commercial features unlocked

dbenamydd:
# Motivation
Reduce user confusion, add polish.


jbachorik:
Hi @dbenamydd , thanks for chipping in!
You need to also update the `ProfilingSystemTest` test which checks for the exact message when running on Oracle JDK 8.

dbenamydd:
Ack!

richardstartin:
I think the message is already quite clear

richardstartin (approved):

PR #4125, opened 2022-10-31
Command Injection detection

manuel-alvarez-alvarez:
# What Does This Do
Add support for command injection detection by instrumenting calls to `Runtime.exec(...)` and `ProcessBuilder.start()`

# Motivation

# Additional Notes

PR #4212, opened 2022-11-11
Path traversal detection

manuel-alvarez-alvarez:
# What Does This Do
Add support for path traversal detection by instrumenting calls to `File`, `Paths`, `Path`, `FileOutputStream` and `FileInputStreamp`

# Motivation

# Additional Notes


dd-java-agent/agent-iast/src/main/java/com/datadog/iast/IastModuleImpl.java
```diff
@@ -289,6 +294,85 @@ public void onProcessBuilderStart(@Nullable final List<String> command) {
...
+
+  @Override
+  public void onPathTraversal(final @Nonnull URI uri) {
+    final IastRequestContext ctx = IastRequestContext.get();
+    if (ctx == null) {
+      return;
+    }
+    final TaintedObjects taintedObjects = ctx.getTaintedObjects();
+    checkInjection(
+        VulnerabilityType.PATH_TRAVERSAL, taintedObjects, singletonList(uri), File.separatorChar);
```
smola, line 340: Could we avoid creating singleton lists here? Maybe we can have a checkInjection with a single parameter instead of a list, just with the inner loop logic.

manuel-alvarez-alvarez: First implementation I did was using specialized versions of the `checkInjection` but having a single one means having much less code to write/test/validate. Taking into account that this is for sinks and not propagation the performance/memory implications are not that strict. 

Let me check if I can do something better without growing the code exponentially :smile:

manuel-alvarez-alvarez: @smola it's now changed (the change starts with the [command injection PR](https://github.com/DataDog/dd-trace-java/pull/4125))

jandro996 (approved):

PR #4223, opened 2022-11-11
[SLS-2706][serverless] Implement secure random trace ID option

DylanLovesCoffee:
## What Does This Do
Allows for the use of SecureRandom to generate a non-deterministic trace ID. This is always disabled unless in a specific case for AWS Lambda or if the option is passed (`trace.secure-random`, `DD_TRACE_SECURE_RANDOM`).


dd-trace-core/src/main/java/datadog/trace/core/CoreTracer.java
```diff
@@ -1288,7 +1290,10 @@ private DDSpanContext buildSpanContext() {
           datadogTags = extractedContext.getDatadogTags();
         } else {
           // Start a new trace
-          traceId = IdGenerationStrategy.RANDOM.generate();
+          traceId =
```
dougqh, line 1293: I think it would be better to separate these into two difference strategies.  Or, have the random strategy pick between the ThreadLocalRandom and SecureRandom.

I do think it is good to isolate this change to just traceIds, but maybe we should do that by having a generateTraceId and generateSpanId routine on the IdGenerationStrategy.

dd-trace-api/src/main/java/datadog/trace/api/IdGenerationStrategy.java
```diff
@@ -9,6 +11,16 @@ public enum IdGenerationStrategy {
     public DDId generate() {
       return DDId.from(ThreadLocalRandom.current().nextLong(1, Long.MAX_VALUE));
     }
+
+    @Override
+    public DDId generateSecure() {
+      try {
+        SecureRandom secureRandom = SecureRandom.getInstanceStrong();
```
dougqh, line 18: This detection should be done in a static block.  If it fails, we'll be repeating it again and impacting application response time unnecessarily.

richardstartin:
can this wait for #4211?

DylanLovesCoffee:
Yup! We can wait 👍

dd-trace-core/src/main/java/datadog/trace/core/CoreTracer.java
bantonsson: The split into _trace id_ and _span id_ is happening as part of a bigger change speak #4211

bantonsson: And I should obviously read all comments before commenting... 😉

DylanLovesCoffee: Will it be likely to have this and #4211 merged before the upcoming freeze so we can meet a deadline?

bantonsson: @DylanLovesCoffee  I'm not sure which freeze you are referring to. When is the deadline?

DylanLovesCoffee: Oops, sorry for the lack of clarity! I believe there's an upcoming prod freeze for the Thanksgiving holiday and re:Invent (tbh, I'm not sure if trace libraries are held to the same freeze schedule). We're hoping to get this in before December if possible. A specific date is TBD for now.

bantonsson: We are not subject to the same freeze schedule, but we are mindful of them. This can definitely get in before December.

bantonsson: #4211 is now merged

dd-trace-api/src/main/java/datadog/trace/api/IdGenerationStrategy.java
DylanLovesCoffee: Updated this! Are the latest changes what you meant? (sorry in advance, first time Java over here 🥲)

One thought I had: with secure random, it's not very performant since I believe it needs to read from `/dev/random`, so having to init secureRandom on every use of IdGenerationStrategy may slow things down for all users rather than those that opt into the feature. I'm curious if you think this is something to be wary of?

bantonsson: So the secure random code should not be mixed into the `Random` class, but be in a separate `SecureRandom` class that will be used if the strategy name is `SECURE_RANDOM` (needs to be added to the `IdGenerationStrategy fromName(...)` method and to the selection logic in `Config`. Then you can skip the check in the `CoreTracer`.

dd-trace-core/src/main/java/datadog/trace/core/CoreTracer.java
bantonsson: As @dougqh said, this check should be removed and the strategies completely separated.

dd-trace-api/src/main/java/datadog/trace/api/IdGenerationStrategy.java
bantonsson: Having a separate strategy also means that you can move the initialization into the creation of that `SecureRandom` instance and not pay any penalty for other strategies.

dd-trace-api/src/main/java/datadog/trace/api/IdGenerationStrategy.java
```diff
@@ -52,4 +56,36 @@ public long generateSpanId() {
...
+      } catch (NoSuchAlgorithmException e) {
+        secureRandomTraceId = null;
+        secureRandomSpanId = 0;
+      }
+    }
+
+    @Override
+    public DDTraceId generateTraceId() {
+      if (secureRandomTraceId != null) {
+        return secureRandomTraceId;
```
bantonsson, line 78: So this will use the same _random_ value every time, making it not so _random_. I added some simple checks in #4286 to catch this.

DylanLovesCoffee: Thanks for catching this 😅

DylanLovesCoffee:
Hmm with the failing jacoco test, what's the preferred way to handle this? Should I exclude the `datadog.trace.api.IdGenerationStrategy.SRandom` class from `dd-trace-api` build?

dd-trace-api/src/main/java/datadog/trace/api/IdGenerationStrategy.java
randomanderson: https://xkcd.com/221/

image

bantonsson:
I actually think that it would be useful to test the avoid returning `0` code, and the only way I can think of doing that is to do some mocking. So the easiest way to do that in Java would be to have an _internal_ constructor that you can use from the tests.

You would also need a type to pass in to the constructor, so you can add something like this to the `IdGenerationStrategy`:

```
  @FunctionalInterface
  interface ThrowingSupplier<T> {
    T get() throws Throwable;
  }
```

And then change the constructors in `SRandom`:

```
    SRandom() {
      this(SecureRandom::getInstanceStrong);
    }

    SRandom(ThrowingSupplier<SecureRandom> supplier) {
      try {
        secureRandom = supplier.get();
      } catch (Throwable e) {
        throw new ExceptionInInitializerError(e);
      }
    }
```

After that you can add test cases to `DDTraceIdTest`:

```
  def "check SRandom avoid 0 loop"() {
    setup:
    def provider = Mock(IdGenerationStrategy.ThrowingSupplier)
    def random = Mock(SecureRandom)

    when:
    def strategy = new IdGenerationStrategy.SRandom(provider)
    strategy.generateTraceId().toLong() == 47
    strategy.generateSpanId() == 11

    then:
    1 * provider.get() >> { random }
    1 * random.nextLong() >> { 0 }
    1 * random.nextLong() >> { 47 }
    1 * random.nextLong() >> { 0 }
    1 * random.nextLong() >> { 11 }
    0 * _
  }

  def "check SRandom creation exception"() {
    setup:
    def provider = Mock(IdGenerationStrategy.ThrowingSupplier)

    when:
    new IdGenerationStrategy.SRandom(provider)

    then:
    1 * provider.get() >> { throw new IllegalArgumentException("whatever") }
    0 * _
    final ExceptionInInitializerError exception = thrown()
    exception.cause.message == "whatever"
  }
```

bantonsson (approved):

PR #3639, opened 2022-07-05
[WIP] ISSUE-1478: Add support for java 11 http client tracing

billgowendavis:
# What Does This Do
Attempts to add support for the Java 11 HttpClient

# Motivation
Address long standing issue - ISSUE-1478

# Additional Notes

PR #4226, opened 2022-11-12
[Help Wanted/ISSUE-1478] - Java http client instrumentation

billgowendavis:
# Why did I open this?
A while back I took a stab at adding java http client instrumentation but I could not seem to get the instrumenter to apply the byte buddy advice - https://github.com/DataDog/dd-trace-java/pull/3639. What I have done here is start from scratch to reduce the number of potential issues. Even still, I cannot seem to get the instrumentation to be applied. 

I have written 2 tests, one that proves out the byte buddy element and type matching `"see it should work"`, and another basic to test to see if the instrumentation is being applied (it is not).

I am not really sure what else to try and would love some help. I have a feeling it is some silly issue, I would really appreciate any 👀 on this.

### DIFF
--- a/dd-java-agent/instrumentation/java-httpclient/build.gradle
+++ b/dd-java-agent/instrumentation/java-httpclient/build.gradle
@@ -1,0 +1,30 @@
+ext {
+  minJavaVersionForTests = JavaVersion.VERSION_11
+}
+
+muzzle {
+  pass {
+    coreJdk()
+  }
+}
+
+apply from: "$rootDir/gradle/java.gradle"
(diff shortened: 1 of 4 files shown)


devinsba:
Make sure to check dd-java-agent/agent-tooling/src/main/resources/datadog/trace/agent/tooling/bytebuddy/matcher/ignored_class_name.trie

There may be an entry there that excludes the classes you need to modify. You'd have to add exceptions for them

billgowendavis:
@devinsba That was totally it! Thank you!!!

PR #4234, opened 2022-11-14
Reworked telemetry communication

ValentinZakharov:
# What Does This Do
Reworked communication - allows to handle various scenarios of available agent. Used feature discovery and fallback strategies with minimum log messages.

# Motivation

# Additional Notes
There are several scenarios that can happens:
## 1. There is no running DD Agent or it`s old version of Agent without telemetry support.
In this case tracer will try to discover agent with feature discovery every 120 sec (getting list of available endpoints). All collected telemetry messages will be dropped.
**Log message:** 
```
WARN: Unable to locate DD Agent with supported telemetry. We will lookup every 120 sec.
```

## 2. There is Agent with supported telemetry, but backend intake can't be reached
Typical case - when backend intake server can't be reached and returns 404. Or Service unavailable 500.
In this case, tracer will try to send request in 3, 9, 27, 81 seconds and then give up - all collected telemetry messages will be dropped.
**Log messages:**
```
WARN: Last attempt to send telemetry failed; will retry in 3.0 seconds (num failures: 1)
WARN: Last attempt to send telemetry failed; will retry in 9.0 seconds (num failures: 2)
WARN: Last attempt to send telemetry failed; will retry in 27.0 seconds (num failures: 3)
WARN: Unable to send telemetry, giving up and drop telemetry. We will try later
```
## 3. Agent have been updated to a version that supports telemetry
In this case customer no need to restart it`s application, tracer automatically will start sending telemetry.
**Log messages:**
```
INFO: Discovered DD Agent with supported telemetry
```


telemetry/src/main/java/datadog/telemetry/DiscoveryRequestBuilderSupplier.java
```diff
@@ -0,0 +1,61 @@
...
+
+    String telemetryEndpoint = featuresDiscovery.getTelemetryEndpoint();
+    if (telemetryEndpoint == null) {
+      log.debug("Telemetry endpoint not found");
+      return null;
+    }
+
+    HttpUrl httpUrl = featuresDiscovery.buildUrl(telemetryEndpoint + RequestBuilder.API_ENDPOINT);
+    if (httpUrl == null) {
+      log.debug("Unable Telemetry httpUrl");
```
cataphract, line 48: missing some verb here

telemetry/src/main/java/datadog/telemetry/DiscoveryRequestBuilderSupplier.java
```diff
@@ -0,0 +1,61 @@
...
+      return null;
+    }
+
+    requestBuilder = new RequestBuilder(httpUrl, config);
+    return requestBuilder;
+  }
+
+  public synchronized void needRediscover() {
+    sco.setFeaturesDiscovery(null);
```
cataphract, line 57: I don't see why you need to reset feature discovery. You can just store the features discovery object and have a flag so it will run discovery the next time around (or even keep that timekeeping inside the supplier)

telemetry/src/main/java/datadog/telemetry/TelemetryRunnable.java
```diff
@@ -73,30 +88,35 @@ public void run() {
...
+      return RequestStatus.NOTING_TO_SEND;
     }
-    return true;
+
+    return sendTelemetry(queue);
   }
 
   private void successWait() {
     consecutiveFailures = 0;
+    if (agentFailure) {
```
cataphract, line 115: if the point of these *failure booleans is not to repeat logging messages, they should be better named. Plus, it would still prob make sense to report these errors on debug level when they happen the second and subsequent times. We have `RatelimitedLogger` that sort of does that already.

telemetry/src/main/java/datadog/telemetry/TelemetryRunnable.java
```diff
@@ -105,43 +125,92 @@ private void successWait() {
...
         log.debug(msg);
       }
       lastRequestSuccess = false;
-
-      // Reached max limit of attempts
-      if (consecutiveFailures > BACKOFF_MAX_EXPONENT) {
-        //  Drop queue
-        telemetryService.prepareRequests().clear();
-      }
+      discoveryRequestBuilderSupplier.needRediscover();
```
smola, line 213: If we're going to do rediscovery, we only need to do it on status `404`, not on other responses such as `500`.

smola (commented):
I see changes in the right direction here, but let's discuss offline. We're introducing quite some additional complexity without addressing some basic issues.

For heartbeats, it just doesn't make sense to retry at all.  Now that we have visibility of request types at all states with `TelemetryData`, it's easier to do this fine-grained control.

telemetry/src/main/java/datadog/telemetry/RequestStatus.java
```diff
@@ -0,0 +1,8 @@
+package datadog.telemetry;
+
+public enum RequestStatus {
+  SUCCESS,
+  NOTING_TO_SEND,
```
smola, line 5:
```suggestion
  NOTHING_TO_SEND,
```

telemetry/src/main/java/datadog/telemetry/DiscoveryRequestBuilderSupplier.java
ValentinZakharov: That will be changed after merge of #4249.

PR #4241, opened 2022-11-15
support resource naming extraction on webmvc 5.3

amarziali:
# What Does This Do

Starting from spring web 5.3, our instrumentation fails gathering the best matching parsed path from the filter `HandlerMappingResourceNameFilter`

In fact, this filter is calling the getHandler function here

```
private boolean findMapping(final HttpServletRequest request) throws Exception {
    for (final HandlerMapping mapping : handlerMappings) {
      final HandlerExecutionChain handler = mapping.getHandler(request);
      if (handler != null) {
        return true;
      }
    }
    return false;
  }
```

And the handler is then calling `org.springframework.web.util.ServletRequestPathUtils #getParsedRequestPath` who's throwing a NPE because the PATH attribute on the servlet request has not been calculated

```
public static RequestPath getParsedRequestPath(ServletRequest request) {
		RequestPath path = (RequestPath) request.getAttribute(PATH_ATTRIBUTE);
		Assert.notNull(path, () -> "Expected parsed RequestPath in request attribute \"" + PATH_ATTRIBUTE + "\".");
		return path;
	}
```

This class is a new entry since 5.3. In fact spring, starting from this version, is not parsing the request path **before** the handle method unless the filter (also new since 5.3) `org.springframework.web.filter.ServletRequestPathFilter` is registered.

In a nominal case it means that the filter we add with the instrumentation is always silently throw a NPE with those spring web version. The name is hence only extracted by the instrumentation we apply on the handle method.

This means that, if the preHandle phase is throwing, the parsed path is never used as resource name.

I added a couple of tests to prove it. We also recently had customer tickets about it.

# Motivation

# Additional Notes


dd-java-agent/instrumentation/spring-webmvc-3.1/src/main/java/datadog/trace/instrumentation/springweb/PathMatchingHttpServletRequestWrapper.java
```diff
@@ -1,35 +1,33 @@
...
 
+import java.util.HashMap;
+import java.util.Map;
 import javax.servlet.http.HttpServletRequest;
 import javax.servlet.http.HttpServletRequestWrapper;
-import org.springframework.web.servlet.HandlerMapping;
 
 class PathMatchingHttpServletRequestWrapper extends HttpServletRequestWrapper {
-  private Object bestMatchingPattern;
+  private final Map<String, Object> localAttributes = new HashMap<>();
```
amarziali, line 9: I needed to change this class also because generally speaking is breaking the contract on a request wrapper. Since it was avoiding attributes to be set, it was causing failures. So I used a map here as a buffer in order to avoid changing the original request (which is also the aim of this class)

devinsba (approved):

devinsba:
Looks like you have an unused variable in the tests: https://output.circle-artifacts.com/output/job/06b6aabc-c4d4-4040-b16f-f32d4fe983bf/artifacts/0/check_reports/dd-java-agent/instrumentation/spring-webmvc-3.1/codenarc/test.html

PR #4249, opened 2022-11-15
Avoid redundant feature discovery requests

smola:
# What Does This Do

* ~Separate feature discovery instance creation from actual discovery.~
* Allow forced discovery (`discovery()`) for cases where we want to force discovery, such as detecting an agent downgrade or config change (e.g. receiving a 404 error on a known endpoint).
* For other cases, use `discoveryIfOld()`. Which will only run discovery if it has not been done in a while.
* Synchronize discovery to debounce requests.

# Motivation

We were doing some ugly hacks in remote config to ensure we didn't send extra feature discovery requests. As we extend this usage to telemetry, we saw ourselves doing the same hack again. Previous discussion is available at https://github.com/DataDog/dd-trace-java/pull/4042#discussion_r998163980

# Additional Notes
* This PR does not address resetting `dataStreamsEndpoint`, which was missing from `reset()`.

### DIFF
--- a/communication/src/main/java/datadog/communication/ddagent/DDAgentFeaturesDiscovery.java
+++ b/communication/src/main/java/datadog/communication/ddagent/DDAgentFeaturesDiscovery.java
@@ -45,6 +45,8 @@ public class DDAgentFeaturesDiscovery implements DroppingPolicy {
 
   public static final String DEBUGGER_ENDPOINT = "debugger/v1/input";
 
+  private static final long MIN_FEATURE_DISCOVERY_INTERVAL_MILLIS = 60 * 1000;
+
   private final OkHttpClient client;
   private final HttpUrl agentBaseUrl;
   private final Recording discoveryTimer;
@@ -62,6 +64,8 @@ public class DDAgentFeaturesDiscovery implements DroppingPolicy {
   private volatile String configEndpoint;
   private volatile String debuggerEndpoint;
   private volatile String version;
+
+  private long lastTimeDiscovered;
 
   public DDAgentFeaturesDiscovery(
       OkHttpClient client,
@@ -87,9 +91,29 @@ public class DDAgentFeaturesDiscovery implements DroppingPolicy {
     configEndpoint = null;
     debuggerEndpoint = null;
     version = null;
+    lastTimeDiscovered = 0;
   }
 
+  /** Run feature discovery, unconditionally. */
   public void discover() {
+    discoverIfOutdated(0);
+  }
+
+  /** Run feature discovery, if it was not run recently. */
+  public void discoverIfOutdated() {
+    discoverIfOutdated(MIN_FEATURE_DISCOVERY_INTERVAL_MILLIS);
+  }
+
+  private synchronized void discoverIfOutdated(final long maxElapsedMs) {
+    final long now = System.currentTimeMillis();
+    final long elapsed = now - lastTimeDiscovered;
+    if (elapsed > maxElapsedMs) {
+      doDiscovery();
+      lastTimeDiscovered = now;
+    }
+  }
+
+  private void doDiscovery() {
     reset();
     // 1. try to fetch info about the agent, if the endpoint is there
     // 2. try to parse the response, if it can be parsed, finish
--- a/communication/src/main/java/datadog/communication/ddagent/SharedCommunicationObjects.java
+++ b/communication/src/main/java/datadog/communication/ddagent/SharedCommunicationObjects.java
@@ -135,8 +135,6 @@ public class SharedCommunicationObjects {
     private String configUrl;
     private final SharedCommunicationObjects sco;
     private final Config config;
-    private long lastTry = 0;
-    private long retryInterval = 5000;
 
     private RetryConfigUrlSupplier(final SharedCommunicationObjects sco, final Config config) {
       this.sco = sco;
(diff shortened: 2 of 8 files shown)


communication/src/main/java/datadog/communication/ddagent/DDAgentFeaturesDiscovery.java
```diff
@@ -87,9 +90,30 @@ private void reset() {
...
+    discoverIfOld(0);
+  }
+
+  /** Run feature discovery, if it was not run recently. */
+  public void discoverIfOld() {
+    final long maxEllapsed = 60 * 1000;
+    discoverIfOld(maxEllapsed);
+  }
+
+  private synchronized void discoverIfOld(final long maxEllapsedMillis) {
```
ValentinZakharov, line 107: Have you considering another naming? For instance `discoverIfOutdated` ?

ValentinZakharov:
It seems we need to reset (set to `null`) fields during re-discovery in `processInfoResponse(...)`
If in first discovery run fields will be initialised, next in second run some of them will be absent - the old values still will be set in fields:
- traceEndpoint
- configEndpoint
- debuggerEndpoint
- foundDatastreamsEndpoint

communication/src/main/java/datadog/communication/ddagent/DDAgentFeaturesDiscovery.java
```diff
@@ -63,6 +63,8 @@ public class DDAgentFeaturesDiscovery implements DroppingPolicy {
   private volatile String debuggerEndpoint;
   private volatile String version;
 
+  private volatile long lastTimeDiscovered;
```
jpbempel, line 66: as `lastTimeDiscovered` is not exposed as a getter, there is no need to have it `volatile` here.

communication/src/main/java/datadog/communication/ddagent/DDAgentFeaturesDiscovery.java
```diff
@@ -77,6 +79,7 @@ public DDAgentFeaturesDiscovery(
             ? new String[] {V5_ENDPOINT, V4_ENDPOINT, V3_ENDPOINT}
             : new String[] {V4_ENDPOINT, V3_ENDPOINT};
     this.discoveryTimer = monitoring.newTimer("trace.agent.discovery.time");
+    this.lastTimeDiscovered = 0;
```
jpbempel, line 82: default value for an int field is 0, so it is useless

smola:
@ValentinZakharov The `reset` method sets fields to `null`, which is the first thing that is called on `discover`. Am I missing something?

ValentinZakharov:
@smola Ah, I'm sorry, my bad. Looks like there is missing clearing `dataStreamsEndpoint` in the `reset()`

ValentinZakharov:
@smola, I'm sorry, I didn't see the `reset()` call.

ValentinZakharov:
There is possible reseting issue.
Currently in `processInfoResponse()` we are reseting `dataStreamsEndpoint`, but not in the `reset()`.
If discovery http request fails - `dataStreamsEndpoint` will have old value, which looks wrong.

jpbempel (approved):

communication/src/main/java/datadog/communication/ddagent/SharedCommunicationObjects.java
```diff
@@ -149,31 +147,13 @@ public String get() {
...
-            sco.featuresDiscovery.discover();
-          }
-          retryInterval = 60000;
-        }
-      } else {
-        return null;
-      }
-      lastTry = now;
-      String configEndpoint = sco.featuresDiscovery.getConfigEndpoint();
+      final DDAgentFeaturesDiscovery discovery = sco.featuresDiscovery(config);
```
cataphract, line 150: Not sure this is what we want. Presumably, the config endpoint can change or disappear.  Is the plan to have to features discovery object reset when there is an error or 404 response?

smola: @cataphract As far as I can tell, the `DDAgentFeaturesDiscovery` object itself is created just only once. However, its internal state, including the config endpoint, can change on subsequent `discover()` calls. A 404 response may trigger discovery again. See, for example `ConflatingMetricsAggregator#disable()`. This PR does not add further handling of 404 responses to subsystems that currently don't do it (e.g. remote config), but I think it's something we should consider for every subsystem relying on feature discovery.

communication/src/main/java/datadog/communication/ddagent/DDAgentFeaturesDiscovery.java
```diff
@@ -112,6 +136,7 @@ public void discover() {
         // disable metrics unless the info endpoint is present, which prevents
         // sending metrics to 7.26.0, which has a bug in reporting metric origin
         metricsEndpoint = null;
+        // disable other endpoints that might be not present or disabled
```
bantonsson, line 139: Is this a `todo` comment?

communication/src/main/java/datadog/communication/ddagent/DDAgentFeaturesDiscovery.java
```diff
@@ -87,9 +91,29 @@ private void reset() {
...
+  }
+
+  /** Run feature discovery, if it was not run recently. */
+  public void discoverIfOutdated() {
+    discoverIfOutdated(MIN_FEATURE_DISCOVERY_INTERVAL_MILLIS);
+  }
+
+  private synchronized void discoverIfOutdated(final long maxEllapsedMillis) {
+    final long now = System.currentTimeMillis();
+    final long ellapsed = now - lastTimeDiscovered;
```
bantonsson, line 109: `ellapsed -> elapsed`

bantonsson (approved):

communication/src/main/java/datadog/communication/ddagent/DDAgentFeaturesDiscovery.java
smola: Kind of. It was a dangling comment to a code block I did not include in the end. I have removed it.

We should do better with initialization, reset, and fallbacks. The current approach, where `reset` sets everything to null, could be potentially a source of race conditions. Also, fallbacks should be part of the constructor, not just of a failed discovery. But taking care of this exceeded the scope of this PR by far.

PR #4252, opened 2022-11-16
Create spans for vulnerabilities outside of requests

manuel-alvarez-alvarez:
# What Does This Do
Creates a new span for vulnerabilities where there is no current context

# Motivation
Vulnerabilities are not only found during the execution of a request, they can be present at boot time, background processes ... this PR tries to enable the creation of custom spans when there is no requests.

# Additional Notes

### DIFF
--- a/dd-java-agent/agent-iast/src/main/java/com/datadog/iast/Reporter.java
+++ b/dd-java-agent/agent-iast/src/main/java/com/datadog/iast/Reporter.java
@@ -7,13 +7,21 @@ import datadog.trace.api.DDTags;
 import datadog.trace.api.TraceSegment;
 import datadog.trace.api.gateway.RequestContext;
 import datadog.trace.api.gateway.RequestContextSlot;
+import datadog.trace.bootstrap.instrumentation.api.AgentScope;
 import datadog.trace.bootstrap.instrumentation.api.AgentSpan;
+import datadog.trace.bootstrap.instrumentation.api.AgentTracer;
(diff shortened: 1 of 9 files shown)


dd-java-agent/agent-iast/src/main/java/com/datadog/iast/Reporter.java
```diff
@@ -7,14 +7,22 @@
...
 import java.util.Set;
 import java.util.concurrent.ConcurrentHashMap;
 import java.util.function.Predicate;
+import javax.annotation.Nonnull;
+import javax.annotation.Nullable;
 
 /** Reports IAST vulnerabilities. */
 public class Reporter {
 
+  private static final CharSequence NEW_SPAN_TYPE = "iast.vulnerability";
```
smola, line 24: Why not `String` here? :thinking:

manuel-alvarez-alvarez: We should continue in the old [PR](https://github.com/DataDog/dd-trace-java/pull/4090)

It's just me trying to understand why I'm getting in CI:
```
> Task :dd-java-agent:agent-iast:compileJava FAILED
/home/circleci/dd-trace-java/dd-java-agent/agent-iast/src/main/java/com/datadog/iast/Reporter.java:82: error: incompatible types: boolean cannot be converted to long
    return AgentTracer.startSpan(NEW_SPAN_TYPE, tagContext, false)
```

PR #4256, opened 2022-11-16
Add transitive dependencies to IAST short-circuits in CI

smola:
# What Does This Do
Add agent-iast transitive dependencies to CI triggers.

# Motivation
Avoid skipping agent-iast tests for anything that can affect it.

# Additional Notes
Generated with:
```
JAVA_HOME=$JAVA_8_HOME ./gradlew :dd-java-agent:agent-iast:dependencies > out
echo $(cat out | grep 'project :' | sed -e 's~.*project :~~g' -e 's~ .*~~g' -e 's~:~/~g'|sort|uniq) | sed -e 's~ ~|~g'
```
plus some manual filtering.


richardstartin:
I would recommend just removing the iast only testing and merging these tests back into the base tests instead. It would create fewer jobs.

richardstartin (commented):
Seems like a good idea, maybe we can find a way to automate this as a build step.

richardstartin (approved):

manuel-alvarez-alvarez (approved):

bantonsson (approved):

richardstartin:
I don't think this is necessary - see #4271

PR #4257, opened 2022-11-16
trigger profiling tests when transitive dependencies change

richardstartin:
# What Does This Do

inspired by #4256 

# Motivation

# Additional Notes


jbachorik (approved):
Now we need a CI check to remind us to change the list when transitive dependencies are going to change ;)

bantonsson (approved):

PR #4260, opened 2022-11-16
Add manual trigger for clean cache in Circle CI

smola:
# What Does This Do

# Motivation

# Additional Notes


richardstartin (approved):

bantonsson (approved):

smola:
Not a great approach, since it flags the whole build as "On Hold" whenever the cache upgrade is not triggered. We should probably rely on an environment variable to invalidate caches (`CACHE_VERSION`). Closing.

PR #4268, opened 2022-11-17
Add dd-api sanity check to dd-trace-ot smoke test

bantonsson:
# What Does This Do

Adds a simple sanity check of the `TraceInterceptor` API to the OpenTracing smoke test that uses `dd-trace-ot`.

# Motivation

Verify that shading and relocation doesn't move the public classes.

# Additional Notes


richardstartin (approved):

mcculls (commented):
Thanks!

PR #4272, opened 2022-11-17
support okhttp context propagation in callbacks

amarziali:
# What Does This Do

When using okhttp `Call.enqueue` the response is provided asynchronously by calling back a function.
In this case we missed propagating the context active at the time `enqueue` was called.

This PR is injecting continuations to mitigate the issue. This works for both okttp 2.x and 3.x

# Motivation

# Additional Notes


dd-java-agent/instrumentation/okhttp-2/src/main/java/datadog/trace/instrumentation/okhttp2/TracingCallbackWrapper.java
```diff
@@ -0,0 +1,31 @@
...
+    try (final AgentScope scope = parentContinuation.activate()) {
+      delegate.onFailure(request, e);
+    }
+  }
+
+  @Override
+  public void onResponse(Response response) throws IOException {
+    try (final AgentScope scope = parentContinuation.activate()) {
+      delegate.onResponse(response);
+    }
```
richardstartin, line 29: I think it would be good to poke a bit deeper than this, because this scope clearly won't be open when the response is being marshalled. This means that any scope aware integrations (e.g. profiling) will miss the response marshalling.

dd-java-agent/instrumentation/okhttp-2/src/main/java/datadog/trace/instrumentation/okhttp2/CallInstrumentation.java
```diff
@@ -0,0 +1,53 @@
...
+  @Override
+  public String[] helperClassNames() {
+    return new String[] {packageName + ".TracingCallbackWrapper"};
+  }
+
+  @Override
+  public void adviceTransformations(AdviceTransformation transformation) {
+    transformation.applyAdvice(
+        isMethod()
+            .and(named("enqueue"))
```
richardstartin, line 37: okhttp2 wraps the callback in `AsyncCall`, which is a `NamedRunnable`, which is a `Runnable` - so it should already have a context field embedded in it. This means that the `RunnableInstrumentation` should already handle this case, and would be more complete because `NamedRunnable.run` activates a scope `AsyncCall.execute` which encloses both `onResponse` and `onFailure` _and_ response marshalling. 

So there are two possibilities (happy to help investigate)
* this context propagation was already happening for okhttp2 and this new instrumentation is unnecessary, and we should just keep the test
* this built in context propagation isn't happening because there are exclusions being applied somewhere, we should remove those exclusions, and keep the test

dd-java-agent/instrumentation/okhttp-3/src/main/java/datadog/trace/instrumentation/okhttp3/CallInstrumentation.java
```diff
@@ -0,0 +1,51 @@
...
+public class CallInstrumentation extends Instrumenter.Tracing
+    implements Instrumenter.ForSingleType {
+
+  public CallInstrumentation() {
+    super("okhttp", "okhttp-3");
+  }
+
+  @Override
+  public String instrumentedType() {
+    return "okhttp3.Call";
```
richardstartin, line 25: this is an interface in okhttp3, so this doesn't look right.

dd-java-agent/instrumentation/okhttp-3/src/main/java/datadog/trace/instrumentation/okhttp3/CallInstrumentation.java
```diff
@@ -0,0 +1,51 @@
...
+
+  @Override
+  public String[] helperClassNames() {
+    return new String[] {packageName + ".TracingCallbackWrapper"};
+  }
+
+  @Override
+  public void adviceTransformations(AdviceTransformation transformation) {
+    transformation.applyAdvice(
+        isMethod().and(named("enqueue")).and(takesArgument(0, named("okhttp3.Callback"))),
```
richardstartin, line 36: okhttp3 has the same pattern - wrap in `AsyncCall`, which is a `NamedRunnable extends Runnable`. This means that the `RunnableInstrumentation` should provide better coverage, so we need to determine whether this instrumentation is unnecessary or if there is an inappropriate exclusion being applied, but keep the tests.

richardstartin (changes requested):
I want to get to the bottom of why `RunnableInstrumentation` doesn't apply here, since it would provide better scope activation for profiler coverage.

richardstartin:
I made some amendments on #4274 to demonstrate that this change is not necessary, so is not addressing the root cause of whatever problem has been observed.

dd-java-agent/instrumentation/okhttp-3/src/main/java/datadog/trace/instrumentation/okhttp3/CallInstrumentation.java
amarziali: yep I removed since unnecessary

dd-java-agent/instrumentation/okhttp-2/src/main/java/datadog/trace/instrumentation/okhttp2/CallInstrumentation.java
amarziali: yep I removed since unnecessary. The customer case was missing a root span so nothing had to be propagated

richardstartin (approved):
looks good!

bantonsson:
Sorry for being a bit late to this, but what does this actually add that is not covered by `"trace request with callback and parent"` in `HttpClientTest`?

richardstartin:
If it's redundant I'm not opposed to reverting.

PR #4279, opened 2022-11-18
Add parent field support collecting fields

jpbempel:
# What Does This Do
Respect max field count starting from the leaf up to the root. if max field count is reached during parent traversal, we stop in the middle.

# Motivation
Exception captured was not collecting relevant data as the exception is message is defined in `Throwable` class.

# Additional Notes

### DIFF
--- a/dd-java-agent/agent-debugger/src/main/java/com/datadog/debugger/util/MoshiSnapshotHelper.java
+++ b/dd-java-agent/agent-debugger/src/main/java/com/datadog/debugger/util/MoshiSnapshotHelper.java
@@ -17,13 +17,14 @@ import java.lang.reflect.Type;
 import java.nio.charset.StandardCharsets;
 import java.util.ArrayList;
 import java.util.Collection;
+import java.util.Collections;
 import java.util.HashMap;
 import java.util.Iterator;
 import java.util.List;
 import java.util.Map;
 import java.util.Set;
 import java.util.function.BiConsumer;
-import java.util.function.ObjIntConsumer;
+import java.util.function.Consumer;
 import okio.Okio;
 import org.slf4j.Logger;
 import org.slf4j.LoggerFactory;
@@ -371,6 +372,8 @@ public class MoshiSnapshotHelper {
                 } else {
                   value = values.stream().map(Snapshot.CapturedValue::getValue).toArray();
                 }
+              } else if (type.equals("java.util.Collections$EmptyList")) {
+                value = Collections.emptyList();
               } else {
                 throw new RuntimeException("Cannot deserialize type: " + type);
               }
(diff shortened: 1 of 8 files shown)


dd-java-agent/agent-debugger/src/main/java/com/datadog/debugger/util/MoshiSnapshotHelper.java
```diff
@@ -358,7 +359,8 @@ public Snapshot.CapturedValue fromJson(JsonReader jsonReader) throws IOException
               }
               jsonReader.endArray();
               if (type.equals(List.class.getTypeName())
-                  || type.equals(ArrayList.class.getTypeName())) {
+                  || type.equals(ArrayList.class.getTypeName())
+                  || type.equals("java.util.Collections$UnmodifiableRandomAccessList")) {
```
ojung, line 363: I'm not sure I understand this, but if we're special casing for `ArrayList` and `UnmodifiableRandomAccessList`, why don't we need special cases for for example for `LinkedList` or other List implementations?

jpbempel: Nobody is using `LinkedList` :)
Jokes apart, this code of deserialization is only used for unit tests

ojung: Ah, right. :) Why is it in `main`?

ojung (approved):

dd-java-agent/agent-debugger/src/main/java/com/datadog/debugger/util/MoshiSnapshotHelper.java
jpbempel: Because it's easier to put it here in the same class

dd-java-agent/agent-debugger/src/test/resources/com/datadog/debugger/sink/snapshotCapturedValueRegex.txt
```diff
@@ -27,6 +27,16 @@
 \{
 "type":"com.sun.management.internal.OperatingSystemImpl.ContainerCpuTicks",
 "notCapturedReason":"java.lang.reflect.InaccessibleObjectException: Unable to make field private com.sun.management.internal.OperatingSystemImpl.ContainerCpuTicks com.sun.management.internal.OperatingSystemImpl.processLoadTicks accessible: module jdk.management does not \\"opens com.sun.management.internal\\" to unnamed module @[0-9a-f]+"
+\},
+"jvm":
+\{
+"type":"sun.management.VMManagement",
+"notCapturedReason":"java.lang.reflect.InaccessibleObjectException: Unable to make field private final sun.management.VMManagement sun.management.BaseOperatingSystemImpl.jvm accessible: module java.management does not \\"opens sun.management\\" to unnamed module @[0-9a-f]+"
```
shatzi, line 34: should we just overwrite this "field not accessible", list all other "notCapturedReason" which are short and not freak the user 😄

jpbempel: We have already discussed that, and I strongly against what you suggest.
"Field not accessible" gives you a reason, but not actionnable advices for resolving the issue, and less information for troubleshooting. we can add more context about the error and why, but definitively not less

ojung: How about communicating the error and the actionable advice in a probe status?

jpbempel: if the probe can't be installed correctly, probe status is the right thing to do, but when it's a runtime error when collecting or evaluating conditions sounds like the snapshot is better suited for that

PR #4283, opened 2022-11-18
implement sampling and labelling for direct allocations

richardstartin:
# What Does This Do
* Events are now sampled
* Events now contain context fields
* MMAP and `ByteBuffer.allocateDirect` events now track their caller
* MMAP and `ByteBuffer.allocateDirect` events now record a duration
* MMAP events are best effort, instrumenting `sun.nio.ch.FileChannelImpl#map`.
* JNI events (which are dubious because they are often just leases on a buffer, it might be sensible to just remove these) obviously do not track their caller, but are not timed either because of limitations in the instrumentation approach.
* I made some attempts to extract commonalities with exception sampling
# Motivation

# Additional Notes


dd-java-agent/agent-bootstrap/src/main/java/datadog/trace/bootstrap/Agent.java
```diff
@@ -272,7 +272,7 @@ public void run() {
       } else {
         startProfilingAgent(false);
         // only enable sampler when we know JFR is ready
```
jbachorik, line 274: Nit: this comment should be updated as this is not about only the exception sampler any more

dd-java-agent/agent-bootstrap/src/main/java/datadog/trace/bootstrap/Agent.java
```diff
@@ -414,7 +414,7 @@ public AgentThread agentThread() {
     public void execute() {
       startProfilingAgent(false);
       // only enable sampler when we know JFR is ready
-      ExceptionSampling.enableExceptionSampling();
```
jbachorik, line 417: ditto

dd-java-agent/instrumentation/java-directbytebuffer/src/main/java/datadog/trace/instrumentation/directbytebuffer/ByteBufferInstrumentation.java
```diff
@@ -0,0 +1,46 @@
...
+public final class ByteBufferInstrumentation extends Instrumenter.Profiling
+    implements Instrumenter.ForBootstrap, Instrumenter.ForSingleType {
+
+  public ByteBufferInstrumentation() {
+    super("allocatedirect", "directallocation");
+  }
+
+  @Override
+  public boolean isEnabled() {
+    return Platform.isJavaVersionAtLeast(11) && super.isEnabled() && Platform.hasJfr();
```
jbachorik, line 23: Why is Java 11 required for the direct allocation related instrumentations?
Is it because of the `StackWalker` usage?

richardstartin: Yes, and the code base has so much language level indirection to ensure that every feature works on every language level even when the feature _can't_ work well at a given language level, I'd like to avoid that.

jbachorik: Ok, fair enough.

jbachorik (approved):

PR #4286, opened 2022-11-21
Test some uniqueness in IdGenerationStrategy

bantonsson:
# What Does This Do

Test that the `IdGenerationStrategy` at least returns some different ids.

# Motivation

# Additional Notes

### DIFF
--- a/dd-trace-api/src/test/groovy/datadog/trace/api/DDSpanIdTest.groovy
+++ b/dd-trace-api/src/test/groovy/datadog/trace/api/DDSpanIdTest.groovy
@@ -86,12 +86,18 @@ class DDSpanIdTest extends DDSpecification {
     ]
   }
 
-  def "generate id with #idGenerator"() {
+  def "generate id with #strategyName"() {
     when:
-    final long spanId = IdGenerationStrategy.fromName(strategyName).generateSpanId()
+    def strategy = IdGenerationStrategy.fromName(strategyName)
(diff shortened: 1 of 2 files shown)


dd-trace-api/src/test/groovy/datadog/trace/api/DDSpanIdTest.groovy
```diff
@@ -86,12 +86,18 @@ class DDSpanIdTest extends DDSpecification {
...
     then:
-    spanId != 0
+    spanIds.forEach { spanId ->
+      assert spanId != 0
+      assert !checked.contains(spanId)
+      checked.add(spanId)
+    }
 
     where:
     strategyName << ["RANDOM", "SEQUENTIAL"]
```
smola, line 103: Won't this be flaky for the `RANDOM` strategy? Or is the assumption here that for 32768 iterations, the probability is low enough?

bantonsson: Yes, a rough approximation of this birthday paradox (with 32768 persons and a 2^64 days long year) comes up with `2.9E-9` percent chance of a collision. Event though the `ThreadLocalRandom` generator isn't perfect, I assume it's at least that good. We could use a smaller range if you prefer.

mcculls (approved):

PR #4287, opened 2022-11-21
Fix contextual profiling events

richardstartin:
# What Does This Do

Puts the context fields back on the exception sample events and makes sure they are present on direct allocation samples. This was broken by #4283 

# Motivation

# Additional Notes


jbachorik (approved):
Argh. This is because of the JFR event bytecode generation, right?

PR #4289, opened 2022-11-21
Telemetry communication rework

ValentinZakharov:
# What Does This Do
Reworked communication - allows to handle various scenarios of available agent. Used feature discovery and fallback strategies with minimum log messages.

# Motivation

# Additional Notes
There are several scenarios that can happens:

## 1. There is no running DD Agent or it's old version of Agent without telemetry support or HTTP response code is 404.
In this case tracer will try to discover agent with feature discovery every 60 sec (getting list of available endpoints). The collected telemetry messages will be kept to send them later. If the telemetry buffers are full - then the oldest messages will be dropped.
Log message:
```
WARN: Unable to locate DD Agent with supported telemetry; will lookup every 60 seconds.
```

## 2. There is Agent with supported telemetry, but backend intake can't be reached
Typical case - when backend intake server returns any unexpected code (except 202 or 404).
In this case, tracer will try to retry in 60 seconds. If the telemetry buffers are full - then the oldest messages will be dropped.
Log messages:
```
WARN: Last attempt to send telemetry failed; will continue retrying every 60 seconds.
```
## 3. Agent have been updated to a version that supports telemetry
In this case customer no need to restart it`s application, tracer automatically will start sending telemetry.
Log messages:
```
INFO: Telemetry back to normal - message sent successfully
```


telemetry/src/main/java/datadog/telemetry/TelemetryRunnable.java
```diff
@@ -48,91 +46,77 @@ public void run() {
...
+          if (status != lastStatus) {
+            log.warn(
+                "Last attempt to send telemetry failed; will continue retrying every {} seconds.",
+                heartbeatIntervalMs / 1000);
+          }
+          break;
+        case SUCCESS:
+          if (status != lastStatus) {
+            log.info("Telemetry back to normal - message sent successfully");
+          }
```
smola, line 91:
```suggestion
          } else {
              log.debug("Telemetry message sent successfully");
          }
```

telemetry/src/main/java/datadog/telemetry/TelemetryServiceImpl.java
```diff
@@ -80,94 +47,178 @@ public boolean addConfiguration(Map<String, Object> configuration) {
...
+          return RequestStatus.SUCCESS;
+
+        case 404:
+          return RequestStatus.ENDPOINT_ERROR;
+
+        default:
+          return RequestStatus.HTTP_ERROR;
+      }
+    } catch (IOException e) {
+      log.warn("IOException on HTTP request to Telemetry Intake Service", e);
```
smola, line 220: The stacktrace is rarely useful in this case, maybe we should stick to this?

```suggestion
      log.warn("IOException on HTTP request to Telemetry Intake Service: {}", e.getMessage());
```

telemetry/src/main/java/datadog/telemetry/TelemetryServiceImpl.java
```diff
@@ -80,94 +47,178 @@ public boolean addConfiguration(Map<String, Object> configuration) {
...
+    try (Response response = okHttpClient.newCall(request).execute()) {
+      switch (response.code()) {
+        case 202:
+          return RequestStatus.SUCCESS;
+
+        case 404:
+          return RequestStatus.ENDPOINT_ERROR;
+
+        default:
+          return RequestStatus.HTTP_ERROR;
```
smola, line 217: A log debug message here would be good, somethig similar to what we had, such as:
```
      log.debug(
          "Telemetry Intake Service responded with: {} {}", response.code(), response.message());
```

telemetry/src/main/java/datadog/telemetry/TelemetryServiceImpl.java
```diff
@@ -80,94 +44,160 @@ public boolean addConfiguration(Map<String, Object> configuration) {
...
-    if (curTime - lastPreparationTimestamp > heartbeatIntervalMs) {
-      Request request = requestBuilderSupplier.get().build(RequestType.APP_HEARTBEAT);
-      queue.offer(request);
-      lastPreparationTimestamp = curTime;
+
+    return status;
+  }
+
+  RequestStatus sendMetrics(RequestBuilder requestBuilder) {
+    List<Metric> mtrs = new LinkedList<>(metrics);
```
jpbempel, line 141: Any particular reason for using `LinkedList` in this case instead of regular `ArrayList`?

telemetry/src/main/java/datadog/telemetry/TelemetryHttpClient.java
```diff
@@ -0,0 +1,41 @@
...
+
+        default:
+          log.debug(
+              "Telemetry Intake Service responded with: {} {}",
+              response.code(),
+              response.message());
+          return RequestStatus.HTTP_ERROR;
+      }
+    } catch (IOException e) {
+      log.warn("IOException on HTTP request to Telemetry Intake ServiceL {}", e.getMessage());
```
jpbempel, line 37: usually I prefer `e.toString()` because message can be `null` and it's useful sometimes to have the type of the exception at least (like for `NullPointerException`)

PR #4307, opened 2022-11-24
Add flush traces and metrics API

PerfectSlayer:
# What Does This Do

This PR creates a new *internal* API in order to control traces and metrics flush.

# Motivation

Its goal is to improve test stability and allow proper `system tests` implementation (no more introspection from Java client code).

# Additional Notes

Tracer will also try to flush at close in order prevent data loss.

### DIFF
--- a/dd-trace-core/src/main/java/datadog/trace/common/metrics/Aggregator.java
+++ b/dd-trace-core/src/main/java/datadog/trace/common/metrics/Aggregator.java
@@ -63,7 +63,7 @@ final class Aggregator implements Runnable {
       try {
         InboxItem item = inbox.take();
         if (item instanceof StopSignal) {
-          report(wallClockTime(), (StopSignal) item);
+          stop((StopSignal) item);
           break;
         } else if (item instanceof ReportSignal) {
           report(wallClockTime(), (ReportSignal) item);
@@ -89,6 +89,16 @@ final class Aggregator implements Runnable {
       }
     }
     log.debug("metrics aggregator exited");
+  }
+
+  private void stop(StopSignal stopSignal) {
+    report(wallClockTime(), stopSignal);
+    for (InboxItem item : inbox) {
+      if (item instanceof SignalItem) {
+        ((SignalItem) item).ignore();
+      }
+    }
+    stopSignal.complete();
   }
 
   private void report(long when, SignalItem signal) {
--- a/dd-trace-core/src/main/java/datadog/trace/common/metrics/Aggregator.java
+++ b/dd-trace-core/src/main/java/datadog/trace/common/metrics/Aggregator.java
@@ -1,11 +1,9 @@
 package datadog.trace.common.metrics;
 
-import static datadog.trace.common.metrics.Aggregator.ReportEvent.SENT;
-import static datadog.trace.common.metrics.Batch.REPORT;
-import static datadog.trace.common.metrics.Batch.REPORT_AND_NOTIFY;
-import static datadog.trace.common.metrics.ConflatingMetricsAggregator.POISON_PILL;
 import static java.util.concurrent.TimeUnit.MILLISECONDS;
 
+import datadog.trace.common.metrics.SignalItem.ReportSignal;
+import datadog.trace.common.metrics.SignalItem.StopSignal;
 import datadog.trace.core.util.LRUCache;
 import java.util.Iterator;
 import java.util.Map;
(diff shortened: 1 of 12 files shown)


dd-trace-core/src/main/java/datadog/trace/common/metrics/Aggregator.java
```diff
@@ -111,6 +116,9 @@ private void report(long when) {
       }
       dirty = false;
     }
+    if (notify) {
+      outbox.add(SENT);
```
bantonsson, line 120: I would use `offer` here, since we can't have the `Aggregator` fail if the queue gets full.

dd-trace-core/src/main/java/datadog/trace/common/metrics/ConflatingMetricsAggregator.java
```diff
@@ -175,6 +181,29 @@ public boolean report() {
...
   }
 
+  @Override
+  public Future<Boolean> forceReport() {
+    while (thread.isAlive() && !report()) {
+      try {
+        Thread.sleep(10);
+      } catch (InterruptedException e) {
+        log.debug("Failed to ask for report");
+        break;
```
bantonsson, line 191: Won't this lead to the thread asking for the `report` to get stuck on the `Future` returned below, if the `Aggregator` never gets around to adding something to the `outbox`?

dd-trace-core/src/main/java/datadog/trace/common/metrics/ConflatingMetricsAggregator.java
```diff
@@ -175,6 +181,29 @@ public boolean report() {
...
+      }
+    }
+    return CompletableFuture.supplyAsync(this::waitForOutboxElement);
```
bantonsson, line 194: It feels a bit weird to communicate via a `queue` and moving execution to a _third_ party just to get a `Future`. Another way to do this, would be to have different types of `Batch` and just have one where you send in the `CompletableFuture` to the `Aggregator` and that is what gets completed at the end of the `report`.

dd-trace-core/src/main/java/datadog/trace/common/metrics/Aggregator.java
PerfectSlayer: Agreed

dd-trace-core/src/main/java/datadog/trace/common/metrics/ConflatingMetricsAggregator.java
PerfectSlayer: Yes it will. I changed the logic. Thanks!

dd-trace-core/src/main/java/datadog/trace/common/metrics/ConflatingMetricsAggregator.java
PerfectSlayer: Agreed. I didn't wanted to change the the `Batch` type, and the `Future` can't fit in it...
On your suggestion, I created an interface for the `inbox` items, which can now be a `Batch` or a `Signal`. Signals being either a `StopSignal` (previous the `POISON_PILL` or `NULL` batch) or `Report`.
`Signal` have a `Future`, completed by `Aggregator`, that API consumer  can await for.

dd-trace-core/src/main/java/datadog/trace/core/CoreTracer.java
```diff
@@ -953,6 +954,15 @@ public void flush() {
     writer.flush();
   }
 
+  @Override
+  public void flushMetrics() {
+    try {
+      metricsAggregator.forceReport().get();
```
bantonsson, line 960: Should there be a timeout here, if there is a race between the publishing in `forceReport` and thread shutdown?

bantonsson: I think that completing all the `Signal` things in the `inbox` when you see `Stop` is good, but it only makes things less likely, since there is still a gap between `thread.isAlive()` and `inbox.offer(...)` where the _stopping_ thread can be finished with the `inbox` but still [_not dead_](https://www.youtube.com/watch?v=grbSQ6O6kbs). Maybe also add a fixed timeout (I mean isn't this just for testing mostly?)

PerfectSlayer: I push an update with pending signal completion (marked as ignored) and a 1s timeout.
As it an internal API for the test, I guess one second is acceptable to wait for (usually at the end of a test suite).

bantonsson (approved):

dd-trace-core/src/main/java/datadog/trace/core/CoreTracer.java
bantonsson: LGTM

PR #4312, opened 2022-11-24
spring mvc filter: complies with servlet < 4 specs

amarziali:
# What Does This Do

Spring Mvc 5.3 instrumentation defines a filter complying with servlet >= 4 spec. If using an older servlet spec defaults methods like init and destroy are not defaulted in the interface.

# Motivation

# Additional Notes


mcculls (approved):

bantonsson:
@amarziali would it be possible to add a test for this?

PR #4313, opened 2022-11-24
Introduced evaluateAt attribute in probe configs

jpbempel:
# What Does This Do
Introduced evaluateAt attribute in probe configs
Refactored probe config builders

# Motivation

# Additional Notes
attribute is not used right now, but will be wired in a next PR


shatzi (approved):
nice

PR #4316, opened 2022-11-25
Remove ClassFormatError handling during initialization

smola:
# What Does This Do

Remove ClassFormatError handling during initialization. This should not be needed since dd-trace-java v1.0.0.

# Motivation

# Additional Notes

### DIFF
--- a/dd-java-agent/agent-bootstrap/src/main/java/datadog/trace/bootstrap/Agent.java
+++ b/dd-java-agent/agent-bootstrap/src/main/java/datadog/trace/bootstrap/Agent.java
@@ -783,12 +783,6 @@ public class Agent {
               }
             });
       }
-    } catch (final ClassFormatError e) {
-      /*
-      Profiling is compiled for Java8. Loading it on Java7 results in ClassFormatError
-      (more specifically UnsupportedClassVersionError). Just ignore and continue when this happens.
-      */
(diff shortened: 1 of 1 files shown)


dd-java-agent/agent-bootstrap/src/main/java/datadog/trace/bootstrap/Agent.java
```diff
@@ -849,12 +849,6 @@ private static synchronized void startDebuggerAgent(
       final Method debuggerInstallerMethod =
           debuggerAgentClass.getMethod("run", Instrumentation.class, scoClass);
       debuggerInstallerMethod.invoke(null, inst, sco);
-    } catch (final ClassFormatError e) {
```
jpbempel, line 852: Can do the same with: 
 * https://github.com/DataDog/dd-trace-java/blob/master/dd-java-agent/agent-bootstrap/src/main/java/datadog/trace/bootstrap/Agent.java#L786-L791
 * https://github.com/DataDog/dd-trace-java/blob/master/dd-java-agent/agent-bootstrap/src/main/java/datadog/trace/bootstrap/Agent.java#L818-L823

jpbempel (approved):
LGTM for me
Profiling team should take a look also if they agree on this also

richardstartin (approved):

PR #4317, opened 2022-11-25
Add support for tainting headers

manuel-alvarez-alvarez:
# What Does This Do
Add support for tainting of header names and values

# Motivation

# Additional Notes

### DIFF
--- a/dd-java-agent/instrumentation/servlet/src/main/java/datadog/trace/instrumentation/servlet/request/HttpServletRequestCallSite.java
+++ b/dd-java-agent/instrumentation/servlet/src/main/java/datadog/trace/instrumentation/servlet/request/HttpServletRequestCallSite.java
@@ -1,0 +1,107 @@
+package datadog.trace.instrumentation.servlet.request;
+
+import datadog.trace.agent.tooling.csi.CallSite;
+import datadog.trace.api.iast.IastAdvice;
+import datadog.trace.api.iast.InstrumentationBridge;
+import datadog.trace.api.iast.source.WebModule;
+import datadog.trace.util.stacktrace.StackUtils;
+import java.util.ArrayList;
+import java.util.Collections;
+import java.util.Enumeration;
+import java.util.List;
+import javax.servlet.http.HttpServletRequest;
+
+@CallSite(spi = IastAdvice.class)
+public class HttpServletRequestCallSite {
+
+  @CallSite.AfterArray({
+    @CallSite.After(
+        "java.lang.String javax.servlet.http.HttpServletRequest.getHeader(java.lang.String)"),
+    @CallSite.After(
+        "java.lang.String javax.servlet.http.HttpServletRequestWrapper.getHeader(java.lang.String)"),
+  })
+  public static String afterGetHeader(
+      @CallSite.This final HttpServletRequest self,
+      @CallSite.Argument final String headerName,
+      @CallSite.Return final String headerValue) {
+    final WebModule module = InstrumentationBridge.WEB;
+    if (module != null) {
+      try {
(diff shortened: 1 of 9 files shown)


jandro996:
Seems that we can reuse the same code for Request hedaders, parameters and cookies as we will  taint name and value

dd-java-agent/instrumentation/servlet/src/test/groovy/HttpServletRequestCallSiteTest.groovy
```diff
@@ -0,0 +1,86 @@
...
+import datadog.trace.agent.test.AgentTestRunner
+import datadog.trace.api.Platform
+import datadog.trace.api.iast.InstrumentationBridge
+import datadog.trace.api.iast.source.WebModule
+import spock.lang.IgnoreIf
+
+import javax.servlet.http.HttpServletRequest
+import javax.servlet.http.HttpServletRequestWrapper
+
+@IgnoreIf({
```
jandro996, line 11: I think we  can remove this kind of validations for < jdk8

jandro996 (approved):

dd-java-agent/instrumentation/servlet/src/test/groovy/HttpServletRequestCallSiteTest.groovy
manuel-alvarez-alvarez: Yep, https://github.com/DataDog/dd-trace-java/pull/4418

let me change it

dd-java-agent/instrumentation/servlet/src/main/java/datadog/trace/instrumentation/servlet/request/HttpServletRequestCallSite.java
```diff
@@ -0,0 +1,94 @@
...
+    if (enumeration == null) {
+      return null;
+    }
+    final WebModule module = InstrumentationBridge.WEB;
+    if (module == null) {
+      return enumeration;
+    }
+    final List<String> headerNames = new ArrayList<>();
+    while (enumeration.hasMoreElements()) {
+      final String headerName = (String) enumeration.nextElement();
```
smola, line 84: This should probably be within the try/catch block?

smola: Maybe also the `hasMoreElements()`...

manuel-alvarez-alvarez: Nice catch, it's a bit more complex like move it withing the try/catch block. If they fail it's servlet code that fails and we should let the exception propagate removing ourselves from the stack first. Check the latest version and the specific tests for that matter:

```java
  def 'test that get headers does not fail when servlet related code fails'(final Class<? extends HttpServletRequest> clazz) {
    setup:
    final iastModule = Mock(WebModule)
    InstrumentationBridge.registerIastModule(iastModule)
    final enumeration = Mock(Enumeration) {
      hasMoreElements() >> { throw new NuclearBomb('Boom!!!')}
    }
    final testSuite = new TestHttpServletRequestCallSiteSuite(Mock(clazz) {
      getHeaders('header') >> enumeration
    })

    when:
    testSuite.getHeaders('header')

    then:
    final bomb = thrown(NuclearBomb)
    bomb.stackTrace.find { it.className == HttpServletRequestCallSite.name } == null

    where:
    clazz                     | _
    HttpServletRequest        | _
    HttpServletRequestWrapper | _
  }
```

smola (approved):

PR #4318, opened 2022-11-25
Add support for tainting cookies

jandro996:
# What Does This Do

Add support for tainting cookies by instrumenting [getCookies](https://javaee.github.io/javaee-spec/javadocs/javax/servlet/http/HttpServletRequest.html#getCookies--)()

# Motivation

# Additional Notes

Quick version using all String properties from [Cookie](https://javaee.github.io/javaee-spec/javadocs/javax/servlet/http/Cookie.html)

### DIFF
--- a/dd-java-agent/agent-iast/src/main/java/com/datadog/iast/IastModuleImpl.java
+++ b/dd-java-agent/agent-iast/src/main/java/com/datadog/iast/IastModuleImpl.java
@@ -443,6 +443,21 @@ public final class IastModuleImpl implements IastModule {
     checkInjection(VulnerabilityType.PATH_TRAVERSAL, rangeProvider);
   }
 
+  @Override
+  public void onCookie(@Nonnull String... cookieStrings) {
+    final IastRequestContext ctx = IastRequestContext.get();
+    if (ctx == null) {
+      return;
+    }
+    final TaintedObjects taintedObjects = ctx.getTaintedObjects();
+    for (String cookieString : cookieStrings) {
+      if (canBeTaintedNullSafe(cookieString)) {
+        taintedObjects.taintInputString(
+            cookieString, new Source(SourceType.REQUEST_COOKIE, cookieString, null));
+      }
+    }
+  }
+
   private <E> void checkInjection(
       @Nonnull final InjectionType type, @Nonnull final RangesProvider<E> rangeProvider) {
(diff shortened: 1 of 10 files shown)


dd-java-agent/agent-iast/src/main/java/com/datadog/iast/IastModuleImpl.java
```diff
@@ -415,6 +443,21 @@ public void onPathTraversal(final @Nullable File parent, final @Nonnull String c
...
+  public void onCookie(@Nonnull String... cookieStrings) {
+    final IastRequestContext ctx = IastRequestContext.get();
+    if (ctx == null) {
+      return;
+    }
+    final TaintedObjects taintedObjects = ctx.getTaintedObjects();
+    for (String cookieString : cookieStrings) {
+      if (canBeTaintedNullSafe(cookieString)) {
+        taintedObjects.taintInputString(
+            cookieString, new Source(SourceType.REQUEST_COOKIE, cookieString, null));
```
smola, line 456: `cookieString` would be the `Source#value`, not `Source#name`.

internal-api/src/main/java/datadog/trace/api/iast/IastModule.java
```diff
@@ -54,4 +58,6 @@ void onStringConcatFactory(
   void onPathTraversal(@Nonnull URI uri);
 
   void onPathTraversal(@Nullable File parent, @Nonnull String child);
+
+  void onCookie(@Nonnull String... cookieStrings);
```
smola, line 62: Maybe we could pass the args for cookies as individual parameters (they are standard, after all)? This would allow us to distinguish the cookie name, to set it in the source.

jandro996: Yes, this is a quick approach for the demo. 

As I've commented in https://github.com/DataDog/dd-trace-java/pull/4317 maybe we can unify the taint of Name/Value elements

dd-java-agent/agent-iast/src/main/java/com/datadog/iast/IastModuleImpl.java
jandro996: After demo I will change it to a Name/Value approach 👍

PR #4320, opened 2022-11-25
Add evaluateAt support for probe instrumentations

jpbempel:
# What Does This Do
 * Snapshot: indicates for method probe where conditions are evaluated before serializing data. Either at entry or exit of the method.
 * Metric: indicates for method probe where the metric is emitted and with which values. For example at Exit of the method you can emit the return value of the method.
 * Log: use the same mechanism than snapshot but can report for example at exit of the method the return value.

# Motivation
following #4313 

# Additional Notes

### DIFF
--- a/dd-java-agent/agent-debugger/debugger-bootstrap/src/main/java/datadog/trace/bootstrap/debugger/Snapshot.java
+++ b/dd-java-agent/agent-debugger/debugger-bootstrap/src/main/java/datadog/trace/bootstrap/debugger/Snapshot.java
@@ -92,7 +92,7 @@ public class Snapshot {
   public void setEntry(CapturedContext context) {
     summaryBuilder.addEntry(context);
     context.setThisClassName(thisClassName);
-    if ((probe.getEvaluateAt() == MethodLocation.NONE
+    if ((probe.getEvaluateAt() == MethodLocation.DEFAULT
             || probe.getEvaluateAt() == MethodLocation.ENTRY)
         && checkCapture(context)) {
       captures.setEntry(context);
@@ -104,7 +104,7 @@ public class Snapshot {
     context.addExtension(ValueReferences.DURATION_EXTENSION_NAME, duration);
     summaryBuilder.addExit(context);
     context.setThisClassName(thisClassName);
-    if ((probe.getEvaluateAt() == MethodLocation.NONE
+    if ((probe.getEvaluateAt() == MethodLocation.DEFAULT
             || probe.getEvaluateAt() == MethodLocation.EXIT)
         && checkCapture(context)) {
       captures.setReturn(context);
@@ -296,7 +296,7 @@ public class Snapshot {
   }
 
   public enum MethodLocation {
-    NONE,
+    DEFAULT,
     ENTRY,
     EXIT
   }
@@ -320,7 +320,7 @@ public class Snapshot {
       this(
           id,
           location,
-          MethodLocation.NONE,
+          MethodLocation.DEFAULT,
           null,
           null,
           new SnapshotSummaryBuilder(location),
--- a/dd-java-agent/agent-debugger/debugger-bootstrap/src/main/java/datadog/trace/bootstrap/debugger/Snapshot.java
+++ b/dd-java-agent/agent-debugger/debugger-bootstrap/src/main/java/datadog/trace/bootstrap/debugger/Snapshot.java
@@ -92,7 +92,9 @@ public class Snapshot {
   public void setEntry(CapturedContext context) {
     summaryBuilder.addEntry(context);
     context.setThisClassName(thisClassName);
-    if (checkCapture(context)) {
+    if ((probe.getEvaluateAt() == MethodLocation.NONE
+            || probe.getEvaluateAt() == MethodLocation.ENTRY)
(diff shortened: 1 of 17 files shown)


dd-java-agent/agent-debugger/src/main/java/com/datadog/debugger/agent/ConfigurationUpdater.java
```diff
@@ -302,6 +303,19 @@ private Snapshot.ProbeDetails convertToProbeDetails(
...
+      ProbeDefinition.MethodLocation methodLocation) {
+    switch (methodLocation) {
+      case NONE:
+        return Snapshot.MethodLocation.NONE;
+      case ENTRY:
+        return Snapshot.MethodLocation.ENTRY;
+      case EXIT:
+        return Snapshot.MethodLocation.EXIT;
+    }
+    return null;
```
ojung, line 316: Will this become a NPE in the caller?

dd-java-agent/agent-debugger/src/main/java/com/datadog/debugger/instrumentation/Instrumentor.java
```diff
@@ -134,6 +135,68 @@ protected void fillLineMap() {
...
+    return node;
+  }
+
+  protected InsnList getBeforeReturnInsnList(AbstractInsnNode node) {
+    return null;
+  }
+
+  protected LabelNode getReturnHandler(AbstractInsnNode exitNode) {
+    // exit node must have been removed from the original instruction list
+    assert exitNode.getNext() == null && exitNode.getPrevious() == null;
```
ojung, line 183: Assertions aren't enabled by default, right?

dd-java-agent/agent-debugger/src/main/java/com/datadog/debugger/agent/ConfigurationUpdater.java
jpbempel: hmmm no, why?

dd-java-agent/agent-debugger/src/main/java/com/datadog/debugger/instrumentation/Instrumentor.java
jpbempel: no, only if you enable with `-ea`

ojung: Soo, this is redundant because nobody enables assertions?

jpbempel: removing...

ojung: How about replacing with a `IllegalArgumentException`?

dd-java-agent/agent-debugger/src/main/java/com/datadog/debugger/instrumentation/MetricInstrumentor.java
```diff
@@ -55,11 +56,31 @@ public void instrument() {
...
+        default:
+          throw new IllegalArgumentException(
+              "Invalid evaluateAt attribute: " + definition.getEvaluateAt());
+      }
     }
   }
 
+  @Override
+  protected InsnList getBeforeReturnInsnList(AbstractInsnNode node) {
+    return callMetric(metricProbe);
```
shatzi, line 81: will this callMetric would have access to @return and @duration?

dd-java-agent/agent-debugger/debugger-bootstrap/src/main/java/datadog/trace/bootstrap/debugger/Snapshot.java
```diff
@@ -290,6 +295,12 @@ public enum Kind {
     AFTER;
   }
 
+  public enum MethodLocation {
+    NONE,
```
shatzi, line 299: I find the name and value confusing.

evalautAt = None means evaluate both at entry and exit for snapshot but entry for metric and logs (I think). maybe change it to DEFAULT or use null to indicate there is nothing there.

Also suggest to rename to EvalauteAtLocation not confusing with method inside ProbleLocation.

dd-java-agent/agent-debugger/src/main/java/com/datadog/debugger/instrumentation/MetricInstrumentor.java
shatzi: would we be able to add condition for metrics in this fashion as well?

shatzi (approved):
looks good. some questions regarding metrics

dd-java-agent/agent-debugger/src/main/java/com/datadog/debugger/instrumentation/MetricInstrumentor.java
jpbempel: > will this callMetric would have access to @return and @duration?

Nope, they are not introduced in this context, yet

> would we be able to add condition for metrics in this fashion as well?

yes it's possible, need some work, though

PR #4324, opened 2022-11-28
do not reference Filter interface

amarziali:
# What Does This Do

Avoid referencing Filter interface but it relies on spring DelegatingFilterProxy for the same.

Also allow disabling the instrumentation by providing an alternate name `spring-path-filter` to inject the `OrderedServletPathRequestFilter`

# Motivation

# Additional Notes

Tested on:
* Spring web 5.3 with tomcat 8 -> working ok
* Spring web 5.3 (boot 2.7.x) -> working ok
* Spring webflux 5.3 (boot 2.7.x) -> Muzzled expected
* Spring webflux 6 (boot 3) -> Muzzled expected
* Spring web 6 (boot 3) -> Muzzled expected


amarziali:
Fix for #4323

mcculls (approved):

PR #4332, opened 2022-11-28
rework profiling context integration

richardstartin:
# What Does This Do

* Rework `ContextThreadListener` interface and registration 
  * renamed as `ProfilingContextIntegration` which is what it is
  * there can only be one of them, and it needs to be registered before creating the tracer
  * its capabilities are expanded, and the activation calls are privileged, being called first on any scope push, and last after any scope cleanup, this means that even datadog activity can be labeled and attributed by the profiler
* Store the native thread id in the scope stack for performance reasons, which eliminates most of the overhead of the profiler integration (setting the context is now very fast, and this eliminates a secondary thread-local) 

# Motivation

# Additional Notes


jbachorik (approved):
Looks reasonable

bantonsson (approved):

PR #4340, opened 2022-11-29
Add pitest support (mutation testing)

smola:
# What Does This Do
Add [pitest](http://pitest.org/) support (mutation testing).

Tested modules: iast, appsec, telemetry.

# Additional notes

Even after https://github.com/DataDog/dd-trace-java/pull/4330, there are larger changes related to `Config` handling, `SpockRunner`, etc, to be able to use pitest in some submodules, such as `instrumentation/*`.

# How it works

Running pitest for the iast module: `./gradlew :dd-java-agent:agent-iast:pitest`

Reports are at `dd-java-agent/agent-iast/build/reports/pitest/index.html`.

Screenshot from 2022-11-29 11-00-00

Screenshot from 2022-11-29 11-01-23

Screenshot from 2022-11-29 11-01-34

### DIFF
--- a/build.gradle
+++ b/build.gradle
@@ -12,6 +12,7 @@ plugins {
   id "com.github.johnrengelman.shadow" version "7.1.2" apply false
   id "me.champeau.jmh" version "0.6.5" apply false
   id 'org.gradle.playframework' version '0.12' apply false
+  id 'info.solidsoft.pitest' version '1.9.11'  apply false
 }
 
 description = 'dd-trace-java'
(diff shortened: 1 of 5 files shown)


richardstartin:
Can you confirm that this will not run on every push?

smola:
@richardstartin That's correct. It will not run on CI, and it won't run in any default gradle task like build or assemble. It will only run if you invoke it explicitly with the `pitest` task.

richardstartin (approved):

PR #4346, opened 2022-11-29
Propagation benchmarks

bantonsson:
# What Does This Do

Adds benchmarks for propagation extraction and injection. Also does a small fix so we don't do a substring of the value string for `x-datadog-tags` that is never used.

# Motivation

# Additional Notes


dd-trace-core/src/jmh/java/datadog/trace/core/propagation/ExtractorBenchmark.java
```diff
@@ -0,0 +1,127 @@
...
+
+  Map<String, String> headers;
+  HttpCodec.Extractor extractor;
+  DDTraceId traceId;
+  long spanId;
+
+  @Setup(Level.Trial)
+  public void setUp() {
+    headers = new HashMap<>();
+    // Yes, the benchmark will be skewed since we reuse String instances for all iterations, but at
```
richardstartin, line 53: I think it would be better to duplicate via `new String(string)` for each invocation, then add a benchmark to just do the duplication as a baseline.

dd-trace-core/src/main/java/datadog/trace/core/propagation/DatadogHttpCodec.java
```diff
@@ -167,23 +167,22 @@ public boolean accept(String key, String value) {
       }
       if (classification != IGNORE) {
         try {
-          String firstValue = firstHeaderValue(value);
-          if (null != firstValue) {
+          if (null != value) {
             switch (classification) {
               case TRACE_ID:
-                traceId = DDTraceId.from(firstValue);
+                traceId = DDTraceId.from(firstHeaderValue(value));
```
richardstartin, line 173: what if the first value is null?

dd-trace-core/src/jmh/java/datadog/trace/core/propagation/ExtractorBenchmark.java
bantonsson: Yeah, I would need to do something other than the `HashMap` to not trigger hash code computation eagerly as well.

richardstartin: I think you should expect for the hash code to be calculated from a fresh string when this code runs for real

dd-trace-core/src/main/java/datadog/trace/core/propagation/DatadogHttpCodec.java
bantonsson: 🤦🏼‍♂️ Thanks for catching it.

dd-trace-core/src/jmh/java/datadog/trace/core/propagation/ExtractorBenchmark.java
bantonsson: Yup, I'll rewrite it.

richardstartin (approved):

PR #4349, opened 2022-11-29
[DSM] Don't send null tags with data streams checkpoint

hokitam:
# What Does This Do
This PR will remove any null tags that we try to send along in a data streams checkpoint.

# Motivation
Null tags cause deserialization issues in the Data Streams backend, causing checkpoints to be dropped.

# Additional Notes


nayeem-kamal (approved):
LGTM

PR #4351, opened 2022-11-30
Split flaky tests to its own CI job

smola:
# What Does This Do
* Introduce a `@Flaky` annotation to mark flaky tests. In Spock, it is managed with a global extension (`FlakySpockExtension`). In JUnit 5, it uses a `@Tag`.
* By default, flaky tests are run (for local development).
* `-PskipFlakyTests` will skip all flaky tests, used for most CI jobs.
* `-PrunFlakyTests` will run only flaky tests, excluding everything else. Added a CI a job to run all flaky tests in JDK 8. Failures in this job will be ignored (always green), but its results will be available both in Circle CI _Tests_ tab as well as in CI Visibility dashboards.
* Mark most flaky tests with `@Flaky`.
* Keep some flaky tests with `@Retry` and without `@Ignore` (e.g. `WeakMapTest`) where testing in multiple JDK versions might be important.
* When a test is defined in a parent class, but it is flaky only in a given subclass, an array of suite names can be given, for example: `@Flaky(value = "flaky on some tests from grizzly-2 module", suites = ["GrizzlyAsyncTest"])`. This has the caveat that it'll mark it as flaky as long as the suite name matches, and in some cases there are suites with the same name in multiple instrumentation modules.

# Motivation
I like green.

# Additional Notes

* Tests that are still ignored after this PR:
```
$ grep -r '@Ignore(' $(find -name 'src')
./dd-smoke-tests/springboot-grpc/src/test/groovy/datadog/smoketest/SpringBootGrpcSpannerTest.groovy:@Ignore("can unignore when strict continuation reference counting is dropped")
./dd-trace-core/src/traceAgentTest/groovy/DataStreamsIntegrationTest.groovy:@Ignore("The agent in CI doesn't have a valid API key. Unlike metrics and traces, data streams fails in this case")
./dd-java-agent/instrumentation/vertx-rx-3.5/src/test/groovy/server/VertxRxHttpServerForkedTest.groovy:@Ignore("This test isn't different from VertxHttpServerTest. Needs rework to meet intent")
./dd-java-agent/instrumentation/junit-4.10/src/test/java/org/example/TestSkipped.java:  @Ignore("Ignore reason in test")
./dd-java-agent/instrumentation/junit-4.10/src/test/java/org/example/TestSkippedClass.java:@Ignore("Ignore reason in class")
./dd-java-agent/instrumentation/kafka-streams-0.11/src/test/groovy/KafkaStreamsTestBase.groovy:  @Ignore("Repeatedly fails with the wrong parent span https://github.com/DataDog/dd-trace-java/issues/3865")
./dd-java-agent/src/test/groovy/datadog/trace/agent/integration/classloading/ShadowPackageRenamingTest.groovy:  //  @Ignore("OT 0.32 removed this field.  Need to find another option.")
./dd-java-agent/src/test/groovy/datadog/trace/agent/integration/classloading/ShadowPackageRenamingTest.groovy:  @Ignore("Agent jar check inaccurate")
```


.circleci/config.yml
```diff
@@ -363,6 +363,9 @@ jobs:
       gradleTarget:
         type: string
         default: ""
+      gradleExtra:
```
PerfectSlayer, line 366: Can this parameter have a more explicit name? 
It is used with ` || true` value which does not really apply to gradle parameters.
What about creating a dedicated parameter instead? Like `continueOnFailure`or something similar?

utils/test-utils/src/main/groovy/datadog/trace/test/util/Flaky.java
```diff
@@ -0,0 +1,25 @@
...
+/**
+ * Use this annotation for suites or test cases that are flaky. When running in CI, these will be
+ * seggregated to a separate job.
+ */
+@Retention(RetentionPolicy.RUNTIME)
+@Target({ElementType.TYPE, ElementType.METHOD})
+@Tag("flaky")
+public @interface Flaky {
+  /** Reason why the test is flaky (optional). */
+  String value() default "";
```
PerfectSlayer, line 18: Why not call it `reason()`?

PerfectSlayer (approved):
Some `@Flacky` annotations are commented but I guess you are still figuring out which ones to comment?

Some comments but great addition! 👍  We all like 💚

utils/test-utils/src/main/groovy/datadog/trace/test/util/Flaky.java
smola: @PerfectSlayer So that it can be used as a positional parameter (e.g. `@Flaky("foo")` rather than `@Flaky(reason="foo")`.

.circleci/config.yml
smola: Done.

smola:
@PerfectSlayer The remaining commented `@Flaky` annotations are those related to `awaitGC`. These tend to be particularly flaky in OpenJ9, and I haven't figured out if we want to:
* Keep a `@Retry` annotation on them.
* Add a `@Flaky` annotation (but lose testing outside of OpenJDK 8).
* Add a `@Flaky` annotation, but add jobs for flaky tests in multiple JDKs.

richardstartin (approved):
For the changes to `CodeHotspotsTest`

jpbempel (approved):

PR #4358, opened 2022-12-01
Adjust notCapturedReason FieldCount location

jpbempel:
# What Does This Do
In snapshot, when field count reached the maximum set, the attribute notCapturedReason was along with the list of fields. We move this up along with value attributes (fields, type, ...)

# Motivation
make snapshot structure more coherent and easier to parse

# Additional Notes


shatzi (approved):
LGTM

PR #4378, opened 2022-12-05
Remove prefixes for value reference in expressions

jpbempel:
# What Does This Do
Value references are now without prefix except field access with `this` or synthetic variable (`@duration`).
Json representation have then changed to use `ref` and `getmember` operator.
The `ValueReferenceResolver` have evolved to have 2 methods instead of
matching those operators: `lookup` and `getmember`

# Motivation
More user-friendly way to express var/field accesses

# Additional Notes


dd-java-agent/agent-debugger/src/main/java/com/datadog/debugger/instrumentation/MetricInstrumentor.java
```diff
@@ -245,71 +260,84 @@ public CompileToInsnList(
...
-      return true;
+      return UndefinedValue.INSTANCE;
+    }
+
+    private Type followReferences(Type currentType, String name, InsnList insnList) {
+      Class<?> clazz;
+      try {
+        String className = currentType.getClassName();
+        clazz = Class.forName(className, true, classLoader);
+        Field declaredField = clazz.getDeclaredField(name); // no parent fields!
```
shatzi, line 302: also this doesn't support interfaces right?

dd-java-agent/agent-debugger/src/main/java/com/datadog/debugger/instrumentation/MetricInstrumentor.java
```diff
@@ -324,61 +352,38 @@ private Type tryRetrieveArgument(String head, InsnList insnList, boolean isLast)
...
-            reportError(
-                String.format(
-                    "Incompatible type for local var %s: %s with expected type: %s",
-                    head, varType.getClassName(), expectedType.getClassName()));
-            return null;
-          }
           VarInsnNode varInsnNode =
               new VarInsnNode(varType.getOpcode(Opcodes.ILOAD), varNode.index);
           insnList.add(varInsnNode);
           convertIfRequired(varType, expectedType, insnList);
```
shatzi, line 372: not sure if convertIfRequired should be called on every tryRetrieveLocalVar but only at the end of expressions.

shatzi (changes requested):
I think there is a bug regarding the resolving references. maybe I'm wrong

jpbempel:
> I think there is a bug regarding the resolving references. maybe I'm wrong

yeah nice catches! need to revisit and add more tests 😞

dd-java-agent/agent-debugger/src/main/java/com/datadog/debugger/instrumentation/MetricInstrumentor.java
jpbempel: we don't support calling method, so interfaces too by consequence

dd-java-agent/agent-debugger/src/main/java/com/datadog/debugger/instrumentation/MetricInstrumentor.java
jpbempel: true

dd-java-agent/agent-debugger/debugger-el/src/test/java/com/datadog/debugger/el/expressions/ValueRefExpressionTest.java
```diff
@@ -47,40 +26,24 @@ void setup() {
...
-    ValueRefExpression valueRef = new ValueRefExpression(".b");
-    ValueRefExpression invalidValueRef = new ValueRefExpression(".x");
+    ValueRefExpression valueRef = new ValueRefExpression("b");
+    ValueRefExpression invalidValueRef = new ValueRefExpression("x");
 
     IsEmptyExpression isEmpty = new IsEmptyExpression(valueRef);
     IsEmptyExpression isEmptyInvalid = new IsEmptyExpression(invalidValueRef);
 
-    ExTester instance = new ExTester(null, "hello");
+    ExTester instance = new ExTester(null, "\"hello\"");
```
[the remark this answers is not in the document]
jpbempel: refactoring leftover

dd-java-agent/agent-debugger/debugger-bootstrap/src/main/java/datadog/trace/bootstrap/debugger/Snapshot.java
```diff
@@ -613,108 +611,91 @@ private CapturedContext(CapturedContext other, Map<String, Object> extensions) {
...
-      return extensions.containsKey(name) ? extensions.get(name) : Values.UNDEFINED_OBJECT;
+      return extensions.getOrDefault(name, Values.UNDEFINED_OBJECT);
     }
 
-    private Object tryRetrieveLocalVar(String name) {
-      if (locals == null || locals.isEmpty()) {
-        return Values.UNDEFINED_OBJECT;
+    private Object tryRetrieve(String name) {
+      Object result = null;
+      if (arguments != null && !arguments.isEmpty()) {
```
shatzi, line 683: do we cap arguments/local/field length when creating a CaptureContext?

dd-java-agent/agent-debugger/debugger-el/src/main/java/com/datadog/debugger/el/ValueScript.java
```diff
@@ -38,53 +44,62 @@ public boolean equals(Object o) {
...
-          jsonReader.endObject();
-          return new ValueScript(obj);
-        } else {
-          throw new IOException("Invalid field: " + fieldName);
+        ValueExpression<?> valueExpression = null;
+        String dsl = null;
+        while (jsonReader.hasNext()) {
+          String fieldName = jsonReader.nextName();
+          switch (fieldName) {
+            case "parsedExpr":
```
shatzi, line 87: what did we decided to use at the end? "expr" + "ParsedExpr" or we are changing everything to "dsl" and "json"?

dd-java-agent/agent-debugger/debugger-el/src/test/java/com/datadog/debugger/el/RefResolverHelper.java
```diff
@@ -0,0 +1,54 @@
...
+import datadog.trace.bootstrap.debugger.el.ValueReferenceResolver;
+import java.lang.reflect.Field;
+import java.util.*;
+
+public class RefResolverHelper {
+
+  public static ValueReferenceResolver createResolver(Object instance) {
+    List<Field> fields = new ArrayList<>();
+    Class<?> clazz = instance.getClass();
+    while (clazz != null) {
```
shatzi, line 13: are we using this trick for CaptureContent.addFields()? 
ans also for ReflectiveFieldResolver?

dd-java-agent/agent-debugger/debugger-el/src/test/java/com/datadog/debugger/el/expressions/Tester.java
```diff
@@ -0,0 +1,25 @@
+package com.datadog.debugger.el.expressions;
+
+class Tester {
```
shatzi, line 3: I find the name confusing.
When I read it the first time I thought it was some kind of a Test, and ExTester was even more confusing.

Suggest HelperObject, or ObjectWithRefAndValue

dd-java-agent/agent-debugger/src/main/java/com/datadog/debugger/instrumentation/MetricInstrumentor.java
shatzi: We might need to rethink the compile logic here.
The UI is adding the ability to create metric from any node in the captured tree.
That mean, there is an expection that we resolve interface/baseClass to the acutal 
object class.

Should we have two different logic pathes:
First, find the field from compiled type. if fails...
Second, use ReflectiveFieldResolver at run-time and report error if failed

Maybe a different PR?

dd-java-agent/agent-debugger/src/main/java/com/datadog/debugger/instrumentation/MetricInstrumentor.java
```diff
@@ -324,87 +355,63 @@ private Type tryRetrieveArgument(String head, InsnList insnList, boolean isLast)
...
-    }
-
     private void convertIfRequired(Type currentType, Type expectedType, InsnList insnList) {
       if (expectedType == Type.LONG_TYPE && currentType == Type.INT_TYPE) {
         insnList.add(new InsnNode(Opcodes.I2L));
       }
     }
   }
+
+  private static class ResolverResult {
```
shatzi, line 408: I like the idea of creating a new insnList for every Ref and append it to with getMember.

Two notes:
1. When it come to do getIndex we would need to merge two lists togther.
2. Each GetMemebr add NULL check and jump. it would be nice if the ResolverResult would have "NullJump" label that all getMembers can jump to and skip the enitre chain.

shatzi (approved):
Great progress. LGTM. There are few points we should address in following PRs. mostly the gap in how we handle interface and base classes in both Script Eval and Compiled expression.

dd-java-agent/agent-debugger/debugger-bootstrap/src/main/java/datadog/trace/bootstrap/debugger/Snapshot.java
```diff
@@ -613,108 +611,93 @@ private CapturedContext(CapturedContext other, Map<String, Object> extensions) {
...
       if (target == Values.UNDEFINED_OBJECT) {
-        addEvalError(path, msg + name);
+        addEvalError(expr, msg + name);
       }
     }
 
+    @Override
+    public Object getMember(Object target, String memberName) {
+      if (target == Values.UNDEFINED_OBJECT) {
+        return target;
```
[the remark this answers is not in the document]
jpbempel: it's a safety net, but the error was already took into account in `lookup` method

dd-java-agent/agent-debugger/debugger-bootstrap/src/main/java/datadog/trace/bootstrap/debugger/Snapshot.java
```diff
@@ -613,108 +611,93 @@ private CapturedContext(CapturedContext other, Map<String, Object> extensions) {
...
+      checkUndefined(memberName, target, memberName, "Cannot dereference to field: ");
+      return target;
+    }
+
     private Object tryRetrieveField(String name) {
       if (fields == null) {
         return Values.UNDEFINED_OBJECT;
       }
-      return fields.containsKey(name) ? fields.get(name) : Values.UNDEFINED_OBJECT;
+      Object field = fields.get(name);
```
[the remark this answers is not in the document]
jpbempel: not here, at serialization time

dd-java-agent/agent-debugger/debugger-bootstrap/src/main/java/datadog/trace/bootstrap/debugger/Snapshot.java
jpbempel: no

dd-java-agent/agent-debugger/debugger-el/src/main/java/com/datadog/debugger/el/ValueScript.java
jpbempel: not decided. and seems the vote are split half-half :(

dd-java-agent/agent-debugger/debugger-el/src/test/java/com/datadog/debugger/el/RefResolverHelper.java
jpbempel: we are not capturing parent fields of class of instance method.

dd-java-agent/agent-debugger/debugger-el/src/test/java/com/datadog/debugger/el/expressions/Tester.java
jpbempel: done

dd-java-agent/agent-debugger/src/main/java/com/datadog/debugger/instrumentation/MetricInstrumentor.java
jpbempel: clearly a different PR, we had the problem before changing the syntax.

jpbempel: Obviously there are some known limitation with the compile approach.
But usually it involves invoking methods, with the all the issues you are pulling with that approach...

dd-java-agent/agent-debugger/debugger-el/src/main/java/com/datadog/debugger/el/ValueScript.java
shatzi: I think you can then be the vote-breaker - and make the decision. it is a good timing as Cimi is update the UI parser.

jpbempel: Already did, we keep dsl/json, and updated the PR accordingly

PR #4388, opened 2022-12-06
Split up B3 propagator into multi and single header

bantonsson:
# What Does This Do

Splits up the B3 propagator into multi header and single header versions

# Motivation

They are separate

# Additional Notes

This is in preparation for new configuration variables and propagation style names.


PerfectSlayer (approved):
LGTM

PR #4390, opened 2022-12-06
fix spring webflux route extraction + improve webflux tests

amarziali:
# What Does This Do

A couple of things:
* Extract http route tag from Handlers (previously it was failing to extract the info for routerFunctions (was only working for method annotated with `@RequestMapping`))
* Simplified testing with spring boot by reusing the source folder
* Add latest 2.x tests

# Motivation

# Additional Notes


dd-java-agent/instrumentation/spring-webflux-5/src/main/java/datadog/trace/instrumentation/springwebflux/server/RouteOnSuccessOrError.java
```diff
@@ -55,10 +60,14 @@ private String parsePredicateString() {
...
-        .trim();
+  @Nonnull
+  private String parseRoute(@Nonnull String routerString) {
+    return METHOD_REGEX
+        .matcher(
+            SPACES_REGEX
+                .matcher(SPECIAL_CHARACTERS_REGEX.matcher(routerString).replaceAll(""))
+                .replaceAll(" ")
+                .trim())
+        .replaceAll("");
```
bantonsson, line 71: This seems like it creates a number of extra objects. Would it make sense to have a fixed size cache here?

bantonsson (approved):

PR #4392, opened 2022-12-07
remove scope events

richardstartin:
# What Does This Do

Entirely removes scope events, which are replaced by async-profiler's contextual events

# Motivation

# Additional Notes


jbachorik (approved):
Bye-bye scope events

PR #4396, opened 2022-12-07
Add correlation id injection to dd-trace-ot

### LINKED ISSUE #2066: Automatically Inject Trace and Span IDs into MDC by dd-trace-ot
Currently, only `dd-java-agent` integration automatically injects Trace/Span IDs into MDC context. When using `dd-trace-ot`, there is no injection support. Without this, integration with Open Tracing significantly looses in value, as logs cannot be correlated with spans and thus are not visible in Traces view in Datadog. As instrumentation for Open Tracing is typically done by the third-party libraries, it is not possible to use [manual injection](https://docs.datadoghq.com/tracing/connect_logs_and_traces/java/?tab=slf4jlogback#manually-inject-trace-and-span-ids) (third-parties will not couple to Datadog API). 

It seems to be feasible and straightforward to add automatic injection for  `dd-trace-ot` as well:
* whenever a new `io.opentracing.Scope` is activated, add Trace/Span ID to the MDC context;
* when `io.opentracing.Scope` is closed, remove the Trace/Span ID from MDC context.

PerfectSlayer:
# What Does This Do

This PR adds automatic injection of Trace / Span Ids to MDC (Mapped Diagnostic Context) to `dd-trace-ot`.

# Motivation

As `ScopeListener` was removed from public API from `1.0.0` version, there is no way for users to do such injection.
I add it as a feature of `dd-trace-ot` but it's implemented in a separated module that can be reused for a coming OTEL support.

# Additional Notes

Fix #2066

### DIFF
--- a/dd-trace-api/src/main/java/datadog/trace/api/internal/InternalTracer.java
+++ b/dd-trace-api/src/main/java/datadog/trace/api/internal/InternalTracer.java
@@ -9,7 +9,7 @@ public interface InternalTracer {
    * Attach callbacks to the global scope manager.
    *
    * @param afterScopeActivatedCallback Callback on scope activation.
-   * @param afterScopeClosedCallback    Callback on scope close.
+   * @param afterScopeClosedCallback Callback on scope close.
    */
   void addScopeListener(Runnable afterScopeActivatedCallback, Runnable afterScopeClosedCallback);
 
--- a/dd-trace-api/src/main/java/datadog/trace/api/internal/InternalTracer.java
+++ b/dd-trace-api/src/main/java/datadog/trace/api/internal/InternalTracer.java
@@ -5,6 +5,14 @@ package datadog.trace.api.internal;
  * at any time.
  */
 public interface InternalTracer {
+  /**
+   * Attach callbacks to the global scope manager.
+   *
+   * @param afterScopeActivatedCallback Callback on scope activation.
+   * @param afterScopeClosedCallback    Callback on scope close.
+   */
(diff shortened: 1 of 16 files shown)


dd-trace-ot/correlation-id-injection/src/main/java/datadog/trace/correlation/CorrectionIdInjectors.java
```diff
@@ -0,0 +1,66 @@
+package datadog.trace.correlation;
+
+import datadog.trace.api.internal.InternalTracer;
+import de.thetaphi.forbiddenapis.SuppressForbidden;
+import org.slf4j.Logger;
+import org.slf4j.LoggerFactory;
+
+/** Helper class to enable correlation identifier injection. */
+public final class CorrectionIdInjectors {
```
devinsba, line 9:
```suggestion
public final class CorrelationIdInjectors {
```

devinsba (commented):
Can you lay out how this is different from what the code looked like before 1.0.0?

PerfectSlayer:
> Can you lay out how this is different from what the code looked like before 1.0.0?

Sure. Previously, we offered the capability to add `ScopeListener` from the public tracer API.
Now `ScopeListener` is no more accessible, I restore a similar feature through the new `InternalTracer` API using default `Runnable` callbacks. (so we don't provide official support for such listener and the `ScopeListener` won't have to move to public API module).

Then, I added trace / span id injection for logging framework, based on those callbacks (as users should not use them themselves). Id injection was only available with automatic instrumentation but not when using `dd-trace-ot`.

Did I fully answer your question?

devinsba (approved):

PR #4397, opened 2022-12-07
Adds process_id to all root spans

zhamza99:
# What Does This Do

Utiltizes PidHelper changes to add process_id tag with the value of PID to all root spans.

# Motivation

Part of larger initiative to have similar tags and value across all tracers. Java tracer did not have process_id compared to other languages and thus has been added.


internal-api/src/main/java/datadog/trace/api/Config.java
```diff
@@ -2435,7 +2451,7 @@ private static String propertyNameToSystemPropertyName(final String setting) {
   }
 
   @Nonnull
-  private static Map<String, String> newHashMap(final int size) {
+  private static <VAL> Map<String, VAL> newHashMap(final int size) {
```
devinsba, line 2454: You don't need this change anymore `getLocalRootSpanTags()` creates the HashMap directly

internal-api/src/main/java/datadog/trace/api/Config.java
```diff
@@ -1257,6 +1259,14 @@ public String getRuntimeId() {
     return runtimeIdEnabled ? RuntimeIdHolder.runtimeId : "";
   }
 
+  public Long getProcessId() {
+    String PID = PidHelper.getPid();
+    if (PID == null || PID.equals("") || PID.trim().equals("")) {
+      return 0L;
+    }
+    return Long.parseLong(PID);
```
devinsba, line 1267: I just want to make sure that we want this to be a number instead of a string? All the other products seem to tag this as a string

devinsba (dismissed):
We should undo the changes to `newHashMap()`

internal-api/src/main/java/datadog/trace/api/Config.java
zhamza99: They used to, but the new expectation is PID should be a numeric for sure.

internal-api/src/main/java/datadog/trace/api/Config.java
zhamza99: Done

zhamza99:
All tests fixed and ready to be merged

internal-api/src/main/java/datadog/trace/api/Config.java
```diff
@@ -1257,6 +1259,14 @@ public String getRuntimeId() {
     return runtimeIdEnabled ? RuntimeIdHolder.runtimeId : "";
   }
 
+  public Long getProcessId() {
+    String PID = PidHelper.getPid();
+    if (PID == null || PID.equals("") || PID.trim().equals("")) {
+      return 0L;
+    }
```
bantonsson, line 1266: Since I assume that you would want to trim before the parse anyway, I would write this as:

```java
    pid = pid == null ? "" : pid.trim();
    if (pid.isEmpty()) {
      return 0L;
    }
```

zhamza99: Done

internal-api/src/main/java/datadog/trace/api/Config.java
```diff
@@ -1257,6 +1259,14 @@ public String getRuntimeId() {
     return runtimeIdEnabled ? RuntimeIdHolder.runtimeId : "";
   }
 
+  public Long getProcessId() {
+    String PID = PidHelper.getPid();
```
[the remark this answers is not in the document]
zhamza99: Done

internal-api/src/main/java/datadog/trace/api/Config.java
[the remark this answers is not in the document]
zhamza99: Done

bantonsson (approved):

PR #4400, opened 2022-12-08
refactor snapshots to report probe version

shatzi:
# What Does This Do

Snapshots now able to report the config-id and version of the probe used to generate the probe. 
This allow the Datadog app to show the right probe details for each snapshot

# Motivation

Support historical versions of probe details when inspecting a snapshot

# Additional Notes


jpbempel (approved):

ojung (commented):
I'm confused about all the span probe changes. Why are they part of this PR?

PR #4401, opened 2022-12-08
Fix rmic Gradle task

monosoul:
# What Does This Do
Makes `rmic` Gradle task use `rmic` binary provided with Java 8 toolchain. Also changes the task type to `Exec` and refactors the command a bit.

Also, this ( 7c875e842adf387d8b162e487ca2356a2d823790 ) makes the paths to resolve properly on Windows (or any other OS). Not sure if the rest of the project will work properly on Windows, but this part will work now. :slightly_smiling_face: 

# Motivation
Without this change the task will fail if `rmic` binary is not on the Path env var. 
This is bad for 2 reasons:

- [Contribution guide](https://github.com/DataDog/dd-trace-java/blob/master/CONTRIBUTING.md) doesn't mention anywhere it has to be on the path.
- Makes the build unpredictable. Since the rest of the modules are built with Java 8, it's better to explicitly specify the toolchain to use here.

# Additional Notes
Without this change I couldn't even import this project in IntelliJ.

### DIFF
--- a/dd-java-agent/instrumentation/rmi/build.gradle
+++ b/dd-java-agent/instrumentation/rmi/build.gradle
@@ -13,13 +13,23 @@ muzzle {
 
 apply from: "$rootDir/gradle/java.gradle"
 
-def rmic = tasks.register('rmic') {
+def rmic = tasks.register('rmic', Exec) {
   dependsOn testClasses
+
+  Provider<JavaLauncher> javaLauncher = getJavaLauncherFor(8)
+  def toolchainPath = javaLauncher.get().metadata.installationPath
+
   def clazz = 'rmi.app.ServerLegacy'
-  String command = """rmic -g -keep -classpath ${sourceSets.test.output.classesDirs.asPath} -d ${buildDir}/classes/java/test ${clazz}"""
-  command.execute().text
+  commandLine(
+    "${toolchainPath}/bin/rmic",
+    "-g",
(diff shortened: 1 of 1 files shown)


monosoul:
@bantonsson here's the new one

bantonsson:
That is just so weird. Why is CircleCI basing its decisions on your fork of the repo.

monosoul:
@bantonsson looks like now it works :grin: I disabled Circle CI for my fork

bantonsson:
@monosoul where did you do that? Was it something on the forked repo that was copied when you forked it?

bantonsson (approved):

monosoul:
@bantonsson did that in the Circle CI's UI. Here's what happened:

- Initially I wasn't authenticated with Circle CI, so the workflow was failing because of that.
- Then I signed in to Circle CI like you suggested AND also enabled Circle CI for my fork (I did that in Circle CI's UI, that was a mistake apparently).
- Then I reopened the PR, which didn't help.
- And now I disabled Circle CI for the fork in Circle CI's UI.

So, the only thing I should've done to make it work with the first PR - is sign in to Circle CI, nothing else. I've never used Circle CI before, so never had an OAuth token for GitHub there.

bantonsson:
Thanks @monosoul we'll need to update the contribution guidelines/pr-template.

Issue #1123, opened 2019-12-02
Kotlin awaiting not traced correctly.


ikovalyov:
If you await for the deferred returned by coroutine while being in another coroutine your span information is lost and whole span is being recreated as if it was parent span. See example project:

https://github.com/ikovalyov/dd-tracing-coroutines-issue-v2

Application logs:
```
> Task :run
[dd.trace 2019-12-02 09:23:39:910 +0100] [main] INFO datadog.trace.agent.jmxfetch.JMXFetch - JMXFetch config: null [] [] [] null null {runtime-id=fedd6bdd-2499-451c-bb36-0f662f3e8648, service=unnamed-java-app} statsd:localhost:8125 System.err INFO
[dd.trace 2019-12-02 09:23:40:502 +0100] [main] INFO datadog.trace.agent.ot.DDTraceOTInfo - dd-trace - version: 0.38.0~1cb704e7
[dd.trace 2019-12-02 09:23:40:571 +0100] [main] INFO datadog.trace.agent.ot.DDTracer - New instance: DDTracer-77bb48d5{ serviceName=unnamed-java-app, writer=DDAgentWriter { api=DDApi { tracesUrl=http://localhost:8126/v0.3/traces } }, sampler=datadog.trace.agent.common.sampling.RateByServiceSampler@181d8899, defaultSpanTags={}}
[dd.trace 2019-12-02 09:23:40:600 +0100] [main] INFO datadog.trace.agent.tooling.VersionLogger - dd-trace-ot - version: 0.38.0~1cb704e7
[dd.trace 2019-12-02 09:23:40:600 +0100] [main] INFO datadog.trace.agent.tooling.VersionLogger - dd-trace-api - version: 0.38.0~1cb704e7
[dd.trace 2019-12-02 09:23:40:601 +0100] [main] INFO datadog.trace.agent.tooling.VersionLogger - dd-java-agent - version: 0.38.0~1cb704e7
2019-12-02 09:23:41.840  INFO 3998 [           main] Main$main$1                              : main DDSpan [ t_id=2654422379082193946, s_id=5753717729389507729, p_id=0] trace=unnamed-java-app/trace.annotation/Main.main metrics={} tags={component=trace, language=jvm, runtime-id=fedd6bdd-2499-451c-bb36-0f662f3e8648, thread.id=1, thread.name=main}, duration_ns=0
[...]

PR #4405, opened 2022-12-08
Kotlin coroutines instrumentation

### LINKED ISSUE #931: Kotlin coroutines support
Hi.
I can see there is closed issue already: https://github.com/DataDog/dd-trace-java/issues/652
But it looks like agent is still having issues with coroutines support. See code/logs below:

```
import datadog.opentracing.DDSpan
import io.opentracing.util.GlobalTracer
import kotlinx.coroutines.*
import org.slf4j.LoggerFactory

object Main {
    val logger = LoggerFactory.getLogger(Main.javaClass)
    @JvmStatic
    fun main(args: Array<String>) = runBlocking {
        val tracer = GlobalTracer.get()
        tracer.buildSpan("main").startActive(true)
        val span = tracer.activeSpan()
        logger.info("${Thread.currentThread().name} $span")
        coroutineScope {
            repeat(10) {
                launch(Dispatchers.IO) {
                    waitFor(1000)
                }
            }
        }
    }

    suspend fun waitFor(millis: Long) {
        val tracer = GlobalTracer.get()
        val span = tracer.activeSpan()
        logger.info("${Thread.currentThread().name} $span")
        delay(millis)
    }
}
```

```
> Task :run
[dd.tracing.agent - 2019-07-26 12:06:40:165 +0200] [main] INFO datadog.trace.agent.ot.DDTraceOTInfo - dd-trace - version: 0.31.0~e9872762
[dd.tracing.agent - 2019-07-26 12:06:40:320 +0200] [main] INFO datadog.trace.agent.ot.DDTracer - New instance: DDTracer-64a9d48c{ serviceName=unnamed-java-app, writer=DDAgentWriter { api=DDApi { tracesUrl=http://localhost:8126/v0.3/traces } }, sampler=datadog.trace.agent.common.sampling.RateByServiceSampler@365a6a43, defaultSpanTags={}}
[dd.tracing.agent - 2019-07-26 12:06:40:333 +0200] [main] INFO datadog.trace.agent.tooling.VersionLogger - dd-trace-ot - version: 0.31.0~e9872762
[dd.tracing.agent - 2019-07-26 12:06:40:333 +0200] [main] INFO datadog.trace.agent.tooling.VersionLogger - dd-trace-api - version: 0.31.0~e9872762
[dd.tracing.agent - 2019-07-26 12:06:40:333 +0200] [main] INFO datadog.trace.agent.tooling.VersionLogger - dd-java-agent - version: 0.31.0~e9872762
[dd.tracing.agent - 2019-07-26 12:06:40:348 +0200] [main] INFO datadog.trace.agent.jmxfetch.JMXFetch - JMXFetch config: null [] [] [] null null {runtime-id=f6a7f3e0-beef-42bb-ae40-901125a3fe94, language=jvm, service=unnamed-java-app} statsd:localhost:8125 System.err INFO
2019-07-26 12:06:41.718  INFO 4149 [           main] Main$main$1                              : main DDSpan [ t_id=4386340172421044579, s_id=2509487107151721448, p_id=0] trace=unnamed-java-app/main/main metrics={_sampling_priority_v1=1} tags={language=jvm, runtime-id=f6a7f3e0-beef-42bb-ae40-901125a3fe94, thread.id=1, thread.name=main}, duration_ns=0
2019-07-26 12:06:41.752  INFO 4149 [atcher-worker-1] Main                                     : DefaultDispatcher-worker-1 null
2019-07-26 12:06:41.763  INFO 4149 [atcher-worker-3] Main                                     : DefaultDispatcher-worker-3 null
2019-07-26 12:06:41.763  INFO 4149 [atcher-worker-2] Main                                     : DefaultDispatcher-worker-2 null
2019-07-26 12:06:41.765  INFO 4149 [atcher-worker-4] Main                                     : DefaultDispatcher-worker-4 null
2019-07-26 12:06:41.768  INFO 4149 [atcher-worker-8] Main                                     : DefaultDispatcher-worker-8 null
2019-07-26 12:06:41.767  INFO 4149 [atcher-worker-6] Main                                     : DefaultDispatcher-worker-6 null
2019-07-26 12:06:41.771  INFO 4149 [atcher-worker-7] Main                                     : DefaultDispatcher-worker-7 null
2019-07-26 12:06:41.774  INFO 4149 [atcher-worker-5] Main                                     : DefaultDispatcher-worker-5 null
2019-07-26 12:06:41.775  INFO 4149 [tcher-worker-10] Main                                     : DefaultDispatcher-worker-10 null
2019-07-26 12:06:41.776  INFO 4149 [tcher-worker-12] Main                                     : DefaultDispatcher-worker-12 null
[dd.tracing.agent - 2019-07-26 12:06:42:815 +0200] [Thread-1] INFO datadog.trace.agent.common.writer.DDAgentWriter - Flushing any remaining traces.
```

As you can see, parent span is lost in the coroutine. Same issue happens with automated instrumentario (which is discovered on our datadog APM dashboard). Right now only spans created within the same thread are being displaying.

Also if we create spans in the coroutine we suggest it to be a child of the parent span. But span is not set yet (null in the log). So I assume they are created as parent spans and either wrongly reported as parent ones or ignored as expired ones.

Do you have a solution for it?

Thanks.

monosoul:
# What Does This Do
Provides a proper instrumentation for Kotlin coroutines. Solves #931 , Solves #1123 .

Introduces:
- `ScopeState` (probably might have a better name) <-- an abstraction for **holding** and **managing** current scope state.
- `ContinuableManagedScope` <-- implementation that holds a `ScopeStack`. On `activate` it sets the scope stack it holds into the thread local scope stack of `ContinuableScopeManager`. On `fetchFromActive` it fetches the value of thread local scope stack into it's local variable.
- `ScopeStateAware` <-- provides a method to get a new scope state. Extended by `AgentScopeManager` and `TracerAPI`
- `ScopeStateCoroutineContext` <-- coroutine context element that allows to store/restore scope state on suspension/continuation.
- `AbstractCoroutineInstrumentation` <-- instrumentation for `AbstractCoroutine`, adds 3 advices:
   - `AbstractCoroutineConstructorAdvice` <-- advice around `AbstractCoroutine` constructor that creates a new instance of `ScopeStateCoroutineContext` and puts it into the `CoroutineContext` ensuring each coroutine has it's own instance of the context element and it is never inherited. If the coroutine is not lazily started, then also captures the current active scope.
   - `AbstractCoroutineOnStartAdvice` <-- advice around `AbstractCoroutine#onStart` to capture active scope on if/when a coroutine starts. Will do nothing for eagerly started coroutines (scope already got captured there on construction), only applicable to lazily started coroutines.
   - `JobSupportAfterCompletionInternalAdvice` <-- advice around `AbstractCoroutine#onCompletionInternal` to guarantee the scope captured by coroutine is closed.
- `CoroutineContextHelper` <-- a set of helper functions to access elements of `CoroutineContext`.

Basically, what it does - is makes each coroutine have it's own scope stack.

# Motivation

At the moment DataDog agent doesn't work properly with Kotlin coroutines, especially in the case when a coroutine gets suspended first and then gets continued. This is because `ContinuableScopeManager` stores the scope stack in a thread local variable.
Imagine this situation:

1. We have a coroutine dispatcher that is an executor pool of 2 threads.
2. We start a coroutine on the thread `#1` and then it gets suspended.
3. When it gets continued, the thread `#1` is busy with another coroutine, so our coroutine gets scheduled to the thread `#2`
4. Because `ContinuableScopeManager` uses [a thread local scope stack](https://github.com/DataDog/dd-trace-java/blob/master/dd-trace-core/src/main/java/datadog/trace/core/scopemanager/ContinuableScopeManager.java#L44), all the scopes that we have activated in our coroutine got left on the thread `#1`. Moreover, the thread `#2` might also have some leftovers in the scope stack from another coroutine, leading to missing spans and broken span hierarchy.

Having this instrumentation will make DD customers using Kotlin coroutines way happier with the product.

# Additional Notes
The test case to demonstrate the issue can be found in this commit: df6d7224ad00441c07aa5625a202e4255fd7e48d (for easier review).

You can find more information about coroutines in Kotlin here:

- [General info](https://kotlinlang.org/docs/coroutines-guide.html#table-of-contents)
- [Context and dispatchers](https://kotlinlang.org/docs/coroutine-context-and-dispatchers.html)

# Feature flags
The feature is disabled by default.

- System property `dd.integration.kotlin_coroutine.experimental.enabled`
- Environment variable `DD_INTEGRATION_KOTLIN_COROUTINE_EXPERIMENTAL_ENABLED`

Either of the 2 should be set to `true` to enable the feature.


vsingal-p:
Thanks for raising this 💯

monosoul:
Hey @mcculls @bantonsson , sorry for pinging, but would you mind having a look here? I believe having a proper instrumentation for coroutines will make lots of DD customers (including myself) happy. In case if you think the changes here are a bit too intrusive, there's also an option to do it like this: c7b4468a1aabc2af394c6a5393d1046511aa15d0 , with no changes to the API.
Thanks!

bantonsson:
Thanks @monosoul. I'll try to review this tomorrow.

bantonsson:
@monosoul I need to dig a bit deeper into this. The `Scope` instances are not really intended to be shared across threads and I want to make sure that I understand the underlying mechanisms in Kotlin.

monosoul:
@bantonsson yeah, the change here basically makes sure that each coroutine has it's own scope that doesn't interfere with the thread's scope and other coroutines. I.e. while the thread runs the coroutine - the scope manager will use the scope stored in the `ScopeStateCoroutineContext` instance unique for each coroutine. Once the coroutine is suspended or finished - the original thread's scope will be restored. Without this change multiple coroutines might share the same scope which leads to lost scopes and scope inheritance issues. You can think of `ThreadContextElement`'s as of `ThreadLocal`'s of coroutine world. Not sure if that explanation makes it any more clear :sweat_smile:

bantonsson:
Hey @monosoul. I've been reading up on this a bit. This whole solution would need to use continuations and activate/deactivate scopes in a different way to handle separating the creation of coroutines and the running of them.

I have a changed test that splits the reported spans into two traces which is not correct and I'll open a PR against your branch with it.

Would love to have a bigger discussion about how one would expect the UI to show these coroutines, since I think there is where we need to start to get this right.

vsingal-p:
@bantonsson Why would we need a change in the UI? The main problem is that traces getting split with new trace_ids because scope/span are not propagated correctly with coroutines. If we do, the flow would look exactly same as it does now (in a normal/other app).

monosoul:
Hey @bantonsson , thanks for looking into it!

> I have a changed test that splits the reported spans into two traces which is not correct and I'll open a PR against your branch with it.

I'm not sure if I got that part, but I guess it will be easier to discuss when you open the PR. 

Here are my expectations from coroutines instrumentation:
- All calls of instrumented methods from within coroutines should work the same way as they work outside of coroutines. I.e. if I start a span (let's call it span A) before calling a coroutine and then inside the coroutine I call a method that was instrumented automatically, for example fetch messages with Kafka client, then each span created with Kafka client call should have span A as it's parent regardless of the thread the coroutine got scheduled to. If there are multiple parallel calls, they all should have span A as it's parent.
- The instrumentation should just work without any changes to coroutines invocations.
- And of course running coroutines should not mess up other spans.

As for how it looks in the UI - I'd also expect the span hierarchy to be respected there. If I got it right - you want to use continuations so that the spans will have breaks in them when coroutine gets suspended, right?

monosoul:
> Would it be the span around the creation of the coroutine, or the span around the starting of the coroutine that the coroutine inherits?

Ah, that's a good question! Span around the creation of the coroutine makes more sense to me, but there might be other opinions :slightly_smiling_face:

bantonsson:
@monosoul The continuations are not about introducing breaks, when suspending, but rather ensuring that spans propagate _properly_ between the continuation contexts. They are use in the Datadog tracer to hold off sending spans until all spans related to a trace are finished.

vsingal-p:
> Span around the creation of the coroutine makes more sense to me, but there might be other opinions 🙂

+1

bantonsson:
Opened the PR with the changed test here https://github.com/monosoul/dd-trace-java/pull/1

monosoul:
@bantonsson ah, got it, thanks. If I got it right, I think to use continuations we'll need something more than `ThreadContextElement`, as we'll need a way to "subscribe" to some sort of coroutine finish event and I'm not sure if there's a way to do that, but I might be wrong.

monosoul:
Hey @bantonsson , thanks for the PR, it's a very good catch. I think the approach I took here wouldn't let me make lazy started coroutines work easily, but I have a few ideas how to fix it while also using continuations. I'll change this PR back to draft for now and ping you again once it's ready.

bantonsson:
Thanks for working on this @monosoul. I'll be happy to discuss any ideas you have for making this work.

bantonsson:
Hey @monosoul, there seems to be some issue with task dependencies in the build. A number of the runs failed on this:

```
FAILURE: Build failed with an exception.

* What went wrong:
A problem was found with the configuration of task ':dd-java-agent:instrumentation:kotlin-coroutines:instrumentKotlin' (type 'ByteBuddyTask').
  - Type 'net.bytebuddy.build.gradle.ByteBuddyTask' property '$2' specifies directory '/home/circleci/dd-trace-java/workspace/dd-java-agent/instrumentation/kotlin-coroutines/build/classes/kotlin/raw' which doesn't exist.
    
    Reason: An input file was expected to be present but it doesn't exist.
    
    Possible solutions:
      1. Make sure the directory exists before the task is called.
      2. Make sure that the task which produces the directory is declared as an input.
```

I can reproduce this by changing anything in `dd-trace-core` so it needs to be recompiled, and then running `./gradlew :dd-java-agent:instrumentation:kotlin-coroutines:test`. I see that the `build.gradle` file try to create these. Do you have any idea why it's not working?

monosoul:
> I can reproduce this by changing anything in dd-trace-core so it needs to be recompiled, and then running ./gradlew :dd-java-agent:instrumentation:kotlin-coroutines:test. I see that the build.gradle file try to create these. Do you have any idea why it's not working?

Hey @bantonsson , yeah, that's a weird issue, ideally the tasks I added should always create the dirs. The root cause is that this module now has 3 different languages to compile from, so there are 3 different main source sets. Probably a better way to solve this issue would be either:

- to explicitly remove `main/kotlin` dir from main source sets
- just rewrite all the classes in the main source set in Kotlin instead of Java (so that there's no Java classes to compile) :sweat_smile: 

I can reproduce this issue locally as well, just didn't pay too much attention to it yet. I'll try to come up with a better way to make it work :+1:

monosoul:
@bantonsson this should fix the task failures: b21c7c4c655293bcbbcb382eaee8d4491b00b706 . At least I can't reproduce it locally anymore after a few tries.

bantonsson:
@monosoul thanks again. I'll try to give this a proper review today.

dd-java-agent/instrumentation/kotlin-coroutines/src/test/groovy/KotlinCoroutineInstrumentationTest.groovy
```diff
@@ -13,6 +13,11 @@ class KotlinCoroutineInstrumentationTest extends AgentTestRunner {
     ThreadPoolDispatcherKt.newSingleThreadContext("Single-Thread"),
   ]
 
+  @Override
+  boolean useStrictTraceWrites() {
+    return false
+  }
```
bantonsson, line 19: Having to disable this means that there is something _leaking_ scopes or continuations.

bantonsson (commented):
I've been looking over this and there is something _leaking_ scopes or continuations right now. I'll try to dig deeper into this next week, and maybe come up with some fixes.

dd-java-agent/instrumentation/kotlin-coroutines/src/test/groovy/KotlinCoroutineInstrumentationTest.groovy
bantonsson: Also, the current way that the scope stacks are duplicated does not work well together with profiling. I have some ideas here that I need to try out.

monosoul: But aren't strict writes disabled by default when instantiating `CoreTracer`?

https://github.com/DataDog/dd-trace-java/blob/0dbfd3ba5d2011474ec87d8add1a9a9e2f607e0c/internal-api/src/main/java/datadog/trace/api/Config.java#L858

I didn't dig to deep into the code, but my assumption was that the traces that should be otherwise reported, but are locked by a continuation get buffered into `PendingTraceBuffer` until continuation is cancelled. And that makes totals sense in our case, because the root span supposed to be reported before the child spans are actually completed. Once the child spans complete and cancel continuations everything gets reported. Or am I wrong?

monosoul: Although, looking now at other usages of `useStrictTraceWrites` I guess that's wrong :sweat_smile:

monosoul: Btw, I only needed this workaround because of 3 tests: `kotlin traced across channels`, `kotlin cancellation prevents trace` and `kotlin either deferred completion`, so maybe I should just have a deeper look into what happens there.

monosoul: I actually had to use `useStrictTraceWrites` only because of a few old test cases in `KotlinCoroutineInstrumentationTest` and I think the issue there might be the test cases themselves.For example `kotlin propagates across nested jobs` doesn't wait for the jobs it launches to finish, so they can finish AFTER the test method completes leading to pending traces. It means continuations will be released by coroutines only AFTER the test method completes.

monosoul: @bantonsson okay, you were right, there were leaking continuations, I have changed the way continuations closed. Now the tests pass without `useStrictTraceWrites`: 7073b6e75ecd823d9149afcac22ce7410d74e572

bantonsson:
Happy Holidays @monosoul. I just wanted to say that I've been looking over this code and working on some more tests as well as digging into the Kotlin Coroutine machinery. I'll get back with an update next week.

monosoul:
@bantonsson thanks! I'll check it tomorrow

monosoul:
@bantonsson I've left a few comments there. Tbh, now I think it's probably impossible/hard to achieve the behavior for lazily started coroutines I mentioned [here](https://github.com/DataDog/dd-trace-java/pull/4405#issuecomment-1355022421). First of all because there are no guarantees a lazily started coroutine will ever be started or cancelled after it was instantiated.

vsingal-p:
Hello folks. Is there any update on this PR?
I have gone through the comments and looks like lazy coroutines are somewhat tricky to handle

Can we keep them out of scope and get the base fix merged? We can mark lazy coroutines as a known bug, but atleast we would be able to solve a lot of other ones (everything basically)

bantonsson:
@vsingal-p I've reviewed the fixes that @monosoul have done in his repo, and I think that with those fixes done, we could move this PR forward if we have the instrumentation as an opt-in until it's been tested and we know what the performance implications are.

vsingal-p:
> @vsingal-p I've reviewed the fixes that @monosoul have done in his repo, and I think that with those fixes done, we could move this PR forward if we have the instrumentation as an opt-in until it's been tested and we know what the performance implications are.

Sounds like a plan. @monosoul Let's get all the changes in and get this fix launched 🤞

monosoul:
Sorry for the delay on my end this time, been a bit busy lately. I'll try to implement the changes suggested by @bantonsson by the end of this week.

monosoul:
@bantonsson I've updated the PR

bantonsson (approved):
Thanks @monosoul for a great contribution.

bantonsson:
@monosoul Would you mind rewording your PR description to what the state of the instrumentation is now, and mention the flags for turning on the instrumentation? Also, could you switch out the use of `fixes #...` to `solves #...` since github auto-close will close the issues on merge and not on release.

monosoul:
@bantonsson done!

kwazii1231:
is it working correctly?
i mean, i'm hard to find out what i'm missing 
i use RUN curl -Lso dd-java-agent.jar https://dtdg.co/latest-java-tracer
to use latest dd-trace-java for my application as datadog agent
and i use implementation("com.datadoghq:dd-trace-api:1.22.0") as gradle dependency.
i don't know what is the root cause that is reason my application can not correctly trace code that i use async of coroutine
anything other is same as what u describe.
if you have any clue to solve my situation.

kwazii1231:
> @kwazii1231 you need to also enable the feature, since it's behind a flag atm

really thanks to reply very quickly
i set up DD_INTEGRATION_KOTLIN_COROUTINE_EXPERIMENTAL_ENABLED=true 
to my container env_variable 
and my docker file is seems like belows

FROM adoptopenjdk:11.0.11_9-jdk-hotspot as app

ARG APP_PROFILE=local
ENV APP_PROFILE=${APP_PROFILE} \
    TZ=Asia/Seoul

WORKDIR /app
RUN curl -Lso dd-java-agent.jar https://dtdg.co/latest-java-tracer

# copy library
WORKDIR /app
COPY build/libs/* /app/

# entry point
ENTRYPOINT java \
  -XX:+UseContainerSupport \
  -XX:InitialRAMPercentage=75 \
  -XX:MaxRAMPercentage=75 \
  -XshowSettings:vm \
  -Dfile.encoding=UTF-8 \
  -Djava.net.preferIPv4Stack=true \
  -jar /app/api.jar \
  --spring.profiles.active=${APP_PROFILE}

monosoul:
@kwazii1231 you need to add the agent to your jvm cmd line: `-javaagent:/path/to/dd-java-agent.jar`

See this documentation: https://docs.datadoghq.com/tracing/trace_collection/dd_libraries/java/?tab=springboot

kwazii1231:
> @kwazii1231 you need to also enable the feature, since it's behind a flag atm

also did at common template of my helm chart like this

- name: JAVA_TOOL_OPTIONS
   value: -javaagent:/app/dd-java-agent.jar

kwazii1231:
> @kwazii1231 you need to add the agent to your jvm cmd line: `-javaagent:/path/to/dd-java-agent.jar`
> 
> See this documentation: https://docs.datadoghq.com/tracing/trace_collection/dd_libraries/java/?tab=springboot

Hi @monosoul
finally i found the reason, and root cause is not from your implementation
but from our custom span append code that was written when this agent didn't support kotlin coroutine tracing.
so i used activeSpan from globalTracer and try to append custom client span.
finally it works!
really thansk to reply very quickly that make me try to find out reason from our code.
have a nice day!

PR #4409, opened 2022-12-09
change metrics queue implementation

richardstartin:
# What Does This Do
Changes the queue implementation and draining strategy in the metrics reporter
# Motivation
Spotchecking the wallclock profiler, I noticed a good amount of time waking the aggregator thread
Screenshot 2022-12-09 at 12 04 34

# Additional Notes


PerfectSlayer (approved):
LGTM 👍

dd-trace-core/src/main/java/datadog/trace/common/metrics/Aggregator.java
```diff
@@ -91,14 +103,32 @@ public void run() {
...
+        // important that it is still *this* batch pending, must not remove otherwise
+        pending.remove(key, batch);
+        AggregateMetric aggregate = aggregates.computeIfAbsent(key, k -> new AggregateMetric());
+        batch.contributeTo(aggregate);
+        dirty = true;
+        // return the batch for reuse
+        batchPool.offer(batch);
       }
     }
-    stopSignal.complete();
```
bantonsson, line 101: I'm pretty sure that there is still code somewhere that is waiting for this to be completed. Would it wait for a timeout now?

richardstartin: I think this was handled somewhere, let me check

richardstartin: It disappeared at some point, good catch

PR #4411, opened 2022-12-09
Improve resource naming for spring data repositories

amarziali:
# What Does This Do

Improve resource naming for spring data repository operations by using the real interface (and not the proxied one like CrudRepository or JpaRepository) for all crud methods.

This is activated by default but can still turned off by setting `spring-data.repository.enhance.resource-name` to false.

Also simplify testing by avoiding flakiness.

# Motivation

# Additional Notes


dd-trace-api/src/main/java/datadog/trace/api/config/TraceInstrumentationConfig.java
```diff
@@ -90,6 +90,9 @@ public final class TraceInstrumentationConfig {
   public static final String SERVLET_ROOT_CONTEXT_SERVICE_NAME =
       "trace.servlet.root-context.service.name";
 
+  public static final String SPRING_DATA_REPOSITORY_ENHANCE_RESOURCE_NAME =
+      "spring-data.repository.enhance.resource-name";
```
bantonsson, line 94: Would it make more sense to call this `spring-data.repository.interface.resource-name`?

amarziali: yes it does let me change it!

bantonsson (approved):

PR #4414, opened 2022-12-12
feat(dd-trace-ot): Add MDC injection config support

PerfectSlayer:
# What Does This Do

This PR makes the MDC injection from `dd-trace-ot` respects the `dd.logs.injection` config.

# Motivation

Requested by @bantonsson. It will match the instrumentation behavior.

# Additional Notes

Complement to #4396

### DIFF
--- a/dd-trace-ot/src/main/java/datadog/opentracing/DDTracer.java
+++ b/dd-trace-ot/src/main/java/datadog/opentracing/DDTracer.java
@@ -407,7 +407,10 @@ public class DDTracer implements Tracer, datadog.trace.api.Tracer, InternalTrace
       this.scopeManager = new OTScopeManager(tracer, converter);
     }
 
-    CorrelationIdInjectors.register(this);
+    if ((config != null && config.isLogsInjectionEnabled())
+        || (config == null && Config.get().isLogsInjectionEnabled())) {
+      CorrelationIdInjectors.register(this);
+    }
   }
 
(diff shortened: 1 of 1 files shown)


dd-trace-ot/src/main/java/datadog/opentracing/DDTracer.java
```diff
@@ -407,7 +407,9 @@ private DDTracer(
       this.scopeManager = new OTScopeManager(tracer, converter);
     }
 
-    CorrelationIdInjectors.register(this);
+    if (Config.get().isLogsInjectionEnabled()) {
```
bantonsson, line 410: If you have a `config` passed into this method, you should use that instead.

PerfectSlayer: There is a null check of the `config` value in this method, meaning it can be `null`.
Should be still try to use it? And disabling the feature if config is `null`?

bantonsson: I think you should use it if it's not `null`, otherwise fall back to the global `Config.get()` since that's what's happening under the hood anyway in `CoreTracer` (yeah, it's messy).

PerfectSlayer: The last amend should do the trick. (yes it is 😞 )

PR #4416, opened 2022-12-13
Add taint tracking support for URLDecoder

manuel-alvarez-alvarez:
# What Does This Do
Add propagation for `URLDecoder.decode` method calls

# Motivation

# Additional Notes


smola (approved):

internal-api/src/test/groovy/datadog/trace/api/iast/InstrumentationBridgeTest.groovy
```diff
@@ -17,10 +18,11 @@ class InstrumentationBridgeTest extends DDSpecification {
     WeakHashModule,
     WebModule,
     StringModule,
+    UrlModule,
     SqlInjectionModule,
     CommandInjectionModule,
     PathTraversalModule,
-    LdapInjectionModule
+    LdapInjectionModule,
```
jandro996, line 25: Why you add an extra **,** ? 😅

jandro996 (approved):

internal-api/src/test/groovy/datadog/trace/api/iast/InstrumentationBridgeTest.groovy
smola: @jandro996 Easier to add/remove elements from the list, since you don't have a special case for the last element.

PR #4417, opened 2022-12-13
fix(ci): Fix CI script for upgraded build image

PerfectSlayer:
# What Does This Do

This PR fixes the CI script to install missing python packages needed for the docker port forwarding, used by `test-containers`.

# Motivation

The new `dd-trace-java-docker-build` image (based on `Ubuntu 22.04` instead of `Ubuntu 20.04`) does not contains the required python package.

# Additional Notes

This PR is required for https://github.com/DataDog/dd-trace-java-docker-build/pull/47


devinsba (approved):

.circleci/config.yml
```diff
@@ -155,6 +155,7 @@ commands:
             chmod -v 0444 ca.pem server-cert.pem cert.pem
             
             cd ~/dd-trace-java
+            sudo pip3 install requests requests-unixsocket
```
bantonsson, line 158: Can’t this be in the CI image build?

PerfectSlayer: Yes it _can_. We already install some python packages from the build image but only the the scripts it bundles (`aws-cli`).

I fear that we loose track of which modules is required by which python scripts...
Personally, I would have created a `requirements.txt` file and added a virtual env before running dependency download and executing this script to ensure it is isolated from the rest of the system.

The questions are:
- Is that a rhetorical question telling me: "move install to `dd-trace-java-docker-build` Docker file?
- Should we move the `autoforward.py` script to `dd-trace-java-docker-build` repo?
- Should I create `venv` for this script or keep it simple?

bantonsson: It is not a rhetorical question. Just curios, since as I see it everything that we can have downloaded/installed on the build image is preferable. 

I agree that it gets messy to keep track of dependencies and risk polluting environments. As for virtual env and Python and Docker, I actually have no experience of that combo.

PerfectSlayer: [Python venv](https://docs.python.org/3/library/venv.html) should not mess with Docker. Even if it's called virtual, it is just way to tell Python to look for dependencies in a specific directory (like if you were running `mvn run` with `-Dmaven.repo.local` to set another location for your `.m2/repository` folder).

The main idea is if the `autoforward.py` script requires specific modules (especially with specific versions), it does not break the system Python packages and the system commands that requires then.

I think there is no good answer it if we choose to keep the python script in the `dd-trace-java` repository:
- I can add the modules to the docker files with a comment and comment why it is required and why the script should be upgraded according the current system package.
- Or, I can declare the requirements of the scripts, create a venv and install them at each run.

But as the script is part of the build tools and use CircleCI `setup_remote_docker` command, I would be in favor to move it to `dd-trace-java-docker-build` repo and copy/install it like we already to for `aws-cli` (no venv, keep it simple).

What do you think about it?

bantonsson: With regards to `venv` I was more thinking about if you can build that as part of a Docker image and _activate_ it when using the Docker image.

It is not completely clear cut since the `dd-trace-java-docker-build` image is used on GitLab as well, but I think you are right that it makes most sense to move the script to the build image.

PerfectSlayer: > With regards to venv I was more thinking about if you can build that as part of a Docker image and activate it when using the Docker image.

I think you would create the venv and grab the dependencies during the docker image build with
`python -m venv venv && pip install -r requirements.txt`.
Then at runtime, we would have to enable the venv first before running the python script: `source venv/bin/active.sh && python autoforward.py`. But for convenience, it should be easier to create a executable script `autoforward` under `/usr/bin/` with those two commands :)

I will move it as part the docker image and keep it simple then.

PR #4421, opened 2022-12-13
DSM - Pathway tags added to spans

kr-igor:
Data Streams Monitoring allows Datadog users to see bottlenecks in their pipelines. However, the troubleshooting process is not trivial. If you know that something is going on within the service, you need to go and search through service's traces / logs. In cases when a single service handles data from multiple upstream queues or sends messages to multiple destinations the search for the data you need is not straightforward.

Changes introduced in this PR link DSM pathways with spans by adding `pathway.hash` tag to the span at the moment DSM performs `checkpoint` operation. This will be used to distinguish traces belonging to different pathways, hence speed up the issue resolution. 

**Affected instrumentations**: gRPC, HTTP (all), Kafka, RabbitMq.
The majority of changed files are tests. The main logic change is in `CoreTracer.java`.


dd-java-agent/instrumentation/kafka-clients-0.11/src/test/groovy/KafkaClientTestBase.groovy
```diff
@@ -47,6 +48,15 @@ abstract class KafkaClientTestBase extends AgentTestRunner {
     injectSysConfig("dd.data.streams.enabled", "true")
   }
 
+  public static final LinkedHashMap<String, String> PRODUCER_PATHWAY_EDGE_TAGS
+
+  static {
+    PRODUCER_PATHWAY_EDGE_TAGS = new LinkedHashMap<>(2)
```
dougqh, line 54: Why is this a LinkedHashMap?  Why is initialized to the wrong size?

dd-trace-core/src/main/java/datadog/trace/core/datastreams/DefaultPathwayContext.java
```diff
@@ -346,6 +346,11 @@ public void addTag(String tag) {
     public long generateHash() {
       return FNV64Hash.generateHash(builder.toString(), FNV64Hash.Version.v1);
     }
+
+    @Override
+    public String toString() {
+      return builder.toString();
```
dougqh, line 352: It would be good to initialize StringBuilder to a more precise size if possible.

dd-java-agent/instrumentation/kafka-clients-0.11/src/test/groovy/KafkaClientTestBase.groovy
kr-igor: I added one more item to the list later and forgot to update the initialization :D Thanks for catching!

dd-trace-core/src/main/java/datadog/trace/core/datastreams/DefaultPathwayContext.java
kr-igor: I don't think we can guess a correct size for this StringBuilder, as it is used in different instrumentations with different sets of tags.

hokitam (approved):

dd-java-agent/testing/src/main/groovy/datadog/trace/agent/test/base/HttpServerTest.groovy
```diff
@@ -36,37 +28,12 @@ import java.util.function.BiFunction
...
-import static datadog.trace.agent.test.base.HttpServerTest.ServerEndpoint.PATH_PARAM
-import static datadog.trace.agent.test.base.HttpServerTest.ServerEndpoint.QUERY_ENCODED_BOTH
-import static datadog.trace.agent.test.base.HttpServerTest.ServerEndpoint.QUERY_ENCODED_QUERY
-import static datadog.trace.agent.test.base.HttpServerTest.ServerEndpoint.QUERY_PARAM
-import static datadog.trace.agent.test.base.HttpServerTest.ServerEndpoint.REDIRECT
-import static datadog.trace.agent.test.base.HttpServerTest.ServerEndpoint.SUCCESS
-import static datadog.trace.agent.test.base.HttpServerTest.ServerEndpoint.TIMEOUT
-import static datadog.trace.agent.test.base.HttpServerTest.ServerEndpoint.TIMEOUT_ERROR
-import static datadog.trace.agent.test.base.HttpServerTest.ServerEndpoint.UNKNOWN
+import static datadog.trace.agent.test.base.HttpServerTest.ServerEndpoint.*
```
ygree, line 31: Looks like you haven't set up your IDE to avoid this type of replacement. Please follow [the contributing guide](https://github.com/DataDog/dd-trace-java/blob/master/CONTRIBUTING.md#intellij-idea) to avoid inconsistent formatting.

kr-igor: Good catch! I forgot to configure my IDE for .groovy :)

amarziali (approved):
looks good from my standpoint

PR #4423, opened 2022-12-14
Use fixed version of Google Java Format

bantonsson:
# What Does This Do

Ensures that building with different JDK versions does not change formatting of code.

# Motivation

# Additional Notes


PerfectSlayer (approved):
Might be useful to add a comment that `1.7` denotes the Google Format version, not the Java one, that it's the last one to support Java 8.
(current version is `1.15` and `1.8` requires Java 11)

PR #4428, opened 2022-12-14
Avoid EnvAndService allocation on lookup triggered by setSamplingPriority call

ygree:
# What Does This Do

Avoids allocation in `RateByServiceTraceSampler.setSamplingPriority`:

Before:

```java
RateSampler sampler = rates.getSampler(new EnvAndService(env, serviceName));
```

After:

```java
RateSampler sampler = rates.getSampler(env, serviceName);
```

# Motivation

`RateByServiceTraceSampler.setSamplingPriority` is an allocation hot spot.

# Additional Notes

APMS-8440 Memory allocation issue

- [x] Measure EnvAndService allocation vs extra HashMap lookup
- [ ] Optimizing Sending Sampling Instructions to Tracer

### DIFF
--- a/dd-trace-core/src/main/java/datadog/trace/common/sampling/RateByServiceTraceSampler.java
+++ b/dd-trace-core/src/main/java/datadog/trace/common/sampling/RateByServiceTraceSampler.java
@@ -40,7 +40,7 @@ public class RateByServiceTraceSampler implements Sampler, PrioritySampler, Remo
     final String env = getSpanEnv(span);
 
     final RateSamplersByEnvAndService rates = serviceRates;
-    RateSampler sampler = rates.getSampler(new EnvAndService(env, serviceName));
+    RateSampler sampler = rates.getSampler(env, serviceName);
 
     if (sampler.sample(span)) {
       span.setSamplingPriority(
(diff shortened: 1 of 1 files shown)


ygree:
This is the result of micro benchmark comparison:

```
Benchmark                                                          Mode  Cnt          Score          Error  Units
NestedHashMapLookupVsFlatMapWithAllocation.flatMapWithAllocation  thrpt   25   60585159.001 ±   811716.778  ops/s
NestedHashMapLookupVsFlatMapWithAllocation.nestedMap              thrpt   25  430413068.637 ± 13044523.495  ops/s
```

```java
  @Threads(4)
  @Benchmark
  public String nestedMap() {
    return nestedMap.get("b").get("d");
  }

  @Threads(4)
  @Benchmark
  public String flatMapWithAllocation() {
    return flatMap.get(new AnObject("b", "d"));
  }
```

bantonsson (approved):
I think adding a JMH benchmark that actually calls `setSamplingPriority` would have been better.

ygree:
> I think adding a JMH benchmark that actually calls `setSamplingPriority` would have been better.

Could be so. It just requires more effort. My goal was to do a quick performance comparison of an extra lookup in an inner HashMap lookup vs an extra allocation of a wrapper object and one lookup.

ygree:
I've updated micro benchmarks results as I noticed that AObject in the test had non optimal equals and hashCode implementations. The results are closer now but avoiding an extra allocation still a winner.

richardstartin (approved):

PR #4432, opened 2022-12-15
fix(dd-java-agent): Exclude bogus elasticsearch release

PerfectSlayer:
# What Does This Do

Exclude bogus `org.elasticsearch:elasticsearch` release from muzzle checks

# Motivation

The CI is failing.

# Additional Notes

I reported the issue on their GH repo: https://github.com/elastic/elasticsearch/issues/92383

### DIFF
--- a/dd-java-agent/instrumentation/elasticsearch/transport-5.3/build.gradle
+++ b/dd-java-agent/instrumentation/elasticsearch/transport-5.3/build.gradle
@@ -5,14 +5,14 @@ muzzle {
     module = "transport"
     versions = "[5.3.0,6.0.0)"
     assertInverse = true
-    skipVersions = ["7.11.0"]
+    skipVersions = ["7.11.0", "7.17.8"]
   }
   pass {
     group = "org.elasticsearch"
     module = "elasticsearch"
     versions = "[5.3.0,6.0.0)"
     assertInverse = true
-    skipVersions = ["7.11.0"]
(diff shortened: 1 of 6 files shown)


dd-java-agent/instrumentation/elasticsearch/transport-5.3/build.gradle
```diff
@@ -5,14 +5,14 @@ muzzle {
     module = "transport"
     versions = "[5.3.0,6.0.0)"
     assertInverse = true
-    skipVersions = ["7.11.0"]
+    skipVersions = ["7.11.0", "7.17.8"]
```
PerfectSlayer, line 8: I wonder why `7.11.0` version was excluded first as it does not seem in `[5.3.0,6.0.0)` range?

bantonsson: It's because of the `assertInverse = true` which means that `muzzle` will ensure that the things outside of the `[5.3.0,6.0.0)` range does not match.

bantonsson (approved):

dd-java-agent/instrumentation/elasticsearch/transport-5.3/build.gradle
PerfectSlayer: Thanks for the clarification 👍

PR #4433, opened 2022-12-15
Minor refactoring about trace metrics

PerfectSlayer:
# What Does This Do

Introduce a NOOP implementation of `SingleSpanSampler` to simplify null checks from `TraceProcessingWorker`.

# Motivation

# Additional Notes

### DIFF
--- a/dd-trace-core/src/main/java/datadog/trace/common/writer/RemoteWriter.java
+++ b/dd-trace-core/src/main/java/datadog/trace/common/writer/RemoteWriter.java
@@ -11,7 +11,7 @@ import org.slf4j.LoggerFactory;
 
 /**
  * This writer buffers traces and sends them to the provided DDApi instance. Buffering is done with
- * a distruptor to limit blocking the application threads. Internally, the trace is serialized and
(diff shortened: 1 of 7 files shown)


dd-trace-core/src/main/java/datadog/trace/common/writer/RemoteWriter.java
```diff
@@ -96,19 +96,13 @@ public void write(final List<DDSpan> trace) {
...
       flush();
     }
   }
 
-  private void handleDroppedTrace(final String reason, final List<DDSpan> trace) {
-    log.debug("{}. Counted but dropping trace: {}", reason, trace);
-    healthMetrics.onFailedPublish(UNSET);
-    incrementDropCounts(trace.size());
-  }
-
```
bantonsson, line 111: 🧹

bantonsson (approved):

ygree:
The motivation isn't clear to me, because there used to be only two null check when the worker is built and when it's closed. It'd probably make sense if there was a repetitive null check. Now we have an interface and extra class as opposed to two simple null checks.

PR #4446, opened 2022-12-19
Update overhead controller and add debug info

manuel-alvarez-alvarez:
# What Does This Do
Extract atomic semaphores from overhead controller and tainted objects and adds debug information to troubleshoot issues.

# Motivation
We are getting an increased level of flaky tests in system-tests due to the overhead controller.

# Additional Notes

### DIFF
--- a/dd-java-agent/agent-iast/src/main/java/com/datadog/iast/overhead/OverheadController.java
+++ b/dd-java-agent/agent-iast/src/main/java/com/datadog/iast/overhead/OverheadController.java
@@ -1,92 +1,185 @@
 package com.datadog.iast.overhead;
 
 import com.datadog.iast.IastRequestContext;
+import com.datadog.iast.IastSystem;
+import com.datadog.iast.util.NonBlockingSemaphore;
 import datadog.trace.api.Config;
 import datadog.trace.api.gateway.RequestContext;
 import datadog.trace.api.gateway.RequestContextSlot;
 import datadog.trace.bootstrap.instrumentation.api.AgentSpan;
+import datadog.trace.bootstrap.instrumentation.api.AgentTracer;
 import datadog.trace.util.AgentTaskScheduler;
 import java.util.concurrent.TimeUnit;
 import java.util.concurrent.atomic.AtomicInteger;
+import org.slf4j.Logger;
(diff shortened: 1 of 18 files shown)


dd-java-agent/agent-iast/src/main/java/com/datadog/iast/overhead/OverheadController.java
```diff
@@ -1,92 +1,188 @@
...
-  public void releaseRequest() {
-    availableRequests.updateAndGet(x -> (x < maxConcurrentRequests) ? x + 1 : x);
-  }
+    @Override
+    public void releaseRequest() {
+      delegate.releaseRequest();
+      if (LOGGER.isDebugEnabled()) {
+        LOGGER.debug(
+            "releaseRequest: availableRequests={}, span={}",
+            getAvailableRequests(),
```
smola, line 63: Keep in mind that this won't be fully accurate unless you pick the available requests returning from the `releaseRequest` call. But I guess we can live with that.

dd-java-agent/agent-iast/src/main/java/com/datadog/iast/util/AtomicSemaphore.java
```diff
@@ -0,0 +1,83 @@
+package com.datadog.iast.util;
+
+import java.util.concurrent.atomic.AtomicBoolean;
+import java.util.concurrent.atomic.AtomicInteger;
+
+public interface AtomicSemaphore {
```
smola, line 6: I'd probably call it something like `NonBlockingSemaphore`, which is more related to its semantics, rather than its implementation.

manuel-alvarez-alvarez: Yep, it's a better name, changed!

dd-java-agent/agent-iast/src/main/java/com/datadog/iast/overhead/OverheadController.java
manuel-alvarez-alvarez: I've checked the code, but it's true that even getting the result from `releaseRequest` it will never be fully accurate.

smola (approved):

PR #4448, opened 2022-12-19
Fix serialization of telemetric metrics

cataphract:
Serializer didn't know how to handle `Number` objects.

### DIFF
--- a/telemetry/src/main/java/datadog/telemetry/RequestBuilder.java
+++ b/telemetry/src/main/java/datadog/telemetry/RequestBuilder.java
@@ -1,7 +1,11 @@
 package datadog.telemetry;
 
+import com.squareup.moshi.FromJson;
 import com.squareup.moshi.JsonAdapter;
+import com.squareup.moshi.JsonReader;
+import com.squareup.moshi.JsonWriter;
 import com.squareup.moshi.Moshi;
+import com.squareup.moshi.ToJson;
 import datadog.common.container.ContainerInfo;
 import datadog.communication.ddagent.TracerVersion;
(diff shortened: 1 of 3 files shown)


internal-api/build.gradle
```diff
@@ -8,6 +8,13 @@ ext {
   enableJunitPlatform = true
 }
 
+// sun.misc.SharedSecrets is gone in later versions
+compileJava {
```
smola, line 12: @cataphract Would you mind splitting this out to a separate PR?

telemetry/src/main/java/datadog/telemetry/RequestBuilder.java
```diff
@@ -98,4 +105,17 @@ public Request build(RequestType requestType, Payload payload) {
...
+  private static final class NumberJsonAdapter {
+    @Nullable
+    @FromJson
+    public Number fromJson(JsonReader reader) throws IOException {
+      throw new UnsupportedOperationException();
+    }
+
+    @ToJson
+    public void toJson(JsonWriter writer, @Nullable Number value) throws IOException {
+      writer.value(value);
```
smola, line 118: That's it? Is it _that_ simple? I guess this will work as long as it's a standard one?
I would add tests using types we expect here (BigDecimal? Float?)

cataphract: `JsonWriter` accepts any `Number` and will call `toString` on it. Technically, it will not work, in the sense of producing valid json, on any `Number` (e.g. a too big `BigDecimal`), but it should be good enough™.

smola (approved):

PR #4449, opened 2022-12-19
Add metric for serialization failure

PerfectSlayer:
# What Does This Do

Count the number of dropped traces by serialization failure.

# Motivation

Where a trace is sampled, it can be lost if something unexpected happens during serialization and it was not counted as a dropped trace.

# Additional Notes


nayeem-kamal:
LGTM

nayeem-kamal (approved):

PR #4454, opened 2022-12-20
Reduce Telemetry warnings

smola:
# What Does This Do

Reduce Telemetry warnings. In particular, log all network and HTTP errors to debug, since there are many scenarios where this is frequent (but not harmful).

# Motivation

These warnings are too alarming for customers, for scenarios that are common and safe.

# Additional Notes

There are still some possible warning messages in the module, but the ones I reviewed seem to be unexpected conditions that we do want customers to notice.


ValentinZakharov (approved):

PerfectSlayer (approved):

mcculls (approved):
+1

my usual rule of thumb is "is this something the customer can/should act on" - if it is then log it as a warning, otherwise log it as debug

PR #4459, opened 2022-12-20
Fix instrumentation problem in jetty 9.4

cataphract:
ASM doesn't like getting two frame declared in a row, without any intervening code (with one exception that's not relevant). So add a NOP instruction.

### DIFF
--- a/dd-java-agent/instrumentation/jetty-common/src/main/java/datadog/trace/instrumentation/jetty/HandleRequestVisitor.java
+++ b/dd-java-agent/instrumentation/jetty-common/src/main/java/datadog/trace/instrumentation/jetty/HandleRequestVisitor.java
@@ -127,6 +127,8 @@ public class HandleRequestVisitor extends MethodVisitor {
       super.visitLabel(afterHandle);
       if (needsStackFrames()) {
         super.visitFrame(Opcodes.F_SAME, 0, null, 0, null);
+        // add NOP because if this is followed by another frame asm may not like it
+        super.visitInsn(Opcodes.NOP);
       }
       this.success = true;
       return;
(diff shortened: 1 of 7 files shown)


dd-java-agent/instrumentation/java-concurrent/java-completablefuture/build.gradle
```diff
@@ -5,8 +5,17 @@ muzzle {
   }
 }
 
+ext {
+  skipSettingCompilerRelease = true
+}
 apply from: "${rootDir}/gradle/java.gradle"
 
+compileJava {
```
smola, line 13: Separate PR? If we really want to handle it this way, we should do it for all required modules at once?

smola (approved):

dd-java-agent/instrumentation/jetty-common/src/main/java/datadog/trace/instrumentation/jetty/HandleRequestVisitor.java
```diff
@@ -127,6 +127,8 @@ public void visitMethodInsn(
       super.visitLabel(afterHandle);
       if (needsStackFrames()) {
         super.visitFrame(Opcodes.F_SAME, 0, null, 0, null);
+        // add NOP because if this is followed by another frame asm may not like it
+        super.visitInsn(Opcodes.NOP);
```
dougqh, line 131: Injecting nop-s to fix a problem elsewhere seems wrong to me.

This seems like we're creating coupling at distance which is going to make the overall system rather brittle.

cataphract: The alternative is to look ahead in the visitor and check if the next visitXXX call is one that adds instructions to the method. This is significantly more difficult to do with the asm library. We would need to create a wrapper that held off on emitting visitFrame until it saw the next visitation, something like we have in `DelayLoadsMethodVisitor`. Even then, though this is unlikely to be a problem, we would rely on the method visitor below actually producing the code associated with the visitation and not it doing a transformation itself that would place a frame before emitting some code. So while inserting a NOP doesn't look great, I still think it's better than the alternative.

dougqh: That doesn't sound terribly complicated to me -- just normal things that need to be done in stream processing / code gen.  I cannot say I agree with the conclusion, but I'll let you decide.

cataphract: I've changed the strategy in the new commit

PR #4464, opened 2022-12-21
Deprecate dd.propagaion.style for dd.trace.propagation.style

bantonsson:
# What Does This Do

This deprecates the `dd.propagation.style.extract` and `dd.propagation.style.inject` settings in favor of the three new settings `dd.trace.propagation.style`, `dd.trace.propagation.style.extract`, and `dd.trace.propagation.style.inject`.

The new `dd.trace.propagation.style*` settings makes a distinction between `B3` which is [single header b3 propagation](https://github.com/openzipkin/b3-propagation/tree/master#single-header) and `B3MULTI` which is [multiple header b3 propagation](https://github.com/openzipkin/b3-propagation/tree/master#multiple-headers).

The change is backwards compatible and using the deprecated settings will only result in a deprecation warning being logged, and the settings being translated into the new ones, including the old `B3` setting being translated into both the new `B3` and `B3MULTI`.

# Motivation

Alignment between tracer implementations and preparation for w3c header propagation.

# Additional Notes

### DIFF
--- a/dd-trace-api/src/main/java/datadog/trace/api/TracePropagationStyle.java
+++ b/dd-trace-api/src/main/java/datadog/trace/api/TracePropagationStyle.java
@@ -1,0 +1,39 @@
+package datadog.trace.api;
+
+/** Trace propagation styles for injecting and extracting trace propagation headers. */
+public enum TracePropagationStyle {
+  // Datadog context propagation style
+  DATADOG,
+  // B3 single header context propagation style
+  // https://github.com/openzipkin/b3-propagation/tree/master#single-header
+  B3,
+  // B3 multi header context propagation style
+  // https://github.com/openzipkin/b3-propagation/tree/master#multiple-headers
+  B3MULTI,
+  // Haystack context propagation style
+  // https://github.com/ExpediaDotCom/haystack
+  HAYSTACK,
+  // Amazon X-Ray context propagation style
+  // https://docs.aws.amazon.com/xray/latest/devguide/xray-concepts.html#xray-concepts-tracingheader
+  XRAY;
+
+  public static TracePropagationStyle valueOfDisplayName(String displayName) {
+    String convertedName = displayName.toUpperCase().replace(' ', '_');
+    // Another name for B3 for cross tracer compatibility
+    if (convertedName.equals("B3_SINGLE_HEADER")) {
+      return B3;
+    }
+    return TracePropagationStyle.valueOf(convertedName);
+  }
+
+  private String displayName;
+
+  @Override
+  public String toString() {
+    String string = displayName;
+    if (displayName == null) {
+      string = displayName = name().toLowerCase().replace('_', ' ');
+    }
+    return string;
+  }
+}
--- a/internal-api/src/main/java/datadog/trace/api/Config.java
+++ b/internal-api/src/main/java/datadog/trace/api/Config.java
@@ -51,8 +51,6 @@ import static datadog.trace.api.ConfigDefaults.DEFAULT_PERF_METRICS_ENABLED;
 import static datadog.trace.api.ConfigDefaults.DEFAULT_PRIORITY_SAMPLING_ENABLED;
 import static datadog.trace.api.ConfigDefaults.DEFAULT_PRIORITY_SAMPLING_FORCE;
 import static datadog.trace.api.ConfigDefaults.DEFAULT_PROPAGATION_EXTRACT_LOG_HEADER_NAMES_ENABLED;
-import static datadog.trace.api.ConfigDefaults.DEFAULT_PROPAGATION_STYLE_EXTRACT;
-import static datadog.trace.api.ConfigDefaults.DEFAULT_PROPAGATION_STYLE_INJECT;
 import static datadog.trace.api.ConfigDefaults.DEFAULT_REMOTE_CONFIG_ENABLED;
 import static datadog.trace.api.ConfigDefaults.DEFAULT_REMOTE_CONFIG_INITIAL_POLL_INTERVAL;
 import static datadog.trace.api.ConfigDefaults.DEFAULT_REMOTE_CONFIG_INTEGRITY_CHECK_ENABLED;
@@ -275,6 +273,9 @@ import static datadog.trace.api.config.TracerConfig.TRACE_ANALYTICS_ENABLED;
 import static datadog.trace.api.config.TracerConfig.TRACE_CLIENT_IP_HEADER;
 import static datadog.trace.api.config.TracerConfig.TRACE_CLIENT_IP_RESOLVER_ENABLED;
 import static datadog.trace.api.config.TracerConfig.TRACE_HTTP_SERVER_PATH_RESOURCE_NAME_MAPPING;
+import static datadog.trace.api.config.TracerConfig.TRACE_PROPAGATION_STYLE;
+import static datadog.trace.api.config.TracerConfig.TRACE_PROPAGATION_STYLE_EXTRACT;
+import static datadog.trace.api.config.TracerConfig.TRACE_PROPAGATION_STYLE_INJECT;
 import static datadog.trace.api.config.TracerConfig.TRACE_RATE_LIMIT;
 import static datadog.trace.api.config.TracerConfig.TRACE_REPORT_HOSTNAME;
 import static datadog.trace.api.config.TracerConfig.TRACE_RESOLVER_ENABLED;
@@ -325,6 +326,7 @@ import java.util.Properties;
 import java.util.Set;
 import java.util.SortedSet;
 import java.util.UUID;
+import java.util.function.Function;
 import java.util.regex.Pattern;
 import javax.annotation.Nonnull;
 import org.slf4j.Logger;
@@ -421,6 +423,8 @@ public class Config {
   private final boolean logExtractHeaderNames;
   private final Set<PropagationStyle> propagationStylesToExtract;
   private final Set<PropagationStyle> propagationStylesToInject;
+  private final Set<TracePropagationStyle> tracePropagationStylesToExtract;
+  private final Set<TracePropagationStyle> tracePropagationStylesToInject;
   private final int clockSyncPeriod;
 
   private final String dogStatsDNamedPipe;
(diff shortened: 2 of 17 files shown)


internal-api/src/main/java/datadog/trace/api/Config.java
```diff
@@ -2533,14 +2676,41 @@ private static Set<String> parseStringIntoSetOfNonEmptyStrings(final String str)
...
+            "Cannot recognize config string value: {} for setting {}",
+            value,
+            propertyNameToSystemPropertyName(setting));
+      }
+    }
+    return Collections.unmodifiableSet(result);
+  }
+
+  @Nonnull
+  private static Set<TracePropagationStyle> convertStringSetToTracePropagationStyleSet(
```
PerfectSlayer, line 2699: Can this method be factorized with `convertStringSetToPropagationStyleSet()`?
Like `Set<T> convertStringSetToEnumSet(String settings, Set<String> input, Fuction<String, T> enumMapper)`?

internal-api/src/main/java/datadog/trace/api/Config.java
```diff
@@ -2533,14 +2676,41 @@ private static Set<String> parseStringIntoSetOfNonEmptyStrings(final String str)
 
   @Nonnull
   private static Set<PropagationStyle> convertStringSetToPropagationStyleSet(
```
PerfectSlayer, line 2678: Should this be deprecated too?

internal-api/src/main/java/datadog/trace/api/Config.java
```diff
@@ -862,12 +865,106 @@ private Config(final ConfigProvider configProvider, final InstrumenterConfig ins
...
+        // This setting is here for backwards compatibility until it can be removed in a major
+        // release of the java tracer
+        propagationStylesToInject =
+            convertStringSetToPropagationStyleSet(
+                PROPAGATION_STYLE_INJECT, Collections.singleton(PropagationStyle.DATADOG.name()));
+      }
+      // Now we can check if we should pick the default injection/extraction
+      if (extract.isEmpty()) {
+        extract =
+            convertStringSetToTracePropagationStyleSet(
```
PerfectSlayer, line 955: Why not simply use `Set.of(DATADOG)`?

internal-api/src/main/java/datadog/trace/api/Config.java
```diff
@@ -862,12 +865,106 @@ private Config(final ConfigProvider configProvider, final InstrumenterConfig ins
...
+                TRACE_PROPAGATION_STYLE_EXTRACT,
+                Collections.singleton(TracePropagationStyle.DATADOG.toString()));
+      }
+      if (inject.isEmpty()) {
+        inject =
+            convertStringSetToTracePropagationStyleSet(
```
PerfectSlayer, line 961: Why not simply use `Set.of(DATADOG)`?

internal-api/src/main/java/datadog/trace/api/Config.java
```diff
@@ -862,12 +865,106 @@ private Config(final ConfigProvider configProvider, final InstrumenterConfig ins
...
+          getTracePropagationStyleSetSettingFromEnvironment(TRACE_PROPAGATION_STYLE);
+      Set<TracePropagationStyle> extract =
+          getTracePropagationStyleSetSettingFromEnvironment(TRACE_PROPAGATION_STYLE_EXTRACT);
+      Set<TracePropagationStyle> inject =
+          getTracePropagationStyleSetSettingFromEnvironment(TRACE_PROPAGATION_STYLE_INJECT);
+      String extractOrigin = TRACE_PROPAGATION_STYLE_EXTRACT;
+      String injectOrigin = TRACE_PROPAGATION_STYLE_INJECT;
+      // The more specific settings will override the common setting, so log warnings if they are
+      // both set
+      if (!common.isEmpty()) {
```
PerfectSlayer, line 890: Open question: I wonder if there is a way to avoid nested `if` by checking settings in the priority order.
If a setting has value, define the config from its value, or log issue if already defined (according to the current priority of the settings itself: overidden/deprecated/etc...).

dd-trace-api/src/main/java/datadog/trace/api/TracePropagationStyle.java
```diff
@@ -0,0 +1,36 @@
...
+  // Datadog context propagation style
+  DATADOG,
+  // B3 single header context propagation style
+  // https://github.com/openzipkin/b3-propagation/tree/master#single-header
+  B3,
+  // B3 multi header context propagation style
+  // https://github.com/openzipkin/b3-propagation/tree/master#multiple-headers
+  B3MULTI,
+  // Another name for B3 for cross tracer compatibility
+  B3_SINGLE_HEADER,
```
PerfectSlayer, line 14: I would prevent creating a duplicate value for compatibility.
Where does this value could come from? If it's from config, handling it from parsing might be a better solution than adding to into API (by creating its enum value).

PerfectSlayer (commented):
Few minor comments and questions :)

internal-api/src/main/java/datadog/trace/api/Config.java
bantonsson: Nope, since one uses the `valueOf` and the other the special method `valueOfDisplayName`

internal-api/src/main/java/datadog/trace/api/Config.java
bantonsson: Sure, that makes sense.

internal-api/src/main/java/datadog/trace/api/Config.java
bantonsson: 🤦🏼‍♂️ Absolutely

internal-api/src/main/java/datadog/trace/api/Config.java
bantonsson: Cut'n'paste  🤦🏼‍♂️

dd-trace-api/src/main/java/datadog/trace/api/TracePropagationStyle.java
bantonsson: Sure, I can remove that one. It only materializes via parsing of config settings.

internal-api/src/main/java/datadog/trace/api/Config.java
bantonsson: And `Set.of(...)` is Java 9+ 😢 But `Collections.singleton(...)` works fine.

internal-api/src/main/java/datadog/trace/api/Config.java
bantonsson: Yes, the code is a bit convoluted. I'll try to make it more straight forward.

internal-api/src/main/java/datadog/trace/api/Config.java
PerfectSlayer: And using `PropagationStyle::valueOf` and `TracePropagationStyle::valueOfDisplayName` as mapper does not do the trick?
-- Just saw there is a call to uppercase too... So require to compose both functions 😞

bantonsson: I'll try to clean it up a bit

bantonsson: I've collapsed all of these duplicate methods now.

dd-trace-api/src/main/java/datadog/trace/api/TracePropagationStyle.java
```diff
@@ -0,0 +1,39 @@
...
+  HAYSTACK,
+  // Amazon X-Ray context propagation style
+  // https://docs.aws.amazon.com/xray/latest/devguide/xray-concepts.html#xray-concepts-tracingheader
+  XRAY;
+
+  public static TracePropagationStyle valueOfDisplayName(String displayName) {
+    String convertedName = displayName.toUpperCase().replace(' ', '_');
+    // Another name for B3 for cross tracer compatibility
+    if (convertedName.equals("B3_SINGLE_HEADER")) {
+      convertedName = B3.name();
```
ygree, line 24: It's minor but it caught my eye. Could just return B3 value at this point?

dd-trace-api/src/main/java/datadog/trace/api/TracePropagationStyle.java
```diff
@@ -0,0 +1,39 @@
...
+    return TracePropagationStyle.valueOf(convertedName);
+  }
+
+  private String displayName;
+
+  @Override
+  public String toString() {
+    String string = displayName;
+    if (displayName == null) {
+      string = displayName = name().toLowerCase().replace('_', ' ');
```
ygree, line 35: Also minor but, maybe extract a method for ".toLowerCase().replace('_', ' ')", e.g. normalizeName to have this logic in one place.

dd-trace-api/src/main/java/datadog/trace/api/TracePropagationStyle.java
```diff
@@ -0,0 +1,39 @@
+package datadog.trace.api;
+
+/** Trace propagation styles for injecting and extracting trace propagation headers. */
+public enum TracePropagationStyle {
+  // Datadog context propagation style
+  DATADOG,
+  // B3 single header context propagation style
+  // https://github.com/openzipkin/b3-propagation/tree/master#single-header
```
ygree, line 8: Thanks for providing useful comments!

internal-api/src/main/java/datadog/trace/api/Config.java
```diff
@@ -862,12 +866,92 @@ private Config(final ConfigProvider configProvider, final InstrumenterConfig ins
...
             DEFAULT_PROPAGATION_EXTRACT_LOG_HEADER_NAMES_ENABLED);
 
-    propagationStylesToExtract =
-        getPropagationStyleSetSettingFromEnvironmentOrDefault(
-            PROPAGATION_STYLE_EXTRACT, DEFAULT_PROPAGATION_STYLE_EXTRACT);
-    propagationStylesToInject =
-        getPropagationStyleSetSettingFromEnvironmentOrDefault(
-            PROPAGATION_STYLE_INJECT, DEFAULT_PROPAGATION_STYLE_INJECT);
+    {
+      // The dd.propagation.style.(extract|inject) settings have been deprecated in
```
ygree, line 870: Nice to have this comment explaining this change!

ygree (approved):
LGTM with a couple minor suggestions

dd-trace-api/src/main/java/datadog/trace/api/TracePropagationStyle.java
bantonsson: Yes definitley

dd-trace-api/src/main/java/datadog/trace/api/TracePropagationStyle.java
bantonsson: The two code snippets do the opposite of each other.

PR #4465, opened 2022-12-21
Fix NotEqual operator

jpbempel:
# What Does This Do
add support for `ne` json operator in expression Language as specified (instead of `neq`).
Fix also the way we parse Json numbers for separating integers from decimals

# Motivation

# Additional Notes


ojung (approved):
Thanks 👍

PR #4469, opened 2022-12-22
Log JSON sampling rules when failed to parse for easier debugging.

ygree:
# What Does This Do

Adds logging for JSON sampling rules when failed to parse for easier debugging.

# Motivation

When it couldn't parse provided JSON sampling rules it would be useful the rules logged to simplify debugging.

# Additional Notes

APMS-8472

### DIFF
--- a/dd-trace-core/src/main/java/datadog/trace/common/sampling/SpanSamplingRules.java
+++ b/dd-trace-core/src/main/java/datadog/trace/common/sampling/SpanSamplingRules.java
@@ -128,7 +128,7 @@ public class SpanSamplingRules {
       List<Rule> rules = LIST_OF_RULES_ADAPTER.fromJson(json);
       return filterOutNullRules(rules);
     } catch (Throwable ex) {
-      log.error("Couldn't parse Span Sampling Rules from JSON", ex);
+      log.error("Couldn't parse Span Sampling Rules from JSON: " + json, ex);
       return null;
     }
   }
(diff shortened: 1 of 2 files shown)


dd-trace-core/src/main/java/datadog/trace/common/sampling/SpanSamplingRules.java
```diff
@@ -128,7 +128,7 @@ public static SpanSamplingRules deserialize(String json) {
       List<Rule> rules = LIST_OF_RULES_ADAPTER.fromJson(json);
       return filterOutNullRules(rules);
     } catch (Throwable ex) {
-      log.error("Couldn't parse Span Sampling Rules from JSON", ex);
+      log.error("Couldn't parse Span Sampling Rules from JSON: " + json, ex);
```
bantonsson, line 131:
```suggestion
      log.error("Couldn't parse Span Sampling Rules from JSON: {}", json, ex);
```

ygree: I wasn't sure if the exception object would be properly logged this way, since there is a method for `void error(String var1, Throwable var2);` that takes error that must be Throwable but if passed two params it'd use `void error(String var1, Object var2, Object var3);`.  I've checked and it seems to work as expected.

bantonsson (approved):

PR #4470, opened 2022-12-22
Added NoOp implementation of HealthMetrics and modified tests

nayeem-kamal:
Modified Healthmetrics to implement a NoOp version and modified test to reflect the new NoOpHealthMetrics implementation

# What Does This Do
Adds an interface for health metrics which will now be implemented by TracerHealthMetrics and NoOpHealthMetrics
# Motivation
Needed a NoOp implementation of HealthMetrics to be utilized when Health metrics is not enabled in the config or for testing.
# Additional Notes


dd-trace-core/src/main/java/datadog/trace/core/monitor/HealthMetrics.java
```diff
@@ -30,272 +16,44 @@
...
-
-  public void onPartialPublish(final int numberOfDroppedSpans) {
-    partialTraces.inc();
-    droppedSpans.inc(numberOfDroppedSpans);
-  }
-
-  public void onScheduleFlush(final boolean previousIncomplete) {
-    // not recorded
-  }
+public interface HealthMetrics extends AutoCloseable {
```
ygree, line 19: I'd suggest an abstract class for HealthMetrics instead of an interface for two reasons. First, less boilerplate as the method declaration can be placed together with an empty implementation. Second, unoptimized interface calls can be less efficient.

dd-trace-core/src/main/java/datadog/trace/core/CoreTracer.java
```diff
@@ -482,8 +483,8 @@ private CoreTracer(
             : Monitoring.DISABLED;
     this.healthMetrics =
         config.isHealthMetricsEnabled()
-            ? new HealthMetrics(this.statsDClient)
-            : new HealthMetrics(StatsDClient.NO_OP);
+            ? new TracerHealthMetrics(this.statsDClient)
```
ygree, line 486: Do we ever need combination of TracerHealthMetrics and StatsDClient.NO_OP? If not, do we still need StatsDClient.NO_OP for other purposes? Otherwise we could remove it completely, but I didn't check if this is the case.

dd-trace-core/src/main/java/datadog/trace/core/monitor/HealthMetrics.java
ygree: In that case we could also keep HealthMetrics name. No need to introduce TracerHealthMetrics, that could be a bit misleading.

dd-trace-core/src/main/java/datadog/trace/core/CoreTracer.java
nayeem-kamal: It looks like the StatsDClient.NO_OP implementation is still used in the tracer and in tests, so I think we would need to keep it.

ygree (approved):
LGTM

PR #4471, opened 2022-12-22
DDIntakeWriter: Do not flush after every single trace

albertvaka:
# What Does This Do
* Change the `DDIntakeWriter` default for `alwaysFlush` so we batch traces together.
* Fixes the `CiTestCycleMapperV1` mapper so it doesn't add N headers when sending N traces. It now generates the header separately and sends it only once at the beginning of every payload.
* Updates the test so we only expect one header followed by any number of traces in a payload.

Generating the header once in the mapper this way is similar to what we do in `TraceMapperV0_4` and `TraceMapperV0_5`.
# Motivation
Using the `DDIntakeWriter` was painfully slow because it was making one request per trace (ie: one per test in CI Visibility).

This only affected CI Visibility users in Agentless mode before, but after #4369 would also affect Agent users.

### DIFF
--- a/dd-trace-core/src/main/java/datadog/trace/common/writer/DDIntakeWriter.java
+++ b/dd-trace-core/src/main/java/datadog/trace/common/writer/DDIntakeWriter.java
@@ -40,7 +40,7 @@ public class DDIntakeWriter extends RemoteWriter {
     RetryPolicy retryPolicy = RetryPolicy.builder().withMaxRetry(5).withBackoff(100).build();
     private int flushTimeout = 5;
     private TimeUnit flushTimeoutUnit = TimeUnit.SECONDS;
-    private boolean alwaysFlush = true;
+    private boolean alwaysFlush = false;
 
     private RemoteApi intakeApi;
     private String apiKey;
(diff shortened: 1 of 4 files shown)


dd-trace-core/src/main/java/datadog/trace/common/writer/DDIntakeWriter.java
```diff
@@ -40,7 +40,7 @@ public static class DDIntakeWriterBuilder {
     RetryPolicy retryPolicy = RetryPolicy.builder().withMaxRetry(5).withBackoff(100).build();
     private int flushTimeout = 5;
     private TimeUnit flushTimeoutUnit = TimeUnit.SECONDS;
-    private boolean alwaysFlush = true;
+    private boolean alwaysFlush = false;
```
bantonsson, line 43: I assume that flushing during _shutdown_ is now handled by the _closing_ of the tracer via the shudown hook. Is that good enough?

bantonsson (approved):

dd-trace-core/src/main/java/datadog/trace/common/writer/DDIntakeWriter.java
albertvaka: It has been enough in my tests.

drodriguezhdez (commented):
LGTM

PR #4472, opened 2022-12-23
apache httpclient5 async instrumentation

amarziali:
# What Does This Do

Provides instrumentation and distributed tracing for apache httpclients5 async

# Motivation

# Additional Notes

### DIFF
--- a/dd-java-agent/instrumentation/apache-httpclient-5/src/main/java/datadog/trace/instrumentation/apachehttpclient5/ApacheHttpAsyncClientInstrumentation.java
+++ b/dd-java-agent/instrumentation/apache-httpclient-5/src/main/java/datadog/trace/instrumentation/apachehttpclient5/ApacheHttpAsyncClientInstrumentation.java
@@ -1,0 +1,117 @@
+package datadog.trace.instrumentation.apachehttpclient5;
+
+import static datadog.trace.agent.tooling.bytebuddy.matcher.HierarchyMatchers.implementsInterface;
+import static datadog.trace.agent.tooling.bytebuddy.matcher.NameMatchers.named;
+import static datadog.trace.bootstrap.instrumentation.api.AgentTracer.activeScope;
+import static datadog.trace.bootstrap.instrumentation.api.AgentTracer.startSpan;
+import static datadog.trace.instrumentation.apachehttpclient5.ApacheHttpClientDecorator.DECORATE;
+import static datadog.trace.instrumentation.apachehttpclient5.ApacheHttpClientDecorator.HTTP_REQUEST;
+import static net.bytebuddy.matcher.ElementMatchers.isMethod;
+import static net.bytebuddy.matcher.ElementMatchers.takesArgument;
+import static net.bytebuddy.matcher.ElementMatchers.takesArguments;
+
+import com.google.auto.service.AutoService;
+import datadog.trace.agent.tooling.Instrumenter;
+import datadog.trace.bootstrap.instrumentation.api.AgentScope;
+import datadog.trace.bootstrap.instrumentation.api.AgentSpan;
+import net.bytebuddy.asm.Advice;
+import net.bytebuddy.description.type.TypeDescription;
+import net.bytebuddy.matcher.ElementMatcher;
+import org.apache.hc.core5.concurrent.FutureCallback;
+import org.apache.hc.core5.http.nio.AsyncRequestProducer;
+import org.apache.hc.core5.http.protocol.HttpContext;
+
+@AutoService(Instrumenter.class)
+public class ApacheHttpAsyncClientInstrumentation extends Instrumenter.Tracing
+    implements Instrumenter.CanShortcutTypeMatching {
+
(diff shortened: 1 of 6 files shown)


dd-java-agent/instrumentation/apache-httpclient-5/src/main/java/datadog/trace/instrumentation/apachehttpclient5/HttpAsyncClientExchangeHandlerInstrumentation.java
```diff
@@ -0,0 +1,88 @@
...
+      packageName + ".DelegatingRequestChannel",
+      packageName + ".DelegatingRequestProducer",
+      packageName + ".TraceContinuedFutureCallback"
+    };
+  }
+
+  @Override
+  public void adviceTransformations(AdviceTransformation transformation) {
+    transformation.applyAdvice(
+        namedOneOf("consumeContent", "produceContent"), getClass().getName() + "$RecordActivity");
```
PerfectSlayer, line 65: I struggle to find which methods from which classes match from the httpclient5 library. Can you help me with this?

dd-java-agent/instrumentation/apache-httpclient-5/src/main/java/datadog/trace/instrumentation/apachehttpclient5/HttpAsyncClientExchangeHandlerInstrumentation.java
```diff
@@ -0,0 +1,88 @@
...
+      if (callback instanceof TraceContinuedFutureCallback) {
+        TraceContinuedFutureCallback<?> tracedCallback = (TraceContinuedFutureCallback<?>) callback;
+        tracedCallback.resume();
+        return tracedCallback;
+      }
+      return null;
+    }
+
+    @Advice.OnMethodExit(onThrowable = Throwable.class)
+    public static void after(@Advice.Enter TraceContinuedFutureCallback<?> callback) {
```
PerfectSlayer, line 82: Quick question: Is it implicit you will have the right type (`TraceContinuedFutureCallback`) from the `OnMethodExit`? Because it tested in the `OnMethodEnter` but not `OnMethodExit`.

amarziali: yes because otherwise on the onEnter will return null. The same works safely on apache httpclient 4 instrumentation: datadog.trace.instrumentation.apachehttpasyncclient.HttpAsyncClientExchangeHandlerInstrumentation

dd-java-agent/instrumentation/apache-httpclient-5/src/main/java/datadog/trace/instrumentation/apachehttpclient5/HttpAsyncClientExchangeHandlerInstrumentation.java
amarziali: I struggle as well. Sorry I used bad names. I'm fixing it. Thanks for having spotted it out

dd-java-agent/instrumentation/apache-httpclient-5/src/main/java/datadog/trace/instrumentation/apachehttpclient5/HttpAsyncClientExchangeHandlerInstrumentation.java
PerfectSlayer: Ok thanks!

PerfectSlayer (approved):

dd-java-agent/instrumentation/apache-httpclient-5/src/main/java/datadog/trace/instrumentation/apachehttpclient5/ApacheHttpAsyncClientInstrumentation.java
```diff
@@ -0,0 +1,117 @@
...
+
+      requestProducer = new DelegatingRequestProducer(clientSpan, requestProducer);
+      futureCallback =
+          new TraceContinuedFutureCallback<>(parentScope, clientSpan, context, futureCallback);
+
+      return clientSpan;
+    }
+
+    @Advice.OnMethodExit(onThrowable = Throwable.class, suppress = Throwable.class)
+    public static void methodExit(
```
richardstartin, line 106: If `methodEnter` produced and opened a scope instead of a span here, if an error is thrown we (profiling) can label our exception samples with the span id, and any allocation/cpu/walltime sampled here can be attributed to the span too. 

This is a general point about writing instrumentation, having scopes open where possible leads to better profile context, so please bear it in mind when writing instrumentaiton.

richardstartin:
Sorry I didn't mean to merge, I thought I was looking at a different PR 🤦🏻

PR #4473, opened 2022-12-23
Ignore org.springframework.web.util.WebUtils from exclusion list

jandro996:
# What Does This Do

- Ignore org.springframework.web.util.WebUtils from exclusion list
- Add IAST smoke test for springframework parameter binding

# Motivation

IAST needs to instrument this class to taint source parameters when parameter binding is used in  spring framework 

# Additional Notes


dd-java-agent/agent-tooling/src/main/resources/datadog/trace/agent/tooling/bytebuddy/matcher/ignored_class_name.trie
```diff
@@ -261,6 +261,7 @@
 0 org.springframework.web.context.support.XmlWebApplicationContext
 0 org.springframework.web.reactive.*
 0 org.springframework.web.servlet.*
+0 org.springframework.web.util.WebUtils
```
smola, line 264: We could adda comment here:
```
# Included for IAST
```

smola (approved):

PR #4474, opened 2022-12-23
Guard against agent being initialized twice

smola:
# What Does This Do

Guard against agent being initialized twice. Instead of producing later errors, it will just print a warning and stop any further work.

# Motivation

Some customers are defining `-javaagent` more than once. This produces later failures, in very non-obvious ways.

# Additional Notes


albertvaka (approved):
LGTM

bantonsson (approved):

PR #4478, opened 2022-12-27
Add null literal support

jpbempel:
# What Does This Do
Can check accessing reference against null

# Motivation
Able to write `field != null` in expression language

# Additional Notes


cimi (approved):
:+1: Thanks!

PR #4479, opened 2022-12-27
Fail process when running in Agentless mode and API key is not set

nikita-tkachenko-datadog:
# What Does This Do
When in Agentless mode, the Datadog API key is required to send payloads to the intake.
If we detect there’s no API Key, currently we silently continue without sending payloads.
We want to fail the process instead.

# Motivation
It is easy to overlook such misconfiguration, so we want to make it more obvious by failing the process rather than silently continuing.

# Additional Notes
This only applies to scenarios where CI visibility feature is enabled.
If CI visibility is disabled, previous behaviour remains unchanged (proceed with starting the process even if the API key is missing).

### DIFF
--- a/internal-api/src/main/java/datadog/trace/api/Config.java
+++ b/internal-api/src/main/java/datadog/trace/api/Config.java
@@ -297,6 +297,7 @@ import datadog.trace.bootstrap.config.provider.ConfigProvider;
 import datadog.trace.bootstrap.config.provider.SystemPropertiesConfigSource;
 import datadog.trace.util.PidHelper;
 import datadog.trace.util.Strings;
+import datadog.trace.util.throwable.FatalAgentMisconfigurationError;
 import edu.umd.cs.findbugs.annotations.SuppressFBWarnings;
 import java.io.BufferedReader;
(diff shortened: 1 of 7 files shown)


bantonsson (approved):

internal-api/src/main/java/datadog/trace/api/Config.java
```diff
@@ -1255,6 +1256,14 @@ private Config(final ConfigProvider configProvider, final InstrumenterConfig ins
           "Agentless profiling activated but no api key provided. Profile uploading will likely fail");
     }
 
+    if (isCiVisibilityEnabled()
+        && ciVisibilityAgentlessEnabled
+        && (apiKey == null || apiKey.isEmpty())) {
+      throw new FatalAgentMisconfigurationError(
+          "Attempt to start in Agentless mode without API key. "
+              + "Please ensure that either API key is configured, or the tracer is set up to work with Agent");
```
bantonsson, line 1264:
```suggestion
              + "Please ensure that either an API key is configured, or the tracer is set up to work with the Agent");
```

PR #4480, opened 2022-12-28
Handle HTTP response 429: Rate Limit Exceeded

nikita-tkachenko-datadog:
# What Does This Do
This change updates retry policy for HTTP requests to the Intake API.
Special handling is added for response code 429 (Too Many Requests): if the backend sets "x-ratelimit-reset" header, which contains time in seconds to wait, the wait time before retry will be the value returned + rand(0, 0.4) seconds, and only one retry attempt will be made. If the delay is greater than a maximum of 10 seconds, we won’t retry.

# Motivation
If we hit the rate limit imposed by the backend, the "x-ratelimit-reset" header will tell us specifically how long to wait before the next successful request can be made. We want to wait for that amount of time rather than rely on exponential backoff delays (which can be either too short or too long).

# Additional Notes
See [McNulty Rate Limiting](https://datadoghq.atlassian.net/wiki/spaces/API/pages/1349682095/McNulty+Rate+Limiting) for details on how rate limiting is done on the backend


communication/src/main/java/datadog/communication/http/HttpRetryPolicy.java
```diff
@@ -0,0 +1,101 @@
...
+      return true;
+
+    } else {
+      return false;
+    }
+  }
+
+  public long backoff() {
+    long currentDelay = delay;
+    delay *= delayFactor;
```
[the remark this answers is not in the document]
bantonsson: @nikita-tkachenko-datadog Could you write out the casts explicitly to avoid triggering CodeQL warnings?

communication/src/main/java/datadog/communication/http/HttpRetryPolicy.java
```diff
@@ -0,0 +1,101 @@
...
+import java.util.concurrent.ThreadLocalRandom;
+import java.util.concurrent.TimeUnit;
+import javax.annotation.Nullable;
+import javax.annotation.concurrent.NotThreadSafe;
+import org.slf4j.Logger;
+import org.slf4j.LoggerFactory;
+
+/**
+ * A policy which encapsulates retry rules for HTTP calls. Whether to retry and how long to wait
+ * before the next attempt is determined by received HTTP response.
```
bantonsson, line 12: It would be great if the logic described in the PR was added here for clarification.

communication/src/main/java/datadog/communication/http/HttpRetryPolicy.java
```diff
@@ -0,0 +1,101 @@
...
+      return false;
+    }
+
+    int responseCode = response != null ? response.code() : NO_RESPONSE_RECEIVED;
+    if (responseCode == TOO_MANY_REQUESTS_HTTP_CODE) {
+      retriesLeft = 0;
+
+      String waitTime = response.header(X_RATELIMIT_RESET_HTTP_HEADER);
+      if (waitTime == null) {
+        return false;
```
albertvaka, line 49: I'm not 100% sure the backend will always set the `x-ratelimit-reset` when there's a 429 (too many requests), so I think we should handle this case as a regular retry.
```suggestion
      retriesLeft--;
      return true;
```

albertvaka, line 49: I'm not 100% sure the backend will always set the `x-ratelimit-reset` when there's a 429 (too many requests), so I think we should handle this case as a regular retry.
```suggestion
      retriesLeft--;
      return true;
```

albertvaka (dismissed):
Left one suggestion but it looks good to me.

communication/src/main/java/datadog/communication/http/RetryPolicy.java
```diff
@@ -2,7 +2,7 @@
 
 /**
  * Used to define simple retry policies based on exponential backoff mechanism. Typically used for
- * retrying HTTP calls in the DDIntakeApi class.
+ * retrying HTTP calls.
  */
 public class RetryPolicy {
```
bantonsson, line 7: Is this class used anymore?

communication/src/main/java/datadog/communication/http/HttpRetryPolicy.java
bantonsson: Don't forget to move `retriesLeft = 0;` down to after this check then.

dd-trace-core/src/main/java/datadog/trace/common/writer/ddintake/DDIntakeApi.java
```diff
@@ -92,31 +87,29 @@ public DDIntakeApi build() {
...
     }
   }
 
   private final OkHttpClient httpClient;
   private final HttpUrl intakeUrl;
   private final String apiKey;
-  private final RetryPolicy retryPolicy;
+  private final HttpRetryPolicy.Factory retryPolicyFactory;
 
   private DDIntakeApi(
```
albertvaka, line 99: You should also update the `DDEvpProxyApi` class which also uses `RetryPolicy`.

communication/src/main/java/datadog/communication/http/RetryPolicy.java
albertvaka: It is used in `DDEvpProxyApi` but I think we should use the new logic in both places, so we probably don't need this class. Good catch.

communication/src/main/java/datadog/communication/http/HttpRetryPolicy.java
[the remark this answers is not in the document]
nikita-tkachenko-datadog: sure, done

communication/src/main/java/datadog/communication/http/HttpRetryPolicy.java
nikita-tkachenko-datadog: Makes sense, changed the logic to do regular retries if reset time is not provided

communication/src/main/java/datadog/communication/http/HttpRetryPolicy.java
nikita-tkachenko-datadog: I was a bit reluctant to do it at first, since a comment that describes internal implementation details can easily get outdated.
Although maybe in this particular case it's more of a specification.
Added description in the doc.

communication/src/main/java/datadog/communication/http/RetryPolicy.java
nikita-tkachenko-datadog: Ok, definitely reasonable if `DDEvpProxyApi` is talking to the same rate limiter.
Remove old `RetryPolicy` altogether.

dd-trace-core/src/main/java/datadog/trace/common/writer/ddintake/DDIntakeApi.java
nikita-tkachenko-datadog: done!

bantonsson (approved):

albertvaka (approved):

PR #4481, opened 2022-12-29
Add validation for git metadata env vars

nikita-tkachenko-datadog:
# What Does This Do
This change introduces additional checks for [user-supplied git metadata](https://docs.datadoghq.com/continuous_integration/troubleshooting/#data-appears-in-test-runs-but-not-tests).
An error will be logged if the user is passing any git metadata env var but the following conditions are not met:
* `DD_GIT_REPOSITORY_URL` is not set or is empty, and we could not resolve the repo URL automatically either
* `DD_GIT_COMMIT_SHA is not set`, is not a hex string or is shorter than 40 chars, and we could not resolve the SHA automatically either

# Motivation
We allow users to pass git metadata to the tracer in env vars instead of the tracer auto-detecting them.
Some of this git metadata is required, so if the user passes some variables but they don’t pass `DD_GIT_REPOSITORY_URL` and `DD_GIT_COMMIT_SHA`, the backend will silently discard everything else.

This is can be confusing, so we want to log an error if the supplied metadata is not complete.

# Additional Notes
As part of this change `datadog.trace.bootstrap.instrumentation.ci.CIProviderInfo` was split into several classes:
* the logic that detects which CI provider is used was extracted into `datadog.trace.bootstrap.instrumentation.ci.CIProviderInfoFactory`
* the logic that populates tags based on provided metadata was extracted into `datadog.trace.bootstrap.instrumentation.ci.CITagsProvider`
* the logic that detects user-supplied git metadata was extracted into `datadog.trace.bootstrap.instrumentation.ci.git.info.UserSuppliedGitInfoBuilder`
* the logic that detects local git metadata (available in ".git" folder) was extracted into `datadog.trace.bootstrap.instrumentation.ci.git.info.CILocalGitInfoBuilder`
* the the CI-provider-specific logic remained in the children of `datadog.trace.bootstrap.instrumentation.ci.CIProviderInfo` (now changed to interface)

The motivation behind this refactoring was to split a large class into several cohesive parts and to replace Template Method pattern with Strategy pattern, as the latter is more in line with the Single Responsibility Principle and favours composition over inheritance, which results in code that is easier to cover with unit tests.


albertvaka (approved):
LGTM!

internal-api/src/main/java/datadog/trace/bootstrap/instrumentation/ci/git/PersonInfo.java
```diff
@@ -26,13 +26,13 @@ public PersonInfo(final String name, final String email) {
   public PersonInfo(final String name, final String email, final String iso8601date) {
     this.name = name;
     this.email = email;
-    this.ISO8601Date = iso8601date;
+    ISO8601Date = iso8601date;
```
albertvaka, line 29: This field could be renamed to use camelCase.

nikita-tkachenko-datadog: Yup! Checked the [definition of camel case](https://google.github.io/styleguide/javaguide.html#s5.3-camel-case) in Google Java Style Guide (since this is what is used in the project), and acronyms are considered words, so `ISO` should be `iso` according to the standard.
Renamed

smola (approved):

PR #4501, opened 2023-01-04
Support request body interception in servlet 5+

cataphract:
Package names changed in Servlet 5. The code is otherwise the same as for servlet 3.1, so if possible classes are relocated using the shadow plugin.


ygree (approved):
LGTM

smola (approved):

PR #4510, opened 2023-01-09
Add line probe span instrumentation

jpbempel:
# What Does This Do
Uses line range for defining span

# Motivation
Be able to add span in the middle of a method for a specific call/operation

# Additional Notes

### DIFF
--- a/dd-java-agent/agent-debugger/src/main/java/com/datadog/debugger/instrumentation/SpanInstrumentor.java
+++ b/dd-java-agent/agent-debugger/src/main/java/com/datadog/debugger/instrumentation/SpanInstrumentor.java
@@ -5,6 +5,7 @@ import static com.datadog.debugger.instrumentation.Types.DEBUGGER_SPAN_TYPE;
 import static com.datadog.debugger.instrumentation.Types.STRING_TYPE;
 
 import com.datadog.debugger.probe.SpanProbe;
+import com.datadog.debugger.probe.Where;
 import datadog.trace.bootstrap.debugger.DiagnosticMessage;
 import java.util.List;
 import org.objectweb.asm.Opcodes;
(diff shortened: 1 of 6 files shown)


dd-java-agent/agent-debugger/src/test/java/com/datadog/debugger/agent/SpanProbeInstrumentationTest.java
```diff
@@ -51,12 +51,43 @@ public void simpleSpanWithTags() throws IOException, URISyntaxException {
...
+    Class<?> testClass = compileAndLoadClass(CLASS_NAME);
+    int result = Reflect.on(testClass).call("main", "1").get();
+    assertEquals(3, result);
+    assertEquals(1, tracer.spans.size());
+    assertEquals("main-span", tracer.spans.get(0).name);
+    assertTrue(tracer.spans.get(0).isFinished());
+  }
+
+  @Test
+  public void invalidLineSimpleSpan() throws IOException, URISyntaxException {
```
shatzi, line 67: do we have an example of what happen if someone create span probe line range that stop in a middle of a loop block? or a span inside a loop block - does it work to generate multiple spans?

dd-java-agent/agent-debugger/src/test/java/com/datadog/debugger/agent/SpanProbeInstrumentationTest.java
```diff
@@ -116,17 +147,22 @@ private static SpanProbe createSpan(
...
         .build();
   }
 
-  private static SpanProbe createSpan(String id, String spanName, String sourceFile, int line) {
-    return SpanProbe.builder().probeId(id).where(sourceFile, line).name(spanName).build();
+  private static SpanProbe createSpan(
+      String id, String spanName, String sourceFile, int lineFrom, int lineTill, String[] tags) {
+    return SpanProbe.builder()
+        .probeId(id)
+        .where(sourceFile, lineFrom, lineTill)
```
shatzi, line 163: do we want to test multiple line segments as well?

shatzi (approved):
we might have a bug with after-label position. but beside that this looks good.

dd-java-agent/agent-debugger/src/main/java/com/datadog/debugger/instrumentation/SpanInstrumentor.java
```diff
@@ -41,19 +68,43 @@ public void instrument() {
...
+    }
+    if (lineMap.isEmpty()) {
+      reportError("Missing line debug information.");
+      return;
+    }
+    for (Where.SourceLine sourceLine : targetLines) {
+      int from = sourceLine.getFrom();
+      int till = sourceLine.getTill();
+      LabelNode beforeLabel = lineMap.getLineLabel(from);
+      LabelNode afterLabel = lineMap.getLineLabel(till);
```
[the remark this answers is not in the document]
jpbempel: a probe can't be single line, so it's only `till`
where is the bug?

dd-java-agent/agent-debugger/src/test/java/com/datadog/debugger/agent/SpanProbeInstrumentationTest.java
jpbempel: for now, no.

dd-java-agent/agent-debugger/src/test/java/com/datadog/debugger/agent/SpanProbeInstrumentationTest.java
jpbempel: a span inside a loop block, yes it will generate multiple spans

PR #4513, opened 2023-01-09
Handle mapped baggage keys as case sensitive

Xyaren:
# What Does This Do
Solves #4511
During the cleanup of the map, derived from the configuration, the baggage keys were converted to lowercase.
This PR skips the conversion to lowercase strings for baggage keys.

# Motivation
Keep case sensitivity of baggage keys throughout the library.

# Additional Notes

PR #4520, opened 2023-01-10
Consider Baggage when creating new root traces

### LINKED ISSUE #4516: baggage via header config is not honoured in root trace
Continuing the conversation from #3542

in a recent release there's a config to set baggage via headers similar to tags `dd.trace.header.baggage`, however it is not honoured and is dropped on root trace because of following code:

```
TagContext build() {
    if (valid) {
      if (!DDTraceId.ZERO.equals(traceId)) {
        final ExtractedContext context;
        context =
            new ExtractedContext(
                traceId,
                spanId,
                samplingPriority,
                origin,
                endToEndStartTime,
                baggage,
                tags,
                httpHeaders,
                datadogTags);
        return context;
      } else if (origin != null || !tags.isEmpty() || httpHeaders != null) {
        return new TagContext(origin, tags, httpHeaders);
      }
    }
    return null;
  }
```

Same is true for other baggages also, i.e. `ot-baggage-*`

Xyaren:
# What Does This Do
Solves #4516

This PR modifies which extracted information is used when creating a new trace.

# Motivation
Due to Baggage mapping, there now is a possible need to map headers into new created traces.

# Additional Notes
No failing test - If someone can point me to the correct test cases to improve to cover this change please do so.

### DIFF
--- a/dd-trace-core/src/main/java/datadog/trace/core/propagation/ContextInterpreter.java
+++ b/dd-trace-core/src/main/java/datadog/trace/core/propagation/ContextInterpreter.java
@@ -39,6 +39,7 @@ public abstract class ContextInterpreter implements AgentPropagation.KeyClassifi
   protected int samplingPriority;
   protected Map<String, String> tags;
   protected Map<String, String> baggage;
+
   protected String origin;
   protected long endToEndStartTime;
   protected boolean valid;
@@ -257,9 +258,10 @@ public abstract class ContextInterpreter implements AgentPropagation.KeyClassifi
       } else if (origin != null
           || !tags.isEmpty()
           || httpHeaders != null
+          || !baggage.isEmpty()
           || samplingPriority != PrioritySampling.UNSET) {
         return new TagContext(
-            origin, tags, httpHeaders, samplingPriorityOrDefault(samplingPriority));
+            origin, tags, httpHeaders, baggage, samplingPriorityOrDefault(samplingPriority));
       }
     }
     return null;
--- a/dd-java-agent/instrumentation/spring-cloud-zuul-2/src/test/groovy/ProxyRequestHelperTest.groovy
+++ b/dd-java-agent/instrumentation/spring-cloud-zuul-2/src/test/groovy/ProxyRequestHelperTest.groovy
@@ -135,7 +135,7 @@ class ProxyRequestHelperTest extends WithHttpServer<ConfigurableApplicationConte
     }
   }
 
-  def "Test baggage headers to be correctly ignored through proxy"() {
+  def "Test baggage headers to be propagated through proxy"() {
     setup:
     def hsBaggageHeader = "Baggage-test"
     def ddBaggageHeader = "ot-baggage-test"
@@ -148,7 +148,7 @@ class ProxyRequestHelperTest extends WithHttpServer<ConfigurableApplicationConte
       .header(PARENT_ID_HEADER, PARENT_ID)
       .header(TRACE_ID_HEADER, TRACE_ID)
       .header(hsBaggageHeader, "shouldnotappear")
-      .header(ddBaggageHeader, "shouldnotappear")
+      .header(ddBaggageHeader, "shouldAppear")
       .header(nonTracerHeader, "shouldAppear")
       .build()
 
@@ -158,7 +158,7 @@ class ProxyRequestHelperTest extends WithHttpServer<ConfigurableApplicationConte
 
     then:
     !receivedHeaders.containsKey(hsBaggageHeader)
-    !receivedHeaders.containsKey(ddBaggageHeader)
+    receivedHeaders.containsKey(ddBaggageHeader)
     receivedHeaders.containsKey(nonTracerHeader)
 
     receivedHeaders.containsKey(defaultParentHeader)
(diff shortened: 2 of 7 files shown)


dd-trace-core/src/main/java/datadog/trace/core/propagation/ContextInterpreter.java
```diff
@@ -251,7 +251,7 @@ TagContext build() {
           || httpHeaders != null
```
bantonsson, line 251: I think there should be a `|| !baggage.isEmpty()` here as well

bantonsson (approved):
Thanks @Xyaren 

I think you need to re-authenticate with CircleCI or push something to the branch to make CI run again.

Xyaren:
> I think you need to re-authenticate with CircleCI or push something to the branch to make CI run again.

image

What should I do here ?

dd-trace-core/src/main/java/datadog/trace/core/propagation/ContextInterpreter.java
Xyaren: Done 👍 
Missed that.

bantonsson:
@Xyaren from the build number in your image, it looks like CircleCI is enabled on your fork of the repo, and it is trying to build it using your account. The CircleCI setup is a bit messy, but you need to disable CircleCI for your fork ([in CircleCI](https://app.circleci.com/projects/project-dashboard/github/Xyaren/)).

Xyaren:
@bantonsson I "unfollowed" the project, can you trigger the CI again ?

bantonsson:
@Xyaren I'm unfortunately unable to trigger anything, since CircleCI seems to think that you want to run CI on your fork https://github.com/Xyaren/dd-trace-java and not on https://github.com/Datadog/dd-trace-java

If you go to the projects for your GitHub account organization [in CircleCI](https://app.circleci.com/projects/project-dashboard/github/Xyaren/), you should see something like this (for your GitHub user):
Screenshot 2023-01-13 at 15 35 45
Make sure that you disable the `dd-trace-java` project there by clicking the `...` on the `dd-trace-java` line (mine is greyed out since I haven't enabled CircleCI for my fork).

Xyaren:
image

That's how it looks like already.
I'll try to re-trigger by adding a new commit.

Xyaren:
Rebased.

bantonsson:
@Xyaren there is one code formatting issue. Running `./gradlew formatCode` will fix it.

Xyaren:
> @Xyaren there is one code formatting issue. Running `./gradlew formatCode` will fix it.

Missed that, sorry for the delay :)
Fixed.

Xyaren:
@bantonsson I really struggle to find the test failures in the CI Logs.
Some ci steps seem to require an API Key ?

bantonsson:
@Xyaren sorry about that.
* You can ignore all the ones named `system-tests-*` since they need an API key secret that is not available to outside contributors.
* If you see several jobs named with a similar pattern, i.e. `z_test_??_base`, then there is probably an issue in that category of tests.

There are two failing tests here:
* In base tests. [HaystackHttpExtractorTest](https://app.circleci.com/pipelines/github/DataDog/dd-trace-java/20829/workflows/bf26cf7b-c5da-4456-9532-776dd27a570b/jobs/549210)
* In instrumentation tests. [ProxyRequestHelperTest](https://app.circleci.com/pipelines/github/DataDog/dd-trace-java/20829/workflows/bf26cf7b-c5da-4456-9532-776dd27a570b/jobs/549230)

Xyaren:
I see.
The Problem is around the Haystack Span,Trace and Parent Id, which is handled as baggage.

https://github.com/Xyaren/dd-trace-java/blob/c99bec69e0a7f72ecc427e0ecb19c877e8bacf81/dd-trace-core/src/main/java/datadog/trace/core/propagation/HaystackHttpCodec.java#L204-L214

This is only in the special case, when the haystack trace Id is zero, as this would normally result in no trace being created.
Due to the addition of haystack trace (and span) ids to the baggage map, this would result in a new context despite a `0` trace id.

I feel like this would still qualify for creating a new trace context including the extracted baggage of the zero-only haystack ids. Thats in fact what was correctly extracted, so why not keep it and propagate it ?

Xyaren:
`ProxyRequestHelperTest#Test baggage headers to be correctly ignored through proxy`
Is probably obsolete after this PR, as it tests exactly the opposite of what this PR is trying to achieve: Creating a new context when baggage is received.

My questions are:
- Does a received trace id of 0 represent a special case ? If so, what is it ?
- Should ot-* baggage be accepted next to configured mapped baggage as well?

bantonsson:
@Xyaren If we receive a traceId of `0` we interpret that as no ingoing trace, so we will create a new Trace starting here. I think it's fine to always propagate the baggage, and you don't need to split up the _different_ baggage types. The test makes the wrong assumptions and instead it should check that the baggage contains the expected items, something like this:

```
    ...
    baggage = [ ... ]

    when:
    final TagContext context = extractor.extract(headers, ContextVisitors.stringValuesMap())

    then:
    if (expectedTraceId) {
      ExtractedContext ec = context as ExtractedContext
      assert ec.traceId == expectedTraceId
      assert ec.spanId == expectedSpanId
    }
    context.baggageItems() == baggage
    ...
```

Xyaren:
@bantonsson Thank you for the explanation.
I decided to split the test up and added one explicitly for the baggage propagation.

bantonsson (approved):

bantonsson:
@Xyaren I think you need to update this PR since there is now a conflict with your changes from #4513

Xyaren:
Rebased & Squashed 👍

PR #4522, opened 2023-01-10
AppSec events tracking API

ValentinZakharov:
# What Does This Do
Implemented Application Security Tracking API, that allows customer to send login/custom events linked to a span.

# Motivation

# Additional Notes


dd-trace-api/src/main/java/datadog/trace/api/EventTracker.java
```diff
@@ -0,0 +1,74 @@
...
+
+    private final InternalTracer tracer;
+
+    EventTracker(InternalTracer tracer) {
+        this.tracer = tracer;
+    }
+
+    /**
+     * Method allowing to track whether a user login was successful or not. A user login event
+     * is made of a user id, along and an optional key-value map of metadata of string types only.
```
bantonsson, line 19:
```
* Method for tracking whether a user login was successful or not. A user login event
* is made up of a user id and an optional metadata key-value map of strings.
```

dd-trace-api/src/main/java/datadog/trace/api/EventTracker.java
```diff
@@ -0,0 +1,74 @@
...
+        if (segment == null) {
+            return;
+        }
+
+        if (success) {
+            segment.setTagTop("appsec.events.users.login.success.track", true);
+            segment.setTagTop("usr.id", userId);
+            metadata.forEach((k, v) -> {
+                segment.setTagTop("appsec.events.users.login.success."+k, v);
+            });
```
bantonsson, line 40: I know that there is no proper way to send structured data to the backend, but this is just horribly inefficient. The `setDataTop` method accepts a `JSON` string, which would be way more space efficient than exploding this into multiple tags. The only issue I can see is that the current implementation wouldn't allow the use of the `appsec` key from both this place and the AppSec gateway bridge. I also can't remember what the backend does when it handles those special json tags. Might be worth looking into.

bantonsson (commented):
Looks god in general

dd-trace-api/src/main/java/datadog/trace/api/internal/TraceSegment.java
```diff
@@ -1,4 +1,4 @@
-package datadog.trace.api;
+package datadog.trace.api.internal;
```
smola, line 1: This is not backwards compatible. Is this intended?

bantonsson: Yes, nobody could ever access an instance of that anyway.

dd-trace-api/src/main/java/datadog/trace/api/EventTracker.java
dougqh: Yeah - I'm also not crazy about putting each key individually.  I'd prefer to see bulk put operation along the lines of setTagsTop.  I also don't love the forEach to loop over the map or what I'm assuming is repeated constructing of the same keys again and again.

ValentinZakharov: I've added bulk put operation (allows to set map of values as tag) with later transformation into flatmap in background thread.

dd-trace-api/src/main/java/datadog/trace/api/EventTracker.java
ValentinZakharov: Thanks, fixed

dd-trace-api/src/main/java/datadog/trace/api/EventTracker.java
ValentinZakharov: Unfortunately `JSON` via `setDataTop` is not supported by backend event handler right now and will take time to implement according to priorities. So I had to skip this proposal for now.

dd-trace-core/src/main/java/datadog/trace/core/DDSpanContext.java
```diff
@@ -703,4 +703,49 @@ private DDSpanContext getTopContext() {
...
+   *    "plain"           ->  "123"
+   *
+   * @param tags    map of tags that can contain sub-maps as values
+   * @return        flat map, after transformation
+   */
+  private Map<String, Object> flatTags(Map<String, Object> tags) {
+    Map<String, Object> extractedTags = new HashMap<>();
+    tags.entrySet().removeIf(entry -> extractValues(entry.getKey(), entry.getValue(), extractedTags));
+    tags.putAll(extractedTags);
+    return tags;
```
bantonsson, line 733: This is way to costly to do on every span even if nothing is _extracted_ and _added_. One option would be to keep track of if there is a _flatable_ type in the tags and only do the flattening if that is the case.

ValentinZakharov:
Did another implementation - directly writing tags during serialisation to avoid any overhead while using maps for bulk tagging

dd-trace-api/src/main/java/datadog/trace/api/EventTracker.java
```diff
@@ -0,0 +1,86 @@
...
+    if (tracer == null) {
+      return;
+    }
+
+    TraceSegment segment = this.tracer.getTraceSegment();
+    if (segment == null) {
+      return;
+    }
+
+    segment.setTagTop("appsec.events." + eventName + ".track", true);
```
cataphract, line 82: I think this should use TagsHelper::sanitize

PR #4523, opened 2023-01-11
Add javac compiler plugin to inject source path into classes

nikita-tkachenko-datadog:
# What Does This Do
This change adds a javac compiler plugin that is supposed to be used by CI visibility clients in combination with the java tracer.
The plugin injects into every compiled class the path to that class' source code (the injection is done by adding a private static field).

# Motivation
CI visibility logic needs to establish correspondence between a test class and its source code file.
For now this is needed in order to determine owners of specific tests (by consulting the CODEOWNERS file), but later it might be used for other purposes, such as determining code coverage per test-case.

# Additional Notes
* This change only adds the plugin. The tracer logic that uses injected data will be added as a separate PR.
* Eclipse JDT compiler does not support plugins, so a separate fallback solution will be developed for the clients who use it

### DIFF
--- a/.github/CODEOWNERS
+++ b/.github/CODEOWNERS
@@ -21,6 +21,7 @@ dd-trace-core/src/main/java/datadog/trace/civisibility/                @DataDog/
 dd-java-agent/instrumentation/junit-4.10/                              @DataDog/ci-app-libraries-java
 dd-java-agent/instrumentation/junit-5.3/                               @DataDog/ci-app-libraries-java
 dd-java-agent/instrumentation/testng-6.4/                              @DataDog/ci-app-libraries-java
+dd-java-compiler-plugin/                                               @DataDog/ci-app-libraries-java
 
 # @DataDog/debugger-java (Live Debugger)
 dd-java-agent/agent-debugger/ @DataDog/debugger-java
--- a/dd-java-compiler-plugin/README.md
+++ b/dd-java-compiler-plugin/README.md
@@ -1,0 +1,126 @@
+# Datadog Java Compiler Plugin
+
+This is a Java compiler (`javac`) [plugin](https://openjdk.org/groups/compiler/processing-code.html#plugin) that
+augments compiled classes with some additional data used by Datadog products.
+
+List of things that the plugin does:
+
+- Inject into every class a `private static final` field with the path to the class' source code
+  (this data is being used by [Datadog's CI Visibility](https://www.datadoghq.com/product/ci-cd-monitoring/)).
+
+## Configuration
+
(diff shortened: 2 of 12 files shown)


albertvaka:
If this plugin becomes a separate artifact with its own release cycle I'm also inclined to move it to its own repo.

However, is there any reason not to make it part of `dd-java-agent` instead of a new artifact?

The main advantage would be that customers won't need to add two dependencies to have this functionality. 

I'm not sure if there's any downside (maybe that we would load as a plugin a bunch of code that isn't actually used?)

bantonsson:
@albertvaka This compiler plugin is its own dependency anyway (as can be seen in the `README.md`). It's not baked into the `dd-java-agent`, and has nothing to do with the release cycle of the java tracer.

albertvaka:
We could change that and make it part of `dd-java-agent`, though: the same artifact could be a Java agent and a compiler plugin, with different entry points. Any thoughts on this idea? Is it too crazy? 😂

nikita-tkachenko-datadog:
I have done some checks, and can confirm that this approach works.

The advantages are that the clients will have to declare only one dependency, and that the plugin will be able to leverage the tracer's CI infrastructure (jobs, gradle logic that facilitates testing with different JVMs, etc)

The disadvantage is that we would have two separate entities bundled together in one artifact. I can't, however, think of any scenarios where this would have any tangible adverse effects.

What do you think? Am I missing something?

PerfectSlayer:
I would be in favor of use a separate repository. Datadog org already has 2.5k repo, it should not hurt to create a new dedicated one.

As Bjorn said, having no relation with the agent will cause issue laters to release versions. Moreover, it will prevent to clutter the repo.

For example, when I created the GitHub action dedicated to the `dd-trace-java` release artifact, I created [a dedicated repository](https://github.com/DataDog/download-release-action), only for this script, even if it is only used here. The same thing applies for [our Docker build image](https://github.com/DataDog/dd-trace-java-docker-build) for example.

jbachorik:
> the same artifact could be a Java agent and a compiler plugin, with different entry points. Any thoughts on this idea? Is it too crazy?

Sorry, this cumulation of unrelated responsibilities sounds like a recipe for maintenance nightmare :(

Also, me as a user would definitely prefer to be able to decide whether I want to apply a javac plugin without also removing the whole agent dependency.

And last but not least, having this as a separate open source repo would be nice publicity - it could be used even when not using the datadog java agent at all.

Just my 2c here ...

nikita-tkachenko-datadog:
The comments are reasonable, I will be moving the plugin to its own repo then

PR #4525, opened 2023-01-11
Set error in instrumented span on exception

jpbempel:
# What Does This Do
When exception is thrown during the instrumented span, we store the throwable inside the span for reporting it as error span.

# Motivation

# Additional Notes


cimi (approved):
:+1: Thanks!

Issue #4527, opened 2023-01-12
Failed to determine dependency


kworkbee:
One of my applications is not running normally because it does not have 'servlet-api-2.5.jar' dependency.
This is a WebFlux application based on Spring Boot 2.7.7, and I would like to ask you to check if there are compatibility issues or if additional configuration settings are required.

Log trace below:
```
OpenJDK 64-Bit Server VM warning: Sharing is only supported for boot loader classes because bootstrap classpath has been appended
[dd.trace 2023-01-12 13:28:40:936 +0900] [main] INFO com.datadog.appsec.AppSecSystem - AppSec is ENABLED_INACTIVE with powerwaf(libddwaf: 1.5.1) no rules loaded
[dd.trace 2023-01-12 13:28:41:125 +0900] [dd-task-scheduler] INFO datadog.trace.agent.core.StatusLogger - DATADOG TRACER CONFIGURATION {"version":"1.3.0~ac8666aa84","os_name":"Linux","os_version":"5.4.219-126.411.amzn2.x86_64","architecture":"amd64","lang":"jvm","lang_version":"17.0.5","jvm_vendor":"Eclipse Adoptium","jvm_version":"17.0.5+8","java_class_version":"61.0","http_nonProxyHosts":"null","http_proxyHost":"null","enabled":true,"service":"eurekube-operator","agent_url":"http://10.192.49.90:8126","agent_error":false,"debug":false,"analytics_enabled":false,"sampling_rules":[{},{}],"priority_sampling_enabled":true,"logs_correlation_enabled":true,"profiling_enabled":true,"remote_config_enabled":true,"debugger_enabled":false,"appsec_enabled":"ENABLED_INACTIVE","telemetry_enabled":true,"dd_version":"","health_checks_enabled":true,"configuration_file":"no config file present","runtime_id":"76b27885-3402-42fd-9bf9-20d8e7a0fff8","logging_settings":{"levelInBrackets":false,"dateTimeFormat":"'[dd.trace 'yyyy-MM-dd HH:mm:ss:SSS Z']'","logFile":"System.err","configurationFile":"simplelogger.properties","showShortLogName":false,"showDateTime":true,"showLogName":true,"showThreadName":true,"defaultLogLevel":"INFO","warnLevelString":"WARN","embedException":false},"cws_enabled":false,"cws_tls_refresh":5000}

  .   ____          _            __ _ _
 /\\ / ___'_ __ _ _(_)_ __  __ _ \ \ \ \
[...]

PR #4530, opened 2023-01-12
Fix issue resolving dependency with empty artifactId

manuel-alvarez-alvarez:
# What Does This Do
Fixes issue in dependency resolution when artifactId is empty

# Motivation

# Additional Notes
See https://github.com/DataDog/dd-trace-java/issues/4527

### DIFF
--- a/telemetry/src/main/java/datadog/telemetry/dependency/DependencyResolver.java
+++ b/telemetry/src/main/java/datadog/telemetry/dependency/DependencyResolver.java
@@ -44,7 +44,7 @@ public class DependencyResolver {
         }
       }
     } catch (RuntimeException rte) {
-      log.warn("Failed to determine dependency for uri {}", uri, rte);
+      log.debug("Failed to determine dependency for uri {}", uri, rte);
     }
     // TODO : moving jboss vfs here is probably a idea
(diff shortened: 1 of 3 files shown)


smola (approved):

bantonsson (approved):

telemetry/src/main/java/datadog/telemetry/dependency/DependencyResolver.java
```diff
@@ -44,7 +44,7 @@ static List<Dependency> extractDependenciesFromURI(URI uri) {
         }
       }
     } catch (RuntimeException rte) {
-      log.warn("Failed to determine dependency for uri {}", uri, rte);
+      log.debug("Failed to determine dependency for uri {}", uri, rte);
```
ValentinZakharov, line 47: Would be nice to report such internal errors via telemetry, when it will be ready

ValentinZakharov (approved):

telemetry/src/main/java/datadog/telemetry/dependency/DependencyResolver.java
manuel-alvarez-alvarez: Yes, I discussed it with Santiago and that's exactly the idea.

PR #4533, opened 2023-01-12
fix and log async-profiler activation

richardstartin:
# What Does This Do

Owing to complexities in configuring the profiler, we haven't been actually enabling async-profiler when we could have been since 1.3.0. This really activates it.

# Motivation

# Additional Notes

### DIFF
--- a/dd-java-agent/agent-profiling/profiling-auxiliary/src/main/java/com/datadog/profiling/auxiliary/AuxiliaryProfiler.java
+++ b/dd-java-agent/agent-profiling/profiling-auxiliary/src/main/java/com/datadog/profiling/auxiliary/AuxiliaryProfiler.java
@@ -1,5 +1,7 @@
 package com.datadog.profiling.auxiliary;
 
+import static datadog.trace.api.Config.isAsyncProfilerSafeInCurrentEnvironment;
+
 import com.datadog.profiling.controller.OngoingRecording;
 import com.datadog.profiling.utils.ProfilingMode;
 import datadog.trace.api.Platform;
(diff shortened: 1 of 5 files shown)


jbachorik (approved):

jbachorik (approved):

dd-java-agent/agent-profiling/profiling-auxiliary/src/main/java/com/datadog/profiling/auxiliary/AuxiliaryProfiler.java
```diff
@@ -41,7 +43,7 @@ private AuxiliaryProfiler() {
         Platform.isLinux()
                 && configProvider.getBoolean(
                     ProfilingConfig.PROFILING_ASYNC_ENABLED,
-                    ProfilingConfig.PROFILING_ASYNC_ENABLED_DEFAULT)
```
PerfectSlayer, line 44: This call should be the same as the result of `Config.isAsyncProfilerEnabled()` isn't it?

internal-api/src/main/java/datadog/trace/api/Config.java
```diff
@@ -1792,13 +1792,12 @@ public boolean isAsyncProfilerEnabled() {
     return isAsyncProfilerEnabled;
   }
 
-  private static boolean isAsyncProfilerSafeInCurrentEnvironment() {
+  public static boolean isAsyncProfilerSafeInCurrentEnvironment() {
```
PerfectSlayer, line 1795: If if is, we don't have to change scope.

internal-api/src/main/java/datadog/trace/api/Config.java
```diff
@@ -1792,13 +1792,12 @@ public boolean isAsyncProfilerEnabled() {
...
     // don't want to put this logic (which will evolve) in the public ProfilingConfig, and can't
     // access Platform there
-    if (Platform.isJ9()) {
-      return true;
-    }
-    return Platform.isJavaVersionAtLeast(18)
+    // don't want to put this logic (which will evolve) in the public ProfilingConfig, and can't
+    // access Platform there
```
PerfectSlayer, line 1799: Duplicate comment

PerfectSlayer (commented):
I was reviewing the PR but it was merged before I had the time to submit.

Can you have a look to my few questions? Thanks! 🙏

PR #4535, opened 2023-01-12
Fix log probe conditions

jpbempel:
# What Does This Do
conditions for log probes were not passed to resolved ProbeDetails in snapshot, so no evaluation of the condition.

# Motivation
No condition evaluation for Log Probes

# Additional Notes
A smoke test for condition evaluation should be added to avoid regression on this part but requires a significant effort to implement (Json serialization of the condition configuration). Will do in a next PR.


cimi (approved):
:+1: Thanks!

PR #4536, opened 2023-01-12
migrate from async-profiler to Datadog java profiler

richardstartin:
# What Does This Do

Migrates profiling integration to [Datadog Java profiler](https://github.com/DataDog/java-profiler) now we have diverged from async-profiler.

# Motivation

# Additional Notes


dd-java-agent/benchmark/build.gradle
```diff
@@ -17,7 +17,7 @@ jmh {
   fork = 1 // How many times to forks a single benchmark. Use 0 to disable forking altogether
   jvmArgs = ["-Ddd.jmxfetch.enabled=false", "-Ddd.writer.type=LoggingWriter"]
   //  jvmArgs += ["-XX:+UnlockDiagnosticVMOptions", "-XX:+DebugNonSafepoints", "-XX:StartFlightRecording=delay=5s,dumponexit=true,name=jmh-benchmark,filename=$rootDir/dd-java-agent/benchmark/build/reports/jmh/jmh-benchmark.jfr"]
-  //  jvmArgs += ["-agentpath:$rootDir/dd-java-agent/benchmark/src/jmh/resources/libasyncProfiler.so=start,collapsed,file=$rootDir/dd-java-agent/benchmark/build/reports/jmh/profiler.txt".toString()]
+  //  jvmArgs += ["-agentpath:$rootDir/dd-java-agent/benchmark/src/jmh/resources/libjavaProfiler.so=start,collapsed,file=$rootDir/dd-java-agent/benchmark/build/reports/jmh/profiler.txt".toString()]
```
jbachorik, line 20: This should stay `libasyncProfiler.so` - it is a proper async-profiler library used by the JMH

dd-java-agent/benchmark/build.gradle
```diff
@@ -43,7 +43,7 @@ tasks.named('jmh').configure {
 }
 
 /*
- If using libasyncProfiler, use the following to generate nice svg flamegraphs.
+ If using libjavaProfiler, use the following to generate nice svg flamegraphs.
```
jbachorik, line 46: This should be `libasyncProfiler` - in `libjavaProfiler` we are not supporting collapsed output

dd-trace-api/src/main/java/datadog/trace/api/config/ProfilingConfig.java
```diff
@@ -58,47 +58,56 @@ public final class ProfilingConfig {
...
-  public static final String PROFILING_ASYNC_CSTACK = "profiling.async.cstack";
-  public static final String PROFILING_ASYNC_CSTACK_DEFAULT = "no";
-  public static final String PROFILING_ASYNC_SAFEMODE = "profiling.async.safemode";
-  public static final int PROFILING_ASYNC_SAFEMODE_DEFAULT = 12; // POP_METHOD|UNWIND_NATIVE
-  public static final String PROFILING_ASYNC_MEMLEAK_ENABLED = "profiling.async.memleak.enabled";
-  public static final boolean PROFILING_ASYNC_MEMLEAK_ENABLED_DEFAULT = false;
-  public static final String PROFILING_ASYNC_MEMLEAK_INTERVAL = "profiling.async.memleak.interval";
-  public static final String PROFILING_ASYNC_MEMLEAK_CAPACITY = "profiling.async.memleak.capacity";
-  public static final int PROFILING_ASYNC_MEMLEAK_CAPACITY_DEFAULT = 1024;
+  public static final String PROFILING_DATADOG_PROFILER_LIBPATH = "profiling.ddprof.lib";
```
jbachorik, line 67: I wonder - should we rename this to `profiling.ddprof.debug.lib`  or similar to make it explicit that this is not something the users should casually do?

dd-java-agent/agent-profiling/profiling-auxiliary/src/main/java/com/datadog/profiling/auxiliary/AuxiliaryProfiler.java
```diff
@@ -42,9 +42,9 @@ private AuxiliaryProfiler() {
     String auxilliaryType =
         Platform.isLinux()
                 && configProvider.getBoolean(
-                    ProfilingConfig.PROFILING_ASYNC_ENABLED,
-                    isAsyncProfilerSafeInCurrentEnvironment())
-            ? "async"
+                    ProfilingConfig.PROFILING_DATADOG_PROFILER_ENABLED,
+                    isDatadogProfilerSafeInCurrentEnvironment())
```
jbachorik, line 46: This should probably keep the `async-profiler` version as an alias for for backward compatibility.

Something like 
```suggestion
                    ProfilingConfig.PROFILING_DATADOG_PROFILER_ENABLED,
                    isDatadogProfilerSafeInCurrentEnvironment(),
                    ProfilingConfig.PROFILING_DEPRECATED_ASYNC_ENABLED)
```

dd-java-agent/agent-profiling/profiling-controller/src/main/java/com/datadog/profiling/controller/ProfilerSettingsSupport.java
```diff
@@ -79,8 +79,8 @@ protected ProfilerSettingsSupport() {
         !"no"
             .equalsIgnoreCase(
                 configProvider.getString(
-                    ProfilingConfig.PROFILING_ASYNC_CSTACK,
-                    ProfilingConfig.PROFILING_ASYNC_CSTACK_DEFAULT));
+                    ProfilingConfig.PROFILING_DATADOG_PROFILER_CSTACK,
+                    ProfilingConfig.PROFILING_DATADOG_PROFILER_CSTACK_DEFAULT));
```
jbachorik, line 83: Should we also add the fallback to the previous config key here?
I would hesitate a bit since this is not something we were telling the users to set and in the worst case they will get the default native stack unwinding, which is 'off' ...

dd-trace-api/src/main/java/datadog/trace/api/config/ProfilingConfig.java
richardstartin: good idea

dd-java-agent/agent-profiling/profiling-controller/src/main/java/com/datadog/profiling/controller/ProfilerSettingsSupport.java
richardstartin: Yes, added

dd-java-agent/agent-profiling/profiling-auxiliary/src/main/java/com/datadog/profiling/auxiliary/AuxiliaryProfiler.java
richardstartin: yes, added back

jbachorik (approved):

PR #4560, opened 2023-01-18
Add telemetry metrics to IAST

manuel-alvarez-alvarez:
# What Does This Do
This PR makes use [Telemetry API generate-metrycs payload](https://github.com/DataDog/instrumentation-telemetry-api-docs/blob/main/GeneratedDocumentation/ApiDocs/v1/producing-telemetry.md#generate-metrics) to send metrics required for IAST, in order to do so, the `call-site-instrumentation-plugin` has been extended with the interface `datadog.trace.plugin.csi.Extension` and a new extension has been provided for IAST that:

- Creates a version of the call sites including telemetry code:
```java
    public static String afterGetParameter(ServletRequest self, String paramName, String paramValue) {
        IastTelemetryCollector.add(IastMetric.EXECUTED_SOURCE, SourceTypes.REQUEST_PARAMETER_VALUE);
...
    }
```

- Modifies the advice in order to switch between the call sites and send metrics about the instrumentation process:
```java
    @Override
    public void apply(MethodHandler handler, int opcode, String owner, String name, String descriptor, boolean isInterface) {
        if (this.telemetry) {
            IastTelemetryCollector.add(IastMetric.INSTRUMENTED_SOURCE, SourceTypes.REQUEST_PARAMETER_VALUE);
        }
...
    }

    @Override
    public void enableTelemetry(boolean enableRuntime) {
        this.telemetry = true;
        if (enableRuntime) {
            this.callSite = "datadog/trace/instrumentation/servlet/request/ServletRequestCallSiteWithTelemetry";
            this.helperClassNames = new String[] { "datadog.trace.instrumentation.servlet.request.ServletRequestCallSiteWithTelemetry" };
        }
    }
```

# Motivation
From IAST we need to be aware of the call sites instrumented by the tracer in a very similar way to the [integrations payload](https://github.com/DataDog/instrumentation-telemetry-api-docs/blob/main/GeneratedDocumentation/ApiDocs/v1/SchemaDocumentation/Schemas/integration.md). We also want to be notified when the call sites are executed to be sure that we are not missing any sources / sinks during the execution of requests.

# Additional Notes


buildSrc/call-site-instrumentation-plugin/src/main/java/datadog/trace/plugin/csi/impl/AsmSpecificationBuilder.java
```diff
@@ -1,54 +1,20 @@
...
-import static datadog.trace.plugin.csi.util.CallSiteConstants.AROUND_ARRAY_ANNOTATION;
-import static datadog.trace.plugin.csi.util.CallSiteConstants.ASM_API_VERSION;
-import static datadog.trace.plugin.csi.util.CallSiteConstants.BEFORE_ANNOTATION;
-import static datadog.trace.plugin.csi.util.CallSiteConstants.BEFORE_ARRAY_ANNOTATION;
-import static datadog.trace.plugin.csi.util.CallSiteConstants.CALL_SITE_ADVICE_CLASS;
-import static datadog.trace.plugin.csi.util.CallSiteConstants.CALL_SITE_ANNOTATION;
-import static datadog.trace.plugin.csi.util.CallSiteConstants.INVOKE_DYNAMIC_CONSTANTS_ANNOTATION;
-import static datadog.trace.plugin.csi.util.CallSiteConstants.RETURN_ANNOTATION;
-import static datadog.trace.plugin.csi.util.CallSiteConstants.THIS_ANNOTATION;
+import static datadog.trace.plugin.csi.util.CallSiteConstants.*;
```
smola: I think this goes against our guidelines: https://github.com/DataDog/dd-trace-java/blob/master/CONTRIBUTING.md#intellij-idea

dd-java-agent/agent-iast/src/main/java/com/datadog/iast/telemetry/taint/TaintedObjectsWithTelemetry.java
```diff
@@ -0,0 +1,90 @@
...
+import static datadog.trace.api.iast.telemetry.IastMetrics.*;
+
+import com.datadog.iast.model.Range;
+import com.datadog.iast.model.Source;
+import com.datadog.iast.taint.TaintedObject;
+import com.datadog.iast.taint.TaintedObjects;
+import datadog.trace.api.iast.telemetry.Verbosity;
+import javax.annotation.Nonnull;
+
+public abstract class TaintedObjectsWithTelemetry {
```
smola, line 12: As much as separation of concerns here is nice, we might want to eventually add this to `TaintedObjects` for performance. But we can defer that until it's measured and profiled.

dd-java-agent/agent-iast/src/main/java/com/datadog/iast/telemetry/taint/TaintedObjectsWithTelemetry.java
```diff
@@ -0,0 +1,90 @@
...
+      return delegate.get(obj);
+    }
+
+    @Override
+    public void release() {
+      delegate.release();
+      if (delegate.isFlat()) {
+        TAINTED_FLAT_MODE.add(1);
+      } else {
+        REQUEST_TAINTED.add(delegate.getEstimatedSize());
```
smola, line 55: Good point that we only bump estimated size if not in flat mode.

buildSrc/call-site-instrumentation-plugin/src/main/java/datadog/trace/plugin/csi/impl/AsmSpecificationBuilder.java
manuel-alvarez-alvarez: Yes, I changed that, don't know why this configuration got removed from my IntelliJ

dd-java-agent/agent-iast/src/main/java/com/datadog/iast/telemetry/taint/TaintedObjectsWithTelemetry.java
manuel-alvarez-alvarez: We could also extend instead of compose, but I don't think that taint operations will be so massive that it really makes a huge difference. But like you said, we have to measure and then make a decision.

internal-api/build.gradle
```diff
@@ -110,6 +110,8 @@ dependencies {
   api deps.slf4j
   api project(":utils:time-utils")
 
+  implementation deps.jctools
```
manuel-alvarez-alvarez, line 113: I am a bit unsure if it's OK to add this dependency to the `internal-api` module

smola (approved):

jandro996 (approved):

dd-trace-api/src/main/java/datadog/trace/api/config/GeneralConfig.java
```diff
@@ -58,6 +58,7 @@ public final class GeneralConfig {
   public static final String TELEMETRY_HEARTBEAT_INTERVAL = "telemetry.heartbeat.interval";
   public static final String TELEMETRY_DEPENDENCY_COLLECTION_ENABLED =
       "telemetry.dependency-collection.enabled";
+  public static final String TELEMETRY_DEBUG_ENABLED = "instrumentation.telemetry.debug.enabled";
```
smola, line 61: @manuel-alvarez-alvarez We probably need to rename this for:
* Dropping the `instrumentation.` part of the prefix.
* Possibly making something more specific, since there are side discussions for a debug flag for logs with pretty different purpose.

smola (changes requested):

dd-trace-api/src/main/java/datadog/trace/api/config/GeneralConfig.java
manuel-alvarez-alvarez: Yes, I've removed the variable for the time being, we can come back to it in the future if needed.

smola (approved):

PR #4563, opened 2023-01-18
Add client span without context health metric

PerfectSlayer:
# What Does This Do

Add a new health metric to track client (span type) spans without context (ie without parent span).

# Motivation

Have a better understanding of incomplete traces.

# Additional Notes

I also included a small refactoring commit for readability (one method extraction). I can remove it if needed.

### DIFF
--- a/dd-trace-core/src/main/java/datadog/trace/core/monitor/TracerHealthMetrics.java
+++ b/dd-trace-core/src/main/java/datadog/trace/core/monitor/TracerHealthMetrics.java
@@ -135,10 +135,9 @@ public class TracerHealthMetrics extends HealthMetrics implements AutoCloseable 
 
   private void checkForClientSpansWithoutContext(final List<DDSpan> trace) {
     for (DDSpan span : trace) {
-      if (span != null) {
-        long parentId = span.getParentId();
-        String tag = span.getTag(SPAN_KIND, "undefined");
-        if (parentId == ZERO && SPAN_KIND_CLIENT.equals(tag)) {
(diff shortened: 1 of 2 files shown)


bantonsson (commented):
Looks good. Only one small comment.

dd-trace-core/src/main/java/datadog/trace/core/monitor/TracerHealthMetrics.java
```diff
@@ -125,6 +130,19 @@ public void onPublish(final List<DDSpan> trace, final int samplingPriority) {
...
     enqueuedSpans.inc(trace.size());
+    checkForClientSpansWithoutContext(trace);
+  }
+
+  private void checkForClientSpansWithoutContext(final List<DDSpan> trace) {
+    for (DDSpan span : trace) {
+      if (span != null) {
+        long parentId = span.getParentId();
+        String tag = span.getTag(SPAN_KIND, "undefined");
+        if (parentId == ZERO && SPAN_KIND_CLIENT.equals(tag)) {
```
bantonsson, line 141: I would move the `span.getTag(...)` to after `parentId == ZERO` to only do it when necessary.

bantonsson (commented):
And now with the comment...

dd-trace-core/src/main/java/datadog/trace/core/monitor/TracerHealthMetrics.java
PerfectSlayer: Agreed. I patched it and will squash the commits.

PR #4565, opened 2023-01-18
implement profiler context tagging API

richardstartin:
# What Does This Do

Let's a user register arbitrary attributes for profiling context by setting:
`-Ddd.profiling.experimental.context.attributes=foo,bar`
Which then allows the user to call 
`ProfilingContext.get().setContextValue("foo", stringValue);`
and have the tags "foo=`stringValue`" appended to each JFR event emitted.

# Motivation

# Additional Notes

### DIFF
--- a/dd-trace-api/src/main/java/datadog/trace/api/config/ProfilingConfig.java
+++ b/dd-trace-api/src/main/java/datadog/trace/api/config/ProfilingConfig.java
@@ -159,5 +159,8 @@ public final class ProfilingConfig {
 
   public static final String PROFILING_DEBUG_DUMP_PATH = "profiling.debug.dump_path";
 
+  public static final String PROFILING_CONTEXT_ATTRIBUTES =
+      "profiling.experimental.context.attributes";
+
   private ProfilingConfig() {}
 }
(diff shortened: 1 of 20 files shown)


dd-trace-api/src/main/java/datadog/trace/api/config/ProfilingConfig.java
```diff
@@ -159,5 +159,7 @@ public final class ProfilingConfig {
 
   public static final String PROFILING_DEBUG_DUMP_PATH = "profiling.debug.dump_path";
 
+  public static final String PROFILING_CONTEXT_ATTRIBUTES = "profiling.context.attributes";
```
jbachorik, line 162: Could this be an experimental key for now?
The registration process might possibly change in the future, correct?

richardstartin: yes, it should be experimental, thanks.

richardstartin: Changed.

richardstartin:
I think we need to merge https://github.com/DataDog/java-profiler-library/pull/46 before this, because the API is a little awkward to use. I'd like to restore some of the modularity we previously broke for the sake of saving a thread-local lookup for the tid (not a good trade off).

bantonsson (approved):
Looks good from the tracer side

PR #4566, opened 2023-01-19
[system-tests poc] upload tracer library to ghcr.io

robertomonteromiguel:
# What Does This Do

Create base image for java tracer and upload to docker registry. It is a POC

# Motivation

Try to standardize release and snapshot publishing. This could be applicable to other language tracers.
This would simplify, in our case, the testing process from system-tests

# Additional Notes


devinsba:
The lib-injection image might already accomplish some of what you are looking for: https://github.com/orgs/DataDog/packages/container/package/dd-trace-java%2Fdd-lib-java-init

PR #4568, opened 2023-01-19
Add report of index expression error

jpbempel:
# What Does This Do
When accessing collection through index expression, we can now report runtime error like index out of bounds with an explicit expression that failed to be resolved.
example:
`index[10] out of bounds: [0-3]`

# Motivation
Gives explicit feedback in case of runtime error evaluating conditions

# Additional Notes

### DIFF
--- a/dd-java-agent/agent-debugger/debugger-el/src/main/java/com/datadog/debugger/el/ValueScript.java
+++ b/dd-java-agent/agent-debugger/debugger-el/src/main/java/com/datadog/debugger/el/ValueScript.java
@@ -11,11 +11,13 @@ import datadog.trace.bootstrap.debugger.el.DebuggerScript;
 import datadog.trace.bootstrap.debugger.el.ValueReferenceResolver;
 import java.io.IOException;
 import java.util.Objects;
+import java.util.regex.Matcher;
 import java.util.regex.Pattern;
 
 /** Implements expression language for capturing values for metric probes */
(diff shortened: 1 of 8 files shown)


dd-java-agent/agent-debugger/debugger-el/src/main/java/com/datadog/debugger/el/ValueScript.java
```diff
@@ -67,7 +69,16 @@ public static ValueExpression<?> parseRefPath(String refPath) {
     } else {
       head = parts[0];
     }
-    ValueExpression<?> current = DSL.ref(head);
+    ValueExpression<?> current;
+    Matcher matcher = INDEX_PATTERN.matcher(head);
```
shatzi, line 73: if this is used only for testing array[12] then we are good.
This doesn't handle this.field.arr[12].field.arr[14] and things like that.

If this is only for testing - suggest take this out from this class and move it to the test code.

shatzi (approved):
nice error message

dd-java-agent/agent-debugger/debugger-el/src/main/java/com/datadog/debugger/el/ValueScript.java
jpbempel: yes only for test code
I will move this also with serialization of the config that needs also to be separated

PR #4571, opened 2023-01-19
Refactor Datadog propagation tags

bantonsson:
# What Does This Do

* Renames `DatadogTags` to `PropagationTags`
* Adds a `HeaderType` and moves out the `PropagationTags` classes to a common `PropagationTagsFactory`

# Motivation

We need to support different header types for Datadog and W3C both for extraction and injection

# Additional Notes

Reviewing the two commits separately is probably easier since otherwise the changes show up as deleted and added files instead of renames and changes.


ygree (approved):
LGTM

PR #4577, opened 2023-01-20
Do not serialize captured data for Log Probes

jpbempel:
# What Does This Do
when `captureSnapshot=false`, don't freeze data and serialize into the snapshot.
Snapshot handling is also refactored by introducing SnapshotStatus which help to handle all combinations of cases when generating snapshots

# Motivation

# Additional Notes

### DIFF
--- a/dd-java-agent/agent-debugger/debugger-bootstrap/src/main/java/datadog/trace/bootstrap/debugger/Snapshot.java
+++ b/dd-java-agent/agent-debugger/debugger-bootstrap/src/main/java/datadog/trace/bootstrap/debugger/Snapshot.java
@@ -8,11 +8,10 @@ import datadog.trace.bootstrap.debugger.el.Values;
 import java.util.ArrayList;
 import java.util.Collections;
 import java.util.HashMap;
-import java.util.HashSet;
+import java.util.LinkedHashMap;
 import java.util.List;
 import java.util.Map;
 import java.util.Objects;
-import java.util.Set;
 import java.util.UUID;
 import java.util.function.BiConsumer;
 import org.slf4j.Logger;
@@ -34,13 +33,14 @@ public class Snapshot {
   private final ProbeDetails probe;
   private final String language;
   private final transient CapturedThread thread;
-  private final transient Set<String> capturingProbeIds = new HashSet<>();
+  private final transient Map<String, SnapshotStatus> snapshotStatuses = new LinkedHashMap<>();
   private final transient Map<String, List<EvaluationError>> errorsByProbeIds = new HashMap<>();
   private final transient String thisClassName;
   private String traceId; // trace_id
   private String spanId; // span_id
   private List<EvaluationError> evaluationErrors;
   private final transient SummaryBuilder summaryBuilder;
+  private transient boolean capturing = true;
 
   public Snapshot(java.lang.Thread thread, ProbeDetails probeDetails, String thisClassName) {
     this.startTs = System.nanoTime();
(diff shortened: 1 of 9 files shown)


dd-java-agent/agent-debugger/src/test/resources/com/datadog/debugger/sink/snapshotWithEvalErrorRegex.txt
```diff
@@ -2,8 +2,9 @@
 "ddsource":"dd_debugger",
 "debugger":\{
 "snapshot":\{
-"captures":\{"entry":\{"arguments":\{\},"locals":\{\}\}\},
+"captures":\{\},
```
shatzi, line 5: why do we have captures here?

dd-java-agent/agent-debugger/debugger-bootstrap/src/main/java/datadog/trace/bootstrap/debugger/Snapshot.java
```diff
@@ -34,13 +33,14 @@ public class Snapshot {
   private final ProbeDetails probe;
   private final String language;
   private final transient CapturedThread thread;
-  private final transient Set<String> capturingProbeIds = new HashSet<>();
+  private final transient Map<String, SnapshotStatus> snapshotStatuses = new LinkedHashMap<>();
```
shatzi, line 36: I like this direction. make the code easier to follow.

I suggest we go even further in future PRs to have SnapshotStatus as factory of the Snapshot object that would be used to commit.

Currently Snapshot act as a collection of snapshots(probes+status) and also used the final data to commit. We can split into SnapshotContext that holds list of SnapshotStatus + CaptureContexts like we talked about, and that would generate Snapshot objects that are much simpler.

shatzi (approved):
Great step forward.

dd-java-agent/agent-debugger/src/test/resources/com/datadog/debugger/sink/snapshotWithEvalErrorRegex.txt
jpbempel: the object instance is always created when creating snapshot object

dd-java-agent/agent-debugger/debugger-bootstrap/src/main/java/datadog/trace/bootstrap/debugger/Snapshot.java
```diff
@@ -243,38 +245,42 @@ private Snapshot copy(ProbeDetails additionalProbe, String newSnapshotId) {
...
-    }
-    boolean ret = isCapturing();
     if (ret) {
       capture.freeze();
     }
+    capturing = ret; // if false, prevent exit part to capture data
     return ret;
   }
 
+  private void updateSendingStatus(String probeId, boolean sending) {
```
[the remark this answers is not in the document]
jpbempel: ah yes left over, I will remove it

dd-java-agent/agent-debugger/debugger-bootstrap/src/main/java/datadog/trace/bootstrap/debugger/Snapshot.java
```diff
@@ -1215,4 +1234,21 @@ public String getMessage() {
       return message;
     }
   }
+
+  private static class SnapshotStatus {
+    boolean capturing;
+    boolean sending;
```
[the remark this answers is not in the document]
jpbempel: nope: log probe are sending=true  and capturing=false && hasErrors=true|false

PR #4579, opened 2023-01-20
docker base image poc

robertomonteromiguel:
# What Does This Do

Poc to create a docker base image with the dd-java-agent.

# Motivation

# Additional Notes


.github/workflows/codeql-analysis.yml
```diff
@@ -46,3 +46,14 @@ jobs:
         with:
           name: CodeQL Analysis Sarif
           path: /home/runner/work/dd-trace-java/results/java.sarif
+
```
robertomonteromiguel, line 49: I think it could be a good idea to attach the generated library to the build. 
In future we can use this artifact from base-docker-image.yml in order to avoid rebuild the artifact. We could use this: https://docs.github.com/en/actions/using-workflows/events-that-trigger-workflows#workflow_run

base-docker-image/Dockerfile
```diff
@@ -0,0 +1,7 @@
+FROM scratch
+COPY dd-java-agent.jar /binaries/
+COPY LIBRARY_VERSION /binaries/
+COPY LIBDDWAF_VERSION /binaries/
+COPY APPSEC_VERSION /binaries/
+COPY IAST_VERSION /binaries/
+COPY APPSEC_EVENT_RULES_VERSION /binaries/
```
pawelchcki, line 7: ideally this should become a single layer.

So:
```
FROM scratch as collect
COPY dd-java-agent.jar /binaries/
COPY LIBRARY_VERSION /binaries/
COPY LIBDDWAF_VERSION /binaries/
COPY APPSEC_VERSION /binaries/
COPY IAST_VERSION /binaries/
COPY APPSEC_EVENT_RULES_VERSION /binaries/
FROM scratch
COPY --from=collect / /
```
Or something of this sort.

lib-injection/Dockerfile
```diff
@@ -1,4 +1,4 @@
-FROM scratch AS java_agent
```
smola, line 1: Is this change needed at all?

.github/workflows/codeql-analysis.yml
```diff
@@ -46,3 +46,14 @@ jobs:
...
+      - name: upload dd-java-agent
+        uses: actions/upload-artifact@82c141cc518b40d92cc801eee768e7aafc9c2fa2 # 2.3.1
+        with:
+          name: dd-java-agent
+          path: |
+            workspace/dd-java-agent/build/libs/
+            workspace/dd-java-agent/build/resources/main/dd-java-agent.version
+            workspace/dd-java-agent/build/resources/main/iast/agent-iast.version
```
smola, line 57: What's this? We have no independent versions for IAST and AppSec. Shouldn't this always be the same as dd-java-agent.version?

PR #4583, opened 2023-01-23
Add well known classes as primitive-like objects

jpbempel:
# What Does This Do
We add two sets of well known classes on which we can safely toString for String representation/serialization.

# Motivation
Having a better string representation than extract fields for classes like `Duration` or `UUID`

Screenshot 2023-01-23 at 11 04 53

Screenshot 2023-01-23 at 11 20 46

# Additional Notes

### DIFF
--- a/dd-java-agent/agent-debugger/src/main/java/com/datadog/debugger/util/WellKnownClasses.java
+++ b/dd-java-agent/agent-debugger/src/main/java/com/datadog/debugger/util/WellKnownClasses.java
@@ -1,0 +1,60 @@
+package com.datadog.debugger.util;
+
+import java.util.Arrays;
+import java.util.HashSet;
+import java.util.Set;
+
+public class WellKnownClasses {
+
+  /** Set of class names which have a toString side effect free and class final */
+  private static Set<String> toStringFinalSafeClasses =
+      new HashSet<>(
+          Arrays.asList(
+              "java.lang.Class",
+              "java.lang.String",
+              "java.lang.Boolean",
+              "java.lang.Integer",
+              "java.lang.Long",
+              "java.lang.Double",
+              "java.lang.Character",
+              "java.lang.Byte",
+              "java.lang.Float",
+              "java.lang.Short",
+              "java.math.BigDecimal",
+              "java.math.BigInteger",
+              "java.time.Duration",
+              "java.time.Instant",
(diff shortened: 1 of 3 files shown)


dd-java-agent/agent-debugger/src/main/java/com/datadog/debugger/util/WellKnownClasses.java
```diff
@@ -0,0 +1,60 @@
...
+            "java.util.concurrent.atomic.AtomicBoolean",
+            "java.util.concurrent.atomic.AtomicInteger",
+            "java.util.concurrent.atomic.AtomicLong"));
+  }
+
+  /**
+   * @return true if type is a final class and toString implementation is well known and side effect
+   *     free
+   */
+  public static boolean isToStringFinalSafe(String type) {
```
shatzi, line 48: Why do we have both? i don't find this been used anywhere - so I wonder how this is going to be used in the future?

dd-java-agent/agent-debugger/src/main/java/com/datadog/debugger/util/WellKnownClasses.java
```diff
@@ -0,0 +1,60 @@
...
+              "java.lang.Short",
+              "java.math.BigDecimal",
+              "java.math.BigInteger",
+              "java.time.Duration",
+              "java.time.Instant",
+              "java.time.LocalTime",
+              "java.time.LocalDate",
+              "java.time.LocalDateTime",
+              "java.util.UUID",
+              "java.net.URI"));
```
shatzi, line 31: This is awesome.

dd-java-agent/agent-debugger/src/main/java/com/datadog/debugger/util/WellKnownClasses.java
jpbempel: because if we are using a declared type and not the dynamic/concrete type (not used yet) we need the final one

Ex: `AtomicInteger` is not final
so it is legal to do:

```
AtomicInteger integer = new AtomicInteger() {
  public String toString() {
    return value + callToDB();
  }
}
```
If we get type of `integer` by reflection and get the declared type we won't see the difference from a classic `AtomicInteger`

PR #4584, opened 2023-01-23
Add string operations for Expression Language

jpbempel:
# What Does This Do
Add following operations:
 - substring
 - startsWith
 - endsWith
 - contains
 - matches

# Motivation

# Additional Notes

### DIFF
--- a/dd-java-agent/agent-debugger/debugger-el/src/main/java/com/datadog/debugger/el/expressions/StringPredicateExpression.java
+++ b/dd-java-agent/agent-debugger/debugger-el/src/main/java/com/datadog/debugger/el/expressions/StringPredicateExpression.java
@@ -1,0 +1,42 @@
+package com.datadog.debugger.el.expressions;
+
+import static com.datadog.debugger.el.Expression.nullSafePrettyPrint;
+
+import com.datadog.debugger.el.Value;
+import com.datadog.debugger.el.values.StringValue;
+import datadog.trace.bootstrap.debugger.el.ValueReferenceResolver;
+import java.util.function.BiPredicate;
+
+public class StringPredicateExpression implements BooleanExpression {
+  private final ValueExpression<?> sourceString;
+  private final StringValue str;
+  private final BiPredicate<String, String> predicate;
+  private final String name;
+
+  public StringPredicateExpression(
+      ValueExpression<?> sourceString,
+      StringValue str,
+      BiPredicate<String, String> predicate,
+      String name) {
+    this.sourceString = sourceString;
+    this.str = str;
+    this.predicate = predicate;
+    this.name = name;
+  }
+
+  @Override
+  public Boolean evaluate(ValueReferenceResolver valueRefResolver) {
+    Value<?> sourceValue =
+        sourceString != null ? sourceString.evaluate(valueRefResolver) : Value.nullValue();
+    if (sourceValue.getValue() instanceof String) {
+      String sourceStr = (String) sourceValue.getValue();
+      return predicate.test(sourceStr, str.getValue());
+    }
+    return Boolean.FALSE;
+  }
+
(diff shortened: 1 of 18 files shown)


dd-java-agent/agent-debugger/debugger-el/src/main/java/com/datadog/debugger/el/expressions/StringPredicateExpression.java
```diff
@@ -0,0 +1,42 @@
...
+
+  @Override
+  public Boolean evaluate(ValueReferenceResolver valueRefResolver) {
+    Value<?> sourceValue =
+        sourceString != null ? sourceString.evaluate(valueRefResolver) : Value.nullValue();
+    if (sourceValue.getValue() instanceof String) {
+      String sourceStr = (String) sourceValue.getValue();
+      return predicate.test(sourceStr, str.getValue());
+    }
+    return Boolean.FALSE;
```
shatzi, line 35: why FALSE and not throw error?

dd-java-agent/agent-debugger/debugger-el/src/main/java/com/datadog/debugger/el/expressions/SubStringExpression.java
```diff
@@ -0,0 +1,37 @@
...
+    Value<?> sourceValue = source != null ? source.evaluate(valueRefResolver) : Value.nullValue();
+    if (sourceValue.getValue() instanceof String) {
+      String sourceStr = (String) sourceValue.getValue();
+      try {
+        return (Value<String>) Value.of(sourceStr.substring(startIndex, endIndex));
+      } catch (StringIndexOutOfBoundsException ex) {
+        valueRefResolver.addEvaluationError(prettyPrint(), ex.getMessage());
+      }
+    }
+    return Value.undefined();
```
shatzi, line 30: Why we return undefined and not throwing an error? not sure what the right thing to do.

dd-java-agent/agent-debugger/debugger-el/src/test/java/com/datadog/debugger/el/expressions/MatchesExpressionTest.java
```diff
@@ -0,0 +1,40 @@
...
+  @Test
+  void undefinedExpression() {
+    MatchesExpression expression =
+        new MatchesExpression(DSL.value(Values.UNDEFINED_OBJECT), new StringValue(null));
+    assertFalse(expression.evaluate(resolver));
+    assertEquals("matches(UNDEFINED, \"null\")", expression.prettyPrint());
+  }
+
+  @Test
+  void stringExpression() {
```
shatzi, line 31: suggest do some tests with some regex stuff like (.*) etc.

shatzi (approved):
LGTM, not sure if we handle argument not string the right way.

dd-java-agent/agent-debugger/debugger-el/src/main/java/com/datadog/debugger/el/expressions/SubStringExpression.java
jpbempel: This is the strategy that we adopt for the other Expressions too.
if there is a type problem we need to be coherent everywhere and if we report it or not. Here this is not a runtime issue but more at language level. We have already discussed that but how do we handle `substring(1, 1)`?

dd-java-agent/agent-debugger/debugger-el/src/test/java/com/datadog/debugger/el/expressions/MatchesExpressionTest.java
jpbempel: done

PR #4591, opened 2023-01-24
profiler availability check

richardstartin:
# What Does This Do

* We should be checking if the auxiliary profiler is available before registering the periodic event, not just that the instance isn't null
* We should make a bit more noise in the logs if the user has installed the profiler and we can't instantiate the profiler, so issues can be resolved faster
* We don't need to rethrow exceptions from the periodic event once they have been logged

# Motivation

# Additional Notes

### DIFF
--- a/dd-java-agent/agent-profiling/profiling-ddprof/src/main/java/com/datadog/profiling/ddprof/DatadogProfiler.java
+++ b/dd-java-agent/agent-profiling/profiling-ddprof/src/main/java/com/datadog/profiling/ddprof/DatadogProfiler.java
@@ -60,11 +60,28 @@ public final class DatadogProfiler {
     private static final DatadogProfiler INSTANCE = newInstance();
   }
 
+  private static void logFailedInstantiation(Throwable error) {
+    if (log.isDebugEnabled()) {
+      log.warn(
+          String.format(
+              "failed to instantiate Datadog profiler on %s %s",
+              OperatingSystem.current(), Arch.current()),
+          error);
+    } else {
+      log.warn(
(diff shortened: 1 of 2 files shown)


dd-java-agent/agent-profiling/profiling-ddprof/src/main/java/com/datadog/profiling/ddprof/DatadogProfiler.java
```diff
@@ -60,11 +60,28 @@ private static final class Singleton {
     private static final DatadogProfiler INSTANCE = newInstance();
   }
 
+  private static void logFailedInstantiation(Throwable error) {
+    if (log.isDebugEnabled()) {
+      log.warn(
+          "failed to instantiate Datadog profiler on {} {}",
+          OperatingSystem.current(),
+          Arch.current(),
+          error);
```
jbachorik, line 69: This seems superfluous - the format string takes only two arguments.

jbachorik (changes requested):

dd-java-agent/agent-profiling/profiling-ddprof/src/main/java/com/datadog/profiling/ddprof/DatadogProfiler.java
jbachorik, line 69: The format string specifies only 2 parameters.

```suggestion
      log.warn(
          "failed to instantiate Datadog profiler on {} {}",
          OperatingSystem.current(),
          Arch.current());
```

jbachorik (changes requested):

jbachorik (approved):

PR #4593, opened 2023-01-24
remove context tracker

richardstartin:
# What Does This Do

Removes the context tracker and associated `AgentSpan` APIs, and any transitive usage of those APIs

# Motivation

# Additional Notes


dd-trace-core/src/main/java/datadog/trace/core/DDSpan.java
```diff
@@ -262,13 +234,6 @@ public final void publish() {
   }
 
   public int storeContextToTag() {
```
jbachorik, line 236: Can this method be completely removed?
It was meant only for storing the context data as a span tag.

jbachorik (approved):

PR #4594, opened 2023-01-24
Move CI visibility logic to a dedicated module

nikita-tkachenko-datadog:
# What Does This Do
This PR moves logic that is specific to CI visibility to its own dedicated module: `dd-java-agent:agent-ci-visibility`

# Motivation
The other systems in the tracer have their own modules and use `internal-api` module as the "bridge" between the instrumentations and the system-specific modules. In this scheme `internal-api` stores the interfaces, while the actual logic mostly resides in the dedicated modules.
The goal of this change is to make CI visibility consistent with the other systems.


.github/CODEOWNERS
```diff
@@ -15,12 +15,12 @@ dd-smoke-tests/profiling-integration-tests/
 dd-java-agent/appsec/  @DataDog/appsec-java @ValentinZakharov
 
 # @DataDog/ci-app-libraries-java
-internal-api/src/main/java/datadog/trace/bootstrap/instrumentation/ci/ @DataDog/ci-app-libraries-java
-dd-trace-core/src/main/java/datadog/trace/civisibility/                @DataDog/ci-app-libraries-java
-dd-java-agent/agent-ci-visibility/                                     @DataDog/ci-app-libraries-java
-dd-java-agent/instrumentation/junit-4.10/                              @DataDog/ci-app-libraries-java
-dd-java-agent/instrumentation/junit-5.3/                               @DataDog/ci-app-libraries-java
-dd-java-agent/instrumentation/testng-6.4/                              @DataDog/ci-app-libraries-java
+internal-api/src/main/java/datadog/trace/api/civisibility/ @DataDog/ci-app-libraries-java
```
devinsba, line 18: This seems fine to me, but moving these out of the bootstrap package will mean they aren't loaded onto the bootstrap classloader. Have you manually tested this? We should probably add a smoke test at some point in this project too. I'll add that as a task for myself

nikita-tkachenko-datadog: Yep, tested this manually, all works fine.

Surprisingly, when I check this manually using the debugger, I can see that the classes in the new package are still loaded by the bootstrap classloader.
As far as I understand the entire agent Jar is added to the bootstrap classpath, but only a selected handful of classes are available to the bootstrap loader as the majority of them are "hidden" by shadowing to different source roots and by changing the extension to ".classdata".
Judging by our Gradle setup these obfuscation techniques are applied to "subproject shadow jars" and the "shared shadow jar", so `internal-api` should not be affected.
I checked the Jar internals and everything adds up, so we should be fine :)

Anyway, a smoke test is a great idea!

dd-java-agent/agent-ci-visibility/src/main/java/datadog/trace/civisibility/CIProviderInfoFactory.java
```diff
@@ -0,0 +1,35 @@
+package datadog.trace.civisibility;
+
+public class CIProviderInfoFactory {
+
+  public static CIProviderInfo createCIProviderInfo() {
+    // CI and Git information is obtained
+    // from different environment variables
+    // depending on which CI server is running the build.
+    if (System.getenv(JenkinsInfo.JENKINS) != null) {
```
smola, line 9: I'm not sure how common this would be, but you might want to account for these environment variables being set to an empty string, or a blank string. For the latter case,  you can use `!Strings.trim(System.getenv(...)).isEmpty())`.

smola (approved):
Looks good to me, provided you have checked possible classloader problems as pointed out by Brian.

PR #4596, opened 2023-01-25
Add formatting object values for log lines

jpbempel:
# What Does This Do
Primitives are printed as is,
collections/maps are printed up to 3 elements/entries Complex classes are printed with fields up to 5 and deep level 1. Refactored how we serializew value for sharing with Json snapshot serializations and applied limits

# Motivation
Coherent log line parameter representation

# Additional Notes

### DIFF
--- a/dd-java-agent/agent-debugger/debugger-bootstrap/src/main/java/datadog/trace/bootstrap/debugger/Snapshot.java
+++ b/dd-java-agent/agent-debugger/debugger-bootstrap/src/main/java/datadog/trace/bootstrap/debugger/Snapshot.java
@@ -182,7 +182,15 @@ public class Snapshot {
   }
 
   public String getSummary() {
-    return summaryBuilder.build();
+    String summary = summaryBuilder.build();
+    List<EvaluationError> errors = summaryBuilder.getEvaluationErrors();
+    if (!errors.isEmpty()) {
+      if (evaluationErrors == null) {
+        evaluationErrors = new ArrayList<>();
+      }
+      evaluationErrors.addAll(errors);
+    }
+    return summary;
   }
 
   public void commit() {
--- a/dd-java-agent/agent-debugger/debugger-bootstrap/src/main/java/datadog/trace/bootstrap/debugger/SnapshotSummaryBuilder.java
+++ b/dd-java-agent/agent-debugger/debugger-bootstrap/src/main/java/datadog/trace/bootstrap/debugger/SnapshotSummaryBuilder.java
@@ -87,6 +87,11 @@ public class SnapshotSummaryBuilder implements SummaryBuilder {
     return sb.toString();
   }
 
+  @Override
+  public List<Snapshot.EvaluationError> getEvaluationErrors() {
+    return Collections.emptyList();
+  }
+
   private static String formatMethod(List<CapturedStackFrame> stack, ProbeLocation probeLocation) {
     if (stack != null && stack.size() > 0) {
       // we first try to use the top frame on the stacktrace, if available
--- a/dd-java-agent/agent-debugger/src/main/java/com/datadog/debugger/agent/LogMessageTemplateSummaryBuilder.java
+++ b/dd-java-agent/agent-debugger/src/main/java/com/datadog/debugger/agent/LogMessageTemplateSummaryBuilder.java
@@ -1,14 +1,27 @@
 package com.datadog.debugger.agent;
 
+import com.datadog.debugger.el.Value;
 import com.datadog.debugger.el.ValueScript;
 import com.datadog.debugger.probe.LogProbe;
+import com.datadog.debugger.util.ValueSerializer;
 import datadog.trace.bootstrap.debugger.CapturedStackFrame;
+import datadog.trace.bootstrap.debugger.FieldExtractor;
+import datadog.trace.bootstrap.debugger.Fields;
+import datadog.trace.bootstrap.debugger.Limits;
 import datadog.trace.bootstrap.debugger.Snapshot;
 import datadog.trace.bootstrap.debugger.SummaryBuilder;
+import java.lang.reflect.Field;
+import java.util.ArrayList;
 import java.util.List;
+import java.util.Map;
+import org.slf4j.Logger;
+import org.slf4j.LoggerFactory;
 
 public class LogMessageTemplateSummaryBuilder implements SummaryBuilder {
+  private static final Logger LOG = LoggerFactory.getLogger(LogMessageTemplateSummaryBuilder.class);
+
   private final LogProbe logProbe;
+  private final List<Snapshot.EvaluationError> evaluationErrors = new ArrayList<>();
 
   public LogMessageTemplateSummaryBuilder(LogProbe logProbe) {
     this.logProbe = logProbe;
@@ -42,10 +55,23 @@ public class LogMessageTemplateSummaryBuilder implements SummaryBuilder {
       if (segment.getStr() != null) {
         sb.append(segment.getStr());
       } else {
-        sb.append(segment.getParsedExpr().getResult().getValue());
+        Value<?> result = segment.getParsedExpr().getResult();
+        if (result.isUndefined()) {
+          sb.append(result.getValue());
+        } else if (result.isNull()) {
+          sb.append("null");
+        } else {
+          Limits limits = new Limits(1, 3, 255, 5);
+          serializeValue(sb, segment.getParsedExpr().getDsl(), result.getValue(), limits);
+        }
       }
     }
     return sb.toString();
+  }
+
+  @Override
+  public List<Snapshot.EvaluationError> getEvaluationErrors() {
+    return evaluationErrors;
   }
 
   private void executeExpressions(Snapshot.CapturedContext entry) {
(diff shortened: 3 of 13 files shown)


dd-java-agent/agent-debugger/src/main/java/com/datadog/debugger/agent/LogMessageTemplateSummaryBuilder.java
```diff
@@ -59,4 +85,177 @@ private void executeExpressions(Snapshot.CapturedContext entry) {
...
+    }
+
+    @Override
+    public void mapEpilogue(Map<?, ?> map, boolean isComplete) throws Exception {
+      sb.append(isComplete ? "}" : ", ...}");
+      inCollection = false;
+    }
+
+    @Override
+    public void objectValue(Object value, ValueSerializer valueSerializer, Limits limits)
```
shatzi, line 206: I think this logic should move to the ValueSeralizer -> so this class doens't need to know bout ProcessFields or limits.

ValueSeralizer can call something like this: objectProluge(value,type). fieldProluge(field) seralizePrimitive(value), fieldEpilogue(field), fieldSeperator() objectEoplouge(value,type,isComplete)

dd-java-agent/agent-debugger/src/main/java/com/datadog/debugger/util/MoshiSnapshotHelper.java
```diff
@@ -543,57 +515,173 @@ public void toJson(JsonWriter jsonWriter, Snapshot.CapturedValue capturedValue)
...
           jsonWriter.name(NOT_CAPTURED_REASON);
           jsonWriter.value(COLLECTION_SIZE_REASON);
         }
         jsonWriter.name(SIZE);
         jsonWriter.value(String.valueOf(map.size()));
-      } else if (limits.maxReferenceDepth > 0) {
+      }
+
+      @Override
+      public void objectValue(Object value, ValueSerializer valueSerializer, Limits limits)
```
shatzi, line 683: same comment here - this looks almost identical to SummaryBuilder StringSerailzer

dd-java-agent/agent-debugger/src/main/java/com/datadog/debugger/agent/LogMessageTemplateSummaryBuilder.java
```diff
@@ -59,4 +85,177 @@ private void executeExpressions(Snapshot.CapturedContext entry) {
...
+    }
+
+    @Override
+    public void arrayEpilogue(Object value, boolean isComplete, int arraySize) throws Exception {
+      sb.append(isComplete ? "]" : ", ...]");
+      inCollection = false;
+    }
+
+    @Override
+    public void primitiveArrayElement(String value, String type) throws Exception {
```
shatzi, line 155: I think the inCollection and initial here are broken as they will break if we have 2 depth array (not now but in the future).

I think there might be a cleaner abstraction by simplify the interface:
1. add `arraySeperator()` function that write the ", ". so the ValueSerailzer can call primitveValue(value) arraySeperator() instead of having special primitiveArrayElement -> that way we can avoid the "inCollection" and "initial"
2. map can then use the same thing. arrayProluge() mapEntryPrologue() value() mapEntrySepertor() value() mapEntryEpilogue() arraySeperator()...
3. also see comment about objectValue()

dd-java-agent/agent-debugger/src/test/java/com/datadog/debugger/agent/LogMessageTemplateSummaryBuilderTest.java
```diff
@@ -52,4 +62,126 @@ public void argTemplate() {
...
+  public void argComplexObjectTemplate() {
+    LogProbe probe = LogProbe.builder().template("{obj}").build();
+    LogMessageTemplateSummaryBuilder summaryBuilder = new LogMessageTemplateSummaryBuilder(probe);
+    Snapshot.CapturedContext capturedContext = new Snapshot.CapturedContext();
+    capturedContext.addArguments(
+        new Snapshot.CapturedValue[] {
+          Snapshot.CapturedValue.of("obj", Level0.class.getTypeName(), new Level0())
+        });
+    summaryBuilder.addEntry(capturedContext);
+    assertEquals("{intField0=0, strField0=foo0, level1=...}", summaryBuilder.build());
```
shatzi, line 163: so if we also have incomplete map it would like `{intField=0, level1=..., ...}`, right?

Can we also say how many items was in the array/map? maybe `{intField=0, level1=..., ...}(10 items)`

Not sure if we that is useful/doable in call cases

dd-java-agent/agent-debugger/src/main/java/com/datadog/debugger/util/ValueSerializer.java
```diff
@@ -0,0 +1,290 @@
...
+
+import datadog.trace.bootstrap.debugger.Limits;
+import java.lang.reflect.Array;
+import java.util.Collection;
+import java.util.Iterator;
+import java.util.Map;
+import java.util.Set;
+
+/** serialize Java Object value with applied {@link Limits} and following references */
+public class ValueSerializer {
```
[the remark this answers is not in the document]
shatzi: Also, suggest to add few unit-tests to ensure this class works as expected with limits and special array types etc.

shatzi (approved):
LGTM - suggest some refactor ideas.

dd-java-agent/agent-debugger/src/main/java/com/datadog/debugger/agent/LogMessageTemplateSummaryBuilder.java
```diff
@@ -59,4 +85,177 @@ private void executeExpressions(Snapshot.CapturedContext entry) {
       }
     }
   }
+
+  private void serializeValue(StringBuilder sb, String expr, Object value, Limits limits) {
+    ValueSerializer serializer =
+        new ValueSerializer(new StringTypeSerializer(sb, evaluationErrors));
+    try {
+      serializer.serialize(
+          value, value != null ? value.getClass().getTypeName() : "java.lang.Object", limits);
```
cimi, line 94: Does the type here matter for nulls? 

If yes, maybe there's a way to get the declared type?

dd-java-agent/agent-debugger/src/main/java/com/datadog/debugger/agent/LogMessageTemplateSummaryBuilder.java
```diff
@@ -59,4 +85,177 @@ private void executeExpressions(Snapshot.CapturedContext entry) {
...
+    }
+
+    @Override
+    public void prologue(Object value, String type) throws Exception {
+      if (inMapEntry && !initial) {
+        sb.append("=");
+      } else if (inCollection && !initial) {
+        sb.append(", ");
+      }
+      initial = false;
```
cimi, line 120: I'm confused 🙂 for  the `initial` element we do nothing in the prologue?

Edit: I see from the tests that this works, but I don't understand the logic here just by reading the code.

dd-java-agent/agent-debugger/src/main/java/com/datadog/debugger/agent/LogMessageTemplateSummaryBuilder.java
```diff
@@ -59,4 +85,177 @@ private void executeExpressions(Snapshot.CapturedContext entry) {
...
+    public void arrayEpilogue(Object value, boolean isComplete, int arraySize) throws Exception {
+      sb.append(isComplete ? "]" : ", ...]");
+      inCollection = false;
+    }
+
+    @Override
+    public void primitiveArrayElement(String value, String type) throws Exception {
+      if (inCollection && !initial) {
+        sb.append(", ");
+      }
```
cimi, line 158: Why do we duplicate logic from `prologue` here?

dd-java-agent/agent-debugger/src/main/java/com/datadog/debugger/util/ValueSerializer.java
```diff
@@ -0,0 +1,290 @@
...
+      }
+    } else if (value.getClass().isArray() && (limits.maxReferenceDepth > 0)) {
+      int arraySize = Array.getLength(value);
+      boolean isComplete = true;
+      typeSerializer.arrayPrologue(value);
+      if (value.getClass().getComponentType().isPrimitive()) {
+        Class<?> componentType = value.getClass().getComponentType();
+        if (componentType == long.class) {
+          isComplete = serializeLongArray((long[]) value, limits.maxCollectionSize);
+        }
```
cimi, line 103: Nit: we could move this to a separate function, `serializeArray` - that way we remove all the casts and if statements from the generic method.

dd-java-agent/agent-debugger/src/main/java/com/datadog/debugger/util/MoshiSnapshotHelper.java
```diff
@@ -543,57 +515,173 @@ public void toJson(JsonWriter jsonWriter, Snapshot.CapturedValue capturedValue)
...
+      public JsonTypeSerializer(JsonWriter jsonWriter) {
+        this.jsonWriter = jsonWriter;
+      }
+
+      @Override
+      public void prologue(Object value, String type) throws Exception {
+        jsonWriter.beginObject();
+        jsonWriter.name(TYPE);
+        jsonWriter.value(type);
+      }
```
cimi, line 576: I assume this is the same logic we used in the original JSON serializer.

cimi (commented):
:+1: Thanks!

I'm late to the party 🙈 I had the page open so I did not see Idan's review. Since I reviewed it anyway, posting my comments in case they help.

dd-java-agent/agent-debugger/debugger-bootstrap/src/main/java/datadog/trace/bootstrap/debugger/Snapshot.java
```diff
@@ -182,7 +182,15 @@ public List<EvaluationError> getEvaluationErrors() {
...
-    return summaryBuilder.build();
+    String summary = summaryBuilder.build();
+    List<EvaluationError> errors = summaryBuilder.getEvaluationErrors();
+    if (!errors.isEmpty()) {
+      if (evaluationErrors == null) {
+        evaluationErrors = new ArrayList<>();
+      }
+      evaluationErrors.addAll(errors);
+    }
+    return summary;
```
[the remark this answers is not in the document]
jpbempel: it's done once here: https://github.com/DataDog/dd-trace-java/blob/c256ff8259be2f404479a23572688bf667b496e8/dd-java-agent/agent-debugger/src/main/java/com/datadog/debugger/agent/JsonSnapshotSerializer.java#L143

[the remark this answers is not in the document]
jpbempel: if it's more clear we can rename it `buildSummary`

dd-java-agent/agent-debugger/debugger-bootstrap/src/main/java/datadog/trace/bootstrap/debugger/SnapshotSummaryBuilder.java
```diff
@@ -87,6 +87,11 @@ public String build() {
     return sb.toString();
   }
 
+  @Override
+  public List<Snapshot.EvaluationError> getEvaluationErrors() {
+    return Collections.emptyList();
+  }
```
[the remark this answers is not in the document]
jpbempel: yes

dd-java-agent/agent-debugger/src/main/java/com/datadog/debugger/agent/LogMessageTemplateSummaryBuilder.java
jpbempel: no we could in this case pass null as argument

dd-java-agent/agent-debugger/src/main/java/com/datadog/debugger/agent/LogMessageTemplateSummaryBuilder.java
jpbempel: yes. It's mimicng the following algorithm:
```
StringBuilder sb = new StringBuilder();
for (int i = 0; i < list.size; i++) {
  if (i > 0) {
    sb.append(", ");
  }
  sb.append(list.get(i));
}
```

or could do the other way around in epilogue testing for the max.

dd-java-agent/agent-debugger/src/main/java/com/datadog/debugger/agent/LogMessageTemplateSummaryBuilder.java
jpbempel: we are not calling `serialize` for primitive array, so the logic is little bit different

dd-java-agent/agent-debugger/src/main/java/com/datadog/debugger/util/ValueSerializer.java
jpbempel: can you elaborate how to remove all the casts?

dd-java-agent/agent-debugger/src/main/java/com/datadog/debugger/util/MoshiSnapshotHelper.java
jpbempel: yes

dd-java-agent/agent-debugger/src/main/java/com/datadog/debugger/util/ValueSerializer.java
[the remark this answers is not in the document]
jpbempel: WE have already a lot of unit tests on both side (Json & string) that covers all of the aspects of the algorithm,
see https://github.com/DataDog/dd-trace-java/blob/master/dd-java-agent/agent-debugger/src/test/java/com/datadog/debugger/agent/SnapshotSerializationTest.java
and
https://github.com/DataDog/dd-trace-java/blob/master/dd-java-agent/agent-debugger/src/test/java/com/datadog/debugger/agent/LogMessageTemplateSummaryBuilderTest.java

dd-java-agent/agent-debugger/src/test/java/com/datadog/debugger/agent/LogMessageTemplateSummaryBuilderTest.java
jpbempel: can be added in the log with EL like `{len(arrayField)}`

jpbempel: > so if we also have incomplete map it would like `{intField=0, level1=..., ...}`, right?

Don't understand

dd-java-agent/agent-debugger/src/main/java/com/datadog/debugger/agent/LogMessageTemplateSummaryBuilder.java
jpbempel: As our limits for log are more restrictive, this case cannot happen, because you can only see an array of primitives
Depth is 1, so an array of depth 2 will not be printed.

PR #4597, opened 2023-01-25
Add support for test.codeowners and test.source.* tags

nikita-tkachenko-datadog:
# What Does This Do
This PR updates CI-visibility part of the tracer, adding support for the following tags:

`test.codeowners`: teams/people who own the test (in accordance with the `CODEOWNERS` file, if one is available in the client's repo)
`test.source.file`: file path where the test is located, relative to the project's source root
`test.source.start`: the line in the test source file where the test definition starts
`test.source.end`: the line of the source code where the test ends

# Motivation
These tags are part of the [CI visibility tracers specification](https://github.com/DataDog/datadog-ci-spec/blob/637677c02391ec9d3cc73892183f8ae4fa0c3b65/spec/citest/test.yaml#L120).
Populating them will allow the clients to use additional features offered by CI visibility: showing source code, notifying code owners about test failures, etc.

# Additional Notes
In scope of this PR a new gradle module was created for logic that is specific to CI visibility: `dd-java-agent:agent-ci-visibility`.
For now, in order to keep the scope of the PR smaller, the new module only contains logic introduced to support the codeowners/source tag features.
The intention is to later move all the CI visibility logic to the new module, with only the interfaces remaining in `internal-api` module - that way CI visibility will be consistent with the other subsystems residing in the tracer.


internal-api/src/main/java/datadog/trace/bootstrap/instrumentation/ci/codeowners/Codeowners.java
```diff
@@ -30,9 +31,9 @@ private Codeowners(String repoRoot, Iterable<Entry> entries) {
    * @param path <b>Absolute</b> path to a file/folder inside the repository
    * @return the list of teams/people who own the provided path
    */
-  public Collection<String> getOwners(String path) {
+  public @Nullable Collection<String> getOwners(String path) {
     if (!path.startsWith(repoRoot)) {
-      return Collections.emptyList();
+      return null;
```
ValentinZakharov: What was the reason using `null` ?
I'd rather keep empty list to avoid possible NPE.

dd-java-agent/agent-ci-visibility/src/main/java/datadog/trace/bootstrap/instrumentation/ci/source/RepoIndexSourcePathResolver.java
```diff
@@ -0,0 +1,301 @@
...
+  RepoIndexSourcePathResolver(
+      String repoRoot, SourceRootResolver sourceRootResolver, FileSystem fileSystem) {
+    this.repoRoot = repoRoot;
+    this.sourceRootResolver = sourceRootResolver;
+    this.fileSystem = fileSystem;
+  }
+
+  @Nullable
+  @Override
+  public String getSourcePath(Class<?> c) {
```
ValentinZakharov, line 53: I think, there is missing `null` checker for `Class<?> c`

internal-api/src/main/java/datadog/trace/bootstrap/instrumentation/ci/codeowners/Codeowners.java
nikita-tkachenko-datadog: The idea was to use an empty list for files that have no owners (it's possible to have a CODEOWNERS entry that says file/folder is not owned by anyone), and null for the cases when owners could not be determined.

Anyway, this looks like an older version of the code - this PR has a bunch of commits that I will squash during merge - in the PR's last commit this code looks a bit different.

ValentinZakharov (approved):

dd-java-agent/agent-ci-visibility/src/main/java/datadog/trace/bootstrap/instrumentation/ci/source/RepoIndexSourcePathResolver.java
nikita-tkachenko-datadog: From my point of view an appropriate response to `getSourcePath(null)` would be an NPE, since this likely indicates a programming error, which is best signalled in a fail-fast manner. Swallowing a `null` argument might cover this sort of programming issues, making them harder to discover.
Currently this method will never be called with a `null` argument, as the calling code does the necessary checks.

nikita-tkachenko-datadog:
@ValentinZakharov, thank you for the review!

PR #4600, opened 2023-01-25
Support Jakarta servlet in IAST

DDJavierSantos:
# What Does This Do

Instrument the jakarta servlet API

# Motivation
This API is used in a number of frameworks (spring) and web servers. We need to instrument it so we can taint the API output for IAST use.

# Additional Notes

### DIFF
--- a/dd-java-agent/instrumentation/servlet/request-5/src/test/groovy/JakartaCallSiteTest.groovy
+++ b/dd-java-agent/instrumentation/servlet/request-5/src/test/groovy/JakartaCallSiteTest.groovy
@@ -1,0 +1,71 @@
+import datadog.trace.agent.test.AgentTestRunner
+import datadog.trace.api.config.TracerConfig
+import datadog.trace.api.iast.InstrumentationBridge
+import datadog.trace.api.iast.source.WebModule
+import foo.bar.smoketest.JakartaTestSuite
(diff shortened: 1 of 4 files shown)


dd-java-agent/instrumentation/servlet/request-5/src/test/groovy/JakartaCallSiteTest.groovy
```diff
@@ -0,0 +1,71 @@
...
+import datadog.trace.api.iast.source.WebModule
+import foo.bar.smoketest.JakartaTestSuite
+import jakarta.servlet.http.HttpServletRequest
+import jakarta.servlet.http.HttpServletRequestWrapper
+
+class JakartaCallSiteTest extends AgentTestRunner {
+
+  @Override
+  protected void configurePreAgent() {
+    injectSysConfig(TracerConfig.SCOPE_ITERATION_KEEP_ALIVE, "1") // don't let iteration spans linger
```
smola, line 13: Do we need this? What is this for?

DDJavierSantos: It is not really needed. Carried over from another test. I have removed it.

smola (approved):

PR #4602, opened 2023-01-26
Fix CI spec tests

nikita-tkachenko-datadog:
# What Does This Do
This PR updates CI visibility logic that extracts metadata from Azure Pipelines and Github actions CI providers.

# Motivation
The extracted metadata is used to populate tags that are sent to backend along with the rest of the test execution info.
There is a [CI spec](https://github.com/DataDog/datadog-ci-spec) that describes which metadata need to be extracted and from which environment variables.
This update ensures that Java tracer agrees with the spec.


albertvaka (approved):
LGTM. I'm not sure if this has the last version of the fixtures, though (the ones after the change to keep both git tag and branch).

PR #4603, opened 2023-01-26
Add partial key cache to enable cache lookup from parts of keys

bantonsson:
# What Does This Do

Changes the `FixedSizeCache` implementation to allow for a partial key cache, where two extra integers are provided as arguments to hash, compare and produce functions.

# Motivation

This is in preparation for doing value lookups from of parts of strings while not creating any extra objects (i.e. substring instances or wrappers holding the arguments).

# Additional Notes

Naming is hard.

### DIFF
--- a/internal-api/src/main/java/datadog/trace/api/cache/FixedSizeCache.java
+++ b/internal-api/src/main/java/datadog/trace/api/cache/FixedSizeCache.java
@@ -1,8 +1,13 @@
 package datadog.trace.api.cache;
 
 import datadog.trace.api.Pair;
+import datadog.trace.api.cache.DDPartialKeyCache.Comparator;
+import datadog.trace.api.cache.DDPartialKeyCache.Hasher;
+import datadog.trace.api.cache.DDPartialKeyCache.Producer;
 import java.util.Arrays;
 import java.util.function.Function;
+import java.util.function.IntFunction;
+import javax.annotation.Nullable;
 
 /**
  * This is a fixed size cache that only has one operation <code>computeIfAbsent</code>, that is used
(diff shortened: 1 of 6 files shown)


internal-api/src/main/java/datadog/trace/api/IntLeftPair.java
```diff
@@ -0,0 +1,47 @@
+package datadog.trace.api;
+
+import java.util.Objects;
+import javax.annotation.Nullable;
+
+public final class IntLeftPair<U> {
```
richardstartin, line 6: why not pack into a long?

internal-api/src/main/java/datadog/trace/api/function/IntIntBiPredicate.java
```diff
@@ -0,0 +1,5 @@
+package datadog.trace.api.function;
+
+public interface IntIntBiPredicate<T, U> {
```
bantonsson, line 3: I have no idea why I made these into standalone interfaces even though they're rather coupled to the `DDPartialKeyCache`. I'll move them.

internal-api/src/main/java/datadog/trace/api/cache/FixedSizeCache.java
```diff
@@ -46,43 +55,35 @@
...
-   * Look up or create and store a value in the cache.
-   *
-   * @param key the key to look up
-   * @param creator how to create a cached value base on the key if the lookup fails
-   * @return the cached or created and stored value
-   */
-  @Override
-  public final V computeIfAbsent(K key, Function<K, ? extends V> creator) {
+  protected final V internalComputeIfAbsent(
+      K key, int m, int n, H hasher, C comparator, P producer) {
```
jbachorik, line 63: Could you, please, add some docs regarding the new parameters? Especially `m` and `n`
For anyone implementing a specialised subclass in the future it would be nice to have the signature well documented.

bantonsson: @jbachorik documented

richardstartin (approved):

internal-api/src/main/java/datadog/trace/api/cache/FixedSizeCache.java
bantonsson: @jbachorik documented

PR #4604, opened 2023-01-26
Fix inner classes matching for line probes

jpbempel:
# What Does This Do
Allows the DebuggerTransformer to find inner classes when using file+line probe definitions.
Reporting of line not found was removed because it can be found latter by loading inner/top-level class.

# Background
When class are already loaded, we need to retransform them for passing through the `DebuggerTransformer`.
But to detect which class need to be retransformed, we rely on the naming. However inner classes or Top-Level classes are independent class files and the only way to associate them is the sourcefile attribute in each class file.

# Solution
Track all dependent classes by their source file, so it easier to find all related class given a filename.
This avoids scanning all loaded classes to parse their sourcefile and try to match with actual probe definitions.

# Motivation
Support line probe in inner/top-level classes or Kotlin lambdas

# Additional Notes

### DIFF
--- a/dd-java-agent/agent-debugger/src/main/java/com/datadog/debugger/agent/ConfigurationComparer.java
+++ b/dd-java-agent/agent-debugger/src/main/java/com/datadog/debugger/agent/ConfigurationComparer.java
@@ -90,11 +90,8 @@ public class ConfigurationComparer {
           if (key == null || key.equals("")) {
             continue;
           }
-          // remove extension if any
-          int idx = key.lastIndexOf('.');
-          if (idx > -1) {
-            key = key.substring(0, idx);
-          }
+          processAdditionalClasses(key, changedClasses);
+          key = removeExtension(key);
           key = normalizeFilePath(key);
         }
       } else {
@@ -111,6 +108,30 @@ public class ConfigurationComparer {
       changedClasses.insert(reverseStr(typeName));
     }
     return changedClasses;
+  }
+
+  private void processAdditionalClasses(String sourceFile, Trie changedClasses) {
+    int idx = sourceFile.lastIndexOf('/');
+    if (idx != -1) {
+      sourceFile = sourceFile.substring(idx + 1);
+    }
+    List<String> additionalClasses =
+        SourceFileTrackingTransformer.INSTANCE.getClassNameBySourceFile(sourceFile);
+    if (additionalClasses == null) {
+      return;
+    }
+    for (String additionalClass : additionalClasses) {
+      additionalClass = normalizeFilePath(additionalClass);
+      changedClasses.insert(reverseStr(additionalClass));
+    }
+  }
+
+  private String removeExtension(String fileName) {
+    int idx = fileName.lastIndexOf('.');
+    if (idx > -1) {
+      fileName = fileName.substring(0, idx);
+    }
+    return fileName;
   }
 
   private String normalizeFilePath(String filePath) {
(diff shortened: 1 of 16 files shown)


dd-java-agent/agent-debugger/src/main/java/com/datadog/debugger/agent/ConfigurationComparer.java
```diff
@@ -113,6 +110,30 @@ private Trie buildChangedClasses() {
     return changedClasses;
   }
 
+  private void processAdditionalClasses(String sourceFile, Trie changedClasses) {
+    int idx = sourceFile.lastIndexOf('/');
+    if (idx != -1) {
+      sourceFile = sourceFile.substring(idx + 1);
+    }
+    List<String> additionalClasses =
+        SourceFileTrackingTransformer.INSTANCE.getClassNameBySourceFile(sourceFile);
```
shatzi, line 119: I don't like that use static INSTANCE and creating hidden dependencies that complicate how we do unit testing.
Lets clean this up to make more single responsibility classes

My take on this would be:
1. Extract all getLoadedClasses logic from this class. maybe `ClassesToRetansformFinder` that gets list of changes, loaded classes and classFileMapping provider to do all this logic.
    a. that would make the configurationComparer simpler
    b. remote the dependency of SourceFileTrackingTransformer
2. ClassesToRetansformFinder can have a function "registerClassWithCustomSourceName(classname, sourceilfe)
    a. then SourceFileTrackingTransformer can be created debuggerAgent an pass the instance of ClassesToRetansformFinder to it to register classes
    b. configuration updater can use ClassesToRetansformFinder.getAllChangedClasses(changes)
    
WDYT?

dd-java-agent/agent-debugger/src/main/java/com/datadog/debugger/agent/ConfigurationComparer.java
```diff
@@ -113,6 +110,30 @@ private Trie buildChangedClasses() {
...
+    if (additionalClasses == null) {
+      return;
+    }
+    for (String additionalClass : additionalClasses) {
+      additionalClass = normalizeFilePath(additionalClass);
+      changedClasses.insert(reverseStr(additionalClass));
+    }
+  }
+
+  private String removeExtension(String fileName) {
```
shatzi, line 129: Maybe move all those helper functions to ClassFileHelper utility classes - to have a single place when we load file name, normalize it, etc.

dd-java-agent/agent-debugger/src/main/java/com/datadog/debugger/agent/ConfigurationComparer.java
```diff
@@ -151,23 +172,28 @@ public List<Class<?>> getAllLoadedChangedClasses(Class<?>[] allLoadedClasses) {
...
       }
-      classesToBeTransformed.add(clazz);
     }
     return classesToBeTransformed;
   }
 
+  private static boolean lookupClass(Trie changedClasses, Class<?> clazz) {
+    String reversedTypeName = reverseStr(clazz.getName());
+    // try first with FQN (java.lang.String)
+    if (changedClasses.contains(reversedTypeName)) {
```
shatzi, line 185: why not doing containsPrefix? as it the same we do at the end.

dd-java-agent/agent-debugger/src/main/java/com/datadog/debugger/agent/SourceFileTrackingTransformer.java
```diff
@@ -0,0 +1,60 @@
...
+      byte[] classfileBuffer)
+      throws IllegalClassFormatException {
+    if (className == null) {
+      return null;
+    }
+    String sourceFile = ClassFileHelper.extractSourceFile(classfileBuffer);
+    if (sourceFile == null) {
+      return null;
+    }
+    String simpleClassName = className.substring(className.lastIndexOf('/') + 1);
```
shatzi, line 38: look like the normalize or removeExtention functions?

shatzi (approved):
Awesome stuff. few refactor suggestions

dd-java-agent/agent-debugger/src/main/java/com/datadog/debugger/agent/ConfigurationComparer.java
jpbempel: sounds nice!
done

dd-java-agent/agent-debugger/src/main/java/com/datadog/debugger/agent/SourceFileTrackingTransformer.java
jpbempel: done

dd-java-agent/agent-debugger/src/main/java/com/datadog/debugger/agent/ConfigurationComparer.java
jpbempel: done

dd-java-agent/agent-debugger/src/main/java/com/datadog/debugger/agent/ConfigurationComparer.java
jpbempel: done

PR #4608, opened 2023-01-26
AppSec events tracking API

ValentinZakharov:
# What Does This Do
Implemented Application Security Tracking API, that allows customer to send login/custom events linked to a span.

# Motivation

# Additional Notes


dd-trace-api/src/main/java/datadog/trace/api/internal/TraceSegment.java
```diff
@@ -14,15 +14,23 @@ public interface TraceSegment {
    * @param key key of the tag
    * @param value value of the tag
    */
-  void setTagTop(String key, Object value);
+  default void setTagTop(String key, Object value) {
+    setTagTop(key, value, false);
+  }
+
+  void setTagTop(String key, Object value, boolean sanitize);
```
bantonsson, line 21: Document plz

bantonsson (commented):
Looks good, but I would like either some addition to the trace mapper tests and/or a smoke test that validates that the map gets flattened correctly.

dd-trace-api/src/main/java/datadog/trace/api/internal/TraceSegment.java
ValentinZakharov: Done

dd-trace-api/src/main/java/datadog/trace/api/GlobalTracer.java
```diff
@@ -64,6 +66,15 @@ public static Tracer get() {
     return provider;
   }
 
+  public static EventTracker getEventTracker() {
+    if (eventTracker == null) {
+      if (provider instanceof InternalTracer) {
```
smola, line 71: I guess we could have something that does not implement `InternalTracer`, like the initial `NO_OP` value? Maybe we want to return a no-op event tracker in that case, to avoid NPE?

smola (commented):
Looks good. I wonder about the `getEventTracker` behavior though.

dd-trace-api/src/main/java/datadog/trace/api/GlobalTracer.java
smola: `EventTracker.NO_EVENT_TRACKER` would still throw an NPE if called?

ValentinZakharov: Yeah, missed this one. Fixed now!

dd-trace-api/src/main/java/datadog/trace/api/EventTracker.java
```diff
@@ -0,0 +1,109 @@
...
+   * @param metadata custom metadata data represented as key/value map
+   */
+  public void trackLoginFailureEvent(String userId, boolean exists, Map<String, String> metadata) {
+    if (userId == null || userId.isEmpty()) {
+      throw new IllegalArgumentException("UserId is null or empty");
+    }
+
+    if (tracer == null) {
+      return;
+    }
```
bantonsson, line 57: Since the `tracer` is only used in `getTraceSegment` these extra `null` checks could be removed.

bantonsson (approved):

smola (approved):

PR #4610, opened 2023-01-27
Proposal for standardized storage of installable artifacts

robertomonteromiguel:
# What Does This Do

For internal tooling needs (e.g. System Tests), and for having consistent and reliable experience fetching the tracer artifacts. We can leverage a simplified approach of [OCI Artifacts](https://github.com/opencontainers/artifacts) to store the final build artifacts (e.g. dd-java-agent.jar) in GitHub Container Registry.

From GitLab we will generate the docker images with last release and last snapshot on every commit in master branch.

see also: https://docs.google.com/document/d/1HbMGgZFRnIYjfnXGJGyk0LCwWs2op12R0BE58wma448/edit#

# Motivation

Simplify the mechanisms  to download snapshots and releases of the tracer. Simplify tests automation.

# Additional Notes


.gitlab-ci.yml
```diff
@@ -13,6 +13,9 @@ variables:
     value: "false"
     description: "Set to true to override rules in the reliability-env pipeline (e.g. override 'only deploy master')"
 
+include:
+  remote: https://gitlab-templates.ddbuild.io/libdatadog/include/docker-base-image-authenticated.yml
```
pawelchcki, line 17: we could make the name include `ci_authenticated_job` to ensure ppl will understand where the `ci_authenticated_job` comes from

tooling/ci/Dockerfile
```diff
@@ -0,0 +1,20 @@
...
+RUN touch -c -a -m -d @0 /binaries/*
+
+FROM scratch as collect
+COPY --from=java_agent_debug /binaries/dd-java-agent.jar /dd-java-agent.jar
+COPY --from=java_agent_debug /binaries/LIBRARY_VERSION /LIBRARY_VERSION
+COPY --from=java_agent_debug /binaries/LIBDDWAF_VERSION /LIBDDWAF_VERSION
+COPY --from=java_agent_debug /binaries/APPSEC_EVENT_RULES_VERSION /APPSEC_EVENT_RULES_VERSION
+
+FROM scratch
+COPY --from=collect /* /
```
PerfectSlayer, line 20: What is the rationale behind using two more stages from scratch?
Can't we just use one like this:
```
FROM scratch
COPY --from=java_agent_debug /binaries/dd-java-agent.jar /binaries/LIBRARY_VERSION /binaries/LIBDDWAF_VERSION /binaries/APPSEC_EVENT_RULES_VERSION /
```

robertomonteromiguel: you are right. Changed! Thanks!

PerfectSlayer (approved):
LGTM for the tracer part.
Won't judge for the whole OCI artifact initiative :)

pawelchcki (approved):
LGTM too

@PerfectSlayer 
>Won't judge for the whole OCI artifact initiative :)
At this point we're POC'ing it - to see how much it improves our workflow etc. We'll see :man_shrugging:

PR #4616, opened 2023-01-28
data streams: Add Tracking of Kafka commit and produce calls

piochelepiotr:
# What Does This Do

Adds tracking of Kafka commit and Kafka produce calls for the data streams product.
It flushes every 10 seconds the latest offsets of the produce and commit calls for each (topic, partition).

# Motivation

# Additional Notes


dd-java-agent/instrumentation/kafka-clients-0.11/src/main/java/datadog/trace/instrumentation/kafka_clients/ConsumerCoordinatorInstrumentation.java
```diff
@@ -0,0 +1,74 @@
...
+  }
+
+  @Override
+  public void adviceTransformations(AdviceTransformation transformation) {
+    transformation.applyAdvice(
+        isMethod().and(named("sendOffsetCommitRequest")).and(takesArguments(1)),
+        ConsumerCoordinatorInstrumentation.class.getName() + "$CommitOffsetAdvice");
+  }
+
+  public static class CommitOffsetAdvice {
```
amarziali, line 51: I'm wondering if it's not better to track onMethodExit because according to what the method is doing (https://github.com/apache/kafka/blob/6c98544a964b40ede6bbe1b3440f8e5db96a4ad6/clients/src/main/java/org/apache/kafka/clients/consumer/internals/ConsumerCoordinator.java#L1267) we might have situations in which either the offset is invalid either a rebalance is ongoing and the commit is not possible

dd-java-agent/instrumentation/kafka-clients-0.11/src/main/java/datadog/trace/instrumentation/kafka_clients/KafkaConsumerGroupInstrumentation.java
```diff
@@ -85,6 +91,9 @@ public static void muzzleCheck(ConsumerRecord record) {
     }
   }
 
+  // this method transfers the consumer group from the KafkaConsumer class key to the
```
amarziali, line 94: nit: you can use javadoc over a method comment (`/** ... */`)

internal-api/src/main/java/datadog/trace/bootstrap/instrumentation/api/AgentTracer.java
```diff
@@ -174,6 +174,10 @@ private AgentTracer() {}
     AgentSpan.Context notifyExtensionStart(Object event);
 
     void notifyExtensionEnd(AgentSpan span, Object result, boolean isError);
+
+    void trackKafkaProduce(String topic, int partition, long offset);
+
+    void trackKafkaCommit(String consumerGroup, String topic, int partition, long offset);
```
richardstartin, line 180: This feels like the wrong level of abstraction for this interface - it adds integration specific logic into a core service and reduces the `AgentTracer` concept to a registry of globally accessible behaviour. There are two changes I would like to see here:

1. Use `AgentTracer` to locate a DSM specific service, so add `DataStreamsMonitoring getDataStreamsMonitoring()` for the benefit of non-DSM stakeholders in the codebase. This prevents `AgentTracer` from accumulating and churning DSM features, without over-engineering an approach to accessing product integrations. This pattern would follow the profiler integrations I added recently.
2. For the sake of DSM, try to abstract over your future integrations somewhat, or if it's too early to define good abstractions, push more of the Kafka-specific details down into the kafka instrumentation and out of the core.

dougqh: Yeah, I tend to agree.  I think what's missing is a good product agnostic way of tracking context.  DSM needs to somehow tie into header injection & extraction and be carried by the context system, but it needs to do that it a way that doesn't depend on spans -- and is minimally invasive to other products.

I'd hoped to work on context tracking for the last few quarters, but we just haven't had the time.  Maybe we need to make the time before we end up with multiple competing solutions that don't work well together.

amarziali (approved):
Looks good

internal-api/src/main/java/datadog/trace/bootstrap/instrumentation/api/AgentTracer.java
richardstartin: I agree, we don't have perfect abstractions and code will have to be changed in the future, but I'm making a slightly different point. If there are Kafka specific methods here, do we end up with Pulsar and Kinesis methods too? To make the refactoring of the future easier, some kind of modularity and abstraction needs to be maintained, as suggested above.

piochelepiotr: I moved the Kafka specific logic to a `DataStreamsMonitoring` class, and I added a `getDataStreamsMonitoring` method the trace API.
We might refactor the Kafka specific method later for DSM, but for now, it's hard to see if it's something we will end up building for other technologies or no.
What do you think about that approach? At least, we are not complexifying the traceAPI.

richardstartin: That makes sense, as long as it's within a well defined boundary I obviously don't have an opinion about the right level of abstraction. I'm trying to follow similar principles with the profiler integration and don't want to give in to the temptation to make `CoreTracer` a dumping ground for anything and everything global.

richardstartin (approved):
I don't think it was necessary to rename `DataStreamsCheckpointer` to `DataStreamsMonitoring` - that was a just a name I came up with when suggesting how to avoid putting the new logic in `CoreTracer` - but I'm approving just to indicate that my concerns have been addressed. I'm not really a stakeholder in the other details here so please get the relevant approval before merging.

piochelepiotr:
> I don't think it was necessary to rename `DataStreamsCheckpointer` to `DataStreamsMonitoring` - that was a just a name I came up with when suggesting how to avoid putting the new logic in `CoreTracer` - but I'm approving just to indicate that my concerns have been addressed. I'm not really a stakeholder in the other details here so please get the relevant approval before merging.

Thanks! I think I will keep the new name since it's no longer only doing `checkpoints` so the name Checkpointer doesn't fit anymore.

dd-trace-core/src/main/java/datadog/trace/core/CoreTracer.java
```diff
@@ -1083,13 +1087,13 @@ private static String[] generateConstantTags(final Config config) {
   }
 
   @SuppressForbidden
-  private static DataStreamsCheckpointer createDataStreamsCheckpointer(
+  private static DataStreamsMonitoring createDataStreamsCheckpointer(
```
hokitam, line 1090: since we changed the name to DataStreamsMonitoring, do we also want to rename this to createDataStreamsMonitoring?

internal-api/src/main/java/datadog/trace/bootstrap/instrumentation/api/StatsPoint.java
```diff
@@ -67,4 +67,9 @@ public String toString() {
         + edgeLatencyNano
         + '}';
   }
+
+  @Override
```
hokitam, line 71: Instead of having a new `type()` and then checking using something like `if (payload.type() == StatsPayload.Type.StatsGroup)`, I think we can also use `instanceof` in Java if we don't want to maintain the types. Though not sure whether there are any performance implications for either, so feel free to leave as is!

dd-trace-core/src/main/java/datadog/trace/core/datastreams/TopicPartitionGroup.java
```diff
@@ -0,0 +1,39 @@
...
+  public int hashCode() {
+    return partition + topic.hashCode() + group.hashCode();
+  }
+
+  @Override
+  public boolean equals(Object obj) {
+    return obj instanceof TopicPartitionGroup
+        && this.partition == ((TopicPartitionGroup) obj).partition
+        && this.topic == ((TopicPartitionGroup) obj).topic
+        && this.group.equals(((TopicPartitionGroup) obj).group);
```
hokitam, line 37: should we also do null checks on this.group just in case? Not sure if there might be cases in the future where we might try to construct this but have a null group

dd-trace-core/src/main/java/datadog/trace/core/datastreams/TopicPartitionGroup.java
```diff
@@ -0,0 +1,39 @@
...
+
+  public TopicPartitionGroup(String topic, int partition, String group) {
+    this.topic = topic;
+    this.partition = partition;
+    this.group = group;
+  }
+
+  @Override
+  public int hashCode() {
+    return partition + topic.hashCode() + group.hashCode();
```
hokitam, line 29: similar question about null checks for `topic` and `group` just in case

dd-trace-core/src/main/java/datadog/trace/core/datastreams/TopicPartition.java
```diff
@@ -0,0 +1,31 @@
...
+  private final int partition;
+
+  public TopicPartition(String topic, int partition) {
+    this.topic = topic;
+    this.partition = partition;
+  }
+
+  @Override
+  public int hashCode() {
+    return partition + topic.hashCode();
```
hokitam, line 22: maybe null checks for topic here and in the `equals()` method below?

dd-trace-core/src/main/java/datadog/trace/core/datastreams/StatsBucket.java
```diff
@@ -29,6 +32,17 @@ public void addPoint(StatsPoint statsPoint) {
     statsGroup.add(statsPoint.getPathwayLatencyNano(), statsPoint.getEdgeLatencyNano());
   }
 
+  public void addOffset(KafkaOffset offset) {
+    if (offset.getType() == KafkaOffset.Type.COMMIT) {
+      latestKafkaCommitOffsets.put(
```
hokitam, line 37: curious whether there may be cases where an earlier offsets gets called with addOffset later? e.g. we write and somehow get back offset 101 before offset 100?

hokitam: i guess even if that's the case, in the grand scheme of things maybe it's not a huge deal. otherwise I suppose we can do a `get` first and compare the offset value and keep the later one?

dd-trace-core/src/main/java/datadog/trace/core/datastreams/MsgPackDatastreamsPayloadWriter.java
```diff
@@ -129,4 +146,42 @@ private void writeBucket(StatsBucket bucket, Writable packer) {
...
+      Collection<Map.Entry<TopicPartition, Long>> latestProduceOffsets,
+      Writable packer) {
+    packer.writeUTF8(KAFKA);
+    boolean hasCommitOffsets = !latestCommitOffsets.isEmpty();
+    boolean hasProduceOffsets = !latestProduceOffsets.isEmpty();
+    packer.startMap((hasCommitOffsets ? 1 : 0) + (hasProduceOffsets ? 1 : 0));
+    if (hasCommitOffsets) {
+      packer.writeUTF8(LATEST_COMMIT_OFFSETS);
+      packer.startArray(latestCommitOffsets.size());
+      for (Map.Entry<TopicPartitionGroup, Long> entry : latestCommitOffsets) {
```
hokitam, line 161: do we want to clear latestCommitOffsets and latestProduceOffsets after writing them?

hokitam (approved):

dd-trace-core/src/main/java/datadog/trace/core/datastreams/TopicPartitionGroup.java
piochelepiotr: Good point, it looks like it's already handled by the `equals` function: 
```
    public boolean equals(Object anObject) {
        if (this == anObject) {
            return true;
        }
        if (anObject instanceof String) {
            String aString = (String)anObject;
            if (coder() == aString.coder()) {
                return isLatin1() ? StringLatin1.equals(value, aString.value)
                                  : StringUTF16.equals(value, aString.value);
            }
        }
        return false;
    }

```

internal-api/src/main/java/datadog/trace/bootstrap/instrumentation/api/StatsPoint.java
dougqh: instanceof can be moderately expensive when checking against an interface type, but there are speculative optimization within the JVM to handle that.  I think the more important question is why do we need to type sniff.  That seems a bit ugly to me.

dd-trace-core/src/main/java/datadog/trace/core/datastreams/MsgPackDatastreamsPayloadWriter.java
```diff
@@ -75,7 +83,10 @@ public void writePayload(Collection<StatsBucket> data) {
     writer.writeUTF8(STATS);
     writer.startArray(data.size());
     for (StatsBucket bucket : data) {
-      writer.startMap(3);
+      boolean hasKafka =
```
dougqh, line 86: It feels quite invasive to put hasKafka checks directly into the writer.  I feel like this should be abstracted better, but I'm not not quite sure how yet.

dd-trace-core/src/main/java/datadog/trace/core/datastreams/StatsBucket.java
```diff
@@ -9,6 +10,8 @@ public class StatsBucket {
   private final long startTimeNanos;
   private final long bucketDurationNanos;
   private final Map<Long, StatsGroup> hashToGroup = new HashMap<>();
+  private final Map<TopicPartition, Long> latestKafkaProduceOffsets = new HashMap<>();
```
dougqh, line 13: As a general rule, any collection in the tracer needs to have a limit on its size.  Otherwise, we risk unbounded memory consumption.

I think these are limited in a practical sense, so if this is restricted to DSM, I'll let it go for now.

dd-trace-core/src/main/java/datadog/trace/core/datastreams/TopicPartition.java
dougqh: I think you can safely forego null checks.  If you want to be defensive, it might be better to be defensive at construction.

+ alone isn't a great way to compute a hash.
Typically, you'd do something like...

int hash = comp1;
hash = prime * hash + comp2;
return hash;

dd-trace-core/src/main/java/datadog/trace/core/datastreams/TopicPartition.java
```diff
@@ -0,0 +1,31 @@
...
+  }
+
+  @Override
+  public boolean equals(Object obj) {
+    return obj instanceof TopicPartition
```
dougqh, line 27: This probably more complicated than is really required.
This also has an unnecessary cast.  The cast will nicely optimize away, so it isn't a big deal.

if ( !(obj instanceof TopicPartition) ) return false;

TopicPartition that = (TopicPartition)obj;
return this.topic.equals(that.topic);

A fast path for ( obj == this ) can be useful if you reuse objects.

internal-api/src/main/java/datadog/trace/bootstrap/instrumentation/api/KafkaOffset.java
```diff
@@ -0,0 +1,64 @@
+package datadog.trace.bootstrap.instrumentation.api;
+
+public class KafkaOffset implements StatsPayload {
```
dougqh, line 3: StatsPayload doesn't really seem to do much here.  Ultimately, it looks like KafkaOffset is just an immutable object with no special behavior.

For StatsPayload to be useful, I would think it should have one or more abstract methods that encompass how the payload is going to be written.  That would allow the kafka specific code to be removed from the writer as well.

dd-trace-core/src/main/java/datadog/trace/core/datastreams/TopicPartitionGroup.java
piochelepiotr: I changed the constructor to forbid null groups / topics instead. What do you think?

dd-trace-core/src/main/java/datadog/trace/core/datastreams/TopicPartitionGroup.java
piochelepiotr: I changed the constructor to address this.

dd-trace-core/src/main/java/datadog/trace/core/datastreams/StatsBucket.java
piochelepiotr: You should always commit in order, since if you commit up to an offset, not an exact offset. but just in case, I changed the code to handle that

dd-trace-core/src/main/java/datadog/trace/core/datastreams/MsgPackDatastreamsPayloadWriter.java
piochelepiotr: the bucket is removed after being flushed. So I don't think it's important.
https://github.com/DataDog/dd-trace-java/blob/fcee6cdf9ed4fecd00d4d4ea191147b3182b5cbf/dd-trace-core/src/main/java/datadog/trace/core/datastreams/DefaultDataStreamsMonitoring.java#L233

dd-trace-core/src/main/java/datadog/trace/core/datastreams/MsgPackDatastreamsPayloadWriter.java
piochelepiotr: we could abstract logic about offset tracking. So this condition would change to:
hasOffsetTracking
But on the other hand, since we're doing offset tracking only for Kafka right now, I don't know if it's a good idea. What do you think?

dd-trace-core/src/main/java/datadog/trace/core/datastreams/TopicPartition.java
piochelepiotr: ok, I updated the constructor to check for null.
I have to implement these functions just to be able to use TopicPartition as a map key. Do you see a better way of doing it?

internal-api/src/main/java/datadog/trace/bootstrap/instrumentation/api/StatsPoint.java
piochelepiotr: I "have" to type sniff because I have a single queue on which I write both StatsPoints and KafkaOffsets, but then I do different actions with the two. Do you think it would be better to do 2 queues for example? But then, how do you get 1 thread to read 2 queues?

internal-api/src/main/java/datadog/trace/bootstrap/instrumentation/api/KafkaOffset.java
piochelepiotr: I think this is related to https://github.com/DataDog/dd-trace-java/pull/4616/files/b55395d0d1a7c232f023ff44ccb0f21597f8d24d#r1093665724
So depending on what I do for the queues, this might be solved.

internal-api/src/main/java/datadog/trace/bootstrap/instrumentation/api/StatsPoint.java
piochelepiotr: I think I prefer not to group StatsGroup and Offset under a same interface, since they are widely different things.
So I prefer to have the aggregation be different between the two.
But maybe I can do an abstracted offsetTracking mechanism.
So the tricky step is to get the DataStreamsMonitoring thread to handle both StatsGroup and Offset.

I looked at how to have one thread read from two different queues, and people are suggesting putting everything in one queue, and using casting, which I agree is a bit ugly. Do you know of a better way?

piochelepiotr: I checked how it's done for stats: https://github.com/DataDog/dd-trace-java/blob/96aaaf21c8013e149acebe3104d312741b1c4233/dd-trace-core/src/main/java/datadog/trace/common/metrics/Aggregator.java#L111
It looks like it's done via a common interface also. So it doesn't look like there is a much cleaner way to do it. However, I still put some functions in the interface, It looks like it's unnecessary. I will remove it.

dd-trace-core/src/main/java/datadog/trace/core/datastreams/MsgPackDatastreamsPayloadWriter.java
piochelepiotr: I abstracted away the Kafka offsets into backlog information. With a list of tags as the key, and a long as a value.
It can be used for backlog for different technologies, and in different units (bytes for eg).
What do you think of it now?

hokitam (approved):

PR #4619, opened 2023-01-30
Improved logging using placeholders

DDJavierSantos:
# What Does This Do
debug, error and warn logging statements that report both a resource and an exception have been changed so they use placeholders instead of string operations.

# Motivation
Exception reports filed by DDLogger will be sent to our servers using the telemetry API. If the message is composed using placeholders instead of string operations, the placeholders will not be substituted when sending this information over the wire to the telemetry API. This will make sure we don't send any sensitive information to the telemetry service once the telemetry API is merged.

# Additional Notes
This changes are related to the new telemetry functionality implemented by https://github.com/DataDog/dd-trace-java/pull/4521

### DIFF
--- a/dd-java-agent/agent-bootstrap/src/main/java/datadog/trace/bootstrap/Agent.java
+++ b/dd-java-agent/agent-bootstrap/src/main/java/datadog/trace/bootstrap/Agent.java
@@ -640,7 +640,7 @@ public class Agent {
   private static synchronized void initializeJmxSystemAccessProvider(
       final ClassLoader classLoader) {
     if (log.isDebugEnabled()) {
-      log.debug("Initializing JMX system access provider for " + classLoader.toString());
+      log.debug("Initializing JMX system access provider for {}", classLoader);
     }
     try {
       final Class<?> tracerInstallerClass =
@@ -1036,7 +1036,7 @@ public class Agent {
     }
   }
 
-  /** @see datadog.trace.api.ProductActivationConfig#fromString(String) */
+  /** @see datadog.trace.api.ProductActivation#fromString(String) */
   private static boolean isAppSecFullyDisabled() {
     // must be kept in sync with logic from Config!
     final String featureEnabledSysprop = AgentFeature.APPSEC.systemProp;
--- a/dd-java-agent/agent-debugger/src/main/java/com/datadog/debugger/agent/DebuggerTransformer.java
+++ b/dd-java-agent/agent-debugger/src/main/java/com/datadog/debugger/agent/DebuggerTransformer.java
@@ -121,8 +121,10 @@ public class DebuggerTransformer implements ClassFileTransformer {
                   }
                 });
       } catch (IOException ex) {
-        log.warn("Error reading exclude file '{}' for Instrument-The-World: ", fileName, ex);
-      }
+        log.warn(
+            "Error reading exclude file '{}' for Instrument-The-World. Exception: {}",
+            fileName,
+            ex.getMessage());      }
     }
   }
 
@@ -328,7 +330,7 @@ public class DebuggerTransformer implements ClassFileTransformer {
     try {
       classNode.accept(writer);
     } catch (Throwable t) {
-      log.error("Cannot write classfile for class: {}", classFilePath, t);
+      log.error("Cannot write classfile for class: {} Exception: ", classFilePath, t.getMessage());
     }
     byte[] data = writer.toByteArray();
     dumpInstrumentedClassFile(classFilePath, data);
--- a/dd-java-agent/agent-debugger/src/main/java/com/datadog/debugger/agent/DebuggerTransformer.java
+++ b/dd-java-agent/agent-debugger/src/main/java/com/datadog/debugger/agent/DebuggerTransformer.java
@@ -1,8 +1,5 @@
 package com.datadog.debugger.agent;
 
-import static java.util.Collections.singletonList;
-import static java.util.stream.Collectors.toList;
-
 import com.datadog.debugger.instrumentation.DiagnosticMessage;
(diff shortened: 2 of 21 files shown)


dd-java-agent/agent-bootstrap/src/main/java/datadog/trace/bootstrap/blocking/BlockingActionHelper.java
```diff
@@ -252,7 +252,7 @@ private static byte[] readIntoByteArray(File f) {
       log.warn("Template file {} could not be opened", f);
       return null;
     } catch (IOException e) {
-      log.warn("Error reading template file {}", f, e);
+      log.warn("Error reading template file {} Exception: {}", f, e.getMessage());
```
smola, line 255: Exceptions are treated especially, and we generally don't need to add them to the placeholders. I don't think this like (or any like this one) would need any change?

smola (changes requested):

dd-java-agent/agent-bootstrap/src/main/java/datadog/trace/bootstrap/Agent.java
```diff
@@ -579,7 +579,7 @@ private static synchronized void registerDeadlockDetectionEvent() {
   private static synchronized void initializeJmxSystemAccessProvider(
       final ClassLoader classLoader) {
     if (log.isDebugEnabled()) {
-      log.debug("Initializing JMX system access provider for " + classLoader.toString());
+      log.debug("Initializing JMX system access provider for {}", classLoader.toString());
```
smola, line 582: I think we can replace this:
```
    if (log.isDebugEnabled()) {
      log.debug("Initializing JMX system access provider for {}", classLoader.toString());
    }
```

with:

```
log.debug("Initializing JMX system access provider for {}", classLoader);
```

since the `toString()` will be triggered by the logging fornmattibgm and the `isDebugEnabled` call was presumably added to avoid that.

dd-java-agent/agent-profiling/src/main/java/com/datadog/profiling/agent/ProfilingAgent.java
```diff
@@ -159,7 +159,7 @@ public static synchronized void run(final boolean isStartingFirst, ClassLoader a
         log.warn(e.getMessage());
         log.debug("", e);
       } catch (final ConfigurationException e) {
-        log.warn("Failed to initialize profiling agent! " + e.getMessage());
+        log.warn("Failed to initialize profiling agent! {}", e.getMessage());
```
smola, line 162: @richardstartin I guess we might want to log just one or the other? (using the `isDebugEnabled` helper)

dd-smoke-tests/jboss-modules/src/main/java/datadog/smoketest/jbossmodules/common/ServiceSupport.java
```diff
@@ -13,7 +13,7 @@ public final synchronized void start() throws Exception {
       log.info("Started " + this);
       started = true;
     } catch (Exception e) {
-      log.warn("Problem starting " + this, e);
+      log.warn("Problem starting {}" + this, e);
```
smola, line 16:
```suggestion
      log.warn("Problem starting {}", this, e);
```

dd-trace-core/src/main/java/datadog/trace/core/CoreTracer.java
```diff
@@ -630,7 +630,7 @@ private void registerClassLoader(final ClassLoader classLoader) {
         addTraceInterceptor(interceptor);
       }
     } catch (final ServiceConfigurationError e) {
-      log.warn("Problem loading TraceInterceptor for classLoader: " + classLoader, e);
+      log.warn("Problem loading TraceInterceptor for classLoader: {}", classLoader.toString(), e);
```
smola, line 633: Do we need the `toString()` here?

dd-java-agent/agent-bootstrap/src/main/java/datadog/trace/bootstrap/blocking/BlockingActionHelper.java
DDJavierSantos: I have removed those

dd-java-agent/agent-bootstrap/src/main/java/datadog/trace/bootstrap/Agent.java
DDJavierSantos: done

smola (approved):

jpbempel (approved):

richardstartin (approved):

dd-java-agent/agent-debugger/src/main/java/com/datadog/debugger/agent/DebuggerTransformer.java
```diff
@@ -328,7 +331,7 @@ private byte[] writeClassFile(ClassLoader loader, String classFilePath, ClassNod
     try {
       classNode.accept(writer);
     } catch (Throwable t) {
-      log.error("Cannot write classfile for class: {}", classFilePath, t);
+      log.error("Cannot write classfile for class: {} Exception: ", classFilePath, t.getMessage());
```
bantonsson, line 334: Is this missing a placeholder?

DDJavierSantos: removed

jpbempel: Sorry late to the party, but here you have changed the behavior of our log here, I want the full exception and not just the message.
Other places with exception are not changed like here.
Please revert this.

PR #4620, opened 2023-01-30
pin buildx version for publishing to GCP

devinsba:
# What Does This Do

There is a slack thread where lib injection images cannot be pushed to GCR due to changes to buildx

# Motivation

# Additional Notes


smola (approved):

devinsba:
fixes #4658

PR #4621, opened 2023-01-30
Add IAST smoke test to spring modules

jandro996:
# What Does This Do
- Improve code reusability
- Add iast smoke test to springboot-2.6-webmvc 

# Motivation
- Run IAST test with more spring versions

# Additional Notes
- first iteration, we must add IAST tests to springboot-3.0-webmvc


dd-smoke-tests/src/main/groovy/datadog/smoketest/AbstractIastSpringBootSmokeTest.groovy
```diff
@@ -0,0 +1,392 @@
+package datadog.smoketest
```
smola, line 1: Maybe it's better to create a spring-common  module for these?

PR #4622, opened 2023-01-30
Reuse TaintedMap instances in IAST

smola:
# What Does This Do
Use an object pool for `TaintedObject` instances.

# Motivation

Allocating 16k TaintedObject[] tables within TaintedMap creates a considerable GC churn. We use a small object pool for TaintedObjets instead, and use it to reuse instances between requests.

Note that, with asynchronous processing, there might be cases where a request gets a TaintedObject instance that is already being used by next request. This should be no problem: usually get calls will find nothing.

This change means that TaintedMap.clear needs to be thread-safe, so added tests checking that.

# Additional Notes

### DIFF
--- a/dd-java-agent/agent-iast/src/main/java/com/datadog/iast/taint/TaintedObjects.java
+++ b/dd-java-agent/agent-iast/src/main/java/com/datadog/iast/taint/TaintedObjects.java
@@ -4,9 +4,11 @@ import com.datadog.iast.IastSystem;
 import com.datadog.iast.model.Range;
 import com.datadog.iast.model.Source;
 import com.datadog.iast.model.json.TaintedObjectEncoding;
+import datadog.trace.api.Config;
 import java.util.ArrayList;
 import java.util.List;
 import java.util.UUID;
+import java.util.concurrent.ArrayBlockingQueue;
 import javax.annotation.Nonnull;
 import org.slf4j.Logger;
 import org.slf4j.LoggerFactory;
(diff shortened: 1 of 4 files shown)


dd-java-agent/agent-iast/src/main/java/com/datadog/iast/taint/TaintedObjects.java
```diff
@@ -21,20 +23,26 @@ public interface TaintedObjects {
...
+    TaintedObjectsImpl taintedObjects = TaintedObjectsImpl.pool.poll();
+    if (taintedObjects == null) {
+      taintedObjects = new TaintedObjectsImpl();
+    }
     return IastSystem.DEBUG ? new TaintedObjectsDebugAdapter(taintedObjects) : taintedObjects;
   }
 
   class TaintedObjectsImpl implements TaintedObjects {
 
+    private static final ArrayBlockingQueue<TaintedObjectsImpl> pool =
```
manuel-alvarez-alvarez, line 36: Any thoughts on using a non blocking queue from `org.jctools.queues`?

manuel-alvarez-alvarez: It's part of `dd-trace-core` and since this executes just before and after a request probably there won't be any significant gains from it. 

By the way, `Config#getIastMaxConcurrentRequests()` seems more appropriate for me as the upper bound for the queue.

manuel-alvarez-alvarez (approved):

dd-java-agent/agent-iast/src/main/java/com/datadog/iast/taint/TaintedObjects.java
smola: Absolutely. I accidentally picked the wrong variable. I think we can iterate again on the exact queue implementation here.

PR #4623, opened 2023-01-30
Fix log template evaluation

jpbempel:
# What Does This Do
ValueScript is storing result value in a field which can be overwritten by another execution. Now the execute method returns directly the result value and ValueScript became effectively stateless regarding execution.

# Motivation
Bug in log template evaluation for concurrent requests

# Additional Notes

### DIFF
--- a/dd-java-agent/agent-debugger/src/main/java/com/datadog/debugger/agent/LogMessageTemplateSummaryBuilder.java
+++ b/dd-java-agent/agent-debugger/src/main/java/com/datadog/debugger/agent/LogMessageTemplateSummaryBuilder.java
@@ -22,6 +22,7 @@ public class LogMessageTemplateSummaryBuilder implements SummaryBuilder {
 
   private final LogProbe logProbe;
   private final List<Snapshot.EvaluationError> evaluationErrors = new ArrayList<>();
+  private String message;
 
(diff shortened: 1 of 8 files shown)


dd-java-agent/agent-debugger/src/main/java/com/datadog/debugger/agent/LogMessageTemplateSummaryBuilder.java
```diff
@@ -47,26 +48,10 @@ public void addStack(List<CapturedStackFrame> stack) {}
 
   @Override
   public String build() {
```
shatzi, line 50: find this method name now confusing... with this is not getMessage and executeExpressions is build?

ojung (approved):
👍

dd-java-agent/agent-debugger/src/main/java/com/datadog/debugger/agent/LogMessageTemplateSummaryBuilder.java
jpbempel: This is one implementation of `SummaryBuilder` with the original idea that we need to gather all infos about the snapshot before building the summary, hence the builder.
With LogTemplate we move the building of the summary at context adding. 
If we get rid of the `SnapshotSummaryBuilder` we can revisit it :)

PR #4626, opened 2023-01-31
Add version to probe config objects

ojung:
# What Does This Do

Add probe version to snapshots and probe statuses.

# Motivation

Probes are mutable in the new model.

# Additional Notes

### DIFF
--- a/dd-java-agent/agent-debugger/src/main/java/com/datadog/debugger/agent/ProbeStatus.java
+++ b/dd-java-agent/agent-debugger/src/main/java/com/datadog/debugger/agent/ProbeStatus.java
@@ -7,6 +7,7 @@ import com.squareup.moshi.JsonWriter;
 import com.squareup.moshi.Moshi;
 import datadog.trace.api.Config;
 import datadog.trace.bootstrap.debugger.CapturedStackFrame;
+import datadog.trace.bootstrap.debugger.ProbeId;
 import datadog.trace.util.TagsHelper;
 import java.io.IOException;
 import java.lang.annotation.Annotation;
@@ -104,19 +105,24 @@ public class ProbeStatus {
 
   /** Stores status information for a probe */
   public static class Diagnostics {
+
     private final String probeId;
+
+    private final int probeVersion;
+
     private final Status status;
 
     private final ProbeException exception;
 
-    public Diagnostics(String probeId, Status status, ProbeException exception) {
-      this.probeId = probeId;
+    public Diagnostics(ProbeId probeId, Status status, ProbeException exception) {
+      this.probeId = probeId != null ? probeId.getId() : null;
+      this.probeVersion = probeId != null ? probeId.getVersion() : 0;
       this.status = status;
       this.exception = exception;
     }
 
-    public String getProbeId() {
-      return probeId;
+    public ProbeId getProbeId() {
+      return new ProbeId(probeId, probeVersion);
     }
 
     public Status getStatus() {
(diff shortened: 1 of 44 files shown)


dd-java-agent/agent-debugger/src/main/java/com/datadog/debugger/agent/ProbeStatus.java
```diff
@@ -104,19 +105,19 @@ public String toString() {
...
-      this.probeId = probeId;
+    public Diagnostics(ProbeId probeId, Status status, ProbeException exception) {
+      this.probe = probeId;
       this.status = status;
       this.exception = exception;
     }
 
-    public String getProbeId() {
-      return probeId;
+    public ProbeId getProbe() {
```
jpbempel, line 119: same here, keep probeId

dd-java-agent/agent-debugger/src/main/java/com/datadog/debugger/probe/ProbeDefinition.java
```diff
@@ -52,12 +53,13 @@ public static Snapshot.MethodLocation convert(MethodLocation methodLocation) {
   public ProbeDefinition(
       String language,
       String id,
+      int version,
```
jpbempel, line 56: Can we maybe use ProbeId class here also as well?

dd-java-agent/agent-debugger/src/main/java/com/datadog/debugger/probe/ProbeDefinition.java
```diff
@@ -66,7 +68,11 @@ public ProbeDefinition(
   }
 
   public String getId() {
```
jpbempel, line 70: do we need to keep this method?

dd-java-agent/agent-debugger/src/main/java/com/datadog/debugger/probe/ProbeDefinition.java
```diff
@@ -153,6 +160,11 @@ public T probeId(String probeId) {
       return (T) this;
     }
 
+    public T version(int version) {
+      this.version = version;
+      return (T) this;
+    }
+
```
jpbempel, line 167: suggestion: having a `probeId` method taking both `id` & `version`?

jpbempel: and another taking directly `ProbeId`

jpbempel (commented):
Thanks Looks good, small suggestions

dd-java-agent/agent-debugger/src/main/java/com/datadog/debugger/agent/ProbeStatus.java
```diff
@@ -104,19 +105,19 @@ public String toString() {
 
   /** Stores status information for a probe */
   public static class Diagnostics {
-    private final String probeId;
+    private final ProbeId probe;
```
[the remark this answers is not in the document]
ojung: Yeah, I was modeling the class after what it should look like in json. With jackson I would have used `@JsonProperty`, but since this isn't jackson I didn't spend the time figuring out how to do it with moshi. I'll look into it 👍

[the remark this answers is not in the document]
jpbempel: you can use `@Json` annotation
https://github.com/DataDog/dd-trace-java/blob/master/dd-java-agent/agent-debugger/src/main/java/com/datadog/debugger/agent/JsonSnapshotSerializer.java#L115

dd-java-agent/agent-debugger/src/main/java/com/datadog/debugger/probe/ProbeDefinition.java
ojung: I think it's difficult because of the Inheritance. The only way I see is having an adapter per probe type (classes extending `ProbeDefinition`) that handles this. Can you think of a simpler way? I haven't used Moshi before so I might be overcomplicating this in my head.

jpbempel (approved):

jpbempel (approved):

PR #4634, opened 2023-02-01
Add support for undertow-jakarta-servlet

amarziali:
# What Does This Do

Add compatibility with undertow-jakarta-servlet (undertow-2.2+)

# Motivation

# Additional Notes


dd-java-agent/instrumentation/undertow-2.2/src/main/java/datadog/trace/instrumentation/undertow/JakartaServletInstrumentation.java
```diff
@@ -0,0 +1,65 @@
...
+      packageName + ".UndertowExtractAdapter$Response"
+    };
+  }
+
+  public static class DispatchAdvice {
+    @Advice.OnMethodEnter(suppress = Throwable.class)
+    public static void enter(
+        @Advice.Argument(0) final HttpServerExchange exchange,
+        @Advice.Argument(1) final ServletRequestContext servletRequestContext) {
+      AgentSpan undertow_span = exchange.getAttachment(DD_UNDERTOW_SPAN);
```
richardstartin, line 54: sorry to nitpick 😅 but this should be camel case

amarziali: lol completely right richard. I copied paste from the undertow-2.0. I'll change also here so!

devinsba (approved):
Looks right to me. I assume this is mostly a direct copy of the servlet related code from 2.0 with just the servlet packages changed?

amarziali:
> Looks right to me. I assume this is mostly a direct copy of the servlet related code from 2.0 with just the servlet packages changed?

Thanks for the review. It's a copy from the undertow package rather since it was tied to javax.servlet. But undertow can also work with the jakarta variant and this was not supported

PR #4635, opened 2023-02-01
Add IAST smoke test to springboot 2.X

jandro996:
# What Does This Do
- Improve code reusability
- Add iast smoke test to springboot-2.6-webmvc 

# Motivation
- Run IAST test with more spring versions

# Additional Notes
- first iteration, we must add IAST tests to springboot-3.0-webmvc


dd-smoke-tests/iast/build.gradle
```diff
@@ -0,0 +1,7 @@
+apply from: "$rootDir/gradle/java.gradle"
```
smola, line 1: Can we name this submodule as something like `iast-util`?

smola (approved):

PR #4640, opened 2023-02-02
activate grpc.client span in ClientCall.request and ClientCall.halfClose

richardstartin:
# What Does This Do

Activates the client span in the request method, which ensures context propagation throughout the call. 
# Motivation

Better profiler integration for code hotspots.

# Additional Notes

### DIFF
--- a/dd-java-agent/instrumentation/grpc-1.5/src/main/java/datadog/trace/instrumentation/grpc/client/ClientCallImplInstrumentation.java
+++ b/dd-java-agent/instrumentation/grpc-1.5/src/main/java/datadog/trace/instrumentation/grpc/client/ClientCallImplInstrumentation.java
@@ -48,6 +48,12 @@ public final class ClientCallImplInstrumentation extends Instrumenter.Tracing
     transformation.applyAdvice(named("start").and(isMethod()), getClass().getName() + "$Start");
     transformation.applyAdvice(named("cancel").and(isMethod()), getClass().getName() + "$Cancel");
     transformation.applyAdvice(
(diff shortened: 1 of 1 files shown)


PerfectSlayer (approved):
LGTM

dd-java-agent/instrumentation/grpc-1.5/src/main/java/datadog/trace/instrumentation/grpc/client/ClientCallImplInstrumentation.java
```diff
@@ -48,8 +48,11 @@ public void adviceTransformations(AdviceTransformation transformation) {
     transformation.applyAdvice(named("start").and(isMethod()), getClass().getName() + "$Start");
     transformation.applyAdvice(named("cancel").and(isMethod()), getClass().getName() + "$Cancel");
     transformation.applyAdvice(
-        named("request").and(isMethod()).and(takesArguments(int.class)),
-        getClass().getName() + "$Request");
+        named("request")
+            .and(isMethod())
+            .and(takesArguments(int.class))
+            .or(named("halfClose").and(takesArguments(0))),
```
PerfectSlayer, line 54: Does the 2nd matcher also need `isMethod()` like the first one?

richardstartin: Probably, I don't think it actually matters but I'll add it in for the sake of being thorough

PR #4644, opened 2023-02-02
Double the size of CI worker for builds

smola:
# What Does This Do

# Motivation

# Additional Notes


smola:
It does not seem worth it for the gain (build ~12:30 -> ~10:40).

PR #4647, opened 2023-02-03
Use fixed commits for system-tests/lib-injection

smola:
# What Does This Do

# Motivation

# Additional Notes


mcculls (approved):
I assume we'll only need to update these runners occasionally (if at all)

smola:
> I assume we'll only need to update these runners occasionally (if at all)

Hopefully.

PR #4649, opened 2023-02-03
Fix mixing capturing and non-capturing log probes

jpbempel:
# What Does This Do
Creates a copy of the current snapshot and depending of the probe definition indicates capturing or not, includes the captured data (`Captures` object)

# Motivation
When creating 2 log probes on the same method (duplicated), one is capturing the other not, snapshots with captured data included are sent in both cases.

# Additional Notes

### DIFF
--- a/dd-java-agent/agent-debugger/debugger-bootstrap/src/main/java/datadog/trace/bootstrap/debugger/Snapshot.java
+++ b/dd-java-agent/agent-debugger/debugger-bootstrap/src/main/java/datadog/trace/bootstrap/debugger/Snapshot.java
@@ -216,31 +216,26 @@ public class Snapshot {
        * - Snapshot.commit()
        */
       recordStackTrace(3);
-      if (currentProbeId.equals(probe.id)) {
-        if (status.hasErrors) {
-          evaluationErrors = errorsByProbeIds.get(currentProbeId);
-        }
-      }
-      DebuggerContext.addSnapshot(copy(status.probeDetails, UUID.randomUUID().toString()));
+      DebuggerContext.addSnapshot(duplicateSnapshotForProbe(status.probeDetails));
     }
   }
 
-  private Snapshot copy(ProbeDetails additionalProbe, String newSnapshotId) {
+  private Snapshot duplicateSnapshotForProbe(ProbeDetails probe) {
     Snapshot snapshot =
         new Snapshot(
-            newSnapshotId,
+            UUID.randomUUID().toString(),
             version,
             timestamp,
             duration,
             stack,
-            additionalProbe.captureSnapshot ? captures : new Captures(),
-            additionalProbe,
+            probe.captureSnapshot ? captures : new Captures(),
+            probe,
             language,
             thread,
             thisClassName,
             traceId,
             spanId);
-    List<EvaluationError> evalErrors = errorsByProbeIds.get(additionalProbe.id);
+    List<EvaluationError> evalErrors = errorsByProbeIds.get(probe.id);
     if (evalErrors != null) {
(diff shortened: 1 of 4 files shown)


dd-java-agent/agent-debugger/debugger-bootstrap/src/main/java/datadog/trace/bootstrap/debugger/Snapshot.java
```diff
@@ -237,7 +233,7 @@ private Snapshot copy(ProbeDetails additionalProbe, String newSnapshotId) {
             timestamp,
             duration,
             stack,
-            captures,
+            additionalProbe.captureSnapshot ? captures : new Captures(),
```
shatzi, line 236: suggest to change this variable name to "newProbe" or "forProbe".
while it was additional Probe before, now it the current probe for the copied version of the probe.

maybe also change the name of the function duplicateSnapshotForProbe
Do we need the second argument? as we always just generate a random UUID for duplicated snapshot

dd-java-agent/agent-debugger/debugger-bootstrap/src/main/java/datadog/trace/bootstrap/debugger/Snapshot.java
```diff
@@ -222,10 +220,8 @@ public void commit() {
         if (status.hasErrors) {
```
shatzi, line 220: I'm not sure we need this now as it also happen on the copy function. line no. 243
So we might just remove this?

dd-java-agent/agent-debugger/src/test/resources/com/datadog/debugger/sink/snapshotWithEvalErrorRegex.txt
```diff
@@ -4,7 +4,6 @@
 "snapshot":\{
 "captures":\{\},
 "evaluationErrors":\[\{"expr":"obj.field","message":"Cannot dereference obj"\}\],
-"id":"[^"]+",
```
shatzi, line 7: not sure why we don't have an ID as the snapshot copy always add id.

shatzi (approved):
LGTM.

dd-java-agent/agent-debugger/src/test/resources/com/datadog/debugger/sink/snapshotWithEvalErrorRegex.txt
jpbempel: because of the way we unit test this. we are not using a sink for that, so no id generated.

dd-java-agent/agent-debugger/debugger-bootstrap/src/main/java/datadog/trace/bootstrap/debugger/Snapshot.java
jpbempel: right, thanks

dd-java-agent/agent-debugger/debugger-bootstrap/src/main/java/datadog/trace/bootstrap/debugger/Snapshot.java
jpbempel: done

dd-java-agent/agent-debugger/src/test/resources/com/datadog/debugger/sink/snapshotWithEvalErrorRegex.txt
jpbempel: actually I have changed the unit test to really use the sink, and adapted the regex

PR #4653, opened 2023-02-03
Profiler API to record time spent in queues

richardstartin:
# What Does This Do

# Motivation

# Additional Notes

### DIFF
--- a/dd-java-agent/agent-profiling/profiling-ddprof/src/main/java/com/datadog/profiling/ddprof/DatadogProfiler.java
+++ b/dd-java-agent/agent-profiling/profiling-ddprof/src/main/java/com/datadog/profiling/ddprof/DatadogProfiler.java
@@ -38,6 +38,7 @@ import java.util.ArrayList;
 import java.util.EnumSet;
 import java.util.List;
 import java.util.Set;
+import java.util.concurrent.TimeUnit;
 import java.util.concurrent.atomic.AtomicBoolean;
 import java.util.concurrent.locks.LockSupport;
 import javax.annotation.Nullable;
@@ -343,5 +344,11 @@ public final class DatadogProfiler {
       return contextSetter.clearContextValue(attribute);
     }
     return false;
+  }
+
+  public void recordQueueingTime(long time, TimeUnit unit) {
+    if (profiler != null) {
+      profiler.recordQueueingTime(time, unit);
+    }
   }
 }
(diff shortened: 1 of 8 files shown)


dd-java-agent/agent-profiling/profiling-ddprof/src/main/java/com/datadog/profiling/ddprof/DatadogProfiler.java
```diff
@@ -344,4 +349,12 @@ public boolean clearContextValue(String attribute) {
     }
     return false;
   }
+
+  public boolean isQueueingTimeEnabled() {
+    return isQueueingTimeEnabled;
+  }
+
+  public void recordQueueingTime(long time, TimeUnit unit) {
+    // TODO
```
richardstartin, line 358: this requires a profiler update

dd-trace-core/src/test/groovy/datadog/trace/core/scopemanager/ScopeAndContinuationLayoutTest.groovy
```diff
@@ -14,11 +14,11 @@ class ScopeAndContinuationLayoutTest extends DDSpecification {
   }
 
   def "single continuation layout"() {
-    expect: layoutAcceptable(ContinuableScopeManager.SingleContinuation, 32)
+    expect: layoutAcceptable(ContinuableScopeManager.SingleContinuation, 40)
```
richardstartin, line 17: this could stay at 32 if, for example, the reference to `AgentTrace` were removed. We can get away with just 32 bits for the timestamp because we know the queuing duration will be less than 25 days.

PerfectSlayer: How would you make `Continuation` work without an `AgentTrace` instance?
For the record, I replace it by a _Context_ instance in my PoC.

dd-trace-core/src/main/java/datadog/trace/core/scopemanager/ContinuableScopeManager.java
```diff
@@ -659,18 +670,22 @@ private static final class SingleContinuation extends Continuation {
...
       super(scopeManager, spanUnderScope, source);
+      timestampMillis = scopeManager.getContinuationTimestamp();
     }
 
     @Override
     public AgentScope activate() {
       if (USED.compareAndSet(this, 0, 1)) {
-        return scopeManager.continueSpan(this, spanUnderScope, source);
+        AgentScope scope = scopeManager.continueSpan(this, spanUnderScope, source);
+        scopeManager.recordContinuationLifeSpan(timestampMillis);
```
PerfectSlayer, line 687: Any implicit reason to call `ScopeManager.continueSpan()` before `ScopeManager.recordContinuationLifeSpan()`?

richardstartin: as discussed offline, I think we can wait for changes in this area and keep this PR scoped to the profiler changes

jbachorik (approved):

PR #4654, opened 2023-02-03
Split check job out of the critical path

smola:
# What Does This Do
* Split `check` job out of the critical path. If it fails, it cancels the whole workflow.
* Add a `spotless` job that runs early in the pipeline. No tests are run if formatting fails.
* Skip `buildSrc` tests in every job except in `check`.

# Motivation
So we don't delay starting tests, but if check fails, we still cancel the whole workflow.

# Additional Notes

### DIFF
--- a/.circleci/cancel_workflow.sh
+++ b/.circleci/cancel_workflow.sh
@@ -1,4 +1,4 @@
 #!/usr/bin/env bash
 set -eu
-echo "Canceliing workflow ${CIRCLE_WORKFLOW_ID}"
+echo "Cancelling workflow ${CIRCLE_WORKFLOW_ID}"
 curl --request POST --url https://circleci.com/api/v2/workflow/$CIRCLE_WORKFLOW_ID/cancel --header "Circle-Token: ${CIRCLE_TOKEN}"
--- a/.circleci/cancel_workflow.sh
+++ b/.circleci/cancel_workflow.sh
@@ -1,0 +1,4 @@
+#!/usr/bin/env bash
+set -eu
+echo "Canceliing workflow ${CIRCLE_WORKFLOW_ID}"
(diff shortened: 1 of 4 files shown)


richardstartin (changes requested):
If I've got a formatting error, I want to know straight away

richardstartin (approved):

.circleci/cancel_workflow.sh
```diff
@@ -0,0 +1,4 @@
+#!/usr/bin/env bash
+set -eu
+echo "Canceliing workflow ${CIRCLE_WORKFLOW_ID}"
```
bantonsson, line 3: Cancel*ii*ng

bantonsson (approved):

.circleci/config.yml
```diff
@@ -275,6 +275,47 @@ jobs:
 
       - display_memory_usage
 
+  spotless:
```
bantonsson, line 278: Is there a particular reason that the `spotless` job doesn't use `.circleci/cancel_workflow.sh` the same way as `check`?

smola: While `check` runs concurrently to all test jobs, all downstream jobs depend on both `build` and `spotless`, so a spotless failure already prevents every job from running, except `build`. I thought it wouldn't be as relevant in that case, but I have no strong opinion on that.

bantonsson: No strong opinion either. It was only that the number of dependency lines was getting a bit cluttered. Another option would be to have `build` and `spotless` combine into `ok to test` and all tests depend on that.

smola: Sounds good, I think we'll survive to the extra 3 seconds of the fan-in job :wink: 
Here's an example: https://app.circleci.com/pipelines/github/DataDog/dd-trace-java/21488/workflows/4bf34acd-8ff0-48bb-9b7b-e116bb49fd6b

Issue #4661, opened 2023-02-05
GraphQL query sanitizer is overzealous


theJC:
The dd-trace-java GraphQL query sanitizer is overzealous in how much it changes.

For example, queries that use input variables, the use of the input variables within the query gets wiped out and requires one to go find the real query to understand what it really was doing.

GraphQLQuerySanitizer should be changed to only sanitize literals that could contain sensitive information, ie: leave Enums and variable references alone.

Use dd-trace-js's implementation as reference, as that has worked perfectly for us, and it basically copies the Apollo sanitization/normalization algorithm which is used throughout their product offering, and has worked well.

https://github.com/DataDog/dd-trace-js/blob/7b27059fe3b9de6788c0517a2b499ff6b78da5ff/packages/datadog-plugin-graphql/src/tools/transforms.js#L13

PR #4662, opened 2023-02-05
Improve graphql sanitizer to only sanitize value types that are neces…

theJC:
…sary/could contain sensitive info

# What Does This Do

Fixes https://github.com/DataDog/dd-trace-java/issues/4661

# Motivation

Currently the GraphQL sanitization implementation changes too much/wipes out useful information that would not be considered sensitive/needing sanitization, such as variable references and use of Enums.

# Additional Notes

dd-trace-js provides a good implementation which was borrowed from Apollo.  Visitors in this MR align with what visitors are used in it:

https://github.com/DataDog/dd-trace-js/blob/7b27059fe3b9de6788c0517a2b499ff6b78da5ff/packages/datadog-plugin-graphql/src/tools/transforms.js#L13

### DIFF
--- a/dd-java-agent/instrumentation/graphql-java-14.0/src/main/java/datadog/trace/instrumentation/graphqljava/GraphQLQuerySanitizer.java
+++ b/dd-java-agent/instrumentation/graphql-java-14.0/src/main/java/datadog/trace/instrumentation/graphqljava/GraphQLQuerySanitizer.java
@@ -42,5 +42,4 @@ public final class GraphQLQuerySanitizer extends NodeVisitorStub {
   public TraversalControl visitStringValue(StringValue node, TraverserContext<Node> context) {
     return TreeTransformerUtil.changeNode(context, new EnumValue("{String}"));
   }
-
 }
--- a/dd-java-agent/instrumentation/graphql-java-14.0/src/main/java/datadog/trace/instrumentation/graphqljava/GraphQLQuerySanitizer.java
+++ b/dd-java-agent/instrumentation/graphql-java-14.0/src/main/java/datadog/trace/instrumentation/graphqljava/GraphQLQuerySanitizer.java
@@ -1,18 +1,21 @@
 package datadog.trace.instrumentation.graphqljava;
 
+import graphql.language.ArrayValue;
 import graphql.language.AstPrinter;
 import graphql.language.AstTransformer;
-import graphql.language.BooleanValue;
-import graphql.language.EnumValue;
+import graphql.language.FloatValue;
+import graphql.language.IntValue;
 import graphql.language.Node;
 import graphql.language.NodeVisitor;
 import graphql.language.NodeVisitorStub;
-import graphql.language.NullValue;
-import graphql.language.Value;
-import graphql.language.VariableReference;
+import graphql.language.ObjectValue;
+import graphql.language.StringValue;
 import graphql.util.TraversalControl;
 import graphql.util.TraverserContext;
 import graphql.util.TreeTransformerUtil;
+import java.math.BigDecimal;
+import java.math.BigInteger;
+import java.util.Collections;
 
 public final class GraphQLQuerySanitizer extends NodeVisitorStub {
   private static final NodeVisitor REPLACE_VALUES_VISITOR = new GraphQLQuerySanitizer();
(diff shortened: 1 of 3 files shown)


dd-java-agent/instrumentation/graphql-java-14.0/src/main/java/datadog/trace/instrumentation/graphqljava/GraphQLQuerySanitizer.java
```diff
@@ -23,25 +26,34 @@ public static String sanitizeQuery(Node<?> node) {
...
   }
 
+  private static final IntValue sanitizedIntValue = new IntValue(BigInteger.ZERO);
+  private static final FloatValue sanitizedFloatValue = new FloatValue(BigDecimal.ZERO);
+  private static final StringValue sanitizedStringValue = new StringValue("");
+  private static final ObjectValue sanitizedObjectValue = new ObjectValue(Collections.EMPTY_LIST);
+  private static final ArrayValue santitizedArrayValue = new ArrayValue(Collections.EMPTY_LIST);
+
+  @Override
+  public TraversalControl visitIntValue(IntValue node, TraverserContext<Node> context) {
```
ygree, line 36: Would be nice to have these cases also covered in the tests.

dd-java-agent/instrumentation/graphql-java-14.0/src/test/groovy/GraphQLTest.groovy
```diff
@@ -57,8 +58,8 @@ class GraphQLTest extends AgentTestRunner {
 
   def "successful query produces spans"() {
     setup:
-    def query = 'query findBookById {\n' +
-      '  bookById(id: "book-1") {\n' +
+    def query = 'query findBookById($id: ID) {\n' +
```
ygree, line 61: Good to have this test case for named parameters. I think that it will work with current implementation the same way though preserving the named parameters as is in the sanitized queries.

dd-java-agent/instrumentation/graphql-java-14.0/src/test/groovy/GraphQLTest.groovy
```diff
@@ -177,7 +179,7 @@ class GraphQLTest extends AgentTestRunner {
       '  }\n' +
       '}'
     def expectedQuery = 'query findBookById {\n' +
-      '  bookById(id: ?) {\n' +
+      '  bookById(id: "") {\n' +
```
ygree, line 182: Now sanitized values will be replaced with defaults which makes them indistinguishable from the real values. Is it possible to use `?` for all sanitized values to make this more explicit?

ygree (changes requested):
Thanks for your contribution. Could you provide an example where current implementation is overzealous? Would be great to have it as a test case. From current changes I only see that `?` is now `""`.

dd-java-agent/instrumentation/graphql-java-14.0/src/test/groovy/GraphQLTest.groovy
theJC: I dont see why not.

I like your suggestion, its a blend of what I'm try to improve and what was being done before - best of both worlds.  I'll give it a whirl, thanks for the 👀 and suggestion!

theJC: One thing that this updated method of sanitization does for the reader is they already have an idea of what type the constant is, versus if we replace the values with a Enum ?, then they have to go reference something else to refresh/review what type it is.

theJC: Based on this feedback, how about using different enum values that provide hints to the customer as to which type.  I've pushed an update showing an example of this.

dd-java-agent/instrumentation/graphql-java-14.0/src/main/java/datadog/trace/instrumentation/graphqljava/GraphQLQuerySanitizer.java
theJC: Updated, created a separate test class that just tests sanitization by itself and hit all the cases.

theJC:
Added an additional test case that showed the variable references being sanitized which doesn't seem like it was necessary and lost information that could be useful.

dd-java-agent/instrumentation/graphql-java-14.0/src/test/groovy/GraphQLTest.groovy
theJC: @ygree -- Whats your take on my proposal to preserve some type fingerprint on the nodes that are sanitized?  

If we can get alignment on that front, I think everything else mentioned/discussed in this MR has been addressed.

ygree (approved):

PR #4665, opened 2023-02-06
Fix duplicated probe not instrumented

jpbempel:
# What Does This Do
copy a log probe when receive new configuration to detect changes for a duplicated probe

# Motivation
Since we keep the same instance of deserialized probe and for duplicated we add them in the log definition as additional probes, changes are not detected with ConfigurationComparer

# Additional Notes


ojung (approved):
Nice, I think this will help down the line 👍

PR #4666, opened 2023-02-06
Add agent connection reliability tests

PerfectSlayer:
# What Does This Do
 
Add some agent connection reliability tests to ensure the tracer is capable of sending traces on late agent start or after agent restart.

# Motivation

Ensure reliable tracer connection to the agent. 

# Additional Notes

### DIFF
--- a/communication/src/main/java/datadog/communication/ddagent/DDAgentFeaturesDiscovery.java
+++ b/communication/src/main/java/datadog/communication/ddagent/DDAgentFeaturesDiscovery.java
@@ -107,7 +107,11 @@ public class DDAgentFeaturesDiscovery implements DroppingPolicy {
 
   /** Run feature discovery, if it was not run recently. */
   public void discoverIfOutdated() {
-    discoverIfOutdated(MIN_FEATURE_DISCOVERY_INTERVAL_MILLIS);
+    discoverIfOutdated(getFeaturesDiscoveryMinDelayMillis());
+  }
+
+  protected long getFeaturesDiscoveryMinDelayMillis() {
+    return MIN_FEATURE_DISCOVERY_INTERVAL_MILLIS;
   }
 
   private synchronized void discoverIfOutdated(final long maxElapsedMs) {
--- a/dd-trace-core/src/test/groovy/datadog/trace/TracerConnectionReliabilityTest.groovy
+++ b/dd-trace-core/src/test/groovy/datadog/trace/TracerConnectionReliabilityTest.groovy
@@ -1,0 +1,155 @@
+package datadog.trace
+
+import com.squareup.moshi.JsonAdapter
+import com.squareup.moshi.Moshi
+import com.squareup.moshi.Types
+import datadog.communication.ddagent.DDAgentFeaturesDiscovery
+import datadog.communication.ddagent.SharedCommunicationObjects
+import datadog.communication.monitor.Monitoring
+import datadog.trace.agent.test.utils.PortUtils
+import datadog.trace.api.IdGenerationStrategy
+import datadog.trace.core.CoreTracer
+import datadog.trace.test.util.DDSpecification
+import okhttp3.OkHttpClient
+import okhttp3.Request
+import org.testcontainers.containers.FixedHostPortGenericContainer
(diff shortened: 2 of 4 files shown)


dd-trace-core/src/test/groovy/datadog/trace/TracerConnectionReliabilityTest.groovy
```diff
@@ -0,0 +1,155 @@
...
+  def createSpans(int count, int delay) {
+    for (def index: 1..count) {
+      def span = tracer.buildSpan("operation-${index}").start()
+      Thread.sleep(delay)
+      span.finish()
+    }
+  }
+
+  static waitForDiscoveryTimeout() {
+    Thread.sleep(75_000) // Discovery timeout is 1min
```
bantonsson, line 129: Could we make the `MIN_FEATURE_DISCOVERY_INTERVAL_MILLIS` in the `DDAgentFeaturesDiscovery` configurable to avoid waiting a minute here?

PerfectSlayer: I am on it!

dd-trace-api/src/main/java/datadog/trace/api/config/TracerConfig.java
```diff
@@ -75,6 +75,7 @@ public final class TracerConfig {
   public static final String TRACE_PROPAGATION_STYLE_EXTRACT = "trace.propagation.style.extract";
   public static final String TRACE_PROPAGATION_STYLE_INJECT = "trace.propagation.style.inject";
 
+  public static final String FEATURES_DISCOVERY_MIN_DELAY = "feature.discovery.delay";
```
bantonsson, line 78: I think the setting should have a name relating to the trace agent, i.e. `trace.agent.feature.discovery.delay`

dd-trace-core/src/main/java/datadog/trace/common/writer/DDAgentWriter.java
```diff
@@ -135,7 +136,12 @@ public DDAgentWriter build() {
       if (null == featureDiscovery) {
         featureDiscovery =
             new DDAgentFeaturesDiscovery(
-                client, monitoring, agentUrl, traceAgentV05Enabled, metricsReportingEnabled);
+                client,
+                monitoring,
+                agentUrl,
+                DEFAULT_FEATURES_DISCOVERY_MIN_DELAY,
```
bantonsson, line 142: This does not use the setting from `Config`.

communication/src/main/java/datadog/communication/ddagent/DDAgentFeaturesDiscovery.java
```diff
@@ -75,10 +74,12 @@ public DDAgentFeaturesDiscovery(
       OkHttpClient client,
       Monitoring monitoring,
       HttpUrl agentUrl,
+      int featureDiscoveryMinDelay,
```
bantonsson, line 77: I would have kept the original constructor and let it call this one with the `DEFAULT_FEATURES_DISCOVERY_MIN_DELAY` to avoid changing so many test files that don't really care about this.

PerfectSlayer: Sure. Ideally, I would have create an interface for `FeaturesDiscovery` and a builder to optionally configure all features… (monitoring, V5, metrics, timing). It would have simplify most of tests that use it.

bantonsson: I'm even willing to not have the option exposed in config, but only override it in the test from the subclass 🤷🏼‍♂️

PerfectSlayer: I went for a lighter approach: no more config, only exposing interval value through inheritance.
I still think the whole thing deserve some love (or at least a proper interface) but that is out of the scope of this change.

bantonsson (approved):

PR #4668, opened 2023-02-06
Disable Docker layer caching in Circle CI

smola:
# What Does This Do
Disable Docker layer caching in Circle CI

# Motivation
Setting up the remote Docker with DLC enabled has about 1 minute of extra time. Also it has an extra cost for every job using it, and I don't think we're leveraging it.

# Additional Notes
* This PR is currently on top of https://github.com/DataDog/dd-trace-java/pull/4651

### DIFF
--- a/.circleci/config.yml
+++ b/.circleci/config.yml
@@ -87,7 +87,7 @@ commands:
           version: 20.10.18
           # DLC shares Docker layers across jobs (at an extra cost).
           # But its time to setup (~1min) exceeds the time required to prefetch all images we use.
-          docker_layer_caching: true
+          docker_layer_caching: false
 
       - run:
           name: Prepare testcontainers environment


.circleci/config.yml
```diff
@@ -126,41 +126,23 @@ commands:
...
-            chmod -v 0400 ca-key.pem key.pem server-key.pem
-            chmod -v 0444 ca.pem server-cert.pem cert.pem
-            
-            autoforward --port 60906 --remote $FORWARDED_DOCKER_HOST --forward remote-docker --secure --server-key $DOCKER_CERT_PATH/server-key.pem --server-cert $DOCKER_CERT_PATH/server-cert.pem --remote-key $DOCKER_CERT_PATH/remote-key.pem --remote-cert $DOCKER_CERT_PATH/remote-cert.pem --remote-ca $DOCKER_CERT_PATH/remote-ca.pem
+          command: .circleci/start_docker_autoforward.sh
+
+      - run:
+          name: Prefetch Docker images
+          background: true
+          command: .circleci/fetch_docker_images.sh
```
manuel-alvarez-alvarez, line 145: Way better!

.circleci/fetch_docker_images.sh
```diff
@@ -0,0 +1,29 @@
...
+)
+# smoke-tests
+IMAGES+=(
+  mongo:4.0.10
+  #rabbitmq:3.9.20-alpine
+)
+
+echo "Waiting for Docker to be available"
+while ! docker system info &>/dev/null; do
+  sleep 1
```
manuel-alvarez-alvarez, line 23: Perhaps we can add a limit to prevent an infinite loop if something goes wrong with docker?

manuel-alvarez-alvarez (approved):

.circleci/fetch_docker_images.sh
smola: This runs in the background, so it's somewhat safe. But I'll add a 1 minute limit. If 1 minute has passed, it doesn't even make sense to prefetch images anymore...

smola:
Already done in https://github.com/DataDog/dd-trace-java/pull/4831

PR #4669, opened 2023-02-06
Add Open Telemetry tracer API instrumentation

PerfectSlayer:
# What Does This Do

This PR brings support to Open Telemetry tracing features.
It should allow users to use Open Telemetry tracing API backed by DataDog tracing agent.

# Motivation

It aims to provide an initial support of current Open Telemetry initiative.

# Additional Notes

This PR replaces #4528 as I can't rewrite `/project/*` branch history.
The commits will be squashed at merged.


dd-java-agent/instrumentation/opentelemetry-1.4/src/main/java/datadog/trace/opentelemetry14/OtelSpan.java
```diff
@@ -0,0 +1,105 @@
...
+
+  @Override
+  public SpanContext getSpanContext() {
+    return OtelSpanContext.fromLocalSpan(this.delegate);
+  }
+
+  @Override
+  public boolean isRecording() {
+    // TODO Should we use DDSpan.isFinished()?
+    return this.recording;
```
bantonsson, line 95: Yes, I think you could switch to `isFinished` instead.

settings.gradle
```diff
@@ -274,7 +274,8 @@ include ':dd-java-agent:instrumentation:netty-promise-4'
 include ':dd-java-agent:instrumentation:okhttp-2'
 include ':dd-java-agent:instrumentation:okhttp-3'
 include ':dd-java-agent:instrumentation:ognl-appsec'
-include ':dd-java-agent:instrumentation:opentelemetry'
+include ':dd-java-agent:instrumentation:opentelemetry-0.3'
+include ':dd-java-agent:instrumentation:opentelemetry-1.4'
```
bantonsson, line 278: Minor nitpick. It would be nice if `opentelemetry-0.3` and `opentelemetry-1.4` both ended up below `opentelemtry` much like `opentracing`

dd-java-agent/instrumentation/opentelemetry-1.4/build.gradle
```diff
@@ -0,0 +1,26 @@
...
+    dirName = 'test'
+  }
+}
+
+dependencies {
+  compileOnly group: 'io.opentelemetry', name: 'opentelemetry-api', version: openTelemetryVersion
+  compileOnly group: 'com.google.auto.value', name: 'auto-value-annotations', version: '1.6.6'
+
+  testImplementation group: 'io.opentelemetry', name: 'opentelemetry-api', version: openTelemetryVersion
+  latestDepTestImplementation group: 'io.opentelemetry', name: 'opentelemetry-api', version: '1.4+'
```
bantonsson, line 25: I don't think `'1.4+'` does you want. I think it will select `1.4.1`, while `'1.+'` will select the latest `1.x`

bantonsson (commented):
Looks good

settings.gradle
PerfectSlayer: Will do 👍 
I read the contributing guidelines (https://github.com/DataDog/dd-trace-java/blob/master/CONTRIBUTING.md#adding-instrumentations) about the expected layout but it feels not a lot of instrumentations follow it so I kept the wrong layout by conformity 😅

dd-java-agent/instrumentation/opentelemetry-1.4/build.gradle
PerfectSlayer: Good catch!
I don't even really know what will be the behavior... According [the documentation](https://docs.gradle.org/current/userguide/single_versions.html), it should match everything that starts with `1.4`, so `1.4.1`, `1.4.2` but also `1.41.0` (!!) but not `1.5.0` nor `1.6.0`... 😒

dd-java-agent/instrumentation/opentelemetry-1.4/src/main/java/datadog/trace/opentelemetry14/OtelSpan.java
PerfectSlayer: `DDSpan` is part of `:dd-trace-core` module so it should not be accessed from `:dd-java-agent:instrumentations`.
I will stick with `#recording` for now.

dd-java-agent/instrumentation/opentelemetry/opentelemetry-1.4/src/test/groovy/OpenTelemetry14Test.groovy
```diff
@@ -0,0 +1,303 @@
...
+    def builder = tracer.spanBuilder("some-name")
+    def result = builder.startSpan()
+
+    expect:
+    tracer instanceof OtelTracer
+    builder instanceof OtelSpanBuilder
+    result instanceof OtelSpan
+  }
+
+  def "test parent span using active span"() {
```
bantonsson, line 32: You should split this into two different tests.
* Create Datadog `Span` and activate it, and use OTel API to create child `Span` and activate it
* Create OTel `Span` and activate it, and use Datadog API to create child `Span` and activate it

PerfectSlayer: The two tests `test parent span using active span` and `test parent span using reference` was created to test the expected behavior from the OTel API point of view. (by the way, there was an issue in the first one I fixed).

I created a dedicated test `test mixing manual and OTel instrumentation` where:
1. OTel Span creation ==> Check DD active span
2. DD Span creation (child) ==> Check OTel active span
3. OTel Span creation (great child) ==> Check DD active span
4. End ==> Check the trace and spans links.

Would you prefer having a dedicated test for each case (DD as parent with OTel as child + OTel as parent with DD as child)?

I also added a test to check `Span.current()`/`span.makeCurrent()` behavior.

PerfectSlayer:
The last update marks the instrumentation as `experimental` and disabled by default, includes a smoke test application and will warn if both _Open Tracing_ and _Open Telemetry_ instrumentations are enabled at the same time.
It should be ready to be merged @bantonsson

dd-java-agent/instrumentation/opentelemetry/opentelemetry-1.4/src/test/groovy/OpenTelemetry14Test.groovy
```diff
@@ -0,0 +1,398 @@
...
+    when:
+    def ddChildSpan = TEST_TRACER.startSpan("other-name")
+    TEST_TRACER.activateSpan(ddChildSpan, MANUAL)
+    def current = Span.current()
+
+    then:
+    DDSpanId.toHexString(ddChildSpan.spanId) == current.getSpanContext().spanId
+
+    when:
+    def otelGreatChildSpan = tracer.spanBuilder("another-name").startSpan()
```
bantonsson, line 367: I think the word you're looking for is grandchild 😉

telemetry/src/main/java/datadog/telemetry/TelemetryServiceImpl.java
```diff
@@ -154,6 +168,13 @@ public int getHeartbeatInterval() {
     return heartbeatIntervalMs;
   }
 
+  private void warnAboutExclusiveIntegrations() {
+    if (this.openTelemetryIntegrationEnabled && this.openTracingIntegrationEnabled) {
+      log.warn(
+          "Both Open Tracing and Open Telemetry integrations are enabled whereas mutually exclusive. Tracing performance can be degraded.");
```
bantonsson, line 174: I don't think you can use _whereas_ this way. Personally I would write _but are_ instead.

bantonsson (approved):

guizmaii:
Hi @PerfectSlayer @bantonsson,

Is there any doc somewhere about this integration, how to configure it, and what it does, please?

PerfectSlayer:
Hi @guizmaii 

The documentation starts rolling out recently. You can find it [here](https://docs.datadoghq.com/tracing/trace_collection/otel_instrumentation/java/) with some other links to OTel Datadog support.

PR #4671, opened 2023-02-06
Naming schema: redis

amarziali:
# What Does This Do

Start implementing naming schema changes for service and operation names for redis instrumentations (lettuce, redisson, redisscala, jedis).
v1 schema is opt-in only by providing the env `DD_TRACE_SPAN_ATTRIBUTE_SCHEMA` set to `1`
If this condition is met:
* service: `{{DD_SERVICE}}-redis`
* operation: `redis.command`

Note on the SchemaNaming interface: it will be enriched in the future with other stuff for (http clients, servers, message consumers, etc..)

# Motivation

# Additional Notes


dd-java-agent/instrumentation/jedis-1.4/src/main/java/datadog/trace/instrumentation/jedis/JedisClientDecorator.java
```diff
@@ -8,17 +11,23 @@
...
+
   @Override
   protected String[] instrumentationNames() {
     return new String[] {"jedis", "redis"};
   }
 
   @Override
   protected String service() {
-    return "redis";
+    return SpanNaming.instance()
```
dougqh, line 27: The performance implications of changing this concerns me.  I'd really prefer that these static final UTF8BytesString instances.

In theory, that could do done by having a complicated expression on the right hand side of a static field declaration.  However, that requires having the Config properly set-up before the Decorator is initialized.

That said, there is also a request for us to work towards supporting more dynamic configuration, but my thoughts on how to do that would require even bigger changes.

Hmm, not sure, what to do at the moment...

dd-java-agent/instrumentation/jedis-1.4/src/main/java/datadog/trace/instrumentation/jedis/JedisInstrumentation.java
```diff
@@ -63,7 +59,7 @@ public static class JedisAdvice {
 
     @Advice.OnMethodEnter(suppress = Throwable.class)
     public static AgentScope onEnter(@Advice.Argument(1) final Command command) {
-      final AgentSpan span = startSpan(REDIS_COMMAND);
+      final AgentSpan span = startSpan(DECORATE.getOperationName());
```
dougqh, line 62: We'll have to check the impact on latency & GC if we do this.

amarziali: I restored all the statically assigned UTF8ByteString

dd-java-agent/instrumentation/jedis-1.4/src/main/java/datadog/trace/instrumentation/jedis/JedisClientDecorator.java
amarziali: I changed as suggested (for service name they are as before statically assigned String).

dd-java-agent/instrumentation/jedis-1.4/src/test/groovy/JedisClientTest.groovy
```diff
@@ -156,3 +158,38 @@ class JedisClientTest extends AgentTestRunner {
...
+  @Override
+  protected String service() {
+    return Config.get().getServiceName() + "-redis"
+  }
+
+  @Override
+  protected String operation() {
+    return "redis.command"
+  }
+}
```
jdgumz, line 195: :+1: Test setup looks good

dd-java-agent/instrumentation/jedis-1.4/src/main/java/datadog/trace/instrumentation/jedis/JedisInstrumentation.java
```diff
@@ -63,7 +59,7 @@ public static class JedisAdvice {
 
     @Advice.OnMethodEnter(suppress = Throwable.class)
     public static AgentScope onEnter(@Advice.Argument(1) final Command command) {
-      final AgentSpan span = startSpan(REDIS_COMMAND);
+      final AgentSpan span = startSpan(JedisClientDecorator.OPERATION_NAME);
```
jdgumz, line 62: What is `JedisClientDecorator.OPERATION_NAME`? If it's a constant, then are we not relying on the schema to set it by specific version?

dd-java-agent/instrumentation/jedis-3.0/src/test/groovy/Jedis30ClientTest.groovy
```diff
@@ -296,3 +297,39 @@ class Jedis30ClientTest extends AgentTestRunner {
...
+  }
+
+  @Override
+  protected String service() {
+    return Config.get().getServiceName() + "-redis"
+  }
+
+  @Override
+  protected String operation() {
+    return "redis.command"
```
jdgumz, line 333: Would it be possible to define something as a more global constant for all `jedis-X` test cases?

dd-java-agent/testing/src/main/groovy/datadog/trace/agent/test/naming/VersionedNamingTestBase.groovy
```diff
@@ -0,0 +1,16 @@
...
+
+abstract class VersionedNamingTestBase extends AgentTestRunner {
+  protected abstract int version()
+  protected abstract String service()
+  protected abstract String operation()
+
+  @Override
+  protected void configurePreAgent() {
+    super.configurePreAgent()
+    injectSysConfig(TracerConfig.TRACE_SPAN_ATTRIBUTE_SCHEMA, Integer.toString(version()))
```
jdgumz, line 14: `Integer.toString(version())` might fail if we do `v0` and `v1` as the input values for the environment variable. It was raised during review that some languages struggle with `0` and `1` because they can be interpreted as false/true.

Should we change input values and validation so that we prefer `v0` and such?

jdgumz (commented):
I'm giving a soft approval as an outsider to the Java team.

The test cases look to be using the right values for v0/v1. Also the implementation of the v0/v1 cache specs looks pretty clean to me.

I've left some questions. One thing we should do prior to merge is carefully ensure that we've updated all the right tests and that we haven't missed any tests or cache frameworks.

dd-java-agent/instrumentation/jedis-1.4/src/main/java/datadog/trace/instrumentation/jedis/JedisInstrumentation.java
amarziali: It's a constant but it's set at runtime depending on the naming schema of the provided version. For performance reason we set only once

dd-java-agent/testing/src/main/groovy/datadog/trace/agent/test/naming/VersionedNamingTestBase.groovy
amarziali: OK at this point the `version` will need to be a String everywhere so. I can do this a better now than later

ygree (approved):
LGTM

PR #4677, opened 2023-02-07
track gRPC scheduling time in SynchronizationContext

richardstartin:
# What Does This Do

Tells the profiler how long gRPC requests spend scheduled in the `SynchronizationContext`. This work is temporary and will eventually be superseded by a generic implementation that will support any kind of asynchronous scheduling.

# Motivation

# Additional Notes

### DIFF
--- a/dd-java-agent/instrumentation/grpc-1.5/src/main/java/datadog/trace/instrumentation/grpc/TimedRunnable.java
+++ b/dd-java-agent/instrumentation/grpc-1.5/src/main/java/datadog/trace/instrumentation/grpc/TimedRunnable.java
@@ -1,0 +1,26 @@
+package datadog.trace.instrumentation.grpc;
+
+import datadog.trace.bootstrap.instrumentation.api.AgentTracer;
+import datadog.trace.bootstrap.instrumentation.api.ProfilingContextIntegration;
+
+/**
+ * Runnable which records how long it waited to execute. Don't use this except when wrapping can be
+ * shown to be safe TODO: generalise this behaviour in the scope manager, then delete this class.
+ */
+public final class TimedRunnable implements Runnable {
+
+  private final long creationTime;
+  private final Runnable delegate;
+
+  public TimedRunnable(Runnable delegate) {
+    this.delegate = delegate;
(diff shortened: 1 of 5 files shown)


dd-java-agent/instrumentation/java-concurrent/src/main/java/datadog/trace/instrumentation/java/concurrent/AbstractExecutorInstrumentation.java
```diff
@@ -40,8 +40,7 @@ public AbstractExecutorInstrumentation(final String... additionalNames) {
         "kotlinx.coroutines.scheduling.CoroutineScheduler",
         "play.api.libs.streams.Execution$trampoline$",
         "scala.concurrent.Future$InternalCallbackExecutor$",
-        "scala.concurrent.impl.ExecutionContextImpl",
-        "io.grpc.SynchronizationContext"
+        "scala.concurrent.impl.ExecutionContextImpl"
```
richardstartin, line 43: We only need context propagation here for the sake of code hotspots, so it seems better to push it into a profiling scoped instrumentation so non-profiling users don't pay for it.

dd-java-agent/instrumentation/grpc-1.5/src/main/java/datadog/trace/instrumentation/grpc/TimedRunnable.java
```diff
@@ -0,0 +1,26 @@
...
+import datadog.trace.bootstrap.instrumentation.api.AgentTracer;
+import datadog.trace.bootstrap.instrumentation.api.ProfilingContextIntegration;
+
+/**
+ * Runnable which records how long it waited to execute. Don't use this except when wrapping can be
+ * shown to be safe TODO: generalise this behaviour in the scope manager, then delete this class.
+ */
+public final class TimedRunnable implements Runnable {
+
+  private final long creationTime;
```
PerfectSlayer, line 12: Hopefully we will find a better solution soon™️

PerfectSlayer (approved):

PR #4686, opened 2023-02-08
Fix race condition with transformer in smoke tests

jpbempel:
# What Does This Do
before loading the class to be instrumented wait for the transformer including the probe configs to be installed

# Motivation
Fix racy smoke tests

# Additional Notes


jpbempel:
> Is this a fix for the flaky tests in `DebuggerIntegrationTest` or is that a separate issue?

yes this is a fix for that

cimi (approved):

PR #4693, opened 2023-02-08
Fix count & isEmpty expression for sets

jpbempel:
# What Does This Do
`ListValue` is not only `List` but more generally `Collection`

# Motivation
make `len(set)` & `isEmpty(set)` works

# Additional Notes

### DIFF
--- a/dd-java-agent/agent-debugger/debugger-el/src/main/java/com/datadog/debugger/el/values/ListValue.java
+++ b/dd-java-agent/agent-debugger/debugger-el/src/main/java/com/datadog/debugger/el/values/ListValue.java
@@ -8,10 +8,11 @@ import java.lang.reflect.Array;
 import java.util.Collection;
 import java.util.List;
 import java.util.Objects;
+import java.util.Set;
 
 /**
  * A list-like {@linkplain Value}.<br>
- * Allows wrapping of arrays as well as {@linkplain List} instances.
+ * Allows wrapping of arrays as well as {@linkplain List} or {@link Set} instances.
  */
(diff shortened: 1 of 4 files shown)


dd-java-agent/agent-debugger/debugger-el/src/main/java/com/datadog/debugger/el/values/ListValue.java
```diff
@@ -92,6 +93,8 @@ public Value<?> get(int index) {
     }
     if (listHolder instanceof List) {
       return Value.of(((List<?>) listHolder).get(index));
+    } else if (listHolder instanceof Set) {
+      throw new UnsupportedOperationException("Cannot access Set by index");
```
cimi, line 97: Related to our recent discussion, I guess this will change if we implement key lookups on sets `mySet["myKey"]`?

In the EL parser we don't know the type of the collection, so we will return `{ "index": [{"ref": "mySet"}, "myKey"] }`.

Nothing to change in this PR, sharing this since it's fresh in my head.

cimi (approved):
:+1: Thanks!

dd-java-agent/agent-debugger/debugger-el/src/main/java/com/datadog/debugger/el/values/ListValue.java
jpbempel: yes that's why I prefer throw an exception in this case to make it explicit, but it needs to be revisited to implement correctly `set["key"]`, right

PR #4707, opened 2023-02-09
Jsantos/resteasy IAST instrumentation

DDJavierSantos:
# What Does This Do
Instrument the resteasy core injector so we can taint request values (path parameters, query parameters, cookies and headers)

# Motivation
RestEasy doesn't use the typical Servlet request mechanism to get http parameters. The Servlet instrumentation is not effective in this case so we need to implement custom instrumentation points for Reseasy so we can taint the http input data for IAST use.

# Additional Notes


dd-java-agent/instrumentation/resteasy-appsec/src/main/java/datadog/trace/instrumentation/resteasy/ValueInjectorAdvice.java
```diff
@@ -0,0 +1,29 @@
+package datadog.trace.instrumentation.resteasy;
+
+import datadog.trace.api.iast.InstrumentationBridge;
+import datadog.trace.api.iast.source.WebModule;
+import net.bytebuddy.asm.Advice;
+
+public class ValueInjectorAdvice {
+  @Advice.OnMethodExit
+  public static void onExit(@Advice.Return(readOnly = true) Object result) {
+    if (result instanceof String) {
```
manuel-alvarez-alvarez, line 10: This might not be enough, for instance a query param can appear twice in the query and it will be bound to a List, Set or other collection. In most of the doc of the @FormParam, @QueryParam annotations they can bind the following values:
```
1. Be a primitive type
2. Have a constructor that accepts a single String argument
3. Have a static method named valueOf or fromString that accepts a single
4. Have a registered implementation of javax.ws.rs.ext.ParamConverterProvider JAX-RS extension SPI that returns a javax.ws.rs.ext.ParamConverter instance capable of a "from string" conversion for the type.
String argument (see, for example, Integer.valueOf(String))
5. Be List<T>, Set<T> or SortedSet<T>, where T satisfies 2, 3 or 4 above. The resulting collection is read-only.
```

At the very least we should support as well collections of strings.

dd-java-agent/instrumentation/resteasy-appsec/src/main/java/datadog/trace/instrumentation/resteasy/ValueInjectorAdvice.java
```diff
@@ -0,0 +1,29 @@
...
+      System.out.println(
+          "Value Injector Instrumentation Called with value ["
+              + result
+              + "] identity hash code: "
+              + System.identityHashCode(result));
+      final WebModule module = InstrumentationBridge.WEB;
+
+      if (module != null) {
+        try {
+          module.onParameterValue(null, (String) result);
```
manuel-alvarez-alvarez, line 20: We are instrumenting all instances of `org.jboss.resteasy.core.ValueInjector`, but the hierarchy contains different types of sources like:

- `org.jboss.resteasy.core.HeaderParamInjector`
- `org.jboss.resteasy.core.QueryParamInjector`
- ...

In this advice we are treating all of them as parameters which is wrong (they can be cookies, headers, path parameters ...). We have two possibilities:

1. Provide a separate advice for each instrumentation that will make it slower to instrument but faster at runtime.
2. Inject the `this` instance and check the type in order to distinguish the actual source.

In both cases you need access to the actual injector in order to extract instance fields that are needed for the source, as an example you need `org.jboss.resteasy.core.QueryParamInjector#encodedName` to known the name of the parameter.

manuel-alvarez-alvarez (changes requested):

dd-java-agent/instrumentation/resteasy-appsec/src/main/java/datadog/trace/instrumentation/resteasy/ValueInjectorAdvice.java
DDJavierSantos: I have added support and tests for List Set and SortedSet.

dd-java-agent/agent-iast/src/main/java/com/datadog/iast/source/WebModuleImpl.java
```diff
@@ -132,4 +132,18 @@ public void onHeaderValue(@Nullable final String headerName, @Nullable final Str
     taintedObjects.taintInputString(
         headerValue, new Source(SourceType.REQUEST_HEADER_VALUE, headerName, headerValue));
   }
+
+  @Override
+  public void onCookieValue(@Nullable final String cookieName, @Nullable final String cookieValue) {
+    if (!canBeTainted(cookieName)) {
```
smola, line 138: This is `onCoockieValue`, so we should short-circuit on `cookieValue`, not `cookieName`.

dd-java-agent/instrumentation/resteasy-appsec/src/main/java/datadog/trace/instrumentation/resteasy/CookieParamInjectorAdvice.java
```diff
@@ -0,0 +1,38 @@
...
+              }
+            }
+          } else {
+            module.onCookieValue(paramName, (String) result);
+          }
+        } catch (final Throwable e) {
+          module.onUnexpectedException("CookieParamInjectorAdvice.onExit threw", e);
+        }
+      } else {
+        System.out.println("Module is null");
```
smola, line 34: If module is null, we just do nothing, not even log.

dd-java-agent/instrumentation/resteasy-appsec/src/main/java/datadog/trace/instrumentation/resteasy/CookieParamInjectorInstrumentation.java
```diff
@@ -0,0 +1,27 @@
+package datadog.trace.instrumentation.resteasy;
+
+import static datadog.trace.agent.tooling.bytebuddy.matcher.NameMatchers.named;
+import static net.bytebuddy.matcher.ElementMatchers.isPublic;
+
+import com.google.auto.service.AutoService;
+import datadog.trace.agent.tooling.Instrumenter;
+
+@AutoService(Instrumenter.class)
+public class CookieParamInjectorInstrumentation extends Instrumenter.Tracing
```
smola, line 10:
```suggestion
public class CookieParamInjectorInstrumentation extends Instrumenter.Iast
```

dd-java-agent/instrumentation/resteasy-appsec/src/main/java/datadog/trace/instrumentation/resteasy/HeaderParamInjectorAdvice.java
```diff
@@ -0,0 +1,38 @@
...
+              }
+            }
+          } else {
+            module.onHeaderValue(null, (String) result);
+          }
+        } catch (final Throwable e) {
+          module.onUnexpectedException("PathParamInjectorAdvice.onExit threw", e);
+        }
+      } else {
+        System.out.println("Module is null");
```
smola, line 34: Same here.

dd-java-agent/instrumentation/resteasy-appsec/src/main/java/datadog/trace/instrumentation/resteasy/HeaderParamInjectorInstrumentation.java
```diff
@@ -0,0 +1,27 @@
+package datadog.trace.instrumentation.resteasy;
+
+import static datadog.trace.agent.tooling.bytebuddy.matcher.NameMatchers.named;
+import static net.bytebuddy.matcher.ElementMatchers.isPublic;
+
+import com.google.auto.service.AutoService;
+import datadog.trace.agent.tooling.Instrumenter;
+
+@AutoService(Instrumenter.class)
+public class HeaderParamInjectorInstrumentation extends Instrumenter.Tracing
```
smola, line 10:
```suggestion
public class HeaderParamInjectorInstrumentation extends Instrumenter.Iast
```

smola (changes requested):

dd-java-agent/instrumentation/resteasy-appsec/src/main/java/datadog/trace/instrumentation/resteasy/HeaderParamInjectorAdvice.java
```diff
@@ -0,0 +1,36 @@
...
+            for (Object o : collection) {
+              if (o instanceof String) {
+                module.onHeaderValue(paramName, (String) o);
+              }
+            }
+          } else {
+            module.onHeaderValue(null, (String) result);
+          }
+        } catch (final Throwable e) {
+          module.onUnexpectedException("PathParamInjectorAdvice.onExit threw", e);
```
smola, line 31:
```suggestion
          module.onUnexpectedException("HeaderParamInjectorAdvice.onExit threw", e);
```

dd-java-agent/instrumentation/resteasy-appsec/src/main/java/datadog/trace/instrumentation/resteasy/HeaderParamInjectorInstrumentation.java
```diff
@@ -0,0 +1,27 @@
...
+
+import com.google.auto.service.AutoService;
+import datadog.trace.agent.tooling.Instrumenter;
+
+@AutoService(Instrumenter.class)
+public class HeaderParamInjectorInstrumentation extends Instrumenter.Iast
+    implements Instrumenter.ForSingleType {
+
+  public HeaderParamInjectorInstrumentation() {
+    super("Header Value Injector for RestEasy");
```
smola, line 14:
```suggestion
    super("resteasy");
```

dd-java-agent/instrumentation/resteasy-appsec/src/main/java/datadog/trace/instrumentation/resteasy/CookieParamInjectorInstrumentation.java
```diff
@@ -0,0 +1,27 @@
...
+
+import com.google.auto.service.AutoService;
+import datadog.trace.agent.tooling.Instrumenter;
+
+@AutoService(Instrumenter.class)
+public class CookieParamInjectorInstrumentation extends Instrumenter.Iast
+    implements Instrumenter.ForSingleType {
+
+  public CookieParamInjectorInstrumentation() {
+    super("Cookie Param Value Injector for RestEasy");
```
smola, line 14:
```suggestion
    super("resteasy");
```

dd-java-agent/instrumentation/resteasy-appsec/src/main/java/datadog/trace/instrumentation/resteasy/PathParamInjectorInstrumentation.java
```diff
@@ -0,0 +1,27 @@
...
+
+import com.google.auto.service.AutoService;
+import datadog.trace.agent.tooling.Instrumenter;
+
+@AutoService(Instrumenter.class)
+public class PathParamInjectorInstrumentation extends Instrumenter.Iast
+    implements Instrumenter.ForSingleType {
+
+  public PathParamInjectorInstrumentation() {
+    super("Path param Value Injector for RestEasy");
```
smola, line 14:
```suggestion
    super("resteasy");
```

dd-java-agent/instrumentation/resteasy-appsec/src/main/java/datadog/trace/instrumentation/resteasy/QueryParamInjectorInstrumentation.java
```diff
@@ -0,0 +1,27 @@
...
+
+import com.google.auto.service.AutoService;
+import datadog.trace.agent.tooling.Instrumenter;
+
+@AutoService(Instrumenter.class)
+public class QueryParamInjectorInstrumentation extends Instrumenter.Iast
+    implements Instrumenter.ForSingleType {
+
+  public QueryParamInjectorInstrumentation() {
+    super("Query Param Value Injector for RestEasy");
```
smola, line 14:
```suggestion
    super("resteasy");
```

dd-smoke-tests/resteasy/src/test/groovy/smoketest/TestUndertow.groovy
```diff
@@ -0,0 +1,207 @@
...
+
+import datadog.smoketest.AbstractServerSmokeTest
+import datadog.trace.api.Platform
+import okhttp3.Request
+import spock.lang.IgnoreIf
+
+@IgnoreIf({
+  System.getProperty("java.vendor").contains("IBM") && System.getProperty("java.version").contains("1.8.")
+})
+class TestUndertow extends AbstractServerSmokeTest {
```
smola, line 11: Let's keep the convention here and call this something like `ResteasySmokeTest`.

dd-smoke-tests/resteasy/src/test/groovy/smoketest/TestUndertow.groovy
```diff
@@ -0,0 +1,207 @@
...
+    List<String> command = new ArrayList<>()
+    command.add(javaPath())
+    command.addAll(defaultJavaProperties)
+    command.addAll([
+      withSystemProperty(datadog.trace.api.config.IastConfig.IAST_ENABLED, true),
+      withSystemProperty(datadog.trace.api.config.IastConfig.IAST_REQUEST_SAMPLING, 100),
+      withSystemProperty(datadog.trace.api.config.IastConfig.IAST_DEBUG_ENABLED, true),
+      withSystemProperty(datadog.trace.api.config.IastConfig.IAST_DEDUPLICATION_ENABLED, false)
+    ])
+    //command.add("-Xrunjdwp:transport=dt_socket,server=y,suspend=y,address=8000")
```
smola, line 32: Remove debugging leftovers.

dd-smoke-tests/resteasy/src/test/groovy/smoketest/TestUndertow.groovy
```diff
@@ -0,0 +1,207 @@
...
+    //command.add("-Xdebug")
+    if (Platform.isJavaVersionAtLeast(17)) {
+      command.addAll((String[]) ["--add-opens", "java.base/java.lang=ALL-UNNAMED"])
+    }
+    command.addAll((String[]) ["-jar", jarPath, httpPort])
+    ProcessBuilder processBuilder = new ProcessBuilder(command)
+    processBuilder.directory(new File(buildDirectory))
+  }
+
+  def "Test path parameter in RestEasy"() {
```
smola, line 42:
```suggestion
  def "test path parameter"() {
```

(more in line with others in the project, also no need for the redundancy given the class name)

dd-java-agent/instrumentation/resteasy-appsec/src/main/java/datadog/trace/instrumentation/resteasy/CookieParamInjectorAdvice.java
```diff
@@ -0,0 +1,36 @@
...
+      @Advice.This CookieParamInjector self,
+      @Advice.Return(readOnly = true) Object result,
+      @Advice.FieldValue("paramName") String paramName) {
+    if (result instanceof String || result instanceof Collection) {
+      final WebModule module = InstrumentationBridge.WEB;
+
+      if (module != null) {
+        try {
+          if (result instanceof Collection) {
+            Collection collection = (Collection) result;
```
manuel-alvarez-alvarez, line 21: Not for this PR but in the one I'm preparing for Vert.x I added support to taint collections, if we do it like this we will access the request context once per element in the list (bad for performance :cry:).

manuel-alvarez-alvarez: I'll update this in my PR once it's merged.

dd-smoke-tests/resteasy/src/test/groovy/smoketest/ResteasySmokeTest.groovy
```diff
@@ -0,0 +1,205 @@
...
+
+  def "Test query parameter in RestEasy"() {
+    setup:
+    def url = "http://localhost:${httpPort}/hello/byqueryparam?param=queryParamValue"
+    boolean sqlInjectionFound = false
+
+    when:
+    def request = new Request.Builder().url(url).get().build()
+    def response = client.newCall(request).execute()
+    checkLog {
```
manuel-alvarez-alvarez, line 71: To me it's a bit overkill to validate sources using vulnerabilities, for the vert.x approach I'm using the logs from the tainted map:

https://github.com/DataDog/dd-trace-java/blob/malvarez/iast-vertx-support/dd-smoke-tests/vertx-3.9/src/test/groovy/datadog/smoketest/IastVertxSmokeTest.groovy#L67

dd-java-agent/instrumentation/resteasy-appsec/src/main/java/datadog/trace/instrumentation/resteasy/CookieParamInjectorAdvice.java
```diff
@@ -0,0 +1,36 @@
...
+import java.util.Collection;
+import net.bytebuddy.asm.Advice;
+import org.jboss.resteasy.core.CookieParamInjector;
+
+public class CookieParamInjectorAdvice {
+  @Advice.OnMethodExit
+  public static void onExit(
+      @Advice.This CookieParamInjector self,
+      @Advice.Return(readOnly = true) Object result,
+      @Advice.FieldValue("paramName") String paramName) {
```
manuel-alvarez-alvarez, line 14: Nice!

manuel-alvarez-alvarez (approved):

dd-java-agent/instrumentation/resteasy-appsec/src/main/java/datadog/trace/instrumentation/resteasy/CookieParamInjectorAdvice.java
smola: Good catch. If the Vert.x PR is not ready, can we get the helpers to taint collections merged? I think we should fix this before releasing.

dd-smoke-tests/resteasy/src/test/groovy/smoketest/ResteasySmokeTest.groovy
smola: No strong opinion here, but we used to have flakiness issues when checking vulnerabilities with the logs. Is this not an issue anymore?

dd-java-agent/instrumentation/resteasy-appsec/src/main/java/datadog/trace/instrumentation/resteasy/FormParamInjectorInstrumentation.java
```diff
@@ -0,0 +1,27 @@
...
+
+import com.google.auto.service.AutoService;
+import datadog.trace.agent.tooling.Instrumenter;
+
+@AutoService(Instrumenter.class)
+public class FormParamInjectorInstrumentation extends Instrumenter.Iast
+    implements Instrumenter.ForSingleType {
+
+  public FormParamInjectorInstrumentation() {
+    super("formparaminjector");
```
smola, line 14:
```suggestion
    super("resteasy");
```

smola (changes requested):

dd-java-agent/instrumentation/resteasy-appsec/src/main/java/datadog/trace/instrumentation/resteasy/HeaderParamInjectorInstrumentation.java
```diff
@@ -0,0 +1,27 @@
...
+    implements Instrumenter.ForSingleType {
+
+  public HeaderParamInjectorInstrumentation() {
+    super("resteasy");
+  }
+
+  @Override
+  public void adviceTransformations(AdviceTransformation transformation) {
+    transformation.applyAdvice(
+        named("inject").and(isPublic()), packageName + ".HeaderParamInjectorAdvice");
```
smola, line 20: Please, add matchers for `takeArguments` here to ensure we select the correct function. Otherewise, the instrumentation will break if the function is overloaded with different parameters in a new version. You should apply this to all instrumenters in the PR.

smola (changes requested):

dd-java-agent/instrumentation/resteasy-appsec/src/main/java/datadog/trace/instrumentation/resteasy/HeaderParamInjectorInstrumentation.java
DDJavierSantos: done

smola (approved):

PR #4708, opened 2023-02-09
ignore null resource name

devinsba:
# What Does This Do

Ignores when a customer uses a null resource name, either through APIs of the tag interceptor

# Motivation

A customer was overwriting a resource name when trying to set it to null. We decided ignoring the call is the best fix


smola (approved):

ygree (approved):
Thanks!

PR #4709, opened 2023-02-09
Refactor Json deserialization for test code only

jpbempel:
# What Does This Do
Snapshot deserialization is only used for test code. Isolate deserialization code from Moshi adapters to test helper class and share it also for smoke tests

# Motivation
Isolate snapshot deserialization code into test only

# Additional Notes


shatzi (approved):
Looks good.

PR #4710, opened 2023-02-10
Refactor ScopeManager for future scoped context improvements

PerfectSlayer:
# What Does This Do

This class splits the `ContinuableScopeManager` to introduce the future changes related to scoped context storage.
It does not change any logic from the original single file.

# Motivation

There will be different scope related changes. Splitting the `ContinuableScopeManager` will make things clearer, and git changes easier to merge.

# Additional Notes

All the commits will be squashed at merge.

### DIFF
--- a/dd-trace-core/src/main/java/datadog/trace/core/monitor/TracerHealthMetrics.java
+++ b/dd-trace-core/src/main/java/datadog/trace/core/monitor/TracerHealthMetrics.java
@@ -212,10 +212,10 @@ public class TracerHealthMetrics extends HealthMetrics implements AutoCloseable 
   }
 
   @Override
-  public void onScopeClose(int scopeSource) {
-    statsd.incrementCounter("scope.close.error");
+  public void onScopeCloseError(int scopeSource) {
+    statsd.incrementCounter("scope.close.error", NO_TAGS);
     if (scopeSource == ScopeSource.MANUAL.id()) {
-      statsd.incrementCounter("scope.user.close.error");
+      statsd.incrementCounter("scope.user.close.error", NO_TAGS);
     }
   }
 
--- a/dd-trace-core/src/main/java/datadog/trace/core/monitor/TracerHealthMetrics.java
+++ b/dd-trace-core/src/main/java/datadog/trace/core/monitor/TracerHealthMetrics.java
@@ -11,6 +11,7 @@ import static java.util.concurrent.TimeUnit.SECONDS;
 
 import datadog.trace.api.StatsDClient;
 import datadog.trace.api.cache.RadixTreeCache;
+import datadog.trace.bootstrap.instrumentation.api.ScopeSource;
 import datadog.trace.common.writer.RemoteApi;
 import datadog.trace.core.DDSpan;
 import datadog.trace.util.AgentTaskScheduler;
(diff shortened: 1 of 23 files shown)


dd-trace-core/src/main/java/datadog/trace/core/scopemanager/ContinuableScope.java
```diff
@@ -0,0 +1,192 @@
...
+    final ScopeStack scopeStack = scopeManager.scopeStack();
+
+    // fast check first, only perform slower check when there's an inconsistency with the stack
+    if (!scopeStack.checkTop(this) && !scopeStack.checkOverdueScopes(this)) {
+      if (ContinuableScopeManager.log.isDebugEnabled()) {
+        ContinuableScopeManager.log.debug(
+            "Tried to close {} scope when not on top.  Current top: {}", this, scopeStack.top);
+      }
+
+      scopeManager.statsDClient.incrementCounter("scope.close.error");
```
bantonsson, line 50: I think that you could move these behind the `HealthMetrics` and get rid of the `StatsDClient` parameter.

bantonsson (approved):

dd-trace-core/src/main/java/datadog/trace/core/scopemanager/ContinuableScope.java
PerfectSlayer: Should this be part of this refactoring or go into another PR?

bantonsson: I think it can go into this PR. You have already cleaned up up some other unnecessary parameters in some calls (like `ScopeSource`).

dd-trace-core/src/main/java/datadog/trace/core/monitor/TracerHealthMetrics.java
```diff
@@ -210,6 +211,14 @@ public void onCreateManualTrace() {
     manualTraces.inc();
   }
 
+  @Override
+  public void onScopeClose(int scopeSource) {
```
bantonsson, line 215: `onScopeCloseError`?

dd-trace-core/src/main/java/datadog/trace/core/scopemanager/ContinuableScope.java
PerfectSlayer: Nice catch. I will get things even cleaner 😊

dd-trace-core/src/main/java/datadog/trace/core/monitor/TracerHealthMetrics.java
PerfectSlayer: Agreed. About the metric itself, I wonder if we are not duplicating things with:
```java
  @Override
  public void onScopeCloseError(int scopeSource) {
    statsd.incrementCounter("scope.close.error");
    if (scopeSource == ScopeSource.MANUAL.id()) {
      statsd.incrementCounter("scope.user.close.error");
    }
  }
```
I will try another PR to use only one counter but with _scope source_ as tag value.

PerfectSlayer: Here is the related PR: #4716

PR #4713, opened 2023-02-10
Remove redundant caches from CI

smola:
# What Does This Do
* Build cache does not need to be restored before the initial build, since we always `--rerun-tasks`.
* Gradle wrapper is already present in the dependency cache, and it does not change often, so it doesn't need to be saved on every build.
* System tests do not require any cache, just the built artifacts which are present in the workspace.

# Motivation

# Additional Notes


bantonsson (approved):
Nice

PR #4714, opened 2023-02-10
Use a ramdisk in CI

smola:
# What Does This Do

# Motivation
See https://support.circleci.com/hc/en-us/articles/360054908812-Speed-up-steps-using-a-RAM-disk

# Additional Notes


smola:
This looked like a good idea, but it doesn't seem to make much difference, and the path change complicates other changes related to caching.

PR #4716, opened 2023-02-10
Improve scope.close.error metric with source tag

PerfectSlayer:
# What Does This Do

This improves health metrics by using tags to keep source of the scope.

# Motivation

Only use one metric instead of two and keep all sources.

# Additional Notes

See original discussion: https://github.com/DataDog/dd-trace-java/pull/4710#discussion_r1102728668

### DIFF
--- a/dd-trace-core/src/main/java/datadog/trace/core/monitor/TracerHealthMetrics.java
+++ b/dd-trace-core/src/main/java/datadog/trace/core/monitor/TracerHealthMetrics.java
@@ -15,7 +15,9 @@ import datadog.trace.bootstrap.instrumentation.api.ScopeSource;
 import datadog.trace.common.writer.RemoteApi;
 import datadog.trace.core.DDSpan;
 import datadog.trace.util.AgentTaskScheduler;
+import java.util.HashMap;
 import java.util.List;
+import java.util.Map;
 import java.util.concurrent.TimeUnit;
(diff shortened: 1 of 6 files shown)


dd-trace-core/src/main/java/datadog/trace/core/monitor/TracerHealthMetrics.java
```diff
@@ -85,6 +89,16 @@ public class TracerHealthMetrics extends HealthMetrics implements AutoCloseable
   private final long interval;
   private final TimeUnit units;
 
+  static {
+    SCOPE_SOURCE_TAGS.put(
+        Byte.toUnsignedInt(ScopeSource.INSTRUMENTATION.id()),
+        new String[] {"source:instrumentation"});
```
ygree, line 95: Should "source:instrumentation" and "source:iteration" be checked in tests similarly to "source:manual"?

PerfectSlayer: For now, I only test it for the tests that check the error was coming from the user with manual instrumentation (`source:manual`). 

I think distinction between `source:instrumentation` and `source:iteration` is an implementation detail than can help us later to diagnosticate issues but we should not enforce some behavior through tests, especially if we plan to improve the `ScopeManager` in a near future to decrease the tracer overhead.

Any thought about this?

PR #4717, opened 2023-02-10
Introduced Visitor in Expression Language

jpbempel:
# What Does This Do
Visitor pattern is introduced in Expression nodes to allow traverse the AST for different purpose like emitting byte code or pretty printing.
Retrofit pretty printing into a visitor.

# Motivation

# Additional Notes

### DIFF
--- a/dd-java-agent/agent-debugger/debugger-el/src/main/java/com/datadog/debugger/el/PrettyPrintVisitor.java
+++ b/dd-java-agent/agent-debugger/debugger-el/src/main/java/com/datadog/debugger/el/PrettyPrintVisitor.java
@@ -1,0 +1,277 @@
+package com.datadog.debugger.el;
+
+import com.datadog.debugger.el.expressions.BinaryExpression;
+import com.datadog.debugger.el.expressions.BinaryOperator;
+import com.datadog.debugger.el.expressions.BooleanExpression;
+import com.datadog.debugger.el.expressions.ComparisonExpression;
+import com.datadog.debugger.el.expressions.ComparisonOperator;
+import com.datadog.debugger.el.expressions.ContainsExpression;
+import com.datadog.debugger.el.expressions.EndsWithExpression;
+import com.datadog.debugger.el.expressions.FilterCollectionExpression;
+import com.datadog.debugger.el.expressions.GetMemberExpression;
+import com.datadog.debugger.el.expressions.HasAllExpression;
+import com.datadog.debugger.el.expressions.HasAnyExpression;
+import com.datadog.debugger.el.expressions.IfElseExpression;
+import com.datadog.debugger.el.expressions.IfExpression;
+import com.datadog.debugger.el.expressions.IndexExpression;
+import com.datadog.debugger.el.expressions.IsEmptyExpression;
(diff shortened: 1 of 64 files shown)


dd-java-agent/agent-debugger/debugger-el/src/main/java/com/datadog/debugger/el/PrettyPrintVisitor.java
```diff
@@ -0,0 +1,277 @@
...
+import com.datadog.debugger.el.values.ObjectValue;
+import com.datadog.debugger.el.values.StringValue;
+import datadog.trace.bootstrap.debugger.el.Values;
+import java.math.BigDecimal;
+import java.math.BigInteger;
+import java.util.List;
+import java.util.Map;
+import java.util.Set;
+
+public class PrettyPrintVisitor implements Visitor<String> {
```
shatzi, line 40: nit: maybe instead of creating string we can simply push into StringBuilder to save memory allocations

dd-java-agent/agent-debugger/debugger-el/src/main/java/com/datadog/debugger/el/PrettyPrintVisitor.java
```diff
@@ -0,0 +1,277 @@
...
+
+  public static String print(Expression<?> expr) {
```
shatzi, line 42: lovely helper function

dd-java-agent/agent-debugger/debugger-el/src/main/java/com/datadog/debugger/el/PrettyPrintVisitor.java
```diff
@@ -0,0 +1,277 @@
...
+    return stringPredicateExpression.getName()
+        + "("
+        + nullSafeAccept(stringPredicateExpression.getSourceString())
+        + ", "
+        + nullSafeAccept(stringPredicateExpression.getStr())
+        + ")";
+  }
+
+  private String nullSafeAccept(Expression<?> expr) {
+    return expr != null ? expr.accept(this) : "null";
```
shatzi, line 275: I think this can also return null (for some reason we call Expression.accept() because we failed to override)

shatzi (approved):
Awesome!

dd-java-agent/agent-debugger/debugger-el/src/main/java/com/datadog/debugger/el/PrettyPrintVisitor.java
jpbempel: true, but this is so little used with small strings, that not sure it really matters here and now.
If we found later that it is high source of allocations then we can revisit this

dd-java-agent/agent-debugger/debugger-el/src/main/java/com/datadog/debugger/el/PrettyPrintVisitor.java
jpbempel: yeah but this is not a problem if it returns null, it will create a string `"null"`

Issue #4718, opened 2023-02-11
Kotlin coroutines instrumentation causes IllegalStateException sometimes


monosoul:
Enabling coroutines instrumentation causes `IllegalStateException` in `kotlinx.coroutines.flow.internal.SafeCollector` sometimes.

E.g.:
Stacktrace:
```
java.lang.IllegalStateException: Flow invariant is violated:
		Flow was collected in [io.ktor.server.engine.DefaultUncaughtExceptionHandler@205ac27b, CoroutineName(call-handler), io.ktor.server.netty.NettyDispatcher$CurrentContext@41c6aff0, datadog.trace.instrumentation.kotlin.coroutines.ScopeStateCoroutineContext@4665ebe3, ScopeCoroutine{Active}@3a4b0a58, NettyDispatcher@6f2a5663],
		but emission happened in [io.ktor.server.engine.DefaultUncaughtExceptionHandler@205ac27b, CoroutineName(call-handler), io.ktor.server.netty.NettyDispatcher$CurrentContext@41c6aff0, datadog.trace.instrumentation.kotlin.coroutines.ScopeStateCoroutineContext@433755fa, ScopeCoroutine{Active}@377acb43, NettyDispatcher@6f2a5663].
		Please refer to 'flow' documentation or use 'flowOn' instead
	at kotlinx.coroutines.flow.internal.SafeCollector_commonKt.checkContext(SafeCollector.common.kt:85)
	at kotlinx.coroutines.flow.internal.SafeCollector.checkContext(SafeCollector.kt:106)
	at kotlinx.coroutines.flow.internal.SafeCollector.emit(SafeCollector.kt:83)
	at kotlinx.coroutines.flow.internal.SafeCollector.emit(SafeCollector.kt:66)
	at io.grpc.kotlin.ClientCalls$rpcImpl$1$1.invokeSuspend(ClientCalls.kt:320)
	at kotlin.coroutines.jvm.internal.BaseContinuationImpl.resumeWith(ContinuationImpl.kt:33)
	at kotlinx.coroutines.DispatchedTask.run(DispatchedTask.kt:106)
	at io.netty.util.concurrent.PromiseTask.runTask(PromiseTask.java:98)
	at io.netty.util.concurrent.PromiseTask.run(PromiseTask.java:106)
	at io.netty.util.concurrent.AbstractEventExecutor.runTask(AbstractEventExecutor.java:174)
[...]

PR #4624, opened 2023-01-31
Instrumentation for Kotlin coroutines 1.5.0+

monosoul:
# What Does This Do
Introduces support for coroutines 1.5.0+. 
Solves #931 , Solves #1123 .

# Motivation
The instrumentation I introduced in #4405 only supports Kotlin coroutines up to 1.4.3. 

# Additional Notes
Updates project-wide Kotlin dependency version to 1.6.21 because latest versions of coroutines require higher API level version (unfortunately Kotlin plugin for Gradle has some breaking changes in version 1.7.0, so I didn't use it here).
Introduces 3 new Gradle modules: `kotlin-coroutines-common`, `kotlin-coroutines-1.3` and `kotlin-coroutines-1.5`.
 Also configures `lastDep` test (was missing before).

PR #4719, opened 2023-02-11
Fix IllegalStateException caused by Flows instrumentation in Kotlin coroutines

monosoul:
# What Does This Do
Makes `ScopeStateCoroutineContext` inheritable by introducing an internal map of a child coroutine (aka Job) (inheriting the context) to the context item holding the continuation. This way each child coroutine will hold the same context element while having individual state internally.
Solves #4718 .

# Motivation
To fix an `IllegalStateException` thrown by [this](https://github.com/Kotlin/kotlinx.coroutines/blob/master/kotlinx-coroutines-core/common/src/flow/internal/SafeCollector.common.kt) Flows collector in case of a modified context.

# Additional Notes
This is a followup to #4405 and  #4624

### DIFF
--- a/dd-java-agent/instrumentation/kotlin-coroutines/src/main/java/datadog/trace/instrumentation/kotlin/coroutines/ScopeStateCoroutineContext.java
+++ b/dd-java-agent/instrumentation/kotlin-coroutines/src/main/java/datadog/trace/instrumentation/kotlin/coroutines/ScopeStateCoroutineContext.java
@@ -1,10 +1,12 @@
 package datadog.trace.instrumentation.kotlin.coroutines;
 
+import datadog.trace.bootstrap.ContextStore;
 import datadog.trace.bootstrap.instrumentation.api.AgentScope;
 import datadog.trace.bootstrap.instrumentation.api.AgentTracer;
 import datadog.trace.bootstrap.instrumentation.api.ScopeState;
 import kotlin.coroutines.CoroutineContext;
 import kotlin.jvm.functions.Function2;
+import kotlinx.coroutines.Job;
 import kotlinx.coroutines.ThreadContextElement;
 import org.jetbrains.annotations.NotNull;
 import org.jetbrains.annotations.Nullable;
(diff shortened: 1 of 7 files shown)


monosoul:
@bantonsson please, have a look. Thanks!

dd-java-agent/instrumentation/kotlin-coroutines/src/main/java/datadog/trace/instrumentation/kotlin/coroutines/ScopeStateCoroutineContext.java
```diff
@@ -3,36 +3,27 @@
...
-  private final ScopeState coroutineScopeState;
-  @Nullable private AgentScope.Continuation continuation;
-  @Nullable private AgentScope continuationScope;
-  private boolean isInitialized = false;
 
-  public ScopeStateCoroutineContext() {
-    coroutineScopeState = AgentTracer.get().newScopeState();
-  }
+  private final Map<Job, ScopeStateCoroutineContextItem> contextItemPerCoroutine =
+      new ConcurrentHashMap<>();
```
bantonsson, line 20: Instead of using a concurrent hash map, there is a helper called `ContextStore<K, C>` and `InstrumentationContext`, that allows you to store instances of `C` directly on the `K`. Simple example usage [here](https://github.com/DataDog/dd-trace-java/blob/8c20fa58d733f6caeeee632cbe00e562018af2f1/dd-java-agent/instrumentation/hazelcast-3.9/src/main/java/datadog/trace/instrumentation/hazelcast39/ClientMessageInstrumentation.java#L69). Please note that you have to declare the context stores that you want to use in the instrumentations.

bantonsson (commented):
Thanks for the contribution @monosoul. I don't know enough about Kotlin Coroutines, but it feels like the jobs should already inherit the same `CoroutineContext`, and only if there is a _new_ job should we add a new `CoroutineContext`. Maybe we could just keep track of the associated job in the `CouroutineContext`?

monosoul:
> but it feels like the jobs should already inherit the same `CoroutineContext`, and only if there is a _new_ job should we add a new `CoroutineContext`. Maybe we could just keep track of the associated job in the CouroutineContext?

By inheritance here I mean that each child job (a new job created within another job) inherits it's context. This is the default behavior in coroutines that we were overriding before to make sure each coroutine has it's own instance of scope stack. And this approach is totally fine in general, but unfortunately [this check here](https://github.com/Kotlin/kotlinx.coroutines/blob/master/kotlinx-coroutines-core/common/src/flow/internal/SafeCollector.common.kt#L29) will make the flow collector [throw an exception](https://github.com/Kotlin/kotlinx.coroutines/blob/master/kotlinx-coroutines-core/common/src/flow/internal/SafeCollector.common.kt#L67) here. I suppose JetBrains added that check there expecting scope context can only be modified manually and it kinda makes sense there, but not in our case. So basically now any coroutine will share scope state context element with all it's child jobs, while still having a unique scope state for every job.

dd-java-agent/instrumentation/kotlin-coroutines/src/main/java/datadog/trace/instrumentation/kotlin/coroutines/ScopeStateCoroutineContext.java
monosoul: I don't think I can use it :thinking: It should only be called from instrumentation, but I need to access this map from `ScopeStateCoroutineContext`

bantonsson: That is only partially true. You can only get access to a generated `ContextStore` in an aspect, but it's perfectly fine to store it in a helper class and use it there. Example [here](https://github.com/DataDog/dd-trace-java/blob/acb275906752bc220630dc268f7dcede14c68805/dd-java-agent/instrumentation/netty-3.8/src/main/java/datadog/trace/instrumentation/netty38/client/HttpClientResponseTracingHandler.java#L20)

monosoul: Hmm, I tried that but it was giving me runtime exceptions :thinking: Okay, let me try again, maybe I missed something

monosoul: Yep, it works, probably I missed something last time. Thanks!

monosoul: Done

bantonsson (approved):

PR #4728, opened 2023-02-13
Remove span name

jpbempel:
# What Does This Do
Use constant for OperationName (spanName)
Use method name eventually + line numbers for resource name 
make it deserializable from remote config

# Motivation

# Additional Notes

### DIFF
--- a/dd-java-agent/agent-debugger/src/main/java/com/datadog/debugger/agent/DebuggerTracer.java
+++ b/dd-java-agent/agent-debugger/src/main/java/com/datadog/debugger/agent/DebuggerTracer.java
@@ -8,13 +8,16 @@ import datadog.trace.bootstrap.instrumentation.api.AgentTracer;
 import datadog.trace.bootstrap.instrumentation.api.ScopeSource;
 
 public class DebuggerTracer implements DebuggerContext.Tracer {
+  private static final String OPERATION_NAME = "dd.dynamic.span";
+
   @Override
-  public DebuggerSpan createSpan(String operationName, String[] tags) {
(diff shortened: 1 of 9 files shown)


dd-java-agent/agent-debugger/src/main/java/com/datadog/debugger/agent/DebuggerTracer.java
```diff
@@ -8,13 +8,16 @@
...
-  public DebuggerSpan createSpan(String operationName, String[] tags) {
+  public DebuggerSpan createSpan(String resourceName, String[] tags) {
     AgentTracer.TracerAPI tracerAPI = AgentTracer.get();
     if (tracerAPI == null) {
       return DebuggerSpan.NOOP_SPAN;
     }
-    AgentSpan dynamicSpan = tracerAPI.startSpan(operationName);
+    AgentSpan dynamicSpan =
+        tracerAPI.buildSpan(OPERATION_NAME).withResourceName(resourceName).start();
     if (tags != null) {
```
shatzi, line 21: I didn't know we add probe tags into the span, that is very cool. will match in python impl

Note sure what is the right thing here: but tags can be only key without ":value". why are we a not adding those?

shatzi (approved):
Looks good.

dd-java-agent/agent-debugger/src/main/java/com/datadog/debugger/agent/DebuggerTracer.java
jpbempel: see https://github.com/DataDog/dd-trace-java/pull/4466#discussion_r1057715740

PR #4729, opened 2023-02-13
Blocking: support redirect action

cataphract:
Support redirecting responses when blocking.

### DIFF
--- a/dd-java-agent/agent-bootstrap/src/main/java/datadog/trace/bootstrap/blocking/BlockingActionHelper.java
+++ b/dd-java-agent/agent-bootstrap/src/main/java/datadog/trace/bootstrap/blocking/BlockingActionHelper.java
@@ -64,6 +64,9 @@ public class BlockingActionHelper {
     if (blockingContentType == BlockingContentType.JSON) {
       return TemplateType.JSON;
     }
+    if (blockingContentType == BlockingContentType.NONE) {
+      throw new IllegalArgumentException("Does not accept BlockingContentType.NONE");
+    }
 
     float jsonPref = 0;
     Specificity curJsonSpecificity = Specificity.UNSPECIFIED;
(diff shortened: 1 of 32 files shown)


dd-java-agent/agent-bootstrap/src/main/java/datadog/trace/bootstrap/blocking/BlockingActionHelper.java
```diff
@@ -64,6 +64,9 @@ public static TemplateType determineTemplateType(
     if (blockingContentType == BlockingContentType.JSON) {
       return TemplateType.JSON;
     }
+    if (blockingContentType == BlockingContentType.NONE) {
+      throw new IllegalArgumentException("Does not accept BlockingContentType.NONE");
```
smola, line 68: Is this really the desired behavior? (rather than falling back to auto-detection?)

dd-java-agent/appsec/src/main/java/com/datadog/appsec/blocking/BlockingServiceImpl.java
```diff
@@ -56,17 +57,20 @@ public BlockingDetails shouldBlockUser(@Nonnull String userId) {
...
     return null;
   }
 
   @Override
-  public boolean tryCommitBlockingResponse(int statusCode, BlockingContentType templateType) {
+  public boolean tryCommitBlockingResponse(
+      int statusCode, BlockingContentType templateType, Map<String, String> extraHeaders) {
     log.info(
-        "Will try to commit blocking response statusCode={} templateType={}",
+        "Will try to commit blocking response statusCode={} templateType={} extraHeaders={}",
```
smola, line 70: Headers might contain sensitive infor I think?
If that's the case, this should be logged to `debug` level, or it should not log the headers.

dd-java-agent/agent-bootstrap/src/main/java/datadog/trace/bootstrap/blocking/BlockingActionHelper.java
cataphract: Yes. On redirect responses, I'm not sending any content in the body, though technically it would be possible to send a body as well.

dd-java-agent/appsec/src/main/java/com/datadog/appsec/blocking/BlockingServiceImpl.java
cataphract: These are the extra headers sent in the blocking response (used to include a `Location` header in the redirect responses), not the headers the request. They're not even the total response headers, if somehow any extra headers were added to the response.

smola (approved):

PR #4733, opened 2023-02-14
Update agent feature keys to refer to corresponding properties

nikita-tkachenko-datadog:
# What Does This Do
This change updates property names used by `AgentFeature` enum instances to refer to constants defined in config classes.

# Motivation
The property names are supposed to correspond to the constants, but it's easy to introduce a mistake (e.g. when a constant changes).

# Additional Notes
There was a mistake in `TRACING` instance, which referred to `"dd.tracing.enabled"` property, while the constant defined in the config was `"dd.trace.enabled"`.
As the result, tracing was enabled by default for customers that used CI visibility.
It was decided to leave it this way, since this is the behaviour that we've had for a while (due to the mismatch in property names).


albertvaka (approved):

albertvaka (approved):
I've reviewed it again more carefully and it seems okay 👍

PR #4736, opened 2023-02-14
Fix eval error in log template preventing capture

jpbempel:
# What Does This Do
introduced hasConditionErrors an hasLogTemplateErrors in SnapshotStatus
do not eval twice the summary builder entry/exit

# Motivation
evaluation errors in log template prevent to include captured data into the snapshot sent

# Additional Notes


dd-java-agent/agent-debugger/debugger-bootstrap/src/main/java/datadog/trace/bootstrap/debugger/Snapshot.java
```diff
@@ -273,14 +270,29 @@ private boolean checkCapture(CapturedContext capture, MethodLocation methodLocat
...
   }
 
+  private void handleEvalErrors(
+      SnapshotStatus status,
+      CapturedContext capture,
+      String currentProbeId,
+      Consumer<SnapshotStatus> setHasError) {
+    if (capture.hasEvaluationErrors()) {
+      setHasError.accept(status);
+      List<EvaluationError> captureEvalErrors = extractEvaluationErrors(capture);
```
shatzi, line 288: why does all errors are recorded inside the context? so we then need to clear that list after each evaluate.
Maybe we can simplify this by return result+error from evaluate script and then add it to the right status.

Another solution would be creating a subclass "CaptureContextForProbe" that would automatically add errors to right list.

dd-java-agent/agent-debugger/debugger-bootstrap/src/main/java/datadog/trace/bootstrap/debugger/Snapshot.java
```diff
@@ -98,7 +99,7 @@ private void addSnapshotStatus(ProbeDetails probe) {
   }
 
   public void setEntry(CapturedContext context) {
-    processSummaries(SummaryBuilder::addEntry, context);
+    processSummaries(SummaryBuilder::addEntry, context, MethodLocation.ENTRY);
```
shatzi, line 102: mmm... why are we creating summary before we check the condition?

If we have a log probe without capture but with a condition -> we should only generate log if condition is true, right?

if we have a log with capture + condition. no need to generate summary if condition is false.

If this is true, aka, we should check condition before summary, then we might also simplify this even more. as errors from condition always happen before errors of summary. so, we no longer need to keep track where the errors came from.

dd-java-agent/agent-debugger/debugger-bootstrap/src/main/java/datadog/trace/bootstrap/debugger/Snapshot.java
```diff
@@ -273,14 +270,29 @@ private boolean checkCapture(CapturedContext capture, MethodLocation methodLocat
...
     }
     return probe.getEvaluateAt() == methodLocation;
   }
 
+  private void handleEvalErrors(
+      SnapshotStatus status,
+      CapturedContext capture,
+      String currentProbeId,
+      Consumer<SnapshotStatus> setHasError) {
+    if (capture.hasEvaluationErrors()) {
```
shatzi, line 286: this function can be refactored into SnapshotStatus I think:

```
status.setConditionErrors(extractEvaluationErrors(capture));

status.setConditionErrors(errors) {
  if (!errors.isEmpty()) {
    this.errors.addAll(errors)
    this.hasConditionErrors = true;
    this.sending = true;
  }
}
```

shatzi (commented):
If I understand this code change right - we might have a better way to resolve this.

shatzi (approved):

PR #4743, opened 2023-02-15
HTTP client path mapping

ygree:
# What Does This Do

Introduces `TRACE_HTTP_CLIENT_PATH_RESOURCE_NAME_MAPPING` to define mapping for HTTP client paths.
It provides a way to define rules to map URL paths matching a pattern to a resource name and thus reduce unwanted cardinality, e.g. to collapse all JSP page calls into one resource name:
`/some/path/main/page.jsp`
`/some/path/settings/user/profile.jsp`
And with the `/some/**/*.jsp:some-jsp` mapping it will group those paths into one resource `some-jsp`.

Also, it supports excluding certain resources from being normalized, it's enabled with a resource name is `*`. E.g.
with mapping`/foo/*/bar:*` it preserves original `/foo/123/bar` whereas without this rule it would be normalized `/foo/?/bar`. This can be useful to tune Java Tracer to disable normalization for defined paths. E.g.
There are two distinct resources:
`/some-path/onlineId2accountId` and `/some-path/accountId2onlineId`
with `SimplePathNormalizer` it's mapped to `/some-path/?`
whereas with a defined mapping `/some-path/*:*` it preserves original path and these resources can be viewed separately.

Note, that this ability to exclude a path from normalization will work not only for client HTTP requests but for server too `TRACE_HTTP_SERVER_PATH_RESOURCE_NAME_MAPPING`, because it's part of shared class `AntPatternPathNormalizer`.

This mechanism only provides a way to configure simple mappings from a ant pattern to a resource name and doesn't allow any rewritings. It can only group paths to some defined names and exclude certain URL paths from being grouped by keeping the original value for the resource name, when matching pattern is mapped to `*`.

# Motivation

By default, Java Tracer normalizes URLs to be used as resource names. Currently, HTTP client calls use `SimplePathNormalizer` that replaces path chunks that look like a variable (has digits, except for v?? pattern) with `?`. It's done this way to reduce cardinality. But sometimes users want to adjust it for certain URL patterns by either keeping original URL path as a resource name or by grouping multiple URL path to the same resource name.

# Additional Notes

APMS-8428
APMS-8165
APMS-6080
APMS-8823

### DIFF
--- a/dd-java-agent/agent-bootstrap/src/main/java/datadog/trace/bootstrap/instrumentation/decorator/http/HttpResourceDecorator.java
+++ b/dd-java-agent/agent-bootstrap/src/main/java/datadog/trace/bootstrap/instrumentation/decorator/http/HttpResourceDecorator.java
@@ -29,27 +29,33 @@ public class HttpResourceDecorator {
   private static final DDCache<Pair<CharSequence, CharSequence>, CharSequence> RESOURCE_NAME_CACHE =
       DDCaches.newFixedSizeCache(64);
 
+  private static final DDCache<Pair<CharSequence, CharSequence>, CharSequence>
+      CLIENT_RESOURCE_NAME_CACHE = DDCaches.newFixedSizeCache(64);
+
   private final boolean shouldSetUrlResourceName =
       Config.get().isRuleEnabled("URLAsResourceNameRule");
 
-  private final AntPatternPathNormalizer antPatternPathNormalizer;
+  private final AntPatternPathNormalizer antPatternServerPathNormalizer;
+  private final AntPatternPathNormalizer antPatternClientPathNormalizer;
   private final SimplePathNormalizer simplePathNormalizer;
 
   private HttpResourceDecorator() {
-    antPatternPathNormalizer =
+    antPatternServerPathNormalizer =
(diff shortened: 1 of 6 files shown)


dd-java-agent/agent-bootstrap/src/main/java/datadog/trace/bootstrap/instrumentation/decorator/http/HttpResourceDecorator.java
```diff
@@ -29,27 +29,33 @@ public class HttpResourceDecorator {
   private static final DDCache<Pair<CharSequence, CharSequence>, CharSequence> RESOURCE_NAME_CACHE =
       DDCaches.newFixedSizeCache(64);
 
+  private static final DDCache<Pair<CharSequence, CharSequence>, CharSequence>
```
ygree, line 32: Not sure if a separate cache for client request resources is needed. Perhaps, one common cache for both server and client resources will be enough.

bantonsson: I think a separate cache is fine.

bantonsson (approved):

PR #4750, opened 2023-02-16
Add 128-bit trace id generation support

PerfectSlayer:
# What Does This Do

Introduce changes to support 128-bit trace ids.

# Motivation

Provide better interoperability with external tracing systems.

# Additional Notes

I wonder if it won't be cleaner to abstract `DDTraceId` and have dedicated types, one for 64-bit id and another for 128-bit id.  
If such types existed, I should be easier to make accessible only 64-bit formatting methods only to the 64-bit type and disambiguate existing ones, like `fromString()` / `fromHex()` and `toString()` / `toHex()`, which should not be different for the 128-bit type as its String representation is a 32 hex chars String.
But such change would be more invasive.

About `Trace128Strategy` enum, I might remove it for a 128-bit generation flag boolean only.
It seems like the code in charge of injecting trace id into logs already heavily rely on `Config` to get other values to inject (like serviceName, env, version) so it should be able to query `Config` 128-bit logging by itself.


PerfectSlayer:
Replaced by #4918

PR #4755, opened 2023-02-16
Add test suite level visibility support for JUnit4 and JUnit5

nikita-tkachenko-datadog:
# What Does This Do
This change adds hook into test lifecycle events reported by JUnit4 and JUnit5 frameworks.
The hooks create additional spans that correspond to starting/finishing a test module or a test suite.

# Motivation
This is needed to support [test suite level visibility](https://docs.google.com/document/d/1Xr7SSi_68rjQ4lbTtEoj2uQEOhD6RJhXltzBa_W6nxA/edit#heading=h.22zs5d9o0pw1) for JUnit4 and JUnit5 projects.

# Additional Notes
Logic that is common to all instrumented test frameworks was extracted into `TestEventsHandler` class.
Future test suite level visibility support for TestNG will also be based on the code available there.


internal-api/src/main/java/datadog/trace/bootstrap/instrumentation/api/Tags.java
```diff
@@ -50,6 +54,11 @@ public class Tags {
   public static final String TEST_SOURCE_FILE = "test.source.file";
   public static final String TEST_SOURCE_START = "test.source.start";
   public static final String TEST_SOURCE_END = "test.source.end";
+  public static final String TEST_TRAITS = "test.traits";
+
+  public static final String TEST_SESSION_ID = "test_session_id";
```
devinsba, line 59: Are these intended to use `_` instead of the `.` above

devinsba (approved):

internal-api/src/main/java/datadog/trace/bootstrap/instrumentation/api/Tags.java
nikita-tkachenko-datadog: Yup, this is intentional, it is what the backend expects to receive

PR #4764, opened 2023-02-17
Add test suite level visibility support for TestNG

nikita-tkachenko-datadog:
# What Does This Do
This change adds hook into test lifecycle events reported by TestNG framework.
The hooks create additional spans that correspond to starting/finishing a test module or a test suite.

# Motivation
This is needed to support [test suite level visibility](https://docs.google.com/document/d/1Xr7SSi_68rjQ4lbTtEoj2uQEOhD6RJhXltzBa_W6nxA/edit#heading=h.22zs5d9o0pw1) for TestNG projects.


devinsba (approved):

dd-java-agent/instrumentation/testng/testng-7.5/src/main/java/datadog/trace/instrumentation/testng/TestNG75ClassListenerInstrumentation.java
```diff
@@ -0,0 +1,89 @@
+package datadog.trace.instrumentation.testng;
```
devinsba, line 1: You might want to consider adding the version numbers to the package names to avoid issues. I see you adding the version number to the class name but if you forget there can be hard to diagnose issues if two classes have the same name

nikita-tkachenko-datadog: Thank you for the comment, I saw it too late unfortunately.
That definitely makes sense, although given that this is merged already I will keep it as is for now.
Should I have to change anything in those modules, I will update the package name.

PR #4765, opened 2023-02-19
Set 'env' to 'none' if not set or empty

nikita-tkachenko-datadog:
# What Does This Do
This change updates span data normalisation logic used by DD Intake writer (i.e. in configurations where the tracer writes directly to backend, with no DD Agent involved).
The logic is updated so that `env` tag is set to `none` if no other value is provided for it (or if provided value is empty).

# Motivation
This was requested by the backend.

# Additional Notes
Normally this data normalisation is done by the agent, but in configurations where the tracer writes directly to the backend, it has to be replicated in the tracer.

### DIFF
--- a/internal-api/src/test/groovy/datadog/trace/util/TraceUtilsTest.groovy
+++ b/internal-api/src/test/groovy/datadog/trace/util/TraceUtilsTest.groovy
@@ -114,8 +114,8 @@ class TraceUtilsTest extends DDSpecification {
 
     where:
     env | expected
-    null | ""
-    "" | ""
+    null             | TraceUtils.DEFAULT_ENV
+    ""               | TraceUtils.DEFAULT_ENV
     "ok" | "ok"
     repeat("a",300)|repeat("a",200)
   }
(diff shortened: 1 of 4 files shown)


internal-api/src/test/groovy/datadog/trace/util/TraceUtilsTest.groovy
```diff
@@ -12,11 +12,11 @@ class TraceUtilsTest extends DDSpecification {
     normalized == expected
 
     where:
-    service | expected
-    null | TraceUtils.DEFAULT_SERVICE_NAME
-    "" | TraceUtils.DEFAULT_SERVICE_NAME
-    "good" | "good"
-    "bad\$service" | "bad_service"
+    service                                                                                                                                | expected
```
smola, line 15: Unless there's an actual required change here, or spotless is complaining in CI, it's better to avoid unrelated formatting changes in PRs.

nikita-tkachenko-datadog: Good point!
I think the Save Actions plugin config suggested in our [contributing guide](https://github.com/DataDog/dd-trace-java/blob/master/CONTRIBUTING.md#intellij-idea) is too "eager".
I had to change `Reformat file` to `Reformat only changed code` to get rid of the extra changes

albertvaka (approved):

PR #4766, opened 2023-02-19
Add source file path detection for Kotlin

nikita-tkachenko-datadog:
# What Does This Do
This change updates logic that detects class source file paths, so that Kotlin classes are supported.

# Motivation
The logic is used by CI Visibility part of the tracer to set source code related tags.
Some of CI Visibility customers are using it with Kotlin-based projects.

# Additional Notes
The change consisted of three main parts:
* update the list of file extensions that are added to repository index, adding `.kt`
* add logic to infer class source language
* update condition for parsing class information to obtain the source file name attribute (previously that was done only for package-private classes, not it is also done for non-Java classes)


PerfectSlayer (approved):
LGTM

PR #4767, opened 2023-02-20
naming schema: couchbase

amarziali:
# What Does This Do

v1 naming schema changes for couchbase:
* service: `{{DD_SERVICE}}-couchbase`
* operation: `couchbase.query`

# Motivation

# Additional Notes

### DIFF
--- a/dd-java-agent/instrumentation/couchbase/couchbase-2.0/src/test/groovy/CouchbaseClientTest.groovy
+++ b/dd-java-agent/instrumentation/couchbase/couchbase-2.0/src/test/groovy/CouchbaseClientTest.groovy
@@ -5,6 +5,7 @@ import com.couchbase.client.java.document.JsonDocument
 import com.couchbase.client.java.document.json.JsonObject
 import com.couchbase.client.java.env.CouchbaseEnvironment
 import com.couchbase.client.java.query.N1qlQuery
+import datadog.trace.api.Config
 import datadog.trace.test.util.Flaky
 import spock.lang.Unroll
 import util.AbstractCouchbaseTest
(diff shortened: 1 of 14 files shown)


dd-java-agent/instrumentation/couchbase/couchbase-2.0/src/test/groovy/util/AbstractCouchbaseTest.groovy
```diff
@@ -158,4 +158,19 @@ abstract class AbstractCouchbaseTest extends AgentTestRunner {
...
+
+  @Override
+  protected String service() {
+    return "couchbase"
+  }
+
+  @Override
+  protected String operation() {
+    return "couchbase.call"
+  }
```
PerfectSlayer, line 175: Why settings default values with v0 name logic here?
I found clearer when v0 values are defined in child classes, near the v1 child classes so you can quickly see the behavior difference. (like you did in all other versioned tests)

dd-java-agent/instrumentation/couchbase/couchbase-2.0/src/test/groovy/CouchbaseClientTest.groovy
```diff
@@ -123,3 +124,24 @@ class CouchbaseClientTest extends AbstractCouchbaseTest {
     cleanupCluster(cluster, environment)
   }
 }
+
+class CouchbaseClientV0ForkedTest extends CouchbaseClientTest {
+
```
PerfectSlayer, line 129: So here, you miss the context of the values you are testing 😞

PerfectSlayer (approved):
Minor comment about one test case. LGTM otherwise 👍

dd-java-agent/instrumentation/couchbase/couchbase-2.0/src/test/groovy/util/AbstractCouchbaseTest.groovy
amarziali: I did this way not to check v0 and v1 on _every_ test. In this way I have the coverage for v1 that I need without changing all the tests. In this way we also ensure that, without changes, v0 is the one used

PR #4770, opened 2023-02-20
spring webflux client 6

amarziali:
# What Does This Do

Instrument spring 6 webflux client. Old instrumentation was not working because the decorator relies on `org.springframework.web.reactive.function.client.ClientResponse#statusCode()Lorg/springframework/http/HttpStatus` but the `HttpStatus` interface is no more present on spring webflux 6 (it has been renamed to `HttpStatusCode`)

Before:

image

After *(now the `http.request` client span is present)*

image

# Motivation

# Additional Notes


devinsba (approved):
Looks good to me

PR #4776, opened 2023-02-21
Fix internal-api test compile with JDK >8 default

smola:
# What Does This Do
Forces JDK 8 when compiling internal-api tests.

# Motivation
Unless JAVA_HOME was set to JDK 8, internal-api test compilation would fail.

# Additional Notes

### DIFF
--- a/internal-api/build.gradle
+++ b/internal-api/build.gradle
@@ -18,6 +18,10 @@ compileJava {
 apply from: "$rootDir/gradle/java.gradle"
 apply from: "$rootDir/gradle/tries.gradle"
 
+tasks.compileTestJava.configure {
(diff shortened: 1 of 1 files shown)


richardstartin (approved):

PerfectSlayer (commented):
Does it imply change on our requirements? https://github.com/DataDog/dd-trace-java/blob/master/CONTRIBUTING.md#requirements

smola:
@PerfectSlayer It shouldn't. JDK 8 should still be the main JDK used in the command line. But some of us use JDK 11 as the JAVA_HOME for IntelliJ (for other compatibility problems I don't remember...) and then Gradle picks it to compile tests in this submodule.

PR #4778, opened 2023-02-22
Test the right version in test-published-dependencies

bantonsson:
# What Does This Do

Removes the dependency on `axion-release` from _external_ gradle projects and makes `test-published-dependencies` use a file with the right version written by the `main` project.

# Motivation

The `scmVersion.repository.directory` setting of the `axion-release` plugin doesn't seem to work in the same way anymore and the `currentVersion` task returns the wrong version, i.e. `1.8.0` instead of `1.9.0-SNAPSHOT`, so we test the wrong version in `test-published-dependencies`.

# Additional Notes

The dependency on the `axion-release` plugin in a number of smoke tests had already been _removed_ in 0e306a78196 by setting `version = ""`.

### DIFF
--- a/build.gradle
+++ b/build.gradle
@@ -1,6 +1,6 @@
 buildscript {
   dependencies {
-    classpath "pl.allegro.tech.build:axion-release-plugin:1.14.2"
+    classpath "pl.allegro.tech.build:axion-release-plugin:1.14.4"
   }
 
   configurations.all {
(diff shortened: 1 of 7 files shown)


build.gradle
```diff
@@ -112,6 +112,28 @@ wrapper {
...
+
+def writeMainVersionFileTask = tasks.register('writeMainVersionFile') {
+  def versionFile = file("${rootProject.buildDir}/main.version")
+  inputs.property "version", scmVersion.version
+  outputs.file versionFile
+
+  doFirst {
+    assert versionFile.parentFile.mkdirs() || versionFile.parentFile.directory
+    versionFile.text = "${inputs.properties.version}"
+  }
```
PerfectSlayer, line 134: Why create a new file (in a new `build` folder) rather than using the already generated files `<project>/build/resources/main/<project>.version` like `dd-java-agent/build/resources/main/dd-java-agent.version`?

PerfectSlayer (commented):
Quick question first about reusing already existing `.version` files.

build.gradle
bantonsson: Because they also include the latest git hash, which the locally published versions don't.

PerfectSlayer: Sure. It should be easy to parse though.

PerfectSlayer (approved):

PR #4780, opened 2023-02-22
Naming schema: http clients

amarziali:
# What Does This Do

v1 changes for http client naming schema:
* operation: `http.client.request`

Service name is unchanged.

# Motivation

# Additional Notes

### DIFF
--- a/dd-java-agent/instrumentation/couchbase/couchbase-2.0/src/test/groovy/CouchbaseAsyncClientTest.groovy
+++ b/dd-java-agent/instrumentation/couchbase/couchbase-2.0/src/test/groovy/CouchbaseAsyncClientTest.groovy
@@ -189,17 +189,17 @@ class CouchbaseAsyncClientV0ForkedTest extends CouchbaseAsyncClientTest {
 
 class CouchbaseAsyncClientV1ForkedTest extends CouchbaseAsyncClientTest {
   @Override
-  protected int version() {
+  int version() {
     return 1
   }
 
   @Override
-  protected String service() {
+  String service() {
     return Config.get().getServiceName() + "-couchbase"
   }
 
   @Override
-  protected String operation() {
+  String operation() {
     return "couchbase.query"
   }
 }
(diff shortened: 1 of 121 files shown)


dd-java-agent/instrumentation/play-2.3/src/test/groovy/client/PlayWSClientTest.groovy
```diff
@@ -10,7 +11,7 @@ import spock.lang.Shared
 
 import java.util.concurrent.TimeUnit
 
-class PlayWSClientTest extends HttpClientTest {
+class PlayWSClientTest extends HttpClientTest implements TestingNettyHttpNamingConventions.ClientV0 {
```
ygree, line 14: I wonder why there is no corresponding test for V1?

amarziali: I gave a default coverage to `v0` to "limit" the number of test to subclass. I should have covered anyway the v1 behaviour on some test for each play version like here: https://github.com/DataDog/dd-trace-java/blob/b72b5cf6ce992809b5f31d892440068e06cb1ffc/dd-java-agent/instrumentation/play-ws-2/src/test/groovy/PlayWSClientTest.groovy#L138

Does it sound reasonable to you?

dd-java-agent/instrumentation/couchbase/couchbase-2.0/src/test/groovy/CouchbaseAsyncClientTest.groovy
```diff
@@ -189,17 +189,17 @@ class CouchbaseAsyncClientV0ForkedTest extends CouchbaseAsyncClientTest {
 
 class CouchbaseAsyncClientV1ForkedTest extends CouchbaseAsyncClientTest {
   @Override
-  protected int version() {
+  int version() {
```
ygree, line 192: What's the goal of this `protected` removal? Looks like a clean up to avoid extra keyword `protected`. I think it would be better to have it in a separate PR or commit. Probably it's not worth doing it now though, when it's already part of the same commit.

amarziali: Yes normally this should have been addressed separately as suggested. The issue is that I create some traits here: `dd-java-agent/testing/src/main/groovy/datadog/trace/agent/test/naming/NamingConventions.groovy` to remove boilerplate and I cannot have protected methods. So I had to change a bit the rest

ygree (approved):
LGTM

PR #4786, opened 2023-02-23
Align DD_TRACE_PROPAGATION_STYLE* names with other tracers

bantonsson:
# What Does This Do

Changes some names of `DD_TRACE_PROPAGATION_STYLE_*` settings to align with other Datadog tracers.

# Motivation

Minimize discrepancies in configuration between Datadog tracers.

# Additional Notes

This is a breaking change if someone started using `DD_TRACE_PROPAGATION_STYLE*` settings and the `B3` propagator since `B3` now mean [B3 multi header](https://github.com/openzipkin/b3-propagation/tree/master#multiple-headers) while before it meant [B3 single header](https://github.com/openzipkin/b3-propagation/tree/master#single-header). The risk of this is considered low, since the setting has not been documented yet.


PerfectSlayer (approved):
Love the `NoneCodec`. We should not find compatibility issue with it 😉

PR #4788, opened 2023-02-23
naming schema: servlet, tomcat and jetty

amarziali:
# What Does This Do

v1 naming schema changes for servlet, tomcat and jetty:
* operation:  `http.server.request`

# Motivation

# Additional Notes

### DIFF
--- a/dd-java-agent/agent-bootstrap/src/main/java/datadog/trace/bootstrap/instrumentation/decorator/HttpServerDecorator.java
+++ b/dd-java-agent/agent-bootstrap/src/main/java/datadog/trace/bootstrap/instrumentation/decorator/HttpServerDecorator.java
@@ -15,6 +15,7 @@ import datadog.trace.api.gateway.Flow;
 import datadog.trace.api.gateway.IGSpanInfo;
 import datadog.trace.api.gateway.RequestContext;
 import datadog.trace.api.gateway.RequestContextSlot;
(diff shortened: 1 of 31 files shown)


cataphract:
These tests take already a long time to run, and you're doubling the amount. I think it's a bit wasteful to rerun everything just to test that the operation name of the span is different. Perhaps just a unit test, something that doesn't require start/stop a server, let alone run all the tests, would be sufficient

amarziali:
> These tests take already a long time to run, and you're doubling the amount. I think it's a bit wasteful to rerun everything just to test that the operation name of the span is different. Perhaps just a unit test, something that doesn't require start/stop a server, let alone run all the tests, would be sufficient

I do not think a unit test on the SpanNaming class is enough because the operation name is assigned when the span is created and this is often done on the advices. Now, my goal was to make sure I do not forget to handle v1 naming and this is the reason why I'm leveraging existing tests to check also for v1. I understand that forking tests is making CI time skyrocketing. The problem is that, since the operation and services name are assigned statically once on the decorator for performance reason, I have to fork the jvm to have a fresh instance taking the new configuration. Also for this reason you will see that I did not fork every test but only most important ones. About the forking, if you have a suggestion to avoid it or to speed up testing please tell me how make the build faster still ensuring the same coverage

ygree (approved):

PR #4789, opened 2023-02-23
Exclude datadog.trace.api.internal.util from shading in dd-trace-ot

bantonsson:
# What Does This Do

Excludes `datadog.trace.api.internal.util` from shading in `dd-trace-ot`

# Motivation

It should not be shaded.

# Additional Notes

Does not affect anything right now by pure luck, since none of the `core` classes access anything in that package directly. The W3C propagator will, and then things break since the class won't exist.

### DIFF
--- a/dd-trace-ot/build.gradle
+++ b/dd-trace-ot/build.gradle
@@ -111,6 +111,7 @@ shadowJar {
     exclude('datadog.trace.api.experimental.*')
     exclude('datadog.trace.api.interceptor.*')
     exclude('datadog.trace.api.internal.*')
+    exclude('datadog.trace.api.internal.util.*')
     exclude('datadog.trace.api.sampling.*')
     exclude('datadog.trace.context.*')
   }


smola (approved):

PerfectSlayer (approved):

mcculls (approved):

dd-trace-ot/build.gradle
```diff
@@ -111,6 +111,7 @@ shadowJar {
     exclude('datadog.trace.api.experimental.*')
     exclude('datadog.trace.api.interceptor.*')
     exclude('datadog.trace.api.internal.*')
+    exclude('datadog.trace.api.internal.util.*')
     exclude('datadog.trace.api.sampling.*')
```
PerfectSlayer, line 115: What about excluding the whole 'datadog.trace.api.**' instead?

bantonsson: There was a rationale about why not [here](https://github.com/DataDog/dd-trace-java/pull/4263#discussion_r1024498699)

PR #2780, opened 2021-06-01
Remove trace processor

devinsba:
The goal is to remove the trace processor, without changing trace structures (except where they were wrong before)

PR #4798, opened 2023-02-25
Add missing URLAsResourceNameRule functionality to TagInterceptor

devinsba:
# What Does This Do

Restores some of the missing functionality from `URLAsResourceNameRule` that was inadvertently lost as part of https://github.com/DataDog/dd-trace-java/pull/2780. Specifically the part that combines the `http.method` and `http.url` tags into a HTTP resource name.

# Motivation

A customer uses clojure which requires manual instrumentation. When they updated to 0.81.0+ their resource names stopped working like they should have.

# Additional Notes

Additionally, there was a part of `URLAsResourceNameRule` that handled the status 404 case so this has also been restored to fix tests that I copied from the previous implementation


dd-java-agent/agent-bootstrap/src/main/java/datadog/trace/bootstrap/instrumentation/decorator/http/HttpResourceDecorator.java
```diff
@@ -2,61 +2,25 @@
...
-              return UTF8BytesString.create(input.getRight());
-            }
-            return UTF8BytesString.create(
-                input.getLeft().toString().toUpperCase() + " " + input.getRight());
-          };
-
-  private static final DDCache<Pair<CharSequence, CharSequence>, CharSequence> RESOURCE_NAME_CACHE =
-      DDCaches.newFixedSizeCache(64);
-
-  private static final DDCache<Pair<CharSequence, CharSequence>, CharSequence>
```
devinsba, line 32: @ygree Can you check this? I had to move the code you just added

dd-trace-core/src/main/java/datadog/trace/core/taginterceptor/TagInterceptor.java
```diff
@@ -91,6 +108,29 @@ public boolean interceptTag(DDSpanContext span, String tag, Object value) {
...
 
+  private boolean interceptUrlResourceAsNameRule(DDSpanContext span, String tag, Object value) {
+    if (shouldSetUrlResourceAsName) {
+      if (HTTP_METHOD.equals(tag)) {
+        if (span.getTags().containsValue(HTTP_URL)) {
+          Pair<String, Byte> normalized =
+              HttpPathNormalizers.serverChainWithPriority(
+                  span.getTags().get(HTTP_URL).toString(), false);
+          span.setResourceName(
+              HttpResourceNames.compute(value.toString(), normalized.getLeft()),
```
ygree, line 119: I'd suggest to move calls to `HttpResourceNames` should be inside `HttpPathNormalizers` (or maybe another abstraction), since they are always used together? Or maybe even pass a span to set the resource name. This way, it will combine three calls normalization, cache and setting the resource name. When all combined it will allow to reduce allocations that are needed now to return both the result and priority. It will also move normalization + caching + resource setting in one place.

dd-trace-core/src/main/java/datadog/trace/core/taginterceptor/TagInterceptor.java
```diff
@@ -49,6 +57,12 @@ public TagInterceptor(
     this.splitServiceTags = splitServiceTags;
     this.ruleFlags = ruleFlags;
     splitByServletContext = splitServiceTags.contains(InstrumentationTags.SERVLET_CONTEXT);
+
+    shouldSet404ResourceName =
+        ruleFlags.isEnabled(URL_AS_RESOURCE_NAME)
+            && ruleFlags.isEnabled(STATUS_404)
```
ygree, line 63: Why do we need two rules STATUS_404 & STATUS_404_DECORATOR if it only makes sense when both are enabled. Wouldn't enough to have just one rule for that?

dd-trace-core/src/main/java/datadog/trace/core/taginterceptor/TagInterceptor.java
```diff
@@ -201,10 +241,16 @@ private boolean interceptServletContext(DDSpanContext span, Object value) {
   private boolean interceptHttpStatusCode(DDSpanContext span, Object statusCode) {
     if (statusCode instanceof Number) {
       span.setHttpStatusCode(((Number) statusCode).shortValue());
+      if (shouldSet404ResourceName && span.getHttpStatusCode() == 404) {
+        span.setResourceName(NOT_FOUND_RESOURCE_NAME, ResourceNamePriorities.HTTP_404);
```
ygree, line 245: Wouldn't it replace the routing path if the existing endpoint return 404 explicitly? Or this case will be handled by the priority?

dd-java-agent/agent-bootstrap/src/main/java/datadog/trace/bootstrap/instrumentation/decorator/http/HttpResourceDecorator.java
ygree: Thanks for the heads up, @devinsba! Moving out this logic makes sense to me. So, now we will only have one cache for both client and server resources. Probably that's fine.

ygree (changes requested):
I'd suggest to also move out cache logic and probably resource setting logic into one abstraction to avoid extra allocations and remove logic duplication since normalization + caching + resource setting are always used together.

dd-java-agent/agent-bootstrap/src/main/java/datadog/trace/bootstrap/instrumentation/decorator/http/HttpResourceDecorator.java
ygree: > # What Does This Do
> # Motivation
> # Additional Notes

Would be nice to have a few sentences about this change motivation for easier navigation.

dd-trace-core/src/main/java/datadog/trace/core/taginterceptor/TagInterceptor.java
devinsba: This is how the rule was implemented before. So I brought back all the settings so we don't run into any more issues with customers updating from a couple years ago

https://github.com/DataDog/dd-trace-java/blob/7cf97e4f3c9ba935e51398ad61e15a4e4891ed1d/dd-trace-core/src/main/java/datadog/trace/core/processor/rule/URLAsResourceNameRule.java#L27

dd-trace-core/src/main/java/datadog/trace/core/taginterceptor/TagInterceptor.java
devinsba: The route priority is higher, so it would not change

dd-java-agent/agent-bootstrap/src/main/java/datadog/trace/bootstrap/instrumentation/decorator/http/HttpResourceDecorator.java
devinsba: To address your first comment, there are still 2 caches. I just moved them both to the same place

dd-trace-core/src/main/java/datadog/trace/core/taginterceptor/TagInterceptor.java
devinsba: I've moved everything into `HttpResourceNames` except for the code needed by `TagInterceptor` since it doesn't have access to internal-api

dd-trace-core/src/main/java/datadog/trace/core/taginterceptor/TagInterceptor.java
```diff
@@ -91,6 +107,25 @@ public boolean interceptTag(DDSpanContext span, String tag, Object value) {
     }
   }
 
+  private boolean interceptUrlResourceAsNameRule(DDSpanContext span, String tag, Object value) {
+    if (shouldSetUrlResourceAsName) {
+      if (HTTP_METHOD.equals(tag)) {
```
ygree, line 112: Could these two calls `computeForServer` and `setResourceName` be replaced with `setForServer`. It would open the opportunity to avoid a Pair allocation altogether.

devinsba: `internal-api` doesn't have access to `DDSpanContext` and there isn't a way to get back to `AgentSpan` from `DDSpanContext`

ygree: I see. Thanks for clarification.

ygree (approved):
LGTM

PR #4800, opened 2023-02-27
Fix evaluateAt resolution for line probe

jpbempel:
# What Does This Do
No need to resolve `evaluateAt` attribute for line probe.
config for the probe can now embed `evaluateAt` with any value, even for line probe

# Motivation
Conditions was not evaluated for line probes

# Additional Notes

### DIFF
--- a/dd-java-agent/agent-debugger/debugger-bootstrap/src/main/java/datadog/trace/bootstrap/debugger/Snapshot.java
+++ b/dd-java-agent/agent-debugger/debugger-bootstrap/src/main/java/datadog/trace/bootstrap/debugger/Snapshot.java
@@ -271,7 +271,11 @@ public class Snapshot {
   }
 
   private boolean resolveEvaluateAt(ProbeDetails probe, MethodLocation methodLocation) {
-    if (methodLocation == MethodLocation.DEFAULT || methodLocation == MethodLocation.ENTRY) {
+    if (methodLocation == MethodLocation.DEFAULT) {
+      // line probe, no evaluation of probe's evaluateAt
+      return true;
+    }
+    if (methodLocation == MethodLocation.ENTRY) {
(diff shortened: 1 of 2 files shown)


dd-java-agent/agent-debugger/debugger-bootstrap/src/main/java/datadog/trace/bootstrap/debugger/Snapshot.java
```diff
@@ -271,7 +271,11 @@ private boolean checkCapture(CapturedContext capture, MethodLocation methodLocat
   }
 
   private boolean resolveEvaluateAt(ProbeDetails probe, MethodLocation methodLocation) {
-    if (methodLocation == MethodLocation.DEFAULT || methodLocation == MethodLocation.ENTRY) {
+    if (methodLocation == MethodLocation.DEFAULT) {
+      // line probe, no evaluation of probe's evaluateAt
+      return true;
+    }
+    if (methodLocation == MethodLocation.ENTRY) {
       return probe.getEvaluateAt() == MethodLocation.DEFAULT
```
cimi, line 279: What should a method probe do when no `evaluateAt` is present?

Given our discussions in the agent sync, shouldn't `probe.evaluateAt == DEFAULT` make it evaluate on `MethodLocation.EXIT`?

cimi (approved):
:+1: Thanks!

dd-java-agent/agent-debugger/debugger-bootstrap/src/main/java/datadog/trace/bootstrap/debugger/Snapshot.java
jpbempel: normally there is always an evaluateAt value, but if default, I think we did not reach consensus on this one, and more generally on to handle evaluateAt for different feature. right now, yes this is arbitrary

PR #4815, opened 2023-02-28
Added tests for measured methods config and annotations

nayeem-kamal:
# What Does This Do
Adds tests for https://github.com/DataDog/dd-trace-java/pull/4738
# Motivation

# Additional Notes


dd-java-agent/instrumentation/trace-annotation/src/test/groovy/MeasuredConfigTest.groovy
```diff
@@ -0,0 +1,73 @@
...
+    if (value) {
+      injectSysConfig("dd.trace.methods", value)
+      injectSysConfig("dd.measure.methods", value)
+    } else {
+      removeSysConfig("dd.trace.methods")
+      removeSysConfig("dd.measure.methods")
+    }
+
+    expect:
+    new TraceConfigInstrumentation().classMethodsToTrace == expected
```
ygree, line 23: Looks like it's static. Does it really need to be static? If so, no need to instantiate it then.

dd-java-agent/instrumentation/trace-annotation/src/test/groovy/MeasuredConfigTest.groovy
```diff
@@ -0,0 +1,73 @@
+import datadog.trace.agent.test.AgentTestRunner
+import datadog.trace.bootstrap.instrumentation.api.Tags
+import datadog.trace.bootstrap.instrumentation.traceannotation.MeasuredMethodFilter
+import datadog.trace.instrumentation.trace_annotation.TraceConfigInstrumentation
+
+import java.util.concurrent.Callable
+
+class MeasuredConfigTest extends AgentTestRunner {
```
ygree, line 8: This name a bit misleading as it tests both traced and measured methods settings. Maybe test `TraceAnnotationConfigParser` directly for common cases and add two small ones for `dd.trace.methods` and `dd.measure.methods` just to make sure that setting is read correctly.

dd-java-agent/instrumentation/trace-annotation/src/test/groovy/MeasuredConfigTest.groovy
nayeem-kamal: The classMethodsToTrace variable was already static I believe so I just maintained that pattern. I fixed this test to utilize the variable as static instead of instantiating.

dd-java-agent/instrumentation/trace-annotation/src/test/groovy/MeasuredConfigTest.groovy
nayeem-kamal: Thanks Yury! I moved the parsing tests into a separate test class for the TraceAnnotationConfigParser and added some tests that test the functionality of measured methods in this class.

nayeem-kamal:
New test PR:https://github.com/DataDog/dd-trace-java/pull/4952

PR #4816, opened 2023-02-28
W3C trace context extractor/injector

bantonsson:
# What Does This Do

Implementation of W3C trace context extractor and injector that tries to be as lazy as possible when parsing the `tracestate` header to avoid request overhead for parts that are not used by the Datadog tracer.

# Motivation

Enable better compatibility with other tracers supporting the W3C trace context standard.

# Additional Notes

This PR tries to separate the changes into multiple commits to make review simpler. They are not the most logical, but hopefully better than a single commit.

### DIFF
--- a/dd-trace-core/src/main/java/datadog/trace/core/propagation/ptags/W3CPTagsCodec.java
+++ b/dd-trace-core/src/main/java/datadog/trace/core/propagation/ptags/W3CPTagsCodec.java
@@ -1,0 +1,685 @@
+package datadog.trace.core.propagation.ptags;
+
+import datadog.trace.api.sampling.PrioritySampling;
+import datadog.trace.core.propagation.PropagationTags;
+import datadog.trace.core.propagation.ptags.PTagsFactory.PTags;
+import datadog.trace.core.propagation.ptags.TagElement.Encoding;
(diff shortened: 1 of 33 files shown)


dd-trace-core/src/main/java/datadog/trace/core/propagation/ptags/W3CPTagsCodec.java
```diff
@@ -0,0 +1,685 @@
...
+    if (pos >= end) {
+      // There needs to be something after the equals sign
+      return -1;
+    }
+
+    return pos;
+  }
+
+  // This is The Fine Specification for the member value https://www.w3.org/TR/trace-context/#value
+  // Please note the wonderful decision to allow ' ' in the value but not let it end with ' ', since
```
PerfectSlayer, line 472: _wonderful_ 👌

PerfectSlayer (approved):
The `PropagationTags`/`PTgas` naming does not make thing easy and there are some unanswered questions on W3C propagation tags codec (but I guess it is the best we can assume?)
LGTM but another pair of eyes won't be too much

dougqh (approved):

PR #4795, opened 2023-02-24
[WIP] Add support for scope context

PerfectSlayer:
# What Does This Do

# Motivation

# Additional Notes

PR #4827, opened 2023-03-02
propagate profiling context as auxiliary context in continuations

richardstartin:
# What Does This Do

This allows profiling context to get carried along with span context if a span is active when a context switch happens, by snapshotting the tags in our native side table and attaching them to the continuation. This snapshot is modelled as a `ContinuableContext` which can be activated and deactivated, and I have structured the changes to the `ContinuableScopeManager` and `Continuation` classes to allow this to be something more generic than a snapshot of profiling context in the future, so that any context source could be snapshotted and attached to a continuation as auxiliary context.

The aim here is to make profiling context work alongside tracing, and propagating when there is no active span is currently a non-goal because it would require either a much more generic context propagation mechanism or a sibling implementation which operates independently of the tracing scope manager. 

# Motivation

# Additional Notes


PerfectSlayer:
Hi Richard,
Did you had a look to #4795 if it is something that could fit your need?
Callback mechanism for activation might be different (based on `ScopeListener`).

Issue #4794, opened 2023-02-24
Enabling Data Streams feature - adds nagative bytes values to kafka headers 


agadek:
After enabling the Data Streams feature of Datadog we have found that kafka/msk messages contain headers with negative int values, while the expected type is uint8. In consequence, clients using strict headers validation fails on parsing. 

ex: 
`dd-pathway-ctx:[94 69 9 87 1 74 101 -11 -82 -125 -34 -31 -49 97 -82 -125 -34 -31 -49 97]`

while normal headers look like: 
` x-datadog-tags -> [95, 100, 100, 46, 112, 46, 100, 109, 61, 45, 49]`

dd-java-agant 0.111.0

PR #4835, opened 2023-03-02
DSM context is now passed via dd-pathway-ctx-base64 in all cases

kr-igor:
This PR switched DSM context to `base64`. Only `dd-pathway-ctx-base64` header will be used going forward. Now clients written in other languages can correctly read the header (signed bytes in Java vs unsigned bytes in other languages).

`dd-pathway-ctx` header will be processed correctly, we can remove the logic later, when all of the clients are updated to the most recent version.

Older versions of the tracer will not see the context passed by never versions. It's OK, as we don't have a "minimal required version" for DSM, because the product is not GA yet.

The change also resolves [this issue](https://github.com/DataDog/dd-trace-java/issues/4794).


richardstartin:
Some historical context about the homegrown Base64 encoder - it was introduced back when we had to support JDK7. It should now be replaced by the JDK implementation, which has AVX2 and AVX512 intrinsic implementations on more recent JVMs. I would recommend using `java.util.Base64` in this PR. The custom implementation can be removed in another PR.

kr-igor:
@richardstartin thanks for your review! I'll switch to java.util.Base64.

dd-trace-core/src/main/java/datadog/trace/core/datastreams/DefaultPathwayContext.java
```diff
@@ -190,7 +190,7 @@ public String strEncode() throws IOException {
     if (bytes == null) {
       return null;
     }
-    return new String(Base64.getEncoder().encode(bytes), UTF_8);
+    return new String(bytes, UTF_8);
```
richardstartin, line 193: nit: if these are already base 64 you could use ISO_8859_1 and skip the negatives check

richardstartin (approved):

PR #4840, opened 2023-03-03
Update Groovy and Spock versions

nikita-tkachenko-datadog:
# What Does This Do
Updates Groovy version (2.5 -> 3.0) and Spock version (1.3 -> 2.1).
Migrates all tests that used to run on JUnit 4 to JUnit 5.

# Motivation
Our current version of Spock is 1.3.
It only support Groovy 2.x, which has some drawbacks:
* our Gradle version uses Groovy 3.0.10, as the result `gradleTestKit()` cannot be used due to dependency conflicts
* Groovy 2 has some annoying bugs (e.g. https://issues.apache.org/jira/browse/GROOVY-10137)

# Additional Notes
`SpockRunner` that we use to run instrumentation tests cannot be properly ported to JUnit 5, since the framework does not provide the hooks / extension points that can be used to shadow the tested class (see [Introduce extension API for providing a custom ClassLoader (e.g. for Powermock)](https://github.com/junit-team/junit5/issues/201)).

In order to mitigate this, `SpockRunner` was changed to extend `JUnitPlatform`, which is a JUnit 4 runner that allows executing JUnit 5 tests in a JUnit 4 environment (i.e. running them as JUnit 4 tests).

So even though Spock 2 tests run on top of JUnit 5, we execute them in "compatibility mode" so that `SpockRunner` could shadow the test class (see [SputnikRunner removed](https://spockframework.org/spock/docs/2.3/all_in_one.html#_sputnik_runner_removed_an_alternative) for more details).

### DIFF
--- a/dd-java-agent/agent-iast/src/test/groovy/com/datadog/iast/sink/PathTraversalModuleTest.groovy
+++ b/dd-java-agent/agent-iast/src/test/groovy/com/datadog/iast/sink/PathTraversalModuleTest.groovy
@@ -79,7 +79,7 @@ class PathTraversalModuleTest extends IastModuleImplTestBase {
     '/==>var<==' | "/==>var<=="
   }
 
-  void 'iast module detects path traversal with String (#parent, #child)'(final String first, final String rest, final String expected) {
+  void 'iast module detects path traversal with String (#first, #rest)'() {
     setup:
     final parent = mapTainted(first)
     final children = mapTainted(rest)
--- a/dd-java-agent/instrumentation/build.gradle
+++ b/dd-java-agent/instrumentation/build.gradle
@@ -74,6 +74,19 @@ subprojects { Project subProj ->
     }
 
     subProj.tasks.withType(Test).configureEach { subTask ->
+      // SpockRunner that we use to run agent tests cannot be properly ported to JUnit 5,
+      // since the framework does not provide the hooks / extension points
+      // that can be used to shadow the tested class.
+
+      // In order to mitigate this, SpockRunner extends JUnitPlatform,
+      // which is a JUnit 4 runner that allows executing JUnit 5 tests in a JUnit 4 environment
+      // (i.e. running them as JUnit 4 tests).
+
+      // So even though Spock 2 tests run on top of JUnit 5,
+      // we execute them in "compatibility mode" so that SpockRunner could shadow the test class
+      // See https://junit.org/junit5/docs/current/user-guide/#running-tests-junit-platform-runner for more details.
+      useJUnit()
+
       onlyIf { !project.rootProject.hasProperty("skipInstTests") }
 
       if (subTask.name == 'latestDepTest') {
--- a/dd-java-agent/testing/src/test/groovy/server/HttpServerTest.groovy
+++ b/dd-java-agent/testing/src/test/groovy/server/HttpServerTest.groovy
@@ -64,11 +64,11 @@ class HttpServerTest extends AgentTestRunner {
       .url("$server.address")
       .get()
       .build()
-    def response = client.newCall(request).execute()
+    def clientResponse = client.newCall(request).execute()
 
     then:
-    response.code() == 404
-    response.body().string().contains("<title>Error 404 </title>")
+    clientResponse.code() == 404
+    clientResponse.body().string().contains("<title>Error 404 </title>")
 
     cleanup:
     server.stop()
@@ -97,20 +97,20 @@ class HttpServerTest extends AgentTestRunner {
     }
 
     when:
-    def request = new Request.Builder()
+    def clientRequest = new Request.Builder()
       .url("$server.address/$path")
 
     if (method == "get") {
-      request."$method"()
+      clientRequest."$method"()
     } else {
-      request."$method"(body())
+      clientRequest."$method"(body())
     }
 
-    def response = client.newCall(request.build()).execute()
+    def clientResponse = client.newCall(clientRequest.build()).execute()
 
     then:
-    response.code() == 200
-    response.body().string().trim() == "/$path response"
+    clientResponse.code() == 200
+    clientResponse.body().string().trim() == "/$path response"
 
     cleanup:
     server.stop()
@@ -154,11 +154,11 @@ class HttpServerTest extends AgentTestRunner {
       request."$method"()
     }
 
-    def response = client.newCall(request.build()).execute()
+    def clientResponse = client.newCall(request.build()).execute()
 
     then:
-    response.code() == code
-    response.body().string().trim() == "$resp"
+    clientResponse.code() == code
+    clientResponse.body().string().trim() == "$resp"
     server.lastRequest.contentType == contentType
     server.lastRequest.text.empty == !hasContent
 
@@ -189,11 +189,11 @@ class HttpServerTest extends AgentTestRunner {
       .header(headerKey, headerValue)
       .build()
 
-    def response = client.newCall(request).execute()
+    def clientResponse = client.newCall(request).execute()
 
     then:
-    response.code() == 200
-    response.body().string().trim() == ""
+    clientResponse.code() == 200
+    clientResponse.body().string().trim() == ""
 
     server.lastRequest.contentType == null
     server.lastRequest.headers.get(headerKey) == headerValue
@@ -245,11 +245,11 @@ class HttpServerTest extends AgentTestRunner {
       .url("$server.address/redirect")
       .get()
       .build()
-    def response = client.newCall(request).execute()
+    def clientResponse = client.newCall(request).execute()
 
     then:
-    response.code() == code
-    response.body().string() == body
+    clientResponse.code() == code
+    clientResponse.body().string() == body
 
     cleanup:
     server.stop()
@@ -276,11 +276,11 @@ class HttpServerTest extends AgentTestRunner {
       .head()
       .build()
 
-    def response = client.newCall(request).execute()
+    def clientResponse = client.newCall(request).execute()
 
     then:
-    response.code() == 200
-    response.body().string().trim() == ""
+    clientResponse.code() == 200
+    clientResponse.body().string().trim() == ""
 
     cleanup:
     server.stop()
@@ -303,11 +303,11 @@ class HttpServerTest extends AgentTestRunner {
       .get()
       .build()
 
-    def response = client.newCall(request).execute()
+    def clientResponse = client.newCall(request).execute()
 
     then:
-    response.code() == 200
-    response.body().string().trim() == "done"
+    clientResponse.code() == 200
+    clientResponse.body().string().trim() == "done"
 
     assertTraces(1) {
       server.distributedRequestTrace(it)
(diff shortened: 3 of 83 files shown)


dd-java-agent/testing/src/test/groovy/server/HttpServerTest.groovy
```diff
@@ -64,11 +64,11 @@ class HttpServerTest extends AgentTestRunner {
       .url("$server.address")
       .get()
       .build()
-    def response = client.newCall(request).execute()
+    def clientResponse = client.newCall(request).execute()
```
smola, line 67: Changes to this file seem superfluous? (I don't mind them, but I wonder if they address any issue)

dd-trace-core/src/test/groovy/datadog/trace/common/writer/DDAgentApiTest.groovy
```diff
@@ -96,9 +96,9 @@ class DDAgentApiTest extends DDCoreSpecification {
     def client = createAgentApi(agent.address.toString())[1]
     Payload payload = prepareTraces("v0.3/traces", [])
     expect:
-    def response = client.sendSerializedTraces(payload)
-    !response.success()
-    response.status() == 404
+    def clientResponse = client.sendSerializedTraces(payload)
```
smola, line 99: Same here: are we avoiding some naming clash here?

dd-java-agent/testing/src/test/groovy/server/HttpServerTest.groovy
nikita-tkachenko-datadog: Another variable named `response` is defined in mock servers' config closures.
Apparently the new Parrot parser introduced in Groovy 3 has a different notion of variable scopes, as there was a clash between the two when I updated the version.

dd-trace-core/src/test/groovy/datadog/trace/common/writer/DDAgentApiTest.groovy
nikita-tkachenko-datadog: Same thing here and in a few other places: all the variable renaming changes are done to avoid clashes after Groovy version update

dd-java-agent/agent-iast/src/test/groovy/com/datadog/iast/sink/PathTraversalModuleTest.groovy
```diff
@@ -79,7 +79,7 @@ class PathTraversalModuleTest extends IastModuleImplTestBase {
     '/==>var<==' | "/==>var<=="
   }
 
-  void 'iast module detects path traversal with String (#parent, #child)'(final String first, final String rest, final String expected) {
+  void 'iast module detects path traversal with String (#first, #rest, #expected)'() {
```
PerfectSlayer, line 82: Is `#expected` really part to the test description? Are inputs (`first` and `rest`) not enough?

gradle/dependencies.gradle
```diff
@@ -1,5 +1,5 @@
 final class CachedData {
-  static groovyVer = "2.5.21"
+  static groovyVer = "3.0.10"
```
PerfectSlayer, line 2: Why not using groovy 4?

gradle/dependencies.gradle
```diff
@@ -10,9 +10,8 @@ final class CachedData {
     okhttp_legacy : "[3.0,3.12.12]", // 3.12.x is last version to support Java7)
     okio          : "1.16.0",
 
-    spock         : "1.3-groovy-$spockGroovyVer",
+    spock         : "2.1-groovy-$spockGroovyVer",
```
PerfectSlayer, line 13: Same question here: why not using latest Spock version `2.3`?

gradle/dependencies.gradle
```diff
@@ -61,11 +60,11 @@ final class CachedData {
 
     spock                : [
       "org.spockframework:spock-core:${versions.spock}",
+      "org.spockframework:spock-junit4:${versions.spock}",
       // Used by Spock for mocking:
       "org.objenesis:objenesis:2.6" // Last version to support Java7
```
PerfectSlayer, line 65: We should be able to update it to `3.3` right?

gradle/dependencies.gradle
```diff
@@ -10,9 +10,8 @@ final class CachedData {
     okhttp_legacy : "[3.0,3.12.12]", // 3.12.x is last version to support Java7)
     okio          : "1.16.0",
 
-    spock         : "1.3-groovy-$spockGroovyVer",
+    spock         : "2.1-groovy-$spockGroovyVer",
     groovy        : groovyVer,
-    junit4        : "4.13.2",
     junit5        : "5.8.1",
```
PerfectSlayer, line 15: Can we bump JUnit to `5.9.2` too?

dd-java-agent/instrumentation/build.gradle
```diff
@@ -74,6 +74,9 @@ subprojects { Project subProj ->
     }
 
     subProj.tasks.withType(Test).configureEach { subTask ->
+      // instrumentation tests use SpockRunner and need to be run as JUnit 4 tests
```
PerfectSlayer, line 77: It might be worth to add a little paragraph to give details and references about all the troubles you went though.
Like the fact JUnit5 does not provides hooks /  extension points to port SpockRunner so you use `PlaformRunner`, a JUnit4 extension mechanism instead.
So tests are run using JUnit4, to boot our `SpockRunner` JUnit4 runner, to then boot a JUnitPlatfrom to run the Spock 2 JUnit 5 engine using a compatibility mode. See https://junit.org/junit5/docs/current/user-guide/#running-tests-junit-platform-runner for more details.

PerfectSlayer (commented):
Great work 👏 

I wonder if it will be an issue for the `junit-4` and `junit-5` instrumentation to not properly test the minimal version supported. I will try to get some feedback on it.
In the meantime, I left you some comments.

dd-java-agent/agent-iast/src/test/groovy/com/datadog/iast/sink/PathTraversalModuleTest.groovy
nikita-tkachenko-datadog: Agree, the inputs should be enough to identify the test case. Dropped `#expected` from the description

gradle/dependencies.gradle
nikita-tkachenko-datadog: One of the reasons for doing this change was to be able to use `gradleTestKit()`, and the one that comes with our Gradle version uses Groovy 3.0.10.
If I update it to Groovy 4 I am afraid the issue will be the same as before, only this time Spock's Groovy will be newer than the test kit's one, not older.

gradle/dependencies.gradle
nikita-tkachenko-datadog: Initial intention was to do the minimal upgrade required to use Groovy 3. I will try the newest version to see if it doesn't require further changes

gradle/dependencies.gradle
nikita-tkachenko-datadog: Interestingly, the latest spock version no longer has `objenesis` among its dependencies. I can probably try dropping it

gradle/dependencies.gradle
nikita-tkachenko-datadog: Spock 2.1 depended on Junit 5.8.1, but the latest version depends on 5.9. I will try updating both of them

dd-java-agent/instrumentation/build.gradle
nikita-tkachenko-datadog: Done

nikita-tkachenko-datadog:
> I wonder if it will be an issue for the `junit-4` and `junit-5` instrumentation to not properly test the minimal version supported.

Yeah, I am not too happy about this, but there isn't an alternative that I can think of.
Test framework instrumentations are only used by CI Visibility, so I suppose I would be the person responsible for supporting them.
I can live with that, since it doesn't seem to be too much of a burden to do basic manual checks to ensure minimum versions work.
Although I am open to discussion of course.

PerfectSlayer:
> Yeah, I am not too happy about this, but there isn't an alternative that I can think of.

We could start thinking of moving away from `SpockRunner` by creating an agent to load bootstrap classes into bootstrap classloader (like the `dd-java-agent.jar` but that only alter the classpath) and run tests with the agent on a forked VM.
It should work but it feels way out of the scope of this PR, especially if you (as JUnit integration maintainer) can live with this PR change 😅

gradle/dependencies.gradle
nikita-tkachenko-datadog: Maybe it is brought in transitively, as removing the dependency altogether breaks some tests.
Upgraded version to 3.3

gradle/dependencies.gradle
nikita-tkachenko-datadog: Bumped to 5.9.2

gradle/dependencies.gradle
nikita-tkachenko-datadog: For some strange reason, upgrading Groovy from `3.0.10` to `3.0.11` breaks tests in `:dd-java-agent:instrumentation:netty-4.1:test` module.
This blocks upgrading Spock.

I assume it's some transitive dependency change in Groovy that is to blame (my prime suspect is asm, which has been bumped from `9.2` to `9.3`), but I wasn't able to prove it.

I cannot dedicate much more time to this change at the moment, so my suggestion would be staying on Groovy `3.0.10` / Spock `2.1` combination for now, and then possibly upgrading them later as a separate change

jpbempel (approved):
LGTM for debugger

smola (approved):
Great job!

gradle/dependencies.gradle
PerfectSlayer: From the [Spock samples](https://github.com/spockframework/spock-example/blob/master/build.gradle), you can see `objenesis` is manually added so I guess it is required.

gradle/configure_tests.gradle
```diff
@@ -68,13 +66,11 @@ tasks.withType(Test).configureEach {
 project.afterEvaluate {
   tasks.withType(Test).configureEach {
     // JUnit 5 support
```
PerfectSlayer, line 68: This comment is now related to line 17 where JUnit5 is enabled.

gradle/dependencies.gradle
PerfectSlayer: Yeah, I tried to bump Groovy version to version 4 but I had to leave it to 3 for Gradle plugins for example.
Is there any specific module(s) that would requires version 3 for your tests or will it impact the whole project?
I wonder if it could be doable (and recommended) to update to 4 while leaving some module to 3 (like we did for JUnit4/5).

gradle/dependencies.gradle
PerfectSlayer: Sure. About my above comment, this question (global groovy 4 with some groovy 3 modules) can be investigate later in another PR in you have other priorities right now.

PerfectSlayer (approved):

gradle/dependencies.gradle
nikita-tkachenko-datadog: It will be just one module that does Gradle instrumentation. Technically it would be possible, of course, to have that specific module use Groovy 3, but it would complicate things too much in my opinion (as that module depends on a couple of common testing modules which themselves depend on Groovy).
Perhaps it would be simpler to migrate to Groovy 4 together with Gradle wrapper update, to avoid the need to have multiple versions.

gradle/configure_tests.gradle
nikita-tkachenko-datadog: I think the intention was to say that this statement enables/disables flaky tests in JUnit 5 (below is a similar comment about flaky tests for Spock).
I have re-worded the comment to clarify this, and added a different one on line 17.

PerfectSlayer:
Seems all good. Can you squash the commits when merging? 🙏

PR #4842, opened 2023-03-03
naming schema: rabbitmq

amarziali:
# What Does This Do

Changes for v1 naming schema:
* Producer:
  - service: `{{DD_SERVICE}}`
  - operation:  `amqp.send`
* Consumer:
  - service: `{{DD_SERVICE}}`
  - operation:  `amqp.process`
* Time in queue :
  - service: `rabbitmq-queue`
  - operation:  `amqp.deliver`

Interactions with other options:
* v1 schema implies legacy tracing disabled by default. However the user can explicitly force the legacy tracing and have service = 'rabbitmq' like v0

# Motivation

# Additional Notes

### DIFF
--- a/internal-api/src/main/java/datadog/trace/api/naming/NamingSchema.java
+++ b/internal-api/src/main/java/datadog/trace/api/naming/NamingSchema.java
@@ -25,6 +25,13 @@ public interface NamingSchema {
   ForDatabase database();
 
   /**
+   * Get the naming policy for messaging.
+   *
+   * @return a {@link NamingSchema.ForMessaging} instance.
+   */
+  ForMessaging messaging();
+
+  /**
    * Get the naming policy for servers.
    *
    * @return a {@link NamingSchema.ForServer} instance.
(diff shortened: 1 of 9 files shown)


internal-api/src/main/java/datadog/trace/api/naming/NamingSchema.java
```diff
@@ -93,6 +100,83 @@ interface ForDatabase {
     String service(@Nonnull String ddService, @Nonnull String databaseType);
   }
 
+  interface ForMessaging {
```
ygree, line 103: I wonder if we could flatten naming interface to avoid nesting that require additional dereferencing?  E.g. instead of having an inner interface InboundNaming just have methods like `inboundOperation` and `inboundService`. It could work if naming schema can either be v0 or v1 for all types of resources. Or do we have a use case when we need to be able to choose naming version per resource type?

ygree (commented):
Overall looks good to me. The only suggestion is to avoid  naming schemas nesting if possible.

internal-api/src/main/java/datadog/trace/api/naming/NamingSchema.java
amarziali: no it's v1 for everything or v0 for everything. I flattened as suggested thanks for proposing

ygree (approved):
LGTM

PR #4843, opened 2023-03-03
Add software.amazon and com.mongodb to exclusions

manuel-alvarez-alvarez:
# What Does This Do
Include two new packages that need to be excluded:

- com.mongodb
- software.amazon.awssdk

# Motivation

We are reporting vulnerabilities on those libraries that are not custom code, and can be considered false positives (weak hash on MD5 usage).

# Additional Notes

### DIFF
--- a/dd-java-agent/instrumentation/iast-instrumenter/src/main/resources/datadog/trace/instrumentation/iastinstrumenter/iast_exclusion.trie
+++ b/dd-java-agent/instrumentation/iast-instrumenter/src/main/resources/datadog/trace/instrumentation/iastinstrumenter/iast_exclusion.trie
@@ -69,6 +69,7 @@
 1 com.jprofiler.*
 1 com.lowagie.*
 1 com.mchange.*
+1 com.mongodb.*
 1 com.mysql.*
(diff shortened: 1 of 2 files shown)


dd-java-agent/instrumentation/iast-instrumenter/src/main/resources/datadog/trace/instrumentation/iastinstrumenter/iast_exclusion.trie
```diff
@@ -78,6 +78,7 @@
 1 com.rabbitmq.*
 1 com.squareup.*
 1 com.tangosol.*
+1 com.timgroup.statsd.*
```
smola, line 81: Please, keep this in the DataDog section, since it's actually from a package within the tracer.

manuel-alvarez-alvarez: Done, I thought we relocated those packages but no.

smola (approved):

PR #4847, opened 2023-03-04
Implement DBM-APM feature for JDBC instrumentation

jmeunier28:
# What Does This Do

**This is an experimental feature**
 
This is a first implementation using sql comment injection to propagate tracing information to the database query activity. The feature has three modes:

* `disabled` (default): Doesn't inject any sql comments on queries
* `service` (static tags): Injects static tags as sql comments. This mode can be used with databases/deployments where dynamic tags would incur a cost by essentially making every query unique and losing benefits of caching. 
    * Service name (the application/DD_SERVER and *not* the database service)
    * Environment
    * Service version (again, the application version and not the database version)
* `full`: Injects both static tags that are shared for the duration of an application (listed above) but also dynamic ones that are specific to a span make each query unique when comments are considered. This mode is the full featured one and can enable more product functionality. Note that for prepared statements, the tracer can't include dynamic tags and reverts back to what is essentially the static tags only mode.
    * Trace ID
    * Span ID
    * Sampling priority
    
 When this the `full` mode is enabled, customers would be able to look at their activity samples and link them directly back to their related traces in the UI. E.g:

Screen Shot 2023-03-06 at 9 31 24 AM

In order to enable the feature, you can do either of the following:

1. Set the system property `dd.dbm.propagation.mode=full,service`

2. Set the environment variable `DD_DBM_PROPAGATION_MODE=full,service`


dd-java-agent/instrumentation/jdbc/src/main/java/datadog/trace/instrumentation/jdbc/StatementInstrumentation.java
```diff
@@ -80,7 +89,22 @@ public static AgentScope onEnter(
         DECORATE.afterStart(span);
         DECORATE.onConnection(
             span, connection, InstrumentationContext.get(Connection.class, DBInfo.class));
-        DECORATE.onStatement(span, DBQueryInfo.ofStatement(sql));
+        if (span != null && DECORATE.injectSQLComment()) {
+          // set the dbm trace injected tag on the span
+          span.setTag(DBM_TRACE_INJECTED, true);
```
alexandre-normand, line 94: To match other tracers and [the spec](https://docs.google.com/document/d/1hlp3CuIRE-os2zYgc0wYOS8kPtCKqlrGL7aepbHHDaU/edit?pli=1#heading=h.v3ga4zojhwao), this tagging should move to the block below (when we're doing _full_ injection). That said, in hindsight, I wish we had just injected a `_dd.dbm_propagation_mode:<mode>` instead because it would have helped debugging in some cases.

jmeunier28: ah duh, thanks

jmeunier28: > That said, in hindsight, I wish we had just injected a _dd.dbm_propagation_mode:<mode> instead because it would have helped debugging in some cases.

Do you want me to set an additional tag on the span so we have the mode?

dd-java-agent/instrumentation/jdbc/src/jmh/java/datadog.trace.instrumentation.jdbc/SQLCommenterBenchmark.java
```diff
@@ -0,0 +1,53 @@
...
+
+  private static final String SQL_STMT = "SELECT * FROM table";
+
+  private final SQLCommenter commenter = new SQLCommenter();
+
+  @Benchmark
+  @BenchmarkMode(Mode.AverageTime)
+  @OutputTimeUnit(TimeUnit.NANOSECONDS)
+  public void encodeComment() {
+    commenter.augmentSQLStatement(SQL_STMT, TAGS);
```
jmeunier28, line 44:
```
Benchmark                                Mode  Cnt     Score    Error  Units
SQLCommenterBenchmark.encodeComment      avgt   25  5124.687 ± 43.768  ns/op
SQLCommenterBenchmark.encodeTraceParent  avgt   25  1123.568 ± 37.793  ns/op
```

I think the comment encoding can be improved. I still have some intermediate `String` creation that I should get rid of

dd-java-agent/instrumentation/jdbc/src/main/java/datadog/trace/instrumentation/jdbc/SQLCommenter.java
```diff
@@ -0,0 +1,288 @@
...
+  }
+
+  /**
+   * toComment takes a map of tags and creates a new sql comment using the sqlcommenter spec. This
+   * is used to inject APM context into sql statements for APM<->DBM linking
+   *
+   * @param tags
+   * @return String
+   */
+  public static String toComment(final SortedMap<String, Object> tags) {
```
jmeunier28, line 53:
```
Benchmark                                Mode  Cnt     Score    Error  Units
SQLCommenterBenchmark.encodeComment      avgt   25   974.606 ± 46.508  ns/op
SQLCommenterBenchmark.encodeTraceParent  avgt   25  1123.753 ± 56.404  ns/op
```

dd-java-agent/agent-bootstrap/src/main/java/datadog/trace/bootstrap/instrumentation/jdbc/DBQueryInfo.java
```diff
@@ -4,33 +4,46 @@
...
+  public static DBQueryInfo ofPreparedStatement(String sql, boolean stripSQLComment) {
     if (sql.length() > MAX_SQL_LENGTH_TO_CACHE) {
-      return NORMALIZE.apply(sql);
+      return NORMALIZE.apply(sql, stripSQLComment);
     } else {
-      return CACHED_PREPARED_STATEMENTS.computeIfAbsent(sql, NORMALIZE);
+      // if a prepared statement has a sql comment injected, we strip the comment
+      // so, it does not get added to the span. This also prevents us from obfuscating more
+      // than necessary.
+      return CACHED_PREPARED_STATEMENTS.computeIfAbsent(
```
dougqh, line 37: This introducing a capturing lambda which we want to avoid for performance reasons.  If "k" is the sql statement, then it would be better to use "k".  (Although, probably with a different name.)

dd-java-agent/agent-bootstrap/src/main/java/datadog/trace/bootstrap/instrumentation/jdbc/DBQueryInfo.java
```diff
@@ -4,33 +4,46 @@
...
     }
   }
 
   private final UTF8BytesString operation;
   private final UTF8BytesString sql;
 
-  public DBQueryInfo(String sql) {
-    this.sql = SQLNormalizer.normalize(sql);
+  public DBQueryInfo(String sql, boolean stripSQLComment) {
+    this.sql = SQLNormalizer.normalize(sql, stripSQLComment);
```
dougqh, line 46: Does this mean we start with the comment and then strip it out again?  My preference would be to only provide the original sql to DBQueryInfo if at all possible.

dd-java-agent/instrumentation/jdbc/src/main/java/datadog/trace/instrumentation/jdbc/AbstractConnectionInstrumentation.java
```diff
@@ -41,21 +46,38 @@ public void adviceTransformations(AdviceTransformation transformation) {
...
   }
 
   public static class ConnectionPrepareAdvice {
-    @Advice.OnMethodExit(suppress = Throwable.class)
+
+    @Advice.OnMethodEnter(suppress = Throwable.class)
+    public static void onEnter(
+        @Advice.This Connection connection,
+        @Advice.Argument(value = 0, readOnly = false) String sql) {
+      if (DECORATE.injectSQLComment()) {
```
dougqh, line 60: I'm on the fence on whether we need to optimize this.  I worry about having a lot of allocation on a potential hot path, but maybe preparation is infrequent enough that this is okay.

We can add a cache later if need be, but hopefully, we won't have to do so.  Often prepared statements are put into a cache at higher level.

dd-java-agent/instrumentation/jdbc/src/main/java/datadog/trace/instrumentation/jdbc/JDBCDecorator.java
```diff
@@ -98,6 +104,21 @@ protected String dbInstance(final DBInfo info) {
     }
   }
 
+  public String dbService(final DBInfo dbInfo) {
+    String dbService;
+    String dbInstance = dbInstance(dbInfo);
+    // by default, the db service is set to the dbType
+    dbService = dbInfo.getType();
```
dougqh, line 111: Is that the service that you want here?  That's really a peer.service and not the service of the Java process.
I suspect that you might really want the service of the local root span instead.

dd-java-agent/agent-bootstrap/src/main/java/datadog/trace/bootstrap/instrumentation/jdbc/DBQueryInfo.java
jmeunier28: Yeah that makes sense, the reason why I have this here is because for prepared statements I can't see any way to mutate the sql on connection (bc now that `PreparedStatement` object's sql text will contain the comments), but avoid adding the comments to the `DBQueryInfo` when that statement is executed, so I strip the comments on normalization.

We don't have to do that for  regular `Statement`s though, we can just keep a copy of the sql text in memory & pass that to the `DBQueryInfo` object we create

dd-java-agent/instrumentation/jdbc/src/main/java/datadog/trace/instrumentation/jdbc/JDBCDecorator.java
```diff
@@ -163,6 +169,27 @@ public AgentSpan onConnection(
     return super.onConnection(span, dbInfo);
   }
 
+  public DBInfo parseDBInfoFromConnection(final Connection connection) {
```
dougqh, line 172: Just breaking out a helper method it appears...  Fine by me.
Does it need to be public?  Should it be static?

dd-java-agent/instrumentation/jdbc/src/main/java/datadog/trace/instrumentation/jdbc/AbstractConnectionInstrumentation.java
```diff
@@ -41,21 +46,38 @@ public void adviceTransformations(AdviceTransformation transformation) {
...
     public static void addDBInfo(
         @Advice.This Connection connection,
-        @Advice.Argument(0) final String sql,
+        @Advice.Argument(value = 0) final String sql,
         @Advice.Return final PreparedStatement statement) {
       ContextStore<Statement, DBQueryInfo> contextStore =
           InstrumentationContext.get(Statement.class, DBQueryInfo.class);
       if (null == contextStore.get(statement)) {
-        DBQueryInfo info = DBQueryInfo.ofPreparedStatement(sql);
+        DBQueryInfo info = DBQueryInfo.ofPreparedStatement(sql, DECORATE.injectSQLComment());
```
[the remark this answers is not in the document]
jmeunier28: ah okay, yeah that might work

dd-java-agent/instrumentation/jdbc/src/main/java/datadog/trace/instrumentation/jdbc/SQLCommentInjectorAdaptor.java
```diff
@@ -0,0 +1,18 @@
+package datadog.trace.instrumentation.jdbc;
+
+import datadog.trace.bootstrap.instrumentation.api.AgentPropagation;
+
+public class SQLCommentInjectorAdaptor implements AgentPropagation.Setter<SQLCommenter.Builder> {
```
dougqh, line 5: At least conceptually, it makes sense to me that SQLCommentInjector is connected to propagation.  Although, we should double check with @bantonsson who was working on injection and propagation recently.

dd-java-agent/instrumentation/jdbc/src/main/java/datadog/trace/instrumentation/jdbc/SQLCommenter.java
```diff
@@ -0,0 +1,292 @@
...
+import java.nio.charset.StandardCharsets;
+import java.util.Map;
+import java.util.Objects;
+import java.util.SortedMap;
+import java.util.TreeMap;
+import java.util.function.Function;
+import org.slf4j.Logger;
+import org.slf4j.LoggerFactory;
+
+public class SQLCommenter {
```
dougqh, line 22: As mentioned elsewhere, I'm debating whether or not we should optimize this some more.

Typically, I'm not a fan of builders because they result in needless allocation that impacts response time latency.
I'd probably prefer this to just be a set of static utility functions.

dd-java-agent/instrumentation/jdbc/src/main/java/datadog/trace/instrumentation/jdbc/SQLCommenter.java
```diff
@@ -0,0 +1,292 @@
...
+  protected static final String DATABASE_SERVICE = "dddbs";
+  protected static final String DD_ENV = "dde";
+  protected static final String DD_VERSION = "ddpv";
+  protected static final String TRACEPARENT = "traceparent";
+  protected static final String W3C_CONTEXT_VERSION = "00";
+  private static final char EQUALS = '=';
+  private static final char COMMA = ',';
+  private static final char QUOTE = '\'';
+  public String commentedSQL;
+  private static final DDCache<SortedMap<String, Object>, String> CACHED_SQL_COMMENTS =
```
dougqh, line 36: Ah, I see there's a cache here -- that I was looking for above.

dougqh, line 36: Why SortedMap?  That seems a bit unnecessary.

dd-java-agent/instrumentation/jdbc/src/main/java/datadog/trace/instrumentation/jdbc/SQLCommenter.java
```diff
@@ -0,0 +1,292 @@
...
+    buffer.put("/*".getBytes(StandardCharsets.UTF_8));
+    buffer.put(commentStr.getBytes(StandardCharsets.UTF_8));
+    buffer.put("*/ ".getBytes(StandardCharsets.UTF_8));
+    buffer.put(sqlStmtBytes);
+
+    return new String(buffer.array(), 0, buffer.position(), StandardCharsets.UTF_8);
+  }
+
+  public static String toComment(final SortedMap<String, Object> tags) {
+    StringBuilder sb = new StringBuilder();
```
dougqh, line 90: It would be good to provide a better default length for this StringBuilder.  We should be able to make a reasonable approximation based on the number of tag entries.

dd-java-agent/instrumentation/jdbc/src/main/java/datadog/trace/instrumentation/jdbc/SQLCommenter.java
```diff
@@ -0,0 +1,292 @@
...
+    }
+
+    int capacity = commentStr.length() + sqlStmt.length() + 5;
+    ByteBuffer buffer = ByteBuffer.allocate(capacity);
+    buffer.put("/*".getBytes(StandardCharsets.UTF_8));
+    buffer.put(commentStr.getBytes(StandardCharsets.UTF_8));
+    buffer.put("*/ ".getBytes(StandardCharsets.UTF_8));
+    buffer.put(sqlStmtBytes);
+
+    return new String(buffer.array(), 0, buffer.position(), StandardCharsets.UTF_8);
```
dougqh, line 86: Why are we converting back and forth from UTF8 here?  Why not just stay in UTF16?

dd-java-agent/instrumentation/jdbc/src/main/java/datadog/trace/instrumentation/jdbc/SQLCommenter.java
```diff
@@ -0,0 +1,292 @@
...
+    }
+    if (obj instanceof Number) {
+      Number number = (Number) obj;
+      return number.doubleValue() == 0.0;
+    }
+    return false;
+  }
+
+  public SortedMap<String, Object> sortedKeyValuePairs(String dbInstance) {
+    SortedMap<String, Object> sortedMap = new TreeMap<>();
```
dougqh, line 178: Why SortedMap / TreeMap?
Or more pointedly, why are we constructing a map?
Couldn't we just build the comment eagerly a wrapper around a StringBuilder?

dougqh (changes requested):
Overall, I think looks pretty good.

I'm concerned about the need to strip comments.  That for me is a must fix.

I'm undecided about the various performance related concerns.
I somewhat willing to ignore them as long as this feature remains off by default.
However if this ever turned on by default, we'll have to revisit it.

dd-java-agent/instrumentation/jdbc/src/main/java/datadog/trace/instrumentation/jdbc/SQLCommenter.java
jmeunier28: oh that was an oversight by me :face-palm:. sorry, i changed this code a bunch of times

jmeunier28:
> I'm concerned about the need to strip comments. That for me is a must fix.

yes this makes sense to me, ill try to implement your suggestion for this. Stripping comments in normalization certainly isn't ideal.

> I'm undecided about the various performance related concerns.

definitely willing to spend more time on performance related concerns, the things you mentioned should be easy fixes.

dd-java-agent/instrumentation/jdbc/src/main/java/datadog/trace/instrumentation/jdbc/SQLCommenter.java
jmeunier28: yes for sure, i can do that

dd-java-agent/instrumentation/jdbc/src/main/java/datadog/trace/instrumentation/jdbc/JDBCDecorator.java
alexandre-normand: We include both although, admittedly, we're definitely getting more value from the parent service (the DD_SERVICE and service name of the Java process). The db service usages is likely going to fade away as more work to migrate towards `peer.service` is implemented in all parts of the stack/product but the reality right now is that the front-end still shows database services and we've got front-end integration that shows database hosts for those (and that relies on this db service to be propagated).

jmeunier28:
Hm I am not sure what could be causing the muzzle check to fail here:
```
Caused by: java.lang.RuntimeException: Instrumentation failed Muzzle validation
	at datadog.trace.agent.tooling.muzzle.MuzzleVersionScanPlugin.assertInstrumentationMuzzled(MuzzleVersionScanPlugin.java:68)
	... 30 more
```

jmeunier28:
hm I the last test that is failing is getting a SockeckTimeout, but it's a test unrelated to my change

```shell
java.net.SocketTimeoutException: timeout
	at okio.Okio$4.newTimeoutException(Okio.java:239)
	at okio.AsyncTimeout.exit(AsyncTimeout.java:286)
	....
	...
	at datadog.trace.agent.test.base.HttpServerTest.test success with #count requests(HttpServerTest.groovy:455)
Caused by: java.net.SocketException: Socket closed
	at java.net.SocketInputStream.read(SocketInputStream.java:204)
	at java.net.SocketInputStream.read(SocketInputStream.java:141)
	at okio.Okio$2.read(Okio.java:140)
	at okio.AsyncTimeout$2.read(AsyncTimeout.java:237)
	... 26 more
```

dd-trace-core/src/main/java/datadog/trace/core/propagation/SqlCommentInjector.java
```diff
@@ -0,0 +1,33 @@
...
+
+  private SqlCommentInjector() {}
+
+  private static class Injector implements HttpCodec.Injector {
+
+    public Injector() {}
+
+    @Override
+    public <C> void inject(DDSpanContext context, C carrier, AgentPropagation.Setter<C> setter) {
+      if (context.lockSamplingPriority()) {
```
bantonsson, line 22: This is a very roundabout way to get hold of the sampling priority. The `lockSamplingPriority` method shouldn't be used and the comment says `this is now effectively a no-op - there is no locking`.
You should just read the sampling priority where you need it, like this.
```
int samplingPriority = span.samplingPriority();
if (samplingPriority != PrioritySampling.UNSET) {
  // do whatever needs to be done
}
```

dd-java-agent/instrumentation/jdbc/src/main/java/datadog/trace/instrumentation/jdbc/SQLCommenter.java
```diff
@@ -0,0 +1,232 @@
...
+        toComment(
+            this.injectionMode, this.dbService, this.traceId, this.spanId, this.samplingPriority);
+    if (comment.isEmpty()) {
+      this.commentedSQL = this.sql;
+      return;
+    }
+
+    int capacity = comment.length() + this.sql.length() + 5;
+    ByteBuffer buffer = ByteBuffer.allocate(capacity);
+    buffer.put("/*".getBytes(StandardCharsets.UTF_8));
```
bantonsson, line 86: These constant short strings could be constant `Utf8ByteStrings` so that the same utf8 bytes can be reused. This will allocate new arrays of size 2 for every call here.

dd-java-agent/instrumentation/jdbc/src/main/java/datadog/trace/instrumentation/jdbc/SQLCommenter.java
```diff
@@ -0,0 +1,232 @@
...
+    if (sb.length() > 0) {
+      sb.deleteCharAt(sb.length() - 1); // remove trailing comma
+    }
+    return sb.toString();
+  }
+
+  private static void append(StringBuilder sb, String key, String value) {
+    if (null != value && !value.isEmpty()) {
+      try {
+        sb.append(URLEncoder.encode(key, UTF8));
```
bantonsson, line 169: Since the keys are controlled constants, then they could already be URL encoded once during initialization.

dd-java-agent/instrumentation/jdbc/src/main/java/datadog/trace/instrumentation/jdbc/SQLCommenter.java
```diff
@@ -0,0 +1,232 @@
...
+      return;
+    }
+    // If the SQL already has a comment, just return it.
+    final byte[] sqlStmtBytes = this.sql.getBytes(StandardCharsets.UTF_8);
+    if (hasDDComment(sqlStmtBytes)) {
+      this.commentedSQL = this.sql;
+      return;
+    }
+
+    final String comment =
```
bantonsson, line 76: Since the `comment` is never used as a string, but the bytes are consumed directly, then it would be more performant to build up the bytes directly instead.

bantonsson (commented):
Mostly performance comments. The only thing that I find confusing is the `SQL_COMMENT` propagation style and what the code is trying to accomplish with it. I think there should be some cleaner way implement it.

dd-java-agent/instrumentation/jdbc/src/main/java/datadog/trace/instrumentation/jdbc/SQLCommenter.java
```diff
@@ -0,0 +1,232 @@
...
+    if (priority != null && priority >= 1) {
+      traceSampledFlag = 1L;
+    }
+    if (null == traceId) {
+      return "";
+    }
+    return encodeTraceParent(Long.parseLong(traceId.toString()), spanId, traceSampledFlag);
+  }
+
+  protected static String encodeTraceParent(long traceID, long spanID, long sampled) {
```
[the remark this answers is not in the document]
jmeunier28: ah good to know 👍

dd-java-agent/instrumentation/jdbc/src/main/java/datadog/trace/instrumentation/jdbc/StatementInstrumentation.java
```diff
@@ -80,7 +89,30 @@ public static AgentScope onEnter(
...
+          } else if (SQL_COMMENT_INJECTION_MODE.equals(SQL_COMMENT_INJECTION_FULL)) {
+            SQLCommenter carrier =
+                new SQLCommenter(
+                    SQL_COMMENT_INJECTION_FULL,
+                    sql,
+                    span.getServiceName(),
+                    span.getTraceId(),
+                    span.getSpanId());
+            // forces a sampling decision & sets the priority on the carrier
+            propagate().inject(span, carrier, SETTER, TracePropagationStyle.SQL_COMMENT);
```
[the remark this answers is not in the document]
jmeunier28: I was just trying to force a sampling decision, because the sampling priority was always null before

dd-java-agent/instrumentation/jdbc/src/main/java/datadog/trace/instrumentation/jdbc/SQLCommenter.java
jmeunier28: ah yeah, good call 👍

dd-java-agent/instrumentation/jdbc/src/main/java/datadog/trace/instrumentation/jdbc/StatementInstrumentation.java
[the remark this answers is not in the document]
jmeunier28: ah I see i can just use the `TRACE_PARENT` propagator instead, that wasn't there when I originally started working on this so that's why I created a new injector so i could set the sampling priority properly

jmeunier28:
OK @bantonsson I think I covered everything here 10f25f95e177e6569b1dbbf66acba49520c70286

I did have one question about setting the service [right here ](https://github.com/DataDog/dd-trace-java/pull/4847/files#diff-2d9c6d46bfc5e23443a77d1e4f7a36426c78021d3ba1f5c4d0dc4805b4400cc9R81-R89), which I was not sure about & so left a `TODO`. 

I did confirm that the changes work, by checking that we are seeing samples with the services set properly in my test app here:
Screen Shot 2023-03-22 at 12 47 29 PM

Traces are being found & properly linked in our test application

Screen Shot 2023-03-22 at 12 46 59 PM

bantonsson (commented):
This is coming together nicely. There is still some cleanup that needs to be done in the `SQLCommenter`.

dd-java-agent/instrumentation/jdbc/src/main/java/datadog/trace/instrumentation/jdbc/SQLCommenter.java
```diff
@@ -0,0 +1,220 @@
...
+    this.dbService = dbService;
+  }
+
+  public void inject() {
+    if (this.sql == null || this.sql.isEmpty()) {
+      this.commentedSQL = this.sql;
+      return;
+    }
+    // If the SQL already has a comment, just return it.
+    final byte[] sqlStmtBytes = this.sql.getBytes(StandardCharsets.UTF_8);
```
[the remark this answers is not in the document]
jmeunier28: true, this code was shuffled around a few times. I'll re-implement it using a string input

dd-java-agent/instrumentation/jdbc/src/main/java/datadog/trace/instrumentation/jdbc/StatementInstrumentation.java
```diff
@@ -80,7 +84,22 @@ public static AgentScope onEnter(
         DECORATE.afterStart(span);
         DECORATE.onConnection(
             span, connection, InstrumentationContext.get(Connection.class, DBInfo.class));
-        DECORATE.onStatement(span, DBQueryInfo.ofStatement(sql));
+        final String copy = sql;
+        if (span != null && INJECT_COMMENT) {
+          SQLCommenter carrier =
```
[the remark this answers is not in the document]
jmeunier28: yes, will fix 👍

dd-java-agent/instrumentation/jdbc/src/main/java/datadog/trace/instrumentation/jdbc/SQLCommenter.java
```diff
@@ -0,0 +1,220 @@
...
+      append(sb, DATABASE_SERVICE, dbService);
+      append(sb, DD_ENV, env);
+      append(sb, DD_VERSION, version);
+      if (injectionMode.equals(SQL_COMMENT_INJECTION_FULL)) {
+        append(sb, TRACEPARENT, traceparent);
+      }
+    }
+    if (sb.length() > 0) {
+      // remove the trailing comment
+      sb.deleteCharAt(sb.length() - 1);
```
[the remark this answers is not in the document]
jmeunier28: yeah i was trying to make it less complicated, since i don't know which one will be the last non-empty value ahead of time

dd-java-agent/instrumentation/jdbc/src/main/java/datadog/trace/instrumentation/jdbc/SQLCommenter.java
```diff
@@ -0,0 +1,220 @@
...
+    }
+    final Config config = Config.get();
+    final String parentService = config.getServiceName();
+    final String env = config.getEnv();
+    final String version = config.getVersion();
+    final int commentSize = capacity(this.traceparent, parentService, this.dbService, env, version);
+    StringBuilder sb = new StringBuilder(this.sql.length() + commentSize);
+    toComment(
+        sb, this.injectionMode, parentService, this.dbService, env, version, this.traceparent);
+    if (sb.length() == 0) {
```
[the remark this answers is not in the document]
jmeunier28: > Or maybe the code should assume that it should inject a comment, and only have a boolean for the trace context part?

yes, this is what I did. Checking the mode again, seemed excessive  since we do this in the advice already

dd-java-agent/instrumentation/jdbc/src/main/java/datadog/trace/instrumentation/jdbc/SQLCommenter.java
```diff
@@ -0,0 +1,187 @@
...
+    if (!(sql.startsWith("/*"))) {
+      return false;
+    }
+    // else check to see if it's a DBM trace sql comment
+    int i = 2;
+    boolean found = false;
+    if (sql.length() > 2) {
+      // check if the next word starts with one of the specified keys
+      if (sql.charAt(i) == PARENT_SERVICE.charAt(0)
+          && hasMatchingSubstring(sql, i, PARENT_SERVICE)) {
```
bantonsson, line 68: This is just `sql.startsWith(PARENT_SERVICE, i)`

dd-java-agent/instrumentation/jdbc/src/main/java/datadog/trace/instrumentation/jdbc/SQLCommenter.java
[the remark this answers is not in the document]
bantonsson: You know which is the first one, and it doesn't need a comma in front of it, but all the other ones do.

dd-java-agent/instrumentation/jdbc/src/main/java/datadog/trace/instrumentation/jdbc/SQLCommenter.java
```diff
@@ -0,0 +1,187 @@
...
+    final String parentService = config.getServiceName();
+    final String env = config.getEnv();
+    final String version = config.getVersion();
+    final int commentSize = capacity(traceParent, parentService, dbService, env, version);
+    StringBuilder sb = new StringBuilder(sql.length() + commentSize);
+    toComment(sb, injectTrace, parentService, dbService, env, version, traceParent);
+    if (sb.length() == 0) {
+      return sql;
+    }
+    sb.insert(0, OPEN_COMMENT);
```
bantonsson, line 51: This is costly, and you can insert the `/*` before calling `toComment(...)` and check if length is `2` instead of `0`.

dd-java-agent/instrumentation/jdbc/src/main/java/datadog/trace/instrumentation/jdbc/SQLCommenter.java
```diff
@@ -116,34 +113,37 @@ protected static void toComment(
       final String env,
       final String version,
       final String traceparent) {
-    append(sb, PARENT_SERVICE, parentService);
-    append(sb, DATABASE_SERVICE, dbService);
-    append(sb, DD_ENV, env);
-    append(sb, DD_VERSION, version);
+    boolean first = false;
+    first = append(sb, PARENT_SERVICE, parentService, first);
+    first = append(sb, DATABASE_SERVICE, dbService, first);
```
bantonsson, line 118: Instead of doing this I would do:
```
int emptySize = sb.length();
append(sb, PARENT_SERVICE, parentService, false);
append(sb, DATABASE_SERVICE, dbService, sb.length() > emptySize);
...
```

dougqh (approved):

PR #4848, opened 2023-03-04
ZIO 2.0 instrumentation

dmytr:
# What Does This Do

Adds instrumentation for  [ZIO 2.0](https://zio.dev/).

# Motivation

ZIO is a popular Scala library for asynchronous and concurrent programming. Its runtime uses green threads (fibers) which breaks instrumentation because scope state is stored in `ThreadLocal` variable. This PR takes care of synchronizing scope state between fibers and JVM threads on which they are executed.

Implementation was tested on a [non-trivial distributed application](https://github.com/dmytr/zio-otel-instrumentation/tree/6de68991ac932473f1d7d8b3151e9cf84e494900).

### DIFF
--- a/dd-java-agent/instrumentation/zio/zio-2.0/src/main/java/datadog/trace/instrumentation/zio/v2_0/FiberContext.java
+++ b/dd-java-agent/instrumentation/zio/zio-2.0/src/main/java/datadog/trace/instrumentation/zio/v2_0/FiberContext.java
@@ -1,0 +1,70 @@
+package datadog.trace.instrumentation.zio.v2_0;
+
+import datadog.trace.bootstrap.instrumentation.api.AgentScope;
+import datadog.trace.bootstrap.instrumentation.api.AgentTracer;
+import datadog.trace.bootstrap.instrumentation.api.ScopeState;
+
+public class FiberContext {
+
+  private final ScopeState state;
+  private AgentScope.Continuation continuation;
+  private AgentScope scope;
+  private ScopeState oldState;
+
+  private FiberContext(ScopeState state, AgentScope.Continuation continuation) {
+    this.state = state;
+    this.continuation = continuation;
+    this.scope = null;
+    this.oldState = null;
+  }
+
+  public static FiberContext create() {
+    final ScopeState state = AgentTracer.get().newScopeState();
+    final AgentScope scope = AgentTracer.get().activeScope();
+    final AgentScope.Continuation continuation;
+
+    if (scope != null && scope.isAsyncPropagating()) {
+      continuation = scope.capture();
+    } else {
+      continuation = null;
+    }
+
+    return new FiberContext(state, continuation);
+  }
+
+  public void onEnd() {
+    if (this.scope != null) {
+      this.scope.close();
+      this.scope = null;
+    }
+
+    if (this.continuation != null) {
+      this.continuation.cancel();
+      this.continuation = null;
+    }
+
+    if (this.oldState != null) {
+      this.oldState.activate();
+      this.oldState = null;
+    }
+  }
+
+  public void onSuspend() {
+    if (this.oldState != null) {
+      this.oldState.activate();
+      this.oldState = null;
+    }
+  }
+
+  public void onResume() {
+    this.oldState = AgentTracer.get().newScopeState();
+    this.oldState.fetchFromActive();
+
+    this.state.activate();
+
+    if (this.continuation != null && this.scope == null) {
+      this.scope = this.continuation.activate();
+      this.continuation = null;
+    }
+  }
+}
--- a/dd-java-agent/instrumentation/zio/zio-2.0/src/main/java/datadog/trace/instrumentation/zio/v2_0/ZioRuntimeInstrumentation.java
+++ b/dd-java-agent/instrumentation/zio/zio-2.0/src/main/java/datadog/trace/instrumentation/zio/v2_0/ZioRuntimeInstrumentation.java
@@ -1,0 +1,59 @@
+package datadog.trace.instrumentation.zio.v2_0;
+
+import static datadog.trace.agent.tooling.bytebuddy.matcher.NameMatchers.named;
+import static java.util.Collections.singletonMap;
+import static net.bytebuddy.matcher.ElementMatchers.isMethod;
+
+import com.google.auto.service.AutoService;
+import datadog.trace.agent.tooling.Instrumenter;
+import datadog.trace.bootstrap.ContextStore;
+import datadog.trace.bootstrap.InstrumentationContext;
+import java.util.Map;
+import net.bytebuddy.asm.Advice;
+import zio.Fiber;
+import zio.Supervisor;
+
+@AutoService(Instrumenter.class)
+public class ZioRuntimeInstrumentation extends Instrumenter.Tracing
+    implements Instrumenter.ForSingleType {
+
+  public ZioRuntimeInstrumentation() {
+    super("zio.experimental");
+  }
+
+  @Override
+  public String instrumentedType() {
+    return "zio.Runtime$";
+  }
+
+  @Override
+  protected final boolean defaultEnabled() {
+    return false;
+  }
+
+  @Override
+  public void adviceTransformations(AdviceTransformation transformation) {
+    transformation.applyAdvice(
+        isMethod().and(named("defaultSupervisor")), getClass().getName() + "$DefaultSupervisor");
+  }
+
+  @Override
+  public String[] helperClassNames() {
+    return new String[] {packageName + ".FiberContext", packageName + ".TracingSupervisor"};
+  }
+
+  @Override
+  public Map<String, String> contextStore() {
+    return singletonMap("zio.Fiber$Runtime", packageName + ".FiberContext");
+  }
+
+  public static final class DefaultSupervisor {
+    @Advice.OnMethodExit(suppress = Throwable.class)
+    public static void onExit(@Advice.Return(readOnly = false) Supervisor<?> supervisor) {
+      @SuppressWarnings("rawtypes")
+      ContextStore<Fiber.Runtime, FiberContext> contextStore =
+          InstrumentationContext.get(Fiber.Runtime.class, FiberContext.class);
+      supervisor = supervisor.$plus$plus(new TracingSupervisor(contextStore));
+    }
+  }
+}
--- a/dd-java-agent/instrumentation/zio/zio-2.0/src/test/groovy/ZioRuntimeInstrumentationTest.groovy
+++ b/dd-java-agent/instrumentation/zio/zio-2.0/src/test/groovy/ZioRuntimeInstrumentationTest.groovy
@@ -1,0 +1,159 @@
+import datadog.trace.agent.test.AgentTestRunner
+
+class ZioRuntimeInstrumentationTest extends AgentTestRunner {
+
+  @Override
+  void configurePreAgent() {
+    super.configurePreAgent()
+    injectSysConfig("dd.integration.zio.experimental.enabled", "true")
+  }
+
+  def "trace is propagated to child fiber"() {
+    setup:
+    fixture().runNestedFibers()
+
+    expect:
+    assertTraces(1) {
+      trace(2, true) {
+        span(0) {
+          operationName "fiber_1_span_1"
+          parent()
+        }
+        span(1) {
+          operationName "fiber_2_span_1"
+          childOf(span(0))
+        }
+      }
+    }
+  }
+
+  def "trace is preserved when fiber is interrupted"() {
+    setup:
+    fixture().runInterruptedFiber()
+
+    expect:
+    assertTraces(1) {
+      trace(2, true) {
+        span(0) {
+          operationName "fiber_1_span_1"
+          parent()
+        }
+        span(1) {
+          operationName "fiber_2_span_1"
+          childOf(span(0))
+        }
+      }
+    }
+  }
+
+  def "synchronized fibers do not interfere with each other's traces"() {
+    setup:
+    fixture().runSynchronizedFibers()
+
+    expect:
+    assertTraces(2, SORT_TRACES_BY_NAMES) {
+      trace(2, true) {
+        span(0) {
+          operationName "fiber_1_span_1"
+          parent()
+        }
+        span(1) {
+          operationName "fiber_1_span_2"
+          childOf(span(0))
+        }
+      }
+      trace(2, true) {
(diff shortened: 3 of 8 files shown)


devinsba:
With the work that @PerfectSlayer is doing on #4795 I think they need to be the one to review this.

dmytr:
Hi @devinsba, it doesn't look like APIs used in this PR are affected by the referenced PR (https://github.com/DataDog/dd-trace-java/pull/4795). Maybe we could make this one move forward somehow?

bantonsson:
Hey @dmytr, just wanted to let you know that I've been looking at this today, but I need to dig a bit deeper into it. Will continue on Monday.

dmytr:
Thank you, @bantonsson! Looking forward to your feedback.

dd-java-agent/instrumentation/zio/zio-2.0/src/main/java/datadog/trace/instrumentation/zio/v2_0/ZioRuntimeInstrumentation.java
```diff
@@ -0,0 +1,54 @@
...
+import net.bytebuddy.asm.Advice;
+import zio.Fiber;
+import zio.Supervisor;
+
+@AutoService(Instrumenter.class)
+public class ZioRuntimeInstrumentation extends Instrumenter.Tracing
+    implements Instrumenter.ForSingleType {
+
+  public ZioRuntimeInstrumentation() {
+    super("zio");
```
bantonsson, line 21: Since this is a rather untested instrumentation I think it would be best to add an _experimental_  ending to the name, i.e. `zio_fiber.experimental`.

Also having it as an opt-in like this would make sense until we get some more feedback.
```
  @Override
  protected final boolean defaultEnabled() {
    return false;
  }
```

dd-java-agent/instrumentation/zio/zio-2.0/src/main/java/datadog/trace/instrumentation/zio/v2_0/FiberContext.java
```diff
@@ -0,0 +1,58 @@
...
+    this.oldState = null;
+  }
+
+  public static FiberContext create() {
+    final ScopeState state = AgentTracer.get().newScopeState();
+    final AgentScope scope = AgentTracer.get().activeScope();
+    final AgentScope.Continuation continuation;
+
+    if (scope != null && scope.isAsyncPropagating()) {
+      continuation = scope.captureConcurrent();
```
bantonsson, line 27: Yes, the naming is a bit strange, but `captureConcurrent` has a very specific use case, that I don't think applies here. From looking at the fiber code it seems that there are no races when calling into the supervisor, so using `scope.capture()` would make more sense.

dd-java-agent/instrumentation/zio/zio-2.0/src/main/java/datadog/trace/instrumentation/zio/v2_0/FiberContext.java
```diff
@@ -0,0 +1,58 @@
...
+    } else {
+      continuation = null;
+    }
+
+    return new FiberContext(state, continuation);
+  }
+
+  public void onEnd() {
+    state.activate();
```
bantonsson, line 36: As far as I can see, the call to `onEnd` can only come from inside the fiber itself, so the state should already have been activated in `onResume`.

dd-java-agent/instrumentation/zio/zio-2.0/src/main/java/datadog/trace/instrumentation/zio/v2_0/FiberContext.java
```diff
@@ -0,0 +1,58 @@
...
+
+    if (scope != null) scope.close();
+    if (continuation != null) continuation.cancel();
```
bantonsson, line 39: I would like the `scope` and `continuation` to be cleared after use.

dd-java-agent/instrumentation/zio/zio-2.0/src/main/java/datadog/trace/instrumentation/zio/v2_0/FiberContext.java
```diff
@@ -0,0 +1,58 @@
...
+
+    if (oldState != null) oldState.activate();
```
bantonsson, line 41: Same goes for the `oldState`. Clear it after use.

dd-java-agent/instrumentation/zio/zio-2.0/src/main/java/datadog/trace/instrumentation/zio/v2_0/FiberContext.java
```diff
@@ -0,0 +1,58 @@
...
+  }
+
+  public void onSuspend() {
+    if (oldState != null) oldState.activate();
```
bantonsson, line 45: Clear `oldState`

dd-java-agent/instrumentation/zio/zio-2.0/src/main/java/datadog/trace/instrumentation/zio/v2_0/FiberContext.java
```diff
@@ -0,0 +1,58 @@
...
+  }
+
+  public void onResume() {
+    oldState = AgentTracer.get().newScopeState();
+    oldState.fetchFromActive();
+
+    state.activate();
+
+    if (continuation != null && scope == null) {
+      scope = continuation.activate();
```
bantonsson, line 55: Clear `continuation`

bantonsson (commented):
Thanks once again @dmytr. Mostly some minor comments based on my limited understanding of the internals of ZIO. Please correct me if I've misunderstood anything.

dd-java-agent/instrumentation/zio/zio-2.0/src/main/java/datadog/trace/instrumentation/zio/v2_0/ZioRuntimeInstrumentation.java
bantonsson: Hey, sorry about the wording. Didn't see that you had tested it on a larger application 🤦🏼‍♂️. Still, I would like to have this flagged as experimental for now, since we don't have any more comprehensive tests in our test suite.

dd-java-agent/instrumentation/zio/zio-2.0/src/main/java/datadog/trace/instrumentation/zio/v2_0/FiberContext.java
dmytr: Done.

dd-java-agent/instrumentation/zio/zio-2.0/src/test/scala/ZioTestFixtures.scala
```diff
@@ -0,0 +1,157 @@
...
+
+  def runConcurrentFibers(): Unit = {
+    def runFiber(
+        fiberNumber: Int,
+        start: Promise[Nothing, Unit]
+    ): ZIO[Any, Nothing, Unit] = {
+      start.await *>
+        childSpan(s"fiber_${fiberNumber}_span_1") {
+          ZIO.yieldNow *>
+            childSpan(s"fiber_${fiberNumber}_span_2") {
```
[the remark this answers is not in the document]
dmytr: `ZIO.yieldNow` tells runtime to suspend current fiber. It's like sleep, just without actually sleeping. So we definitely have calls to `onSuspend`/`onResume` right before span 2 is started and also inside of span 2.

Do you think we should add one more span to this setup?

dd-java-agent/instrumentation/zio/zio-2.0/src/main/java/datadog/trace/instrumentation/zio/v2_0/ZioRuntimeInstrumentation.java
dmytr: No worries, it is indeed experimental at this point. I've disabled this instrumentation by default.

I'm not sure about `zio_fiber` as instrumentation name though. It doesn't seem like there is something else to instrument in ZIO, so simply `zio` seems to be better. 

But I don't have a strong opinion on this one, so please just tell me what you think would be better.

dd-java-agent/instrumentation/zio/zio-2.0/src/main/java/datadog/trace/instrumentation/zio/v2_0/FiberContext.java
dmytr: I've tested how ZIO executes callbacks on supervisor and it seems that it calls `onEnd` before `onSuspend`. So have to activate state in `onEnd`.

dd-java-agent/instrumentation/zio/zio-2.0/src/main/java/datadog/trace/instrumentation/zio/v2_0/FiberContext.java
dmytr: Done.

dd-java-agent/instrumentation/zio/zio-2.0/src/main/java/datadog/trace/instrumentation/zio/v2_0/FiberContext.java
dmytr: Done.

dd-java-agent/instrumentation/zio/zio-2.0/src/main/java/datadog/trace/instrumentation/zio/v2_0/FiberContext.java
dmytr: Done.

dd-java-agent/instrumentation/zio/zio-2.0/src/main/java/datadog/trace/instrumentation/zio/v2_0/FiberContext.java
dmytr: You are right, all calls to `FixberContext` will be done synchronously, as callbacks on supervisor are called from the same thread for the same fiber.

Changed to use `scope.capture()`.

dd-java-agent/instrumentation/zio/zio-2.0/src/main/java/datadog/trace/instrumentation/zio/v2_0/FiberContext.java
dmytr: Sorry, I mixed up `onSuspend` and `onResume` :facepalm: You are right! Changing...

dmytr: Done.

I recall now that I've added call to `state.activate()` because I wasn't sure if `onResume` would be called when fiber is interrupted. But I've tested this and `onResume` is always called before `onEnd`.

dd-java-agent/instrumentation/zio/zio-2.0/src/main/java/datadog/trace/instrumentation/zio/v2_0/ZioRuntimeInstrumentation.java
bantonsson: I guess you're right, `zio.experimental` is fine.

dd-java-agent/instrumentation/zio/zio-2.0/src/main/java/datadog/trace/instrumentation/zio/v2_0/FiberContext.java
bantonsson: I don't see that you are clearing out the `continuation`

dmytr: Sorry, I'm not familiar with continuations API. Do you mean that we need to clear `continuation` after we activate it in `onResume`?

I thought that we need to clear it only in `onEnd`, after we've cancelled it: https://github.com/DataDog/dd-trace-java/pull/4848/files#diff-8892be78dee1cb0bfa34bf6f4a7650bde447fc192d936eef001bc01eaac60227R41-R44.

But if we clear `continuation` right after it was activated in `onResume`, then we don't need to cancel it in `onEnd`, right?

dd-java-agent/instrumentation/zio/zio-2.0/src/test/groovy/ZioRuntimeInstrumentationTest.groovy
```diff
@@ -0,0 +1,159 @@
+import datadog.trace.agent.test.AgentTestRunner
+
+class ZioRuntimeInstrumentationTest extends AgentTestRunner {
+
+  @Override
+  void configurePreAgent() {
+    super.configurePreAgent()
+    injectSysConfig("dd.integration.zio.enabled", "true")
```
bantonsson, line 8: Missing an `experimental` in the integration name.

dd-java-agent/instrumentation/zio/zio-2.0/src/main/java/datadog/trace/instrumentation/zio/v2_0/FiberContext.java
dmytr: Thank you for the explanation. We are clearing `continuation` after the first use now.

Based on my testing and from what I could understand from ZIO's code, it seems that `onResume` will be always called, no mapper if fiber terminates normally or if it was interrupted.

dd-java-agent/instrumentation/zio/zio-2.0/src/test/groovy/ZioRuntimeInstrumentationTest.groovy
dmytr: Thanks! I was wondering why there were more failing checks than usually :sweat_smile:

bantonsson (approved):

bantonsson:
Hey @dmytr would you mind squashing your commits before we merge this PR?

bantonsson:
Thanks for the contribution @dmytr. I'll merge this as soon as I've nursed it through some CI flakiness.

jivanic-demystdata:
Does anyone know if it's still working? It doesn't add any traces in my Scala 3.4.2 / ZIO 2.0.22 project 🤔

PR #4853, opened 2023-03-06
naming schema: graphql

amarziali:
# What Does This Do

v1 naming schema changes for graphql (request only):
* operation: `graphql.server.request`

Note: the resource name will also change on v1 to `graphql.server.request` since the decorator does not set any and thus defaulting to operation name.

# Motivation

# Additional Notes


ygree (commented):
LGTM

ygree (approved):

PR #4870, opened 2023-03-07
New mapping rule for adding trailing slash to the http resource path

ygree:
# What Does This Do

Adds a path mapping rule to force adding trailing slash by adding a special mapping value `*/`.
With this change it's possible to define TRACE_HTTP_SERVER_PATH_RESOURCE_NAME_MAPPING and TRACE_HTTP_CLIENT_PATH_RESOURCE_NAME_MAPPING rules that will force adding a trailing slash if missing.
E.g. next rules will force adding trailing slash for all resource that match `/test/**`
```json
[
      "/test/**": "*/",
]
```
so, `/test/foo` will be changed to `/test/foo/`.

# Motivation

Datadog treats otherwise identical http path resources with and without a trailing slash as two different resource. This addition will help to instruct Java Tracer to group them into one resource.

# Additional Notes

FRAPMS-1972

### DIFF
--- a/dd-java-agent/agent-bootstrap/src/main/java/datadog/trace/bootstrap/instrumentation/decorator/http/AntPatternPathNormalizer.java
+++ b/dd-java-agent/agent-bootstrap/src/main/java/datadog/trace/bootstrap/instrumentation/decorator/http/AntPatternPathNormalizer.java
@@ -16,6 +16,8 @@ final class AntPatternPathNormalizer extends PathNormalizer {
   /** Used to preserve original value as is when it's mapped to this value. */
   private static final String KEEP_AS_IS = "*";
 
+  private static final String ADD_TRAILING_SLASH = "*/";
+
   private final Map<String, String> resourceNameMatchers;
   private final AntPathMatcher matcher = new AntPathMatcher();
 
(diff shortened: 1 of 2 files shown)


nayeem-kamal (approved):

bantonsson (commented):
Would it make sense to have a general setting that doesn't involve the `AntPatternPathNormalizer` where a trailing `/` would always be added?

ygree:
> Would it make sense to have a general setting that doesn't involve the AntPatternPathNormalizer where a trailing / would always be added?

Yeah, I was thinking between that approach as well. My reasoning was that with mapping rules it will be possible to adjust this more granularly, for only specific path patterns. But maybe we don't need it at all, so I'm happy to change it to a single setting if that seems to be more reasonable. What do you think the most beneficial part of having this logic enabled with a single global config option vs mapping rules?

bantonsson:
I'm not sure if we need that level of configurability, but on the other hand, you don't necessarily want to add a `/` on all resources. Maybe it should be the other way around, to remove trailing slash instead?

ygree:
Closing this PR in flavor of #4889

PR #4872, opened 2023-03-08
Telemetry API v1 updated to the latest version

ValentinZakharov:
# What Does This Do
Implemented latest changes according to specification for Telemetry API v1

# Motivation
Requires for metrics over telemetry

# Additional Notes

### DIFF
--- a/telemetry/src/main/java/datadog/telemetry/TelemetryService.java
+++ b/telemetry/src/main/java/datadog/telemetry/TelemetryService.java
@@ -1,8 +1,7 @@
 package datadog.telemetry;
 
-import datadog.telemetry.api.Dependency;
-import datadog.telemetry.api.Integration;
-import datadog.telemetry.api.Metric;
+import datadog.telemetry.api.*;
+
 import java.util.Map;
 import java.util.Queue;
 import okhttp3.Request;
@@ -22,6 +21,10 @@ public interface TelemetryService {
   boolean addIntegration(Integration integration);
 
   boolean addMetric(Metric metric);
+
+  boolean addLogMessage(LogMessage message);
+
+  boolean addDistributionSeries(DistributionSeries series);
 
   Queue<Request> prepareRequests();
 
--- a/telemetry/src/main/java/datadog/telemetry/TelemetryService.java
+++ b/telemetry/src/main/java/datadog/telemetry/TelemetryService.java
@@ -1,7 +1,10 @@
 package datadog.telemetry;
 
-import datadog.telemetry.api.*;
-
+import datadog.telemetry.api.Dependency;
+import datadog.telemetry.api.DistributionSeries;
+import datadog.telemetry.api.Integration;
+import datadog.telemetry.api.LogMessage;
+import datadog.telemetry.api.Metric;
 import java.util.Map;
 import java.util.Queue;
 import okhttp3.Request;
--- a/telemetry/src/main/java/datadog/telemetry/TelemetryServiceImpl.java
+++ b/telemetry/src/main/java/datadog/telemetry/TelemetryServiceImpl.java
@@ -1,6 +1,19 @@
 package datadog.telemetry;
 
-import datadog.telemetry.api.*;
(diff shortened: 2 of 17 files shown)


telemetry/src/main/java/datadog/telemetry/TelemetryService.java
```diff
@@ -1,8 +1,7 @@
 package datadog.telemetry;
 
-import datadog.telemetry.api.Dependency;
-import datadog.telemetry.api.Integration;
-import datadog.telemetry.api.Metric;
+import datadog.telemetry.api.*;
```
smola, line 3: See https://github.com/DataDog/dd-trace-java/blob/master/CONTRIBUTING.md#intellij-idea to avoid this.

telemetry/src/main/java/datadog/telemetry/TelemetryServiceImpl.java
```diff
@@ -29,6 +20,8 @@ public class TelemetryServiceImpl implements TelemetryService {
 
   private static final Logger log = LoggerFactory.getLogger(TelemetryServiceImpl.class);
 
+  private static final int MAX_ELEMENTS_PER_REQUEST = 100;
```
smola, line 23: Is this exact limit from the spec, or an arbitrary one on our side?

smola (changes requested):
Looks good overall, I miss test cases, and an overall description in the PR of what we updated. This is only an update for Telemetry API v1, and not really v2, correct?

telemetry/src/main/java/datadog/telemetry/TelemetryServiceImpl.java
```diff
@@ -151,6 +158,39 @@ public Queue<Request> prepareRequests() {
...
+                            new Logs().messages(messages)
+                        );
+        queue.offer(request);
+      }
+    }
+
+    // New Distributions
+    if (!distributionSeries.isEmpty()) {
+      drainOrEmpty(distributionSeries).stream()
+              .collect(Collectors.groupingBy(DistributionSeries::getNamespace))
```
manuel-alvarez-alvarez, line 180: I think we can group the series and provide a single `Distributions` with mixed namespaces in `DistributionSeries` I'm not sure if this was possible in V1 (the same applies to the metrics).

telemetry/src/main/java/datadog/telemetry/api/Metric.java
```diff
@@ -14,8 +14,8 @@
 
 public class Metric {
 
-  /** Non-serializable field with the namespace of the current metric */
-  private transient String namespace = "appsec";
+  @com.squareup.moshi.Json(name = "namespace")
+  private String namespace;
```
manuel-alvarez-alvarez, line 18: Nice!, I'm not 100% sure but I think this field was not present in the spec when I did the change :disappointed_relieved:

telemetry/src/main/java/datadog/telemetry/TelemetryService.java
ValentinZakharov: Thanks, fixed

telemetry/src/main/java/datadog/telemetry/TelemetryServiceImpl.java
ValentinZakharov: There is no number of entries limit in spec, the value is from our side.

telemetry/src/main/java/datadog/telemetry/TelemetryServiceImpl.java
ValentinZakharov: Thanks, done

telemetry/src/main/java/datadog/telemetry/api/Metric.java
ValentinZakharov: It wasn't before. The spec is still evolving, we might see another small improvements in future.

smola (approved):

PR #4878, opened 2023-03-09
Update client ip resolver

cataphract:
Update behavior to match the changes in the spec. The differences are:

* Forwarded and Via are not to be used anymore.
* Client-ip is changed to X-Client-Ip.
* Added Fast-client-ip, Cf-connecting-ip and Cf-connecting-ipv6
* We report only the first public ip found in the headers.
* We don't report that there are multiple tags presents anymore.


smola (approved):
Looks good.

PR #4882, opened 2023-03-09
Use smoke build cache for check job

smola:
# What Does This Do

# Motivation
This should have a better cache hit rate.
# Additional Notes


smola:
Not good enough.

PR #4889, opened 2023-03-10
A setting to remove http resource name trailing slash

ygree:
# What Does This Do

Adds a setting to remove a trailing slash from the http resource name.

# Motivation

Datadog treats otherwise identical http path resources with and without a trailing slash as two different resource. This addition will help to instruct Java Tracer to group them into one resource by removing a trailing slash.

# Additional Notes

FRAPMS-1972

This change supersedes #4870

### DIFF
--- a/internal-api/src/main/java/datadog/trace/api/normalize/HttpResourceNames.java
+++ b/internal-api/src/main/java/datadog/trace/api/normalize/HttpResourceNames.java
@@ -34,6 +34,7 @@ public class HttpResourceNames {
 
   private final AntPatternHttpPathNormalizer serverAntPatternHttpPathNormalizer;
   private final AntPatternHttpPathNormalizer clientAntPatternHttpPathNormalizer;
+  private final boolean removeTrailingSlash;
 
   private static HttpResourceNames instance() {
     if (null == INSTANCE) {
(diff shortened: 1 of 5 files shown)


internal-api/src/main/java/datadog/trace/api/normalize/HttpResourceNames.java
```diff
@@ -91,6 +93,12 @@ public static AgentSpan setForClient(
   }
 
   public static CharSequence join(CharSequence method, CharSequence path) {
+    if (instance().removeTrailingSlash
+        && path != null
+        && path.length() > 1
+        && path.charAt(path.length() - 1) == '/') {
+      path = path.subSequence(0, path.length() - 1);
```
bantonsson, line 100: I think that it would make sense to move the `JOINER` into the `instance` and this modification to the actual creation of the cached value. Yes, that mean that `path`, and `path/` would take up two different slots in the cache, but you would only do the string copying if there wasn't a cache hit.

ygree: Good idea! I moved removing a trailing slash into the cached value calculation as suggested. It also opened an opportunity to reduce number of String allocation.

internal-api/src/main/java/datadog/trace/api/normalize/HttpResourceNames.java
```diff
@@ -14,13 +14,27 @@ public class HttpResourceNames {
...
+          for (int i = 0; i < sb.length(); i++) {
+            sb.setCharAt(i, Character.toUpperCase(sb.charAt(i)));
+          }
+          sb.append(' ');
+        }
+        sb.append(path);
+        // remove trailing slash from the resource name path if needed
+        if (instance().removeTrailingSlash) {
+          int l = path.length() - 1;
+          if (l > 0 && path.charAt(l) == '/') sb.deleteCharAt(sb.length() - 1);
```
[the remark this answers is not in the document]
ygree: Yep. Changed it. Also preallocated the buffer.

bantonsson (approved):

PR #4894, opened 2023-03-13
CI Visibility Gradle instrumentation

nikita-tkachenko-datadog:
# What Does This Do
Adds a new instrumentation for Gradle Daemon processes.
The instrumentation creates spans for entire build and for each Test task within a build.
The instrumentation passes IDs of created spans to forked test processes (so that the spans created in those processes could refer to the spans from the parent daemon process).

# Motivation
This is needed in order to support Test Session and Test Module levels visibility (a CI Visibility feature) for Gradle builds.
"Test session" is a CI Visibility term that is equivalent to tests run for a single project, "Test module" is equivalent to tests run for a single module within a project.

# Additional Notes
Gradle Daemon is supposed to be a long-lived process potentially reused for multiple builds.
Since multiple builds for different projects can be executed by the same daemon, repository root cannot be calculated during the agent's init phase and stored in a static field anymore.
Instead it is now a part of `TestDecorator` state (the decorator lifetime can now vary depending on whether it is used with a Gradle build or a forked test process run).
Some CI Visibility classes had to be moved from `dd-java-agent/agent-ci-visibility` module to `internal-api` module in order to accommodate this change.


dd-java-agent/instrumentation/gradle/src/main/java/datadog/trace/instrumentation/gradle/GradleBuildListener.java
```diff
@@ -0,0 +1,139 @@
...
+
+  @Override
+  public void projectsEvaluated(Gradle gradle) {
+    if (!Config.get().isCiVisibilityEnabled()) {
+      return;
+    }
+    Project rootProject = gradle.getRootProject();
+    Collection<GradleUtils.TestFramework> testFrameworks =
+        GradleUtils.collectTestFrameworks(rootProject);
+    if (testFrameworks.size() == 1) {
```
devinsba, line 53: Might be good to report back when customers do have multiples, either with a debug log so we can get it in troubleshooting, or a telemetry log message

nikita-tkachenko-datadog: Agree, added a log message

devinsba (approved):

PR #4896, opened 2023-03-14
Properly support async data fetchers returning CompletableFuture

theJC:
# What Does This Do
Properly instruments async data fetchers by adding an additional stage of a completeableFuture that finishes the span at the time the future completes

Fixes: https://github.com/DataDog/dd-trace-java/issues/4051

# Motivation
To allow async fetcher's timeframes be properly represented in APM traces.

# Additional Notes


theJC:
@ygree  -- Can I please ask y'all to take a look at this one and see if there is anything else I've missed that is absolutely required to be added before merging this change?  I've tried to scope it as small and limited to just the minimal capability we need improved for our needs.

I believe after this, there is only one last issue before Indeed can transition off its own instrumentation libraries, and just use Datadog's instrumentation alone.

dd-java-agent/instrumentation/graphql-java-14.0/src/main/java/datadog/trace/instrumentation/graphqljava/InstrumentedDataFetcher.java
```diff
@@ -42,15 +43,27 @@ public Object get(DataFetchingEnvironment environment) throws Exception {
...
       } catch (Exception e) {
-        fieldSpan.addThrowable(e);
-        throw e;
-      } finally {
+        DECORATE.onError(fieldSpan, e);
         DECORATE.beforeFinish(fieldSpan);
         fieldSpan.finish();
+        throw e;
+      }
+      if (dataValue instanceof CompletableFuture<?>) {
```
ygree, line 55: I suggest using CompletionStage instead of CompletableFuture to make this instrumentation work with any CompletionStage implementations other than CompletableFuture.

dd-java-agent/instrumentation/graphql-java-14.0/src/main/java/datadog/trace/instrumentation/graphqljava/InstrumentedDataFetcher.java
```diff
@@ -42,15 +43,27 @@ public Object get(DataFetchingEnvironment environment) throws Exception {
...
+        return ((CompletableFuture<?>) dataValue)
+            .whenComplete(
```
ygree, line 57: Another minor suggestion is to check if it's a completed future (`.isDone()`) and finish the span immediately in that case.

ygree:
> @ygree -- Can I please ask y'all to take a look at this one and see if there is anything else I've missed that is absolutely required to be added before merging this change? I've tried to scope it as small and limited to just the minimal capability we need improved for our needs.

Thank you for your contribution! I suggest using CompletionStage instead of CompletableFuture to support other types that implement the CompletionStage interface. Other than that your solution looks good!

dd-java-agent/instrumentation/graphql-java-14.0/src/main/java/datadog/trace/instrumentation/graphqljava/InstrumentedDataFetcher.java
theJC: Done. Im unsure if there is better sugar syntax to handle this case, but it will now immediately finish out the span instead of adding another stage.  Thanks for the suggestion.

dd-java-agent/instrumentation/graphql-java-14.0/src/main/java/datadog/trace/instrumentation/graphqljava/InstrumentedDataFetcher.java
theJC: Done.  Thanks for the recommendation.

theJC:
I've made updates to address all of your feedback and I think we should be all set unless there is feedback on implementation of the updates.  Thanks Yury!

ygree (approved):
Looks good! Thanks again for your contribution!

PR #4897, opened 2023-03-14
Add Play 2.8 OpenTelemetry smoke test

bantonsson:
# What Does This Do

Adds a Play 2.8 with OpenTelemetry API smoke test.

# Motivation

Need at least one smoke test that tests OTel and Datadog interaction.

# Additional Notes

* Also moves the OTel 1.4 instrumentation to the `datadog.trace.instrumentation.*` package.
* The reason for selecting a Play test was that it was the quickest way to switch existing OpenTracing calls to OpenTelemetry calls.


PerfectSlayer (approved):
LGTM. Any reason `Span.makeCurrent()` is called sometimes within `try-with-resources`, sometimes not?

bantonsson:
The only reason for the `try-with-resource` or not is that was the how the old code looked.

PR #4899, opened 2023-03-14
Added health metrics around trace construction

nayeem-kamal:
# What Does This Do
Adds health metrics around trace and span construction:
`queue.dropped.spans` - number of dropped spans tagged by sampling priority
`span.pending.finished` - number of spans that are finished
`span.sampling.sampled` - number of spans sampled by the single span sampler
`span.sampling.unsampled` - number of spans that are considered unsampled by the single span sampler
`scope.activate.count` - number of activated ContinuableScopes
`scope.close.count` - number of closed ContinuableScopes
`scope.error.stack-overflow` - counts the number of times that the Scope depth limit is exceeded
# Motivation
The goal of implementing these metrics is to increase visibility into potential issues in trace/span creation or delivery.

# Additional Notes

### DIFF
--- a/dd-trace-core/src/main/java/datadog/trace/common/writer/PayloadDispatcher.java
+++ b/dd-trace-core/src/main/java/datadog/trace/common/writer/PayloadDispatcher.java
@@ -62,7 +62,8 @@ public class PayloadDispatcher implements ByteBufferConsumer {
     // introducing an unbound queue and another thread to do the IO
     // however, we can't block the application threads from here.
     if (null == mapper || !packer.format(trace, mapper)) {
-      healthMetrics.onFailedPublish(trace.get(0).samplingPriority(), trace.size());
+      healthMetrics.onFailedPublish(
+          trace.isEmpty() ? 0 : trace.get(0).samplingPriority(), trace.size());
     }
   }
 
--- a/dd-trace-core/src/main/java/datadog/trace/common/writer/PayloadDispatcher.java
+++ b/dd-trace-core/src/main/java/datadog/trace/common/writer/PayloadDispatcher.java
@@ -62,7 +62,7 @@ public class PayloadDispatcher implements ByteBufferConsumer {
     // introducing an unbound queue and another thread to do the IO
     // however, we can't block the application threads from here.
     if (null == mapper || !packer.format(trace, mapper)) {
-      healthMetrics.onFailedPublish(trace.get(0).samplingPriority());
+      healthMetrics.onFailedPublish(trace.get(0).samplingPriority(), trace.size());
     }
   }
 
--- a/dd-trace-core/src/main/java/datadog/trace/common/writer/RemoteWriter.java
+++ b/dd-trace-core/src/main/java/datadog/trace/common/writer/RemoteWriter.java
@@ -138,7 +138,7 @@ public abstract class RemoteWriter implements Writer {
   @Override
   public void close() {
     final boolean flushed = flush();
-    closed = true;
+    closed = true;R
     traceProcessingWorker.close();
     healthMetrics.onShutdown(flushed);
     healthMetrics.close();
--- a/dd-trace-core/src/main/java/datadog/trace/common/writer/RemoteWriter.java
+++ b/dd-trace-core/src/main/java/datadog/trace/common/writer/RemoteWriter.java
@@ -140,8 +140,8 @@ public abstract class RemoteWriter implements Writer {
     final boolean flushed = flush();
     closed = true;
     traceProcessingWorker.close();
-    healthMetrics.close();
     healthMetrics.onShutdown(flushed);
+    healthMetrics.close();
   }
 
   @Override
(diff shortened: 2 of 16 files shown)


dd-trace-core/src/main/java/datadog/trace/common/writer/PayloadDispatcher.java
```diff
@@ -62,7 +62,8 @@ void addTrace(List<? extends CoreSpan<?>> trace) {
     // introducing an unbound queue and another thread to do the IO
     // however, we can't block the application threads from here.
     if (null == mapper || !packer.format(trace, mapper)) {
-      healthMetrics.onFailedPublish(trace.get(0).samplingPriority());
+      healthMetrics.onFailedPublish(
+          trace.isEmpty() ? 0 : trace.get(0).samplingPriority(), trace.size());
```
ygree, line 66: Is this emptiness check really needed?

dd-trace-core/src/main/java/datadog/trace/common/writer/RemoteWriter.java
```diff
@@ -106,7 +106,8 @@ public void write(final List<DDSpan> trace) {
   private void handleDroppedTrace(
       final String reason, final List<DDSpan> trace, final int samplingPriority) {
     log.debug("{}. Counted but dropping trace: {}", reason, trace);
-    healthMetrics.onFailedPublish(samplingPriority);
+    healthMetrics.onFailedPublish(
```
ygree, line 109: Is this emptiness check really needed?

dd-trace-core/src/main/java/datadog/trace/core/monitor/TracerHealthMetrics.java
```diff
@@ -194,13 +233,19 @@ public void onSerialize(final int serializedSizeInBytes) {
   @Override
   public void onFailedSerialize(final List<DDSpan> trace, final Throwable optionalCause) {
     serialFailedDroppedTraces.inc();
+    serialFailedDroppedSpans.inc(trace != null ? trace.size() : 0);
```
ygree, line 236: Looks like it will count `FlushEvent` in case of failure. It either needs to call `serialFailedDroppedTraces.inc()` only if trace is not null or change the call side.

dd-trace-core/src/test/groovy/datadog/trace/core/monitor/HealthMetricsTest.groovy
```diff
@@ -72,18 +72,20 @@ class HealthMetricsTest extends DDSpecification {
...
+  @Flaky
   def "test onFailedPublish"() {
     setup:
     def latch = new CountDownLatch(1)
     def healthMetrics = new TracerHealthMetrics(new Latched(statsD, latch), 100, TimeUnit.MILLISECONDS)
     healthMetrics.start()
 
     when:
-    healthMetrics.onFailedPublish(samplingPriority)
+    healthMetrics.onFailedPublish(samplingPriority,1)
```
ygree, line 83: I'd add a test sampling priority variable to make sure that it's used and assert it, instead of using constant here.

dd-trace-core/src/main/java/datadog/trace/common/writer/PayloadDispatcher.java
nayeem-kamal: There are some tests that test for empty traces that fail without the emptiness check so I think it is needed.

dd-trace-core/src/main/java/datadog/trace/common/writer/SpanSamplingWorker.java
```diff
@@ -135,23 +135,25 @@ public void onEvent(Object event) {
...
+            healthMetrics.onFailedPublish(samplingPriority, trace.size());
             log.debug(
                 "Sampled spans written to overfilled buffer after single span sampling. Counted but dropping trace: {}",
                 trace);
           } else if (unsampledSpans.size() > 0
               && (droppingPolicy.active() || !secondaryQueue.offer(unsampledSpans))) {
             if (sampledSpans.isEmpty()) {
               // dropped all spans because none of the spans sampled
-              healthMetrics.onFailedPublish(samplingPriority);
+              healthMetrics.onFailedPublish(samplingPriority, trace.size());
```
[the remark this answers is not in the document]
nayeem-kamal: Ok that makes sense

dd-trace-core/src/main/java/datadog/trace/core/monitor/TracerHealthMetrics.java
nayeem-kamal: Ok, I think it makes sense to only increment both if the trace is not null because a null trace is not really a serialization error.

dd-trace-core/src/test/groovy/datadog/trace/core/monitor/HealthMetricsTest.groovy
nayeem-kamal: This should be fixed now.

dd-trace-core/src/main/java/datadog/trace/common/writer/PayloadDispatcher.java
ygree: Maybe we don't need such a test at all. My reasoning is that we didn't have this check before and this method should never be called with empty trace according to the call side logic. But that's not a big deal, just an extra check.

ygree (approved):
LGTM!

PR #4900, opened 2023-03-14
Adding RequestContext slot for CI Visibility

devinsba:
# What Does This Do

Adds another RequestContext object type for use in CiVisibility

# Motivation

I plan to build on top of #4894 and add the `TestContext` using this

# Additional Notes


dd-trace-core/src/main/java/datadog/trace/core/CoreTracer.java
```diff
@@ -1143,6 +1143,7 @@ public class CoreSpanBuilder implements AgentTracer.SpanBuilder {
     private boolean errorFlag;
     private CharSequence spanType;
     private boolean ignoreScope = false;
+    private Map<RequestContextSlot, Object> requestContextData;
```
ygree, line 1146: Maybe avoid using Map and just have three variables for existing slots? It would help lower number of allocations.

ygree (changes requested):
LGTM. Only a small suggestion.

dd-trace-core/src/main/java/datadog/trace/core/CoreTracer.java
devinsba: Done

ygree (approved):
LGTM! Thanks!

PR #4904, opened 2023-03-15
Fix Spring Jetty instrumentation

ygree:
# What Does This Do

Fixes Spring Servlet Jetty instrumentation in case of error. It used to override `servlet.request` span's resource name for existing endpoint (first use case). Also it used to use `/error` resource name for non existing endpoint, but it should have been set to `404`.

## Before the fix:
image

## After the fix:
image

# Motivation

Fix issues for Spring Servlet Jetty when `servlet.request` span's resource name is overridden in case of an error.

# Additional Notes

APMS-8958
- [x] implement a test that covers this change
- [ ] fix [corresponding system-test](https://app.circleci.com/pipelines/github/DataDog/dd-trace-java/22757/workflows/281f958e-e9b4-46ea-bdfe-c9aea6cecd8b/jobs/643344)


amarziali (approved):
nice

PR #4905, opened 2023-03-16
WAF Metrics over Telemetry

ValentinZakharov:
# What Does This Do
Implemented mechanism to collect WAF metrics and report them via telemetry 

# Motivation
For internal use at the moment. Allows to get data about request processed by WAF.

# Additional Notes


dd-java-agent/appsec/src/main/java/com/datadog/appsec/powerwaf/PowerWAFModule.java
```diff
@@ -483,6 +485,15 @@ public void onDataAvailable(
...
         reqCtx.reportEvents(events, null);
+
+        if (flow.isBlocking()) {
+          MetricCollector.get().wafRequesBlocked();
+        } else {
+          MetricCollector.get().wafRequesTriggered();
+        }
+
+      } else {
+        MetricCollector.get().wafRequest();
```
smola, line 496: This shouldn't count when it's being blocked too?

PR #4909, opened 2023-03-16
add _dd.trace prefix to built in tracing based profiling attributes

richardstartin:
# What Does This Do

This lets users use these names for custom context.

# Motivation

# Additional Notes


ygree (approved):
LGTM

PR #4925, opened 2023-03-20
[draft] long running spans implementation

raphaelgavache:
First iteration on implementing long running spans missing multiple pieces. It's currently working end to end. I tried to incorporate all [comments of the first POC](https://github.com/DataDog/dd-trace-java/pull/3452) and focused mostly on the overhead

Missing pieces that I'll add after agreeing on the approach
- tests
- span start benchmark 
- flush health stats (number of running spans flushed, number skipped due to maxed buffers)

On a high overview PendingTraceBuffer thread will now track periodically each running span and flush copies (shallow for the moment) of the spans periodically.

### Concerned spans
- if the feature is not enabled a null check is added when a new span is registered
- if enabled the priority value is checked, only priority > 0 spans are concerned

### Hot path
- (span creation overhead) couple atomic int operations and a compareAndSwap (lockless storage) 
- (memory overhead) up 10k DDSpans can be kept alive for more 1 second then necessary

### PendingTraceBuffer thread
Pass every second through each tracked span and newly created spans.

### DIFF
--- a/dd-trace-core/src/main/java/datadog/trace/core/PendingTraceBuffer.java
+++ b/dd-trace-core/src/main/java/datadog/trace/core/PendingTraceBuffer.java
@@ -47,7 +47,10 @@ public abstract class PendingTraceBuffer implements AutoCloseable {
       return;
     }
     if (prio == null && tracer != null) {
-      tracer.sampler.sample(span); // sets sampling priority in the root context
+      // evaluate priority
+      DDSpan rootSpan = span.getLocalRootSpan();
+      DDSpan spanToSample = rootSpan == null ? span : rootSpan;
+      tracer.sampler.sample(spanToSample);
       prio = span.getSamplingPriority();
       if (prio != null && prio <= 0) {
         return;


dd-trace-core/src/main/java/datadog/trace/core/DDSpan.java
```diff
@@ -732,4 +732,25 @@ public void attachWrapper(Object wrapper) {
   public Object getWrapper() {
     return WRAPPER_FIELD_UPDATER.get(this);
   }
+
+  // only for long running spans
+  private DDSpan(DDSpan prevSpan, DDSpanContext newContext, long endTimeNano) {
+    this.durationNano = endTimeNano - prevSpan.startTimeNano;
```
raphaelgavache, line 738: if this is the right place, I'll update to use the [correct duration computation above](https://github.com/DataDog/dd-trace-java/blob/6243acb01a6609b7f45b0def7877ad6a788cec93/dd-trace-core/src/main/java/datadog/trace/core/DDSpan.java#L139-L154)

dd-trace-core/src/main/java/datadog/trace/core/DDSpanContext.java
```diff
@@ -48,7 +48,7 @@
  * the associated Span instance
  */
 public class DDSpanContext
-    implements AgentSpan.Context, RequestContext, TraceSegment, ProfilerContext {
+    implements AgentSpan.Context, RequestContext, TraceSegment, ProfilerContext, Cloneable {
```
dougqh, line 51: Typically, you don't want to implement Cloneable in Java.  It is pretty much considered an anti-pattern.

dd-trace-core/src/main/java/datadog/trace/core/Metadata.java
```diff
@@ -16,6 +16,7 @@ public final class Metadata {
   private final boolean measured;
   private final boolean topLevel;
   private final CharSequence origin;
+  private final int longRunningVersion;
```
dougqh, line 19: Conceptually, I think this is probably the wrong place.
While longRunningVersion is span metadata, it isn't span metadata about the type or content of the span.

It is really metainfo used by the publishing code.  Ideally, it wouldn't be on span, span context, or metadata at all, but I'd be willing to violate functional encapsulation if need be for efficiency reasons.  After all, the real job of the tracer is just to send spans as efficiently as possible.

dd-trace-core/src/main/java/datadog/trace/core/PendingTraceBuffer.java
```diff
@@ -34,6 +35,8 @@ public interface Element {
     boolean setEnqueued(boolean enqueued);
   }
 
+  public RunningSpansBuffer runningSpans = null;
```
dougqh, line 38: Why is this field public?

dd-trace-core/src/main/java/datadog/trace/core/RunningSpansBuffer.java
```diff
@@ -0,0 +1,190 @@
...
+
+    // add is called by any thread generating a span
+    private void add(DDSpan s) {
+      int index = writeIndex.incrementAndGet();
+      // keep the writeIndex value in [0, k*ringSize[ with k*ringSize << maxInt64. Hitting the limit
+      // would require maxInt64 threads incrementing writeIndex and not passing below
+      if (index > ringSize && index % ringSize == 0) {
+        this.writeIndex.addAndGet(-ringSize);
+      }
+      boolean success = this.ring.compareAndSet(index % ringSize, null, new WeakReference<>(s));
```
dougqh, line 61: We don't want to creating WeakReferences on a regular basis.
They create an extra object and they are problematic for the garbage collector.  
We only want to use WeakReferences as last resort when hard references don't work.

Here the ring satisfies the need to restrict the size, the WeakReferences aren't really necessary.

dd-trace-core/src/main/java/datadog/trace/core/DDSpanContext.java
```diff
@@ -786,4 +793,9 @@ public void setDataCurrent(String key, Object value) {
     String tagKey = "_dd." + key + ".json";
     this.setTag(tagKey, value);
   }
+
+  @Override // shallow clone, can this create a race?
+  protected Object clone() throws CloneNotSupportedException {
+    return super.clone();
```
dougqh, line 799: We don't need to cloning.  The publishing code has the ability to do synchronized immutable snapshots as needed already; we should be using that instead.

dd-trace-core/src/main/java/datadog/trace/core/PendingTrace.java
```diff
@@ -161,6 +161,19 @@ void registerSpan(final DDSpan span) {
...
     healthMetrics.onCreateSpan();
+    trackRunningSpan(span);
+  }
+
+  void trackRunningSpan(final DDSpan span) {
+    // feature not enabled
+    if (pendingTraceBuffer.runningSpans == null) {
+      return;
+    }
+    Integer prio = span.getSamplingPriority();
```
dougqh, line 172: We need to double check if this sufficient for sampling checks.  I'm not sure that it is.

dd-trace-core/src/main/java/datadog/trace/core/RunningSpansBuffer.java
```diff
@@ -0,0 +1,190 @@
...
+  }
+
+  /**
+   * RingBuffer is a lockless data struct with multiple threads writing and a single thread reading.
+   *
+   * <p>Writes and flushes in the buffer start at the same position and move to the right by
+   * increments of 1 flush passes through at most newAddition to the buffer When the buffer is full
+   * new spans are not tracked anymore
+   */
+  private static class RingBuffer {
```
dougqh, line 44: There are already Java implementations of right buffers available.  I don't think thing we should be implementing another one.

I also don't think we necessarily need a ring buffer in this case.  It makes sense when the publication model is constantly inserting new spans into the buffer, but we don't really need to be doing that.

All, we really need to do is track which unfinished spans we might want to send.  The inserts and removals of long running spans should by definition be rare.

dd-trace-core/src/main/java/datadog/trace/core/RunningSpansBuffer.java
```diff
@@ -0,0 +1,190 @@
...
+  public void add(DDSpan s) {
+    runningSpans.add(s);
+  }
+
+  public void flushRunningSpans(long nowMilli) {
+    if (nowMilli - lastFlushMilli < TimeUnit.SECONDS.toMillis(1)) {
+      return;
+    }
+    runningTracker.flushAndCompact(nowMilli);
+    runningSpans.flush(runningTracker::add);
```
dougqh, line 33: We typically want to avoid capturing lambdas because they incur allocation.  Admittedly, this method is being called infrequently enough that we can probably live with it.

dd-trace-core/src/main/java/datadog/trace/core/RunningSpansBuffer.java
```diff
@@ -0,0 +1,190 @@
...
+  }
+
+  /**
+   * LongRunningTracker tracks running spans, evicting them if completed and writing them
+   * periodically
+   *
+   * <p>Only called by the PendingTraceBuffer thread.
+   */
+  private class LongRunningTracker {
+    private final int maxTrackedSpans = 10000;
```
dougqh, line 110: maxTrackedSpans should be configurable.  10_000 per trace also seems way too high by default.

dd-trace-core/src/main/java/datadog/trace/core/RunningSpansBuffer.java
```diff
@@ -0,0 +1,190 @@
...
+    private final int flushPeriodMilli = (int) TimeUnit.SECONDS.toMillis(5); // todo get from config
+    private final TrackedSpan[] spansArray = new TrackedSpan[maxTrackedSpans];
+    private int lastElementIndex = -1;
+    private int missedAdd = 0;
+
+    private class TrackedSpan {
```
dougqh, line 116: We shouldn't need another class to wrap a Span.  The startTime is also already in the Span, so I'm not sure what time is for here.

That said if we were going to store a flush version separately, this is actually where I would have expected it to be.

dd-trace-core/src/main/java/datadog/trace/core/DDSpanContext.java
```diff
@@ -138,6 +138,8 @@
 
   private volatile int encodedOperationName;
 
+  private volatile int longRunningVersion = 0;
```
dougqh, line 141: I haven't studied carefully yet, but volatile might not be sufficient.
volatile guarantee an atomic increment if that's what you are looking for.

dougqh (changes requested):
I left a number of comments about overhead, stylistic, and exception handling issues.

However, I'd actually suggest that we revisit the overall approach because I think we're coming at the problem from the wrong angle.

I think that stems from us not thinking through the requirements carefully and jumping to the conclusion that a timer was the appropriate solution.

I think the actual requirement is that we want to "guarantee" that we send the contents of a span before we send the first reference to that span.  Otherwise, we'll still have temporally orphaned spans from the backend perspective.

Ideally, we'd also solve this for cross-service spans as well, but I'll ignore that for the moment.

So my first thought was that when we decide to send a span, we'd check to make sure that all its local ancestors had been sent or were scheduled to send.  However, that requires more book keeping than I'd really like.  Fortunately, I think there's a simpler solution.  

My second thought was that we could simply switch from enqueueing to send on finish to enqueueing to send on start.  Given typical usage patterns, that would generally satisfy the requirement that all ancestor spans were scheduled to send along with their descendants.  Although - arguably, it is a bit wasteful.

I considered the possible refinement of waiting until the first child span is created - OR - the span is finished.  However, that fails if the span starts and then schedules a background job far into the future and then waits for the result.  In our code, we use continuations in those cases, so we could extend this to also include "OR a continuation is created", but that wouldn't work when using OpenTracing or OpenTelemetry APIs which don't have a concept of continuations.

In the end, I think the simplest solution is to just switch to enqueueing on start when long running spans are enabled.  If we must, we can filter out spans that are unfinished and have only been running for a short period of time in the flushing or writing code.

That gives us an easy way to track the unfinished spans with far less machinery than is present in this PR, but it still doesn't cover everything.

First, we don't want to be sending to the trace agent constantly, so we need to gate and batch the flow from a trace in some fashion.  And we still need something to send the long running spans on a regular interval.

One simple way to achieve this would be to track the last publishing time inside PendingTrace.  However, I think a better way is just to leverage the PendingTrace delayed queue mechanism that already exists.

When unfinished span publication is enabled, we should just put the PendingTrace into the delayed state as soon as the first span is created.  When we reach the time limit, we can just send whatever has been accumulated.

As for the actual sending, I do think we'll want to only send updates to spans after they've been hanging around for a few seconds; otherwise, we'll up resending the span every second when the tracer publishes.

That need not break the requirement that an ancestor should be sent with or before its descendants.  We just have to make sure any unsent unfinished spans are sent at the first opportunity.  Only after the first send is the unfinished span eligible to be skipped on a time basis.

I'm fine with storing the necessary book keeping to do that in the spans directly.  It does break encapsulation a bit, but the tracer's job is to send span efficiently and that's the most efficient option.

I think that the outlined approach will fulfill the end user needs better -- and it should do so with fewer data structures, timers, and much less allocation than what is proposed here.

Long term, I'd still really love to just rework all this code and maybe shift to something driven by some tables and a simple duty cycle -- but that doesn't need to be today.

dd-trace-core/src/main/java/datadog/trace/common/writer/ddagent/TraceMapperV0_4.java
```diff
@@ -81,6 +82,15 @@ public void accept(Metadata metadata) {
         writable.writeUTF8(InstrumentationTags.DD_TOP_LEVEL);
         writable.writeInt(1);
       }
+      if (metadata.longRunningversion() != 0) {
```
dougqh, line 85: longRunningversion isn't following naming conventions.  It should be longRunningVersion.

raphaelgavache:
Thanks for the review!
I'll take a step back and reach out with different approaches and their tradeoffs.
I agree that flushing other spans too would provide a better experience. Initially I thought it would be covered by the fact that parents will be flushed / partially flushed before by precedence. But the more complete a trace is, the better it is visually
The complex handling of running spans was to avoid adding a lock between threads generating spans.
I did not find for the moment a good OOTB java struct that satisfies lockless insertion, bounded size and fast read iteration. I'm definitely biased by writing go and will search more
While my spark case is a low throughput, I had in mind that tracking long running spans could be desirable for high throughputs too. In particular to catch spans that are never closed

dd-trace-core/src/main/java/datadog/trace/core/RunningSpansBuffer.java
raphaelgavache: I wasn't sure between the cost of allocating a WeakReference and their impact vs keeping alive spans longer in memory. I removed them

dd-trace-core/src/main/java/datadog/trace/core/DDSpan.java
```diff
@@ -732,4 +732,25 @@ public void attachWrapper(Object wrapper) {
...
+    this.startTimeNano = prevSpan.getStartTime();
+    this.externalClock = prevSpan.externalClock;
+    this.forceKeep = prevSpan.forceKeep;
+    this.context = newContext;
+  }
+
+  public DDSpan cloneLongRunning(int version, long flushTimeNano) {
```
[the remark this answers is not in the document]
raphaelgavache: I did that to set the duration on the DDSpan. Currently the duration is used to know if a [span is finished](https://github.com/DataDog/dd-trace-java/blob/18af25c6027a1dbc257b0ab363e7f443e7ef4ae5/dd-trace-core/src/main/java/datadog/trace/core/DDSpan.java#L72-L79), I went with a copy to avoid impacting this logic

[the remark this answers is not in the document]
dougqh: I'm not quite understanding the connection here.  Yes, the duration shouldn't be set because the span isn't finished.

I guess the intention was to make an artificially finished span to trick the rest of the machinery into sending.  If so, I'm strongly opposed to that because it will just lead to confusion and other surprises later on.

As stated in my other comments, I think we just need to change the sending rules.  Switching to send on start should simplify this considerably.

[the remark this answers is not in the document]
raphaelgavache: We need the time of flush to know when the span was last seen running and for the moment the duration field is used behind for that. I think that using the duration makes sense as it's the duration of the span at time of flush. If you prefer not using the duration field for that we can discuss adding a new tag with the flush date

dd-trace-core/src/main/java/datadog/trace/core/PendingTrace.java
dougqh: We should also make sure we're clear on the rules with how this interacts with single span sampling.  My position is that it shouldn't interact with single span sampling.

In my mind, long running spans is actually making sure we don't have temporally orphaned spans -- e.g. it is about the interaction between the long running span and its descendants.  Meaning that long running span support is really about traces and not about single span sampling.

The other interaction with sampling to consider is if someone switches sampling on manually or if sampling is changed remotely.  In that case, we might have missed the chance to enqueue some span starts.  If that's case, then we probably want to traverse the span chain to late enqueue any ancestors that we might have missed previously.

That's a case that can come later, but I think it is important to handle before this can be enabled by default.

dd-trace-core/src/main/java/datadog/trace/core/DDSpan.java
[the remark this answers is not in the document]
raphaelgavache: Actually if we don't use the duration field here we would have to set it just to have it in the product. Search / analytic and trace view with the partial span duration being the last known duration. This would end up being a trick.

dd-trace-core/src/main/java/datadog/trace/core/PendingTrace.java
raphaelgavache: Nice catch, the sampling check did not work. My test setup was functional as my spans had manual keep.
To make it work properly I had to trigger the [priority evaluation like this](https://github.com/DataDog/dd-trace-java/blob/426c4dd382d8d90aa8944785fab3ea60ad3c0303/dd-trace-core/src/main/java/datadog/trace/core/PendingTraceBuffer.java#L49-L58)

I need to look at how single span sampling is implemented

PR #4927, opened 2023-03-20
service naming: all datastore service is dd_service for v1

amarziali:
# What Does This Do

In v1 we first named service name as `{{DD_SERVICE}}-{DBMS}` but the requirement changed and we will always use `{{DD_SERVICE}}. 
This PR changes the naming for v1 (all a bunch of tests).

# Motivation

# Additional Notes


ygree (approved):
LGTM

PR #4930, opened 2023-03-21
Allow unknown sampling priorities from external sources

bantonsson:
# What Does This Do

Allows sampling priorities outside of the _known_ range from external sources. This is per the spec, where we should just pass those through.

# Motivation

# Additional Notes


PerfectSlayer (approved):
Seems okay for allowing wider range of priority.
What is the current use case if any? (other than being more compliant to the spec)

bantonsson:
No real use case as of know. Just to pass tests, and be compliant.

PR #4936, opened 2023-03-21
Report evaluation error inline in message

jpbempel:
# What Does This Do
instead of reporting UNDEFINED we put the the evaluation error message in curly braces:
ex: this is a log line with arg={Cannot find symbol: var42}

# Motivation
Having evaluation error message directly into the log message

# Additional Notes

### DIFF
--- a/dd-java-agent/agent-debugger/src/main/java/com/datadog/debugger/agent/LogMessageTemplateSummaryBuilder.java
+++ b/dd-java-agent/agent-debugger/src/main/java/com/datadog/debugger/agent/LogMessageTemplateSummaryBuilder.java
@@ -70,13 +70,17 @@ public class LogMessageTemplateSummaryBuilder implements SummaryBuilder {
         sb.append(segment.getStr());
       } else {
         if (parsedExr != null) {
-          Value<?> result = parsedExr.execute(entry);
-          if (result.isUndefined()) {
-            sb.append(result.getValue());
-          } else if (result.isNull()) {
-            sb.append("null");
-          } else {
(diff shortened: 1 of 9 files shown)


dd-java-agent/agent-debugger/src/main/java/com/datadog/debugger/agent/LogMessageTemplateSummaryBuilder.java
```diff
@@ -70,13 +70,17 @@ private void executeExpressions(Snapshot.CapturedContext entry) {
...
-          if (result.isUndefined()) {
-            sb.append(result.getValue());
-          } else if (result.isNull()) {
-            sb.append("null");
-          } else {
-            serializeValue(sb, segment.getParsedExpr().getDsl(), result.getValue());
+          try {
+            Value<?> result = parsedExr.execute(entry);
+            if (result.isUndefined()) {
+              sb.append(result.getValue());
```
shatzi, line 76: when do we get here? do we have a unit-test that show that as well?

shatzi (approved):
Nice!

dd-java-agent/agent-debugger/src/main/java/com/datadog/debugger/agent/LogMessageTemplateSummaryBuilder.java
jpbempel: `UNDEFINED_VALUE` is still in use in the implementation of the EL. for example if you call `isEmpty` on a variable that is not a recognized collection, you end up in this situation

PR #4938, opened 2023-03-21
WIP: Add support for 128-bit TraceId logging (6to7)

PerfectSlayer:
# What Does This Do

# Motivation

# Additional Notes


dd-smoke-tests/log-injection/src/test/groovy/datadog/smoketest/LogInjectionSmokeTest.groovy
```diff
@@ -60,6 +69,10 @@ abstract class LogInjectionSmokeTest extends AbstractSmokeTest {
       command.add("-Ddd.version=" as String)
       command.add("-Ddd.service.name=" as String)
     }
+    if (trace128bits) {
+      command.add("-Ddd.$TRACE_128_BIT_TRACEID_GENERATION_ENABLED=true" as String)
+      command.add("-Ddd.$TRACE_128_BIT_TRACEID_LOGGING_ENABLED=true" as String)
+    }
```
bantonsson, line 75: Maybe there should be some code in `"check raw file injection"` that validates if the `TraceId` is 128 or 64 bit as well?

dd-trace-api/src/main/java/datadog/trace/api/DD64bTraceId.java
```diff
@@ -61,9 +59,13 @@ public static DD64bTraceId fromHex(String s) throws NumberFormatException {
   }
 
   static DD64bTraceId create(long id, String str) {
-    if (id == 0) return ZERO;
-    if (id == -1) return MAX;
-    return new DD64bTraceId(id, str);
+    if (id == 0 && ZERO != null) {
```
bantonsson, line 62: Ah, the `null` check is for the intitialization dance.

jpbempel (approved):

dd-trace-api/src/main/java/datadog/trace/api/DD64bTraceId.java
PerfectSlayer: Yes... It's kind of dirty but it's the best I could find to keep the `DDTraceId` api 😞 
Would you like me to document it with comments for our future selves?

bantonsson: Please add a comment. Could this change be its own PR?

PerfectSlayer: The DDTraceId API fix? Yes, will do.

bantonsson (approved):

PR #4942, opened 2023-03-22
IAST: support Spring Webflux

cataphract:


dd-java-agent/instrumentation/spring-webflux-5/src/main/java/datadog/trace/instrumentation/springwebflux/server/iast/TaintHttpHeadersGetFirstAdvice.java
```diff
@@ -0,0 +1,24 @@
...
+class TaintHttpHeadersGetFirstAdvice {
+  @Advice.OnMethodExit(suppress = Throwable.class)
+  public static void after(@Advice.Argument(0) String arg, @Advice.Return String value) {
+    WebModule module = InstrumentationBridge.WEB;
+    if (module == null || arg == null || value == null) {
+      return;
+    }
+    String lc = arg.toLowerCase(Locale.ENGLISH);
+
+    module.onHeaderName(lc);
```
smola, line 21: Is this header name comming from the request or application code? Note that we should not taint strings coming from user code.

dd-java-agent/instrumentation/spring-webflux-5/src/main/java/datadog/trace/instrumentation/springwebflux/server/iast/TaintHttpHeadersToSingleValueMapAdvice.java
```diff
@@ -0,0 +1,30 @@
...
+  public static void after(@Advice.Return Map<String, String> values) {
+    PropagationModule propModule = InstrumentationBridge.PROPAGATION;
+    WebModule module = InstrumentationBridge.WEB;
+    if (module == null || propModule == null || values == null) {
+      return;
+    }
+
+    for (Map.Entry<String, String> e : values.entrySet()) {
+      String lc = e.getKey().toLowerCase(Locale.ENGLISH);
+      module.onHeaderName(lc);
```
smola, line 26: Same here. And even if it somes from the request, once you've transformed it with toLowerCase, you wouldn't be tainting the right instance anymore?

smola (changes requested):

dd-java-agent/instrumentation/spring-webflux-5/src/main/java/datadog/trace/instrumentation/springwebflux/server/iast/RequestHeaderMapResolveAdvice.java
```diff
@@ -0,0 +1,40 @@
...
+    WebModule module = InstrumentationBridge.WEB;
+    if (module == null || values == null) {
+      return;
+    }
+
+    if (values instanceof MultiValueMap) {
+      for (Map.Entry<String, List<String>> e :
+          ((MultiValueMap<String, String>) values).entrySet()) {
+        String lc = e.getKey().toLowerCase(Locale.ENGLISH);
+        module.onHeaderName(lc);
```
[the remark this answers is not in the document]
cataphract: yes, the tainting of the transformed key makes no sense, I've fixed it.

re: the second point, this advice is precisely for when the user injects a full map (or multimap) of headers, so I think it makes sense to taint it all. If he was interested in only one or two headers, he'd likely inject only those specific ones into string variables.

dd-java-agent/agent-iast/src/main/java/com/datadog/iast/propagation/PropagationModuleImpl.java
```diff
@@ -96,6 +96,19 @@ public void taint(final byte origin, @Nullable final Object... toTaintArray) {
     }
   }
 
+  @Override
+  public boolean isTainted(@Nullable Object obj) {
+    final TaintedObjects taintedObjects = getTaintedObjects();
```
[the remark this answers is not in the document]
cataphract: The current version now checks first if the object is taintable:

```
  @Override
  public boolean isTainted(@Nullable Object obj) {
    if (obj instanceof Taintable) {
      return ((Taintable) obj).$DD$isTainted();
    }

    // ...
  }
```

This is done in https://github.com/DataDog/dd-trace-java/pull/4996/files

dd-java-agent/instrumentation/spring-webflux-5/src/main/java/datadog/trace/instrumentation/springwebflux/server/iast/TaintHttpHeadersToSingleValueMapAdvice.java
cataphract: You're right. Fixed.

dd-java-agent/instrumentation/spring-webflux-5/src/main/java/datadog/trace/instrumentation/springwebflux/server/iast/TaintHttpHeadersGetFirstAdvice.java
cataphract: Fixed

dd-java-agent/instrumentation/spring-webflux-5/src/main/java/datadog/trace/instrumentation/springwebflux/server/iast/HandleMatchAdvice.java
```diff
@@ -0,0 +1,68 @@
...
+import org.springframework.web.reactive.HandlerMapping;
+import org.springframework.web.server.ServerWebExchange;
+
+@RequiresRequestContext(RequestContextSlot.IAST)
+public class HandleMatchAdvice {
+
+  @SuppressWarnings("Duplicates")
+  @Advice.OnMethodExit(suppress = Throwable.class)
+  public static void after(
+      @Advice.Argument(2) ServerWebExchange xchg, @ActiveRequestContext RequestContext reqCtx) {
```
smola, line 20: Can `xchg` possibly be null? If there's the slightest chance of it, I will check it first and do early return. But I guess it would not be the case for onExit...

smola (approved):

PR #4945, opened 2023-03-22
Add padding to B3 propagation headers

bantonsson:
# What Does This Do

Adds padding to B3 propagation headers to be compliant with other tracers behavior.

# Motivation

# Additional Notes

This behavior can be turned off by the setting the java property `-Ddd.trace.propagation.style.b3.padding.enabled=false` or the environment variable `DD_TRACE_PROPAGATION_STYLE_B3_PADDING_ENABLED=false`

### DIFF
--- a/dd-trace-core/src/main/java/datadog/trace/core/CoreTracer.java
+++ b/dd-trace-core/src/main/java/datadog/trace/core/CoreTracer.java
@@ -381,7 +381,9 @@ public class CoreTracer implements AgentTracer.TracerAPI {
       instrumentationGateway(new InstrumentationGateway());
       injector(
           HttpCodec.createInjector(
-              config.getTracePropagationStylesToInject(), invertMap(config.getBaggageMapping())));
+              config,
+              config.getTracePropagationStylesToInject(),
+              invertMap(config.getBaggageMapping())));
       extractor(
           HttpCodec.createExtractor(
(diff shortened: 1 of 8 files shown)


dd-trace-core/src/main/java/datadog/trace/core/CoreTracer.java
```diff
@@ -381,7 +381,9 @@ public CoreTracerBuilder config(final Config config) {
       instrumentationGateway(new InstrumentationGateway());
       injector(
           HttpCodec.createInjector(
-              config.getTracePropagationStylesToInject(), invertMap(config.getBaggageMapping())));
+              config,
+              config.getTracePropagationStylesToInject(),
+              invertMap(config.getBaggageMapping())));
```
PerfectSlayer, line 386: This usage feels weird, giving a full Config instance where the other parameters come from Convif too.

It feels like only a `Config` instance is needed with maybe two other boolean to express the `creation` behavior:
- One for "all styles" rather than using the styles from Config
- One for "empty mapping" rather than using the one from Config.

PerfectSlayer (approved):
A minor comment about the API change.

dd-trace-core/src/main/java/datadog/trace/core/CoreTracer.java
bantonsson: Yeah it the whole config behavior of the propagation should be cleaned up.

PR #4948, opened 2023-03-23
Peer service for client and precursors

amarziali:
# What Does This Do

Implement `peer.service` tag logic for span kind client spans.

Peer service can be:
* Set by the user: in this case it takes precedence to automatic calculation
* Automatically calculated (value derived from other tags)

Calculation logic is (higher in the list, higher in priority):
* `rpc.service` for grpc and rmi
* `db.instance` for databases
* `net.peer.name` as generic fallback

the three above are called precursors since they also need to be set. two of the three are missing hence added:
* `rpc.service`: is the fully qualified name (without method) of the service being called. Added only to client spans (not server for now)
* `net.peer.name`: is the OTEL alias for `peer.hostname`

Once the `peer.service` value is set we also have to know it's provenance. For this we set the tag `_dd.peer.service.source` having as value the tag name which has served as source to calculate peer service.

# Motivation

# Additional Notes

### DIFF
--- a/dd-trace-core/src/main/java/datadog/trace/core/CoreTracer.java
+++ b/dd-trace-core/src/main/java/datadog/trace/core/CoreTracer.java
@@ -558,7 +558,10 @@ public class CoreTracer implements AgentTracer.TracerAPI {
     this.dataStreamsMonitoring.start();
 
     this.tagInterceptor =
-        null == tagInterceptor ? new TagInterceptor(new RuleFlags(config)) : tagInterceptor;
+        null == tagInterceptor
+            ? new TagInterceptor(
+                new RuleFlags(config), (span, tag, value) -> span.unsafeSetTag(tag, value))
+            : tagInterceptor;
 
     if (config.isCiVisibilityEnabled()) {
       addTraceInterceptor(CiVisibilityTraceInterceptor.INSTANCE);
--- a/dd-trace-core/src/main/java/datadog/trace/core/taginterceptor/TagInterceptor.java
+++ b/dd-trace-core/src/main/java/datadog/trace/core/taginterceptor/TagInterceptor.java
@@ -5,10 +5,20 @@ import static datadog.trace.api.DDTags.MEASURED;
 import static datadog.trace.api.DDTags.ORIGIN_KEY;
 import static datadog.trace.api.DDTags.SPAN_TYPE;
 import static datadog.trace.api.sampling.PrioritySampling.USER_DROP;
+import static datadog.trace.bootstrap.instrumentation.api.Tags.DB_INSTANCE;
 import static datadog.trace.bootstrap.instrumentation.api.Tags.HTTP_METHOD;
 import static datadog.trace.bootstrap.instrumentation.api.Tags.HTTP_STATUS;
 import static datadog.trace.bootstrap.instrumentation.api.Tags.HTTP_URL;
-import static datadog.trace.core.taginterceptor.RuleFlags.Feature.*;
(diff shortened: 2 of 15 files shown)


dd-trace-core/src/main/java/datadog/trace/core/CoreTracer.java
```diff
@@ -558,7 +558,10 @@ private CoreTracer(
     this.dataStreamsMonitoring.start();
 
     this.tagInterceptor =
-        null == tagInterceptor ? new TagInterceptor(new RuleFlags(config)) : tagInterceptor;
+        null == tagInterceptor
```
dougqh, line 561: Generally, I'm opposed to use of the current TagInterceptor system.  I do think we need something TagInterceptor like, but better.

dd-trace-core/src/main/java/datadog/trace/core/taginterceptor/TagInterceptor.java
```diff
@@ -269,6 +298,44 @@ private static boolean interceptMeasured(DDSpanContext span, Object value) {
     return false;
   }
 
+  private boolean interceptForPeerService(
```
dougqh, line 301: I think I like what this is trying to do.  I like the idea of setting peer.service outside the instrumentation, but I don't like using a TagInterceptor to do it.  Ideally, I think calculating peer.service should be done lazily on demand rather than eagerly.

I have a dream of a span processing that makes lazy computation easier, but obviously, it doesn't exist yet.

I'm also thinking of having peer.service always be set on the span, but translating it to service or peer.service depending on the versioning scheme.  My reason for proposing that is that it will keep the instrumentation relatively simple and will fix a bug with sampling when context tracking breaks down.

dd-trace-core/src/main/java/datadog/trace/core/taginterceptor/TagInterceptor.java
```diff
@@ -269,6 +298,44 @@ private static boolean interceptMeasured(DDSpanContext span, Object value) {
...
+      final DDSpanContext span, final String tag, final Object value) {
+    // check if the schema support the peer service logic
+    if (!SpanNaming.instance().namingSchema().supportsPeerService()) {
+      return false;
+    }
+    // if the user is setting it, let's set the provenance and we're finished
+    if (Tags.PEER_SERVICE.equals(tag)) {
+      // user is setting it so let set the provenance and exit
+      span.setTag(DDTags.PEER_SERVICE_SOURCE, "_dd.user");
```
dougqh, line 310: Maybe we should have a helper method to handle this.
It looks like we're reading one tag to set peer service and also peer service source -- that could be encapsulated into a static helper or maybe even a method on Span.

I can see we're also handling the user overriding peer.service which is good.  I've also thought about having a system where user overrides are automatically preserved over values provided by auto-instrumentation to avoid this kind of complexity, but again, that doesn't exist right now.

dougqh (changes requested):
I like the overall direction, but I find myself rather torn about the approach.

I think the PR is doing pretty much everything right given the APIs that we have today, but I also feel our APIs aren't well suited to the problem.

If the core span model supported lazy computation and user override more naturally most of this wouldn't be necessary.  I envision these as a replacement for the problematic TagInterceptor which I think is unfortunately the best option at the moment.

I have to think a bit about how far to go right now.

The one thing that I would like to see is cleaning up service in the data model regardless of whether or not the new naming schema is on.  If we do that, then we can fix a bug with sampling not working as expected when context tracking breaks down because of missing integration.

amarziali:
Closed in favour of #4973

PR #4957, opened 2023-03-24
Fix matching source filename with different paths

jpbempel:
# What Does This Do
Separate matching for qualified filenames and simple filenames collect all definitions matching the prefix and not just one as path could be different outside the prefix

# Motivation
issue with multiple probe having different path for the same source code ex: `VetController.java` & `org/springframework/samples/petclinic/vet/VetController.java`

# Additional Notes

### DIFF
--- a/dd-java-agent/agent-debugger/src/main/java/com/datadog/debugger/agent/TransformerDefinitionMatcher.java
+++ b/dd-java-agent/agent-debugger/src/main/java/com/datadog/debugger/agent/TransformerDefinitionMatcher.java
@@ -25,10 +25,7 @@ public class TransformerDefinitionMatcher {
 
   public TransformerDefinitionMatcher(Configuration configuration) {
     this.definitionsByClass = buildDefinitionsMap(configuration.getDefinitions());
-    populateDefinitionFileNamesMap(
-        configuration.getDefinitions(),
-        definitionsBySimpleFileNames,
-        definitionsByQualifiedFileNames);
+    populateDefinitionFileNamesMap(configuration.getDefinitions());
     this.definitionFileNames = buildDefinitionFileNamesTrie(definitionsByQualifiedFileNames);
   }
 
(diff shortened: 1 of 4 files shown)


dd-java-agent/agent-debugger/src/main/java/com/datadog/debugger/agent/TransformerDefinitionMatcher.java
```diff
@@ -53,19 +56,23 @@ private Map<String, List<ProbeDefinition>> buildDefinitionsMap(
     return map;
   }
 
-  private Map<String, List<ProbeDefinition>> buildDefinitionFileNamesMap(
-      Collection<ProbeDefinition> definitions) {
-    Map<String, List<ProbeDefinition>> resultMap = new HashMap<>();
+  private void populateDefinitionFileNamesMap(
+      Collection<ProbeDefinition> definitions,
+      Map<String, List<ProbeDefinition>> definitionsBySimpleFileNames,
+      Map<String, List<ProbeDefinition>> definitionsByQualifiedFileNames) {
```
ojung, line 62: Maybe just use the fields instead of passing this here? 🤔

dd-java-agent/agent-debugger/src/main/java/com/datadog/debugger/agent/TransformerDefinitionMatcher.java
```diff
@@ -53,19 +56,23 @@ private Map<String, List<ProbeDefinition>> buildDefinitionsMap(
...
     for (ProbeDefinition definition : definitions) {
       String fileName = definition.getWhere().getSourceFile();
       if (fileName == null) {
         continue;
       }
+      Map<String, List<ProbeDefinition>> resultMap =
+          fileName.indexOf('/') != -1
+              ? definitionsByQualifiedFileNames
+              : definitionsBySimpleFileNames;
```
ojung, line 71: This is a bit confusing, how about 
```java
if (fileName.indexOf('/') == -1) {
  definitionsByQualifiedFileNames.computeIfAbsent().add();
} else {
  definitionsBySimpleFileNames.computeIfAbsent().add();
}
```
?

dd-java-agent/agent-debugger/src/main/java/com/datadog/debugger/agent/Trie.java
```diff
@@ -85,6 +88,29 @@ public String getStringStartingWith(String prefix) {
     return node.str;
   }
 
+  /**
+   * @param prefix prefix to search into the trie
+   * @return all the strings that matches the given prefix
```
ojung, line 93:
```suggestion
   * @return all the strings that match the given prefix
```

ojung (approved):
Thanks 👍

dd-java-agent/agent-debugger/src/main/java/com/datadog/debugger/agent/TransformerDefinitionMatcher.java
jpbempel: mixed feeling about it, repeating exactly the line here
how it's confusing?

jpbempel: could rename`resultMap` to `targetMap` is it better?

ojung: I don't know, took me a second too long to understand it. The if is clearer IMO. But this is subjective, so not blocking.

dd-java-agent/agent-debugger/src/main/java/com/datadog/debugger/agent/TransformerDefinitionMatcher.java
jpbempel: done

dd-java-agent/agent-debugger/src/main/java/com/datadog/debugger/agent/Trie.java
jpbempel: done

PR #4958, opened 2023-03-24
Enable ddprof allocation profiling for JDK 11+ by default

jbachorik:
# What Does This Do
This modifies the default value of the ddprof based allocation profiling enablement.
The allocation profiling in the DD java profiler lib is now ready to be used as default when possible. This means that when `ddprof` is enabled (either manually or automatically) and we are running on JDK 11+ the ddprof allocation profiling will be activated.

It is still possible to manually disable by config as well as enable for systems where we don't enable it automatically if users are feeling adventurous.

# Motivation
Make the ddprof allocation profiling automatically available where possible.

# Additional Notes


dd-java-agent/agent-profiling/profiling-ddprof/src/main/java/com/datadog/profiling/ddprof/DatadogProfilerConfig.java
```diff
@@ -128,10 +129,11 @@ public static boolean getWallContextFilter(ConfigProvider configProvider) {
   }
 
   public static boolean isAllocationProfilingEnabled(ConfigProvider configProvider) {
-    return getBoolean(
-        configProvider,
-        PROFILING_DATADOG_PROFILER_ALLOC_ENABLED,
-        PROFILING_DATADOG_PROFILER_ALLOC_ENABLED_DEFAULT);
+    return (Platform.isJavaVersionAtLeast(11) && !Platform.isJ9())
```
richardstartin, line 132: I think this needs to be a logical and, because this logic makes it impossible to disable on OpenJDK 11+

richardstartin (approved):

Issue #4210, opened 2022-11-11
GraphQL Instrumentation - operation as resource_name


glenwinters:
I'm interested in graphing trace metrics related to the GraphQL instrumentation by operation name. I can graph latency, hits and errors using `trace.graphql_java`, but there's no tag to view by operation name.

One potential solution is to set `esource_name` (e.g. graphql.request) to the operation name. `resource_name` is already an available tag on trace metrics, so then it could be used to split metrics by operation name.

krosen040:
Making the GraphQL operation name the name of the spans would make the generated trace metrics and the resources on the APM service dashboard a lot more useful. Right now all GraphQL metrics are tagged with `resource_name:graphql.field`, `resource_name:graphql.request`, etc which isn't very helpful.

theJC:
And for anonymous operations, using the sanitized/normalized graphql query text could be an option.

PR #4960, opened 2023-03-25
graphql: Set resource name to operation name (or query when anonymous) to make…

theJC:
… Datadog resource view extremely useful for GraphQL requests

# What Does This Do
Resolves: https://github.com/DataDog/dd-trace-java/issues/4210
https://help.datadoghq.com/hc/en-us/requests/1003318

# Motivation
Having much richer and useful Datadog Resource View for graphql requests.  Performance characteristics vary widely across operations, but are much more likely to similar and SLOs will likely be aligned based on the operation being executed.

https://www.apollographql.com/docs/react/data/operation-best-practices/#name-all-operations

Platforms like Apollo Studio leverage operation names to group and measure performance using operation-level metrics.

# Additional Notes

### DIFF
--- a/dd-java-agent/instrumentation/graphql-java-14.0/src/main/java/datadog/trace/instrumentation/graphqljava/GraphQLInstrumentation.java
+++ b/dd-java-agent/instrumentation/graphql-java-14.0/src/main/java/datadog/trace/instrumentation/graphqljava/GraphQLInstrumentation.java
@@ -99,7 +99,8 @@ public final class GraphQLInstrumentation extends SimpleInstrumentation {
     String operationName = operationDefinition.getName();
 
     requestSpan.setTag("graphql.operation.name", operationName);
-
+    String resourceName = operationName != null ? operationName : state.getQuery();
+    requestSpan.setResourceName(resourceName);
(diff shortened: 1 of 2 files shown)


theJC:
@ygree  -- I believe this is the last one on our list of issues that would allow us to migrate off our custom libraries and just use Datadog instrumentation out of the box.  This is very similar in nature to https://github.com/DataDog/dd-trace-java/pull/4663, only from an graphql execution perspective rather than field perspective.

If you can help us get this reviewed and merged, it would be appreciated (and as an added bonus, I'll leave you alone for a while ;) ) Thanks!

ygree (approved):
Looks good! Thank you for this contribution!

bantonsson:
@ygree @theJC this has started failing some tests intermittently as can be seen here https://app.circleci.com/pipelines/github/DataDog/dd-trace-java/23372/workflows/34d367e4-93bf-49c0-ab54-ddb727e0e0c5/jobs/676542

theJC:
https://github.com/DataDog/dd-trace-java/pull/5008 created to address issue with unit tests due to timing of merges and new unit test from one pull request not coded to expect new behavior from the other pull request.

PR #4968, opened 2023-03-27
Rename GraphQL query->source tag to align it with the dd-trace-js instrumentaion

### LINKED ISSUE #4961: Update graphql query to be in graphql.source tag
Change use of graphql.query -> graphql.source so that:

- is aligned/and the same as what dd-trace-js does
- graphql query is more richly presented to customers in Datadog trace view

ygree:
# What Does This Do

Resolves #4961

# Motivation

Align with dd-trace-js

# Additional Notes

N/A


nayeem-kamal (approved):
LGTM

PR #4969, opened 2023-03-28
DSM - Add DataStreamsCheckpointer to public API

kr-igor:
Added DataStreamsCheckpointer with two methods:

* setConsumeCheckpoint
* setProduceCheckpoint

The main purpose for this change is to allow customers willing to instrument currently unsupported queuing technology manually. An example may be SNS or Kinesis, where we do not support context propagation.

Example usage:
```
import datadog.trace.api.experimental.*;
....
DataStreamsCheckpointer checkpointer = DataStreamsCheckpointer.get();
checkpointer.setProduceCheckpoint("kinesis", "my-stream", new Carrier());
```


dd-trace-core/src/main/java/datadog/trace/core/CoreTracer.java
```diff
@@ -968,6 +976,47 @@ public boolean addTraceInterceptor(final TraceInterceptor interceptor) {
...
+    }
+
+    PathwayContext pathwayContext =
+        extractPathwayContext(carrier, DataStreamsContextCarrierAdapter.INSTANCE);
+    span.mergePathwayContext(pathwayContext);
+
+    LinkedHashMap<String, String> sortedTags = new LinkedHashMap<>();
+    sortedTags.put(DIRECTION_TAG, DIRECTION_IN);
+    sortedTags.put(TYPE_TAG, type);
+    sortedTags.put(TOPIC_TAG, source);
```
hokitam, line 999: nit: can you put topic before type, since this is sorted tags?

dd-trace-core/src/main/java/datadog/trace/core/CoreTracer.java
```diff
@@ -968,6 +976,57 @@ public boolean addTraceInterceptor(final TraceInterceptor interceptor) {
...
+
+  @Override
+  public void setConsumeCheckpoint(String type, String source, DataStreamsContextCarrier carrier) {
+    if (type == null || type.isEmpty() || source == null || source.isEmpty()) {
+      log.warn("setConsumeCheckpoint should be called with non-empty type and source");
+      return;
+    }
+
+    AgentSpan span = activeSpan();
+    if (span == null) {
```
piochelepiotr, line 992: in that case, do you think it's worth / possible to set a checkpoint anyway? but without propagating the context?

piochelepiotr: On the other hand, it might be worth failing here, since it indicates that context propagation will be broken.

piochelepiotr (approved):

internal-api/src/main/java/datadog/trace/bootstrap/instrumentation/api/AgentTracer.java
```diff
@@ -412,6 +424,14 @@ public ScopeState newScopeState() {
     public DataStreamsMonitoring getDataStreamsMonitoring() {
       return dataStreamsMonitoring;
     }
+
+    @Override
+    public void setConsumeCheckpoint(
+        String type, String source, DataStreamsContextCarrier setter) {}
```
piochelepiotr, line 430: nit: `setter` is a bad name here I think.

kr-igor: Good catch! I'll fix it

dd-trace-core/src/main/java/datadog/trace/core/CoreTracer.java
kr-igor: I think it's fine for now to just ignore the checkpoint. We should not throw exceptions here as it may crash the application, the warning should be ok.

dd-trace-core/src/main/java/datadog/trace/core/CoreTracer.java
dougqh: I'm curious why are we using a LinkedHashMap.  That adds extra overhead and the ordering semantics are rather odd.  It sounds like you are just going to sort the tags later anyway, so a HashMap would be sufficient.

dd-trace-core/src/main/java/datadog/trace/core/datastreams/DataStreamsContextCarrierAdapter.java
```diff
@@ -0,0 +1,29 @@
...
+import datadog.trace.api.datastreams.DataStreamsContextCarrier;
+import datadog.trace.bootstrap.instrumentation.api.AgentPropagation;
+import datadog.trace.bootstrap.instrumentation.api.AgentPropagation.KeyClassifier;
+import java.util.Map;
+
+public class DataStreamsContextCarrierAdapter
+    implements AgentPropagation.Setter<DataStreamsContextCarrier>,
+        AgentPropagation.ContextVisitor<DataStreamsContextCarrier> {
+
+  public static final DataStreamsContextCarrierAdapter INSTANCE =
```
dougqh, line 12: Since this looks like it is intended to be a singleton, it would make sense to make the constructor protected or private.
Generally, I'd suggest private because I doubt this class is intended to be extended - in which case it should also be final.
On the off hand that the class can be extended, you'd want to make the constructor protected.

dd-trace-api/src/main/java/datadog/trace/api/GlobalTracer.java
```diff
@@ -28,6 +29,11 @@ public String getSpanId() {
         public boolean addTraceInterceptor(TraceInterceptor traceInterceptor) {
           return false;
         }
+
+        @Override
+        public DataStreamsCheckpointer getDataStreamsCheckpointer() {
```
dougqh, line 34: I'm not sure returning null is the appropriate choice here, but it depends on the other code.
If other code is assuming that the DataStreamsCheckpointer will not be null, it might be better to return a NOP version of DataStreamsCheckpointer instead of null.

dd-trace-api/build.gradle
```diff
@@ -23,6 +23,8 @@ excludedClassesCoverage += [
   'datadog.trace.api.experimental.ProfilingScope',
   'datadog.trace.api.experimental.ProfilingContext',
   'datadog.trace.api.experimental.ProfilingContextSetter.NoOp',
+  'datadog.trace.api.datastreams.DataStreamsCheckpointer',
```
dougqh, line 26: I'm not terribly hung up on code coverage, but I do feel having to add to the coverage exclusion list is testing smell.

dd-trace-core/src/main/java/datadog/trace/core/CoreTracer.java
```diff
@@ -968,6 +976,57 @@ public boolean addTraceInterceptor(final TraceInterceptor interceptor) {
...
+    LinkedHashMap<String, String> sortedTags = new LinkedHashMap<>();
+    sortedTags.put(DIRECTION_TAG, DIRECTION_IN);
+    sortedTags.put(TOPIC_TAG, source);
+    sortedTags.put(TYPE_TAG, type);
+
+    setDataStreamCheckpoint(span, sortedTags);
+  }
+
+  @Override
+  public void setProduceCheckpoint(String type, String target, DataStreamsContextCarrier carrier) {
```
dougqh, line 1010: What are the thread safety requirements for this code?
Is injectPathwayContext thread safe?

dd-trace-core/src/main/java/datadog/trace/core/CoreTracer.java
```diff
@@ -968,6 +976,57 @@ public boolean addTraceInterceptor(final TraceInterceptor interceptor) {
     return interceptors.add(interceptor);
   }
 
+  @Override
+  public DataStreamsCheckpointer getDataStreamsCheckpointer() {
+    return this;
```
dougqh, line 981: I'm not sure how I feel about CoreTracer also being a DataStreamsCheckpointer.

On some level, I think it might be core for our core tracer to just have the extra methods, but in that case, I expect we'd just call them directly inside our own instrumentations.

dd-trace-api/build.gradle
kr-igor: DataStreamsCheckpointer is an interface, not sure it what exactly should be tested inside `trace.api`. I'll add tests for the actual implementation in `dd-trace-core` shortly.

dougqh (commented):
At this point, we're basically just adding the methods onto Tracer directly.  I'm not entirely sure how I feel about that.

One thing that I would really like to see clean-up is the use of LinkedHashMap.  As far as I can tell, HashMap should be sufficient.

dd-trace-core/src/main/java/datadog/trace/core/CoreTracer.java
kr-igor: We can't use HashMap, as ordered tags are used to generate the `hash`. HashMap will cause random hashes to be produced for the same pathway. Also, this PR just follows the LinkedHashMap usage in other places (HTTP / Kafka / RabbitMQ), if we want to perform a cleanup or revisit the logic it should be done in a separate PR, I think.

dd-trace-core/src/main/java/datadog/trace/core/CoreTracer.java
kr-igor: Yes, it is. `pathwayContext.setCheckpoint`, which is called inside, uses a ReentrantLock.

dd-trace-core/src/main/java/datadog/trace/core/CoreTracer.java
kr-igor: Tracer already has all of the methods needed to perform a DSM checkpoint. Specifically, `injectPathwayContext` and `extractPathwayContext`, proxying calls to `DataStreamsMonitoring`. These methods are a part of `AgentPropagation` interface, which core Tracer implements. The two new methods are just a simplification on top of `injectPathwayContext` / `extractPathwayContext` making DSM easier to use. 

I agree this can be done better, but for that we need to refactor the way DSM support was implemented. Which seems like a significant change out of scope of this PR.

dd-trace-core/src/main/java/datadog/trace/core/CoreTracer.java
dougqh: That doesn't fit my understanding from elsewhere -- and I think LinkedHashMap is likely the wrong construct for the requirements.

First, just to be cleared the keys in LinkedHashMap are insertion ordered -- not alphabetically ordered.  Furthermore, they aren't update ordered which may or may not be what you want.

But if really want strict ordering, I think you'd better off just implementing a custom type.  Relying on insertion order into a Map is non-idiomatic and can easily lead to bugs.

The existence of the comments about ordering to me suggests an API design problem.  I think this needs to be addressed, but not in this PR.

dd-trace-api/src/main/java/datadog/trace/api/experimental/DataStreamsCheckpointer.java
```diff
@@ -0,0 +1,49 @@
+package datadog.trace.api.experimental;
+
+import datadog.trace.api.GlobalTracer;
+import datadog.trace.api.Tracer;
+import datadog.trace.api.internal.InternalTracer;
+
+/** An interface to Data Streams checkpointer, allowing passing the context manually. */
+public interface DataStreamsCheckpointer {
+  static DataStreamsCheckpointer get() {
```
dougqh, line 9: private

dd-trace-api/src/main/java/datadog/trace/api/experimental/DataStreamsContextCarrier.java
```diff
@@ -0,0 +1,16 @@
+package datadog.trace.api.experimental;
+
+import java.util.Map.Entry;
+import java.util.Set;
+
+/** An interface representing the context carrier. Typically, message headers. */
+public interface DataStreamsContextCarrier {
```
dougqh, line 7: I have a few API concerns as noted elsewhere.  However since this is experimental and DSM is off by default, I will let it be for now.

dd-trace-core/src/main/java/datadog/trace/core/CoreTracer.java
dougqh: That's fine with me as long as it is something that we can change later.  The public API wasn't something that we can change later, so that required attention.

In general, I feel like in our haste we are resisting clean-up and that concerns me.  The most notable problem is around LinkedHashMap which I know has been raised on multiple reviews previously.  I'd like to see that addressed in a separate PR.

dd-trace-core/src/main/java/datadog/trace/core/CoreTracer.java
dougqh: I think there might be better options than ReentrantLock, but I'm fine with that for now.  If / when we want to turn on DSM more broadly, we're going to have to come back and revisit the performance overhead concerns.

dougqh (approved):
Since these changes are now in an "experimental" package, I'm fine to merge this as is.

That said, there are still a number of internal API choices that I think we need to revisit.  I would like to see those revisited soon.

kr-igor:
> Since these changes are now in an "experimental" package, I'm fine to merge this as is.
> 
> That said, there are still a number of internal API choices that I think we need to revisit. I would like to see those revisited soon.

dd-trace-api/src/main/java/datadog/trace/api/experimental/DataStreamsCheckpointer.java
kr-igor: Not sure what you mean here. This interface should be used outside of the package, which can't be done if it's private (this kinda defeat the purpose of the change).

dougqh: Yeah at first, I didn't realize what it was doing, so I comment on it needing to be private.  Then I realized that it is getting an instance, so I altered my comment.

PR #4973, opened 2023-03-28
lazy calculate peer service for client and precursors

amarziali:
# What Does This Do

Implement `peer.service` tag logic for *span kind producer or client*. Other span kinds are out of scope.

Peer service, its precursors and the source are calculated lazily leveraging the TagPostProcessor mechanism. 

Peer service can be:
* Set by the user: in this case it takes precedence to automatic calculation
* Automatically calculated (value derived from other tags that are called _precursors_)

Calculation logic generally depends to the instrumentation. As a generic fallback we use (in order of priority) 
* `db.instance` 
* `peer.hostname`

Some other specialized precursors are used:
* kafka producers: `messaging.kafka.bootstrap.servers`
* grpc or rmi clients: `rpc.service`

Once the `peer.service` value is set we also have to know it's provenance. For this we set the tag `_dd.peer.service.source` having as value the tag name which has served as source to calculate peer service.

If the user (or a third party instrumentation) sets the tag `peer.service` so that the source will be set to `peer.service`

The computation is done lazily on a tag postprocessor that kicks in if:
* The schema version is v1 -> always enabled
* The schema version is v0 -> enabled only if `DD_TRACE_PEER_SERVICE_DEFAULTS_ENABLED` is set to true

There is also the possibility to remap the value automatically calculated. For this the parameter `DD_TRACE_PEER_SERVICE_MAPPING` has to be supplied by the user. The same format as `DD_SERVICE_MAPPING` is accepted

In terms of testing, the `tagsAssert` grants that, generally speaking, all the span of kind client and producer should define the tag `peer.service` whose value has to be the same as the used precursor. 
In order to test that the value has been calculated using the right precursor, the macro `TagsAssert.peerServiceFrom(String)` should be used.

Since not all the instrumentation set the precursors today, there can be some exceptions and for some of them the value cannot be calculated. On this case, on the related tests the macro `defaultTagsNoPeerService()` should be used explicitely

# Motivation

# Additional Notes

### DIFF
--- a/dd-java-agent/instrumentation/grpc-1.5/src/main/java/datadog/trace/instrumentation/grpc/client/GrpcClientDecorator.java
+++ b/dd-java-agent/instrumentation/grpc-1.5/src/main/java/datadog/trace/instrumentation/grpc/client/GrpcClientDecorator.java
@@ -7,9 +7,12 @@ import static datadog.trace.core.datastreams.TagsProcessor.TYPE_TAG;
 
 import datadog.trace.api.Config;
 import datadog.trace.api.GenericClassValue;
+import datadog.trace.api.cache.DDCache;
+import datadog.trace.api.cache.DDCaches;
 import datadog.trace.api.naming.SpanNaming;
 import datadog.trace.bootstrap.instrumentation.api.AgentSpan;
 import datadog.trace.bootstrap.instrumentation.api.InternalSpanTypes;
+import datadog.trace.bootstrap.instrumentation.api.Tags;
 import datadog.trace.bootstrap.instrumentation.api.UTF8BytesString;
 import datadog.trace.bootstrap.instrumentation.decorator.ClientDecorator;
 import io.grpc.MethodDescriptor;
@@ -50,6 +53,8 @@ public class GrpcClientDecorator extends ClientDecorator {
               return UTF8BytesString.create(input.getName());
             }
           });
+
+  private static final DDCache<String, String> RPC_SERVICE_CACHE = DDCaches.newFixedSizeCache(64);
 
   public UTF8BytesString requestMessageType(MethodDescriptor<?, ?> method) {
     return messageType(method.getRequestMarshaller());
@@ -93,7 +98,12 @@ public class GrpcClientDecorator extends ClientDecorator {
     AgentSpan span =
         startSpan(OPERATION_NAME)
             .setTag("request.type", requestMessageType(method))
-            .setTag("response.type", responseMessageType(method));
+            .setTag("response.type", responseMessageType(method))
+            // method.getServiceName() may not be available on some grpc versions
+            .setTag(
+                Tags.RPC_SERVICE,
+                RPC_SERVICE_CACHE.computeIfAbsent(
+                    method.getFullMethodName(), MethodDescriptor::extractFullServiceName));
     span.setResourceName(method.getFullMethodName());
     return afterStart(span);
   }
--- a/dd-trace-core/src/main/java/datadog/trace/common/writer/ListWriter.java
+++ b/dd-trace-core/src/main/java/datadog/trace/common/writer/ListWriter.java
@@ -1,8 +1,13 @@
 package datadog.trace.common.writer;
 
+import datadog.trace.api.DDTags;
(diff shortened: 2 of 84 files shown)


internal-api/src/main/java/datadog/trace/api/naming/v1/PeerServiceNamingV1.java
```diff
@@ -0,0 +1,49 @@
...
+        value = unsafeTags.get(precursor);
+        if (value != null) {
+          // we have a match. Use the tag name for the source
+          source = precursor;
+          break;
+        }
+      }
+      // if something matched now set the value and the source
+      if (value != null) {
+        final Map<String, Object> ret = new HashMap<>(2);
```
bantonsson, line 41: So the reason for the `unsafeTags` here is so that you can change it directly instead of creating costly extra objects. See how it's used in the `queryObfuscator`

dd-java-agent/instrumentation/grpc-1.5/src/main/java/datadog/trace/instrumentation/grpc/client/GrpcClientDecorator.java
```diff
@@ -93,7 +94,11 @@ public <ReqT, RespT> AgentSpan startCall(MethodDescriptor<ReqT, RespT> method) {
     AgentSpan span =
         startSpan(OPERATION_NAME)
             .setTag("request.type", requestMessageType(method))
-            .setTag("response.type", responseMessageType(method));
+            .setTag("response.type", responseMessageType(method))
+            // method.getServiceName() may not be available on some grpc versions
+            .setTag(
+                Tags.RPC_SERVICE,
+                MethodDescriptor.extractFullServiceName(method.getFullMethodName()));
```
bantonsson, line 101: Would it make sense to cache this?

dd-trace-core/src/main/java/datadog/trace/common/writer/ListWriter.java
```diff
@@ -38,7 +45,17 @@ public void write(List<DDSpan> trace) {
     if (!filter.accept(trace)) {
       return;
     }
-
+    for (DDSpan span : trace) {
```
bantonsson, line 48: Wouldn't it be better to trigger the proper processing in `span.processTagsAndBaggage` with an noop metadata consumer?

Something like this:
```
for (DDSpan span: trace) {
  span.processTagsAndBaggage(NoopMetadataConsumer.INSTANCE);
}
```

internal-api/src/main/java/datadog/trace/bootstrap/instrumentation/api/Tags.java
```diff
@@ -40,7 +41,6 @@ public class Tags {
   public static final String DB_OPERATION = "db.operation";
   public static final String DB_STATEMENT = "db.statement";
   public static final String MESSAGE_BUS_DESTINATION = "message_bus.destination";
-
```
bantonsson, line 43: ?

bantonsson (commented):
Looks good. Only a few questions.

dd-java-agent/instrumentation/grpc-1.5/src/main/java/datadog/trace/instrumentation/grpc/client/GrpcClientDecorator.java
amarziali: yes completely since behind it invokes indexOf/substring stuff

internal-api/src/main/java/datadog/trace/bootstrap/instrumentation/api/Tags.java
amarziali: 🤷 I compared with master again and removed the extra line

dd-trace-core/src/main/java/datadog/trace/common/writer/ListWriter.java
amarziali: As agreed I added a todo to do this way when we'll be able to manage scrambled urls in tests

bantonsson (approved):

PR #4980, opened 2023-03-28
WAF Metrics over Telemetry

ValentinZakharov:
# What Does This Do
Implemented mechanism to collect WAF metrics and report them via telemetry

# Motivation
For internal use at the moment. Allows to get data about request processed by WAF.

# Additional Notes


telemetry/src/main/java/datadog/telemetry/TelemetryRunnable.java
```diff
@@ -101,7 +113,9 @@ private boolean mainLoopIteration() throws InterruptedException {
 
   private void successWait() {
     consecutiveFailures = 0;
-    final int waitMs = telemetryService.getHeartbeatInterval();
+    final int waitMs =
+        Math.min(telemetryService.getMetricsInterval(), telemetryService.getHeartbeatInterval());
```
smola, line 117: This behavior seems surprising. A low metrics interval will also trigger more frequent heartbeats?

smola (approved):

PR #4982, opened 2023-03-29
Report eval error in case of uncaught exception

jpbempel:
# What Does This Do
include context with uncaught exception to the snapshot and add an evaluation error with the type + message of it.

# Motivation
Help to contextualize why evaluation of a condition may fail (locals not available)

# Additional Notes


shatzi (approved):
Looks good.

PR #4983, opened 2023-03-29
Disable flaky tests

jpbempel:
# What Does This Do

# Motivation
Flaky tests

# Additional Notes


bantonsson (approved):

dd-java-agent/agent-debugger/src/test/java/com/datadog/debugger/agent/SnapshotSerializationTest.java
```diff
@@ -769,6 +770,7 @@ public void fieldCount20() throws IOException {
   }
 
   @Test
+  @Disabled("flaky")
```
smola, line 773: You can use `@Flaky` annotation instead. Then these will still run in the `flaky` job, but they will not fail the build. The results and the percentage of flakiness will be visible in the dashboards.

jpbempel: yes thanks I will change it when I will try to fix them to see if my fix will work or not.

PR #4989, opened 2023-03-29
Add IAST propagation for java.net.URI

manuel-alvarez-alvarez:
# What Does This Do
Add required propagation call sites for java.net.URI in IAST

# Motivation

# Additional Notes

### DIFF
--- a/dd-java-agent/agent-iast/src/main/java/com/datadog/iast/propagation/PropagationModuleImpl.java
+++ b/dd-java-agent/agent-iast/src/main/java/com/datadog/iast/propagation/PropagationModuleImpl.java
@@ -119,6 +119,22 @@ public class PropagationModuleImpl implements PropagationModule {
   }
 
   @Override
+  public void taintIfAnyInputIsTainted(
+      @Nullable final Object toTaint, @Nullable final Object... inputs) {
+    if (toTaint == null || inputs == null || inputs.length == 0) {
+      return;
+    }
+    final TaintedObjects taintedObjects = lazyTaintedObjects();
+    for (final Object input : inputs) {
+      final Source source = firstTaintedSource(taintedObjects, input);
+      if (source != null) {
+        taintObject(taintedObjects, toTaint, source);
+        return;
+      }
+    }
+  }
+
+  @Override
   public void taint(final byte origin, @Nullable final Object... toTaintArray) {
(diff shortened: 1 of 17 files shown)


dd-java-agent/agent-iast/src/main/java/com/datadog/iast/propagation/PropagationModuleImpl.java
```diff
@@ -118,6 +118,22 @@ public void taintIfInputIsTainted(
     }
   }
 
+  @Override
+  public void taintIfAnyInputIsTainted(
```
smola, line 122: Great :D

dd-java-agent/instrumentation/java-lang/src/main/java/datadog/trace/instrumentation/java/lang/StringBuilderCallSite.java
```diff
@@ -19,17 +19,17 @@ public class StringBuilderCallSite {
   @CallSite.After("void java.lang.StringBuilder.<init>(java.lang.CharSequence)")
   @Nonnull
   public static StringBuilder afterInit(
-      @CallSite.This @Nonnull final StringBuilder self,
-      @CallSite.Argument @Nullable final CharSequence param) {
+      @CallSite.AllArguments @Nonnull final Object[] params,
```
smola, line 22: Ugly but ok :wink:

smola (approved):

dd-java-agent/instrumentation/java-lang/src/main/java/datadog/trace/instrumentation/java/lang/StringBuilderCallSite.java
manuel-alvarez-alvarez: It's only for after call sites in constructors which are tricky due to how the byte code works (we could maintain the old approach for constructors with an arity lower than 4 but I think it will be uglier :laughing: )

dd-java-agent/agent-iast/src/main/java/com/datadog/iast/propagation/PropagationModuleImpl.java
manuel-alvarez-alvarez: This API reminds me of old times :wink:

PR #4992, opened 2023-03-29
[USM] USMO-259 - Support Async Frameworks for USM

val06:
# What Does This Do

Added SSLEngine and SocketChannel instrumentations as `ForTypedHierarchy` instrumentations
   - We hook write/read/close of any class that extends the `SocketChannel` (`SocketChannelImpl` for sun module)
   - We hook wrap and unwrap of any class that extends the `SSLEngine` class (SSLEngineImpl for sun module)

# Motivation

Support asynchronous https frameworks. Mostly used by the server side, e.g: jetty, spring via Restlet, but also some asynchronous clients like `java.net.http.HttpClient` 

# Additional Notes

   - Need to refactor the USMMessage classes hierarchy, BaseUSMMessage can't contain connection anymore, since we don't have such information inside the SSLEngine instrumentation 
     - Implemented a Visitor pattern to encode each entity (connection, peer, payload) into a given **pre-allocated direct** ByteBuffer (to avoid extra allocations and it must be direct as we pass it to the ioctl)
     - Removed the UsmMessage interface and all related implementations including the MessageFactory, now the only reference to jna exists in the ExtractorImpl and can be easily substituted later by our own tiny library to drop the heavy jna dependency


dd-java-agent/instrumentation/sslengine/src/main/java/datadog/trace/instrumentation/sslengine/SslEngineInstrumentation.java
```diff
@@ -0,0 +1,133 @@
...
+
+    @Advice.OnMethodExit(suppress = Throwable.class)
+    public static void wrap(
+        @Advice.This final SSLEngine thiz,
+        @Advice.Argument(0) final ByteBuffer[] srcs,
+        @Advice.Return SSLEngineResult result) {
+
+      // no source buffer - usually happens during handshake
+      if (srcs.length == 0) {
+        System.out.println("[wrap] no source buffers");
```
Yumasi, line 63: Avoid using `println` for logging, use a logger instead.

Yumasi (changes requested):
A few small comments, but otherwise LGTM ! Also, don't forget to apply spotless to fix the formatting before merging.

dd-java-agent/agent-bootstrap/src/main/java/datadog/trace/bootstrap/instrumentation/usm/Connection.java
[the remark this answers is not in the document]
val06: changed to use `/* */` for function/object documentation
and using `//` for in-function comments

Yumasi (approved):

dd-java-agent/agent-bootstrap/src/main/java/datadog/trace/bootstrap/instrumentation/usm/Connection.java
```diff
@@ -0,0 +1,94 @@
...
+  public int size() {
+    return CONNECTION_INFO_SIZE;
+  }
+
+  @Override
+  /*
+  Encodes the connection into a given buffer (Visitor pattern).
+  */
+  public void encode(ByteBuffer buffer) {
+    log.debug("encoding connection:");
```
dougqh, line 47: debug logging should probably be guarded with isDebugEnabled.
Probably also better to take advantage of interpolation than it is use concatenation.

We also just need to be careful about network in general.  Have you done any performance testing of this instrumentation?

dd-java-agent/agent-bootstrap/src/main/java/datadog/trace/bootstrap/instrumentation/usm/MessageEncoder.java
```diff
@@ -0,0 +1,66 @@
...
+import java.nio.ByteOrder;
+import org.slf4j.Logger;
+import org.slf4j.LoggerFactory;
+
+// This is a utility that allows to serialize different messages for eRPC protocol (over ioctl)
+// between USM instrumentation classes and SystemProbe
+public class MessageEncoder {
+  private static final Logger log = LoggerFactory.getLogger(MessageEncoder.class);
+
+  public enum MessageType {
```
dougqh, line 14: Java enums and especially enum switches are really bloated.  It might be better to just use ints.

dd-java-agent/agent-tooling/src/main/java/datadog/trace/agent/tooling/usm/UsmExtractorImpl.java
```diff
@@ -4,38 +4,42 @@
...
+        "sending ioctl: "
+            + String.format("%08x", USM_IOCTL_ID.intValue())
+            + " with buffer of size: "
+            + buffer.position());
+    NativeLong res =
+        CLibrary.Instance.ioctl(
+            new NativeLong(0),
+            USM_IOCTL_ID,
+            Pointer.nativeValue(Native.getDirectBufferPointer(buffer)));
+    log.debug("ioctl result: " + String.format("%08x", res.intValue()));
```
dougqh, line 39: If a recall correctly, this pattern behaves differently depending on the Locale.  You probably want to specify Locale here.

dd-java-agent/agent-bootstrap/src/main/java/datadog/trace/bootstrap/instrumentation/usm/Extractor.java
```diff
@@ -0,0 +1,21 @@
...
+
+  abstract class Supplier {
+    private static volatile Extractor SUPPLIER;
+
+    public static void send(Buffer message) {
+      SUPPLIER.send(message);
+    }
+
+    public static synchronized void registerIfAbsent(Extractor supplier) {
+      if (SUPPLIER == null) {
```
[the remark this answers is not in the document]
dougqh: I see this is being set by outside.

This could be implemented more efficiently.  You can use double checked locking or an Atomic to make this more efficient.

dd-java-agent/instrumentation/sslengine/src/main/java/datadog/trace/instrumentation/sslengine/SocketChannelImplInstrumentation.java
```diff
@@ -0,0 +1,108 @@
...
+    public static void onMessage(
+        @Advice.FieldValue("localAddress") InetSocketAddress localSocketAddr,
+        @Advice.FieldValue("remoteAddress") InetSocketAddress remoteSocketAddr)
+        throws IOException {
+
+      if (localSocketAddr == null || remoteSocketAddr == null) {
+        return;
+      }
+      boolean isIPv6 = localSocketAddr.getAddress() instanceof Inet6Address;
+      Connection connection =
```
dougqh, line 65: How often is onMessage called?

I'm worry about creating a Connection and Peer on each call.  I'd either pass the address objects directly or find a way to reuse the Connection object.

Do you have any benchmarks that illustrate the overhead?

dd-java-agent/instrumentation/sslengine/src/main/java/datadog/trace/instrumentation/sslengine/SslEngineInstrumentation.java
```diff
@@ -0,0 +1,115 @@
...
+        int consumed = 0;
+        for (int i = 0; i < srcs.length && consumed < b.length; i++) {
+          int oldPos = srcs[i].position();
+          srcs[i].position(srcs[i].arrayOffset());
+          srcs[i].get(b, 0, oldPos);
+          srcs[i].position(oldPos);
+          consumed += oldPos;
+        }
+
+        Peer peer = new Peer(thiz.getPeerHost(), thiz.getPeerPort());
```
dougqh, line 78: Probably don't want to creating extract objects on every wrap call.

dd-java-agent/agent-bootstrap/src/main/java/datadog/trace/bootstrap/instrumentation/usm/Connection.java
val06: switched to parameterized overload of debug 
(according to [this](https://www.slf4j.org/faq.html#logging_performance) should be fast enough, as long as the parameters are simple types and don't require heavy computation

dd-java-agent/agent-bootstrap/src/main/java/datadog/trace/bootstrap/instrumentation/usm/Extractor.java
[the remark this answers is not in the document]
val06: this was copied from WeakMaps [implementation](https://github.com/DataDog/dd-trace-java/blob/master/dd-java-agent/agent-bootstrap/src/main/java/datadog/trace/bootstrap/WeakMap.java), and introduced when implemented the first instrumentation (sslsocket), according to @mcculls advice

dd-java-agent/agent-tooling/src/main/java/datadog/trace/agent/tooling/usm/UsmExtractorImpl.java
val06: this log.debug prints the pointer in a format of 8 digits with a `0x` prefix. Why would locale impact the print of the hex digits?

val06: we still didn't do the performance tests, planning to do them in USM load-test cluster before dog-fooding on staging.

dd-java-agent/instrumentation/sslengine/src/main/java/datadog/trace/instrumentation/sslengine/SocketChannelImplInstrumentation.java
val06: i didn't want to make pre-mature optimizations, we will do benchmarks in our load-test cluster to determine if those needed

bm1549:
Hi @val06 👋 
We noticed that your pull request has been inactive for a while. In order to keep the project clean and organized, we’re going to close it for now.
Feel free to reopen it at any time when you’re ready to continue working on it.

PR #4996, opened 2023-03-30
Improvements to IAST APIs and tests

cataphract:


smola (approved):

internal-api/src/main/java/datadog/trace/api/iast/Taintable.java
```diff
@@ -6,6 +6,10 @@ public interface Taintable {
 
   void $$DD$setSource(final Source source);
 
+  default boolean isTainted() {
```
manuel-alvarez-alvarez, line 9: I'll rather prefix it with `$$DD$` to ensure that there are no collisions in case we instrument arbitrary classes at some point (for instance instrumenting implementations of an interface).

manuel-alvarez-alvarez (approved):

PR #5005, opened 2023-03-31
Add customer-facing API for CI Visibility

nikita-tkachenko-datadog:
# What Does This Do
Adds an API to create CI Visibility events (session, module, test suite, test case) programmatically.
The API is exposed to the customers.

# Motivation
The API is needed to support the needs of customers who are using either test frameworks that we do not instrument, or ad-hoc testing logic.
Such customers would be able to manually demarcate the start and end of their tests, and to supply additional test metadata.

# Additional Notes
In Scope of this PR existing CI Visibility logic was refactored to make test framework instrumentations use the same domain objects that are used by the manual API.


dd-java-agent/agent-bootstrap/src/main/java/datadog/trace/bootstrap/instrumentation/decorator/AbstractTestDecorator.java
```diff
@@ -1,275 +0,0 @@
...
-import datadog.trace.bootstrap.instrumentation.api.UTF8BytesString;
-import java.lang.reflect.Method;
-import java.nio.file.Path;
-import java.nio.file.Paths;
-import java.util.Collection;
-import java.util.Collections;
-import java.util.Map;
-import javax.annotation.Nullable;
-
-public abstract class AbstractTestDecorator extends BaseDecorator implements TestDecorator {
```
albertvaka, line 27: I see that all integrations have a Decorator class that extends `BaseDecorator`. With your change, our integrations don't use the `BaseDecorator` at all (nor the logic in it).  I'm not sure if there's any implication in doing that, have you checked?

dd-java-agent/instrumentation/junit-4.10/src/main/java/datadog/trace/instrumentation/junit4/JUnit4Decorator.java
```diff
@@ -1,27 +0,0 @@
...
-  }
-
-  @Override
-  public String testFramework() {
-    return "junit4";
-  }
-
-  @Override
-  protected String[] instrumentationNames() {
-    return new String[] {"junit", "junit-4", "junit-4-suite-events"};
```
albertvaka, line 20: Related to the removal of the Decorator classes. Weren't those instrumentation names needed? I see there's logic in `BaseDecorator` that uses them.

dd-java-agent/agent-ci-visibility/src/main/java/datadog/trace/civisibility/events/BuildEventsHandlerImpl.java
```diff
@@ -0,0 +1,140 @@
...
+
+  @Override
+  public void onModuleTestFrameworkDetected(
+      final T sessionKey,
+      final String moduleName,
+      final String frameworkName,
+      final String frameworkVersion) {
+    DDTestModule testModule = getTestModule(sessionKey, moduleName);
+    testModule.setTag(Tags.TEST_FRAMEWORK, frameworkName);
+    testModule.setTag(Tags.TEST_FRAMEWORK_VERSION, frameworkVersion);
```
albertvaka, line 99: Why do we have both "detected" and "start"? This seems more complex to me that just passing those two tags and setting them in the "start" function. I'm not asking you to change it, I'm just curious to know the reasons :)

dd-java-agent/agent-ci-visibility/src/main/java/datadog/trace/civisibility/TestDecoratorImpl.java
```diff
@@ -0,0 +1,94 @@
...
+    this.ciTags = ciTags;
+  }
+
+  protected String testType() {
+    return TEST_TYPE;
+  }
+
+  protected String runtimeName() {
+    return System.getProperty("java.runtime.name");
+  }
```
albertvaka, line 37: Not related to this PR, but we could cache those properties in local strings instead of calling `getProperty` each time.

dd-java-agent/agent-ci-visibility/build.gradle
```diff
@@ -8,7 +9,23 @@ apply from: "$rootDir/gradle/version.gradle"
 minimumBranchCoverage = 0.7
 minimumInstructionCoverage = 0.8
 
-excludedClassesCoverage += ["datadog.trace.civisibility.CiVisibilitySystem", "datadog.trace.civisibility.git.GitObject",]
+excludedClassesCoverage += [
```
albertvaka, line 12: Do we have to remember to add every class we create here manually? Can't we use wildcards? What's this for?

dd-java-agent/agent-bootstrap/src/main/java/datadog/trace/bootstrap/instrumentation/decorator/AbstractTestDecorator.java
nikita-tkachenko-datadog: There are components in the tracer that do not use decorators and instead add tags to their spans "manually" (I took inspiration from IAST, as it is similar to CI Visibility from the code organisation point of view).

I have also checked the code in `BaseDecorator`, and the only thing that will be missing is he ability so set spans sampling rate (which is disabled by default and has to be enabled explicitly by providing the instrumentation name - and is not applicable for CI Visibility spans anyway).

dd-java-agent/instrumentation/junit-4.10/src/main/java/datadog/trace/instrumentation/junit4/JUnit4Decorator.java
nikita-tkachenko-datadog: They were used to calculate the names of config properties, that were then used to determine if sampling is enabled for instrumentation and which sampling rate should be used.
We do not apply sampling to CI Visibility spans as it makes little sense from the product point of view.

dd-java-agent/agent-ci-visibility/src/main/java/datadog/trace/civisibility/events/BuildEventsHandlerImpl.java
nikita-tkachenko-datadog: I agree that it looks somewhat clumsy from the API point of view, but there's a reason for it :)

In Gradle we emit the session start event as soon as the build begins. Then follows a (sometimes lengthy) configuration phase when projects are evaluated, and only at the end of that phase we can examine the list of dependencies to determine the test frameworks.

If we fire the session start event after projects evaluation finishes, our total session duration will be skewed since we won't be counting the configuration phase.
Besides, it is possible to fail during configuration phase - before projects are evaluated - in which case we'd have to emit a failed session span.

Having two different events seemed like a simple way to cover the two cases above.

dd-java-agent/agent-ci-visibility/src/main/java/datadog/trace/civisibility/TestDecoratorImpl.java
nikita-tkachenko-datadog: Agree, could be worth doing.

dd-java-agent/agent-ci-visibility/build.gradle
nikita-tkachenko-datadog: There's a logic that evaluates code coverage stats during CI tests and fails the build if coverage percentage criteria are not met.

This list is needed to disable coverage check for specific classes (e.g. DTOs, or those that are covered with integration/smoke tests).

Wildcards are supported, but I think it's already bad enough without them :) at least now every entry has to be added consciously.

albertvaka (approved):
LGTM

PR #5001, opened 2023-03-31
Mark GraphQLVxForkedTest as flaky

bantonsson:
# What Does This Do

Marks GraphQLVxForkedTest as flaky.

# Motivation

Flaky tests are flaky.

# Additional Notes
This is just one example. They break a lot.

https://app.circleci.com/pipelines/github/DataDog/dd-trace-java/23372/workflows/34d367e4-93bf-49c0-ab54-ddb727e0e0c5/jobs/676542

PR #5008, opened 2023-04-03
Update resource name expected by recently added GraphQL CompletionStage support test

theJC:
…upport to align with updated behavior

# What Does This Do

Because https://github.com/DataDog/dd-trace-java/pull/4896 and https://github.com/DataDog/dd-trace-java/pull/4960 were being worked on independently and merged on around the same time, the changes in behavior introduced in https://github.com/DataDog/dd-trace-java/pull/4960 were not updated in the newly added unit test in https://github.com/DataDog/dd-trace-java/pull/4896.

This MR changes the test to properly expect the right value now.

# Motivation

To help address https://github.com/DataDog/dd-trace-java/pull/5001/files where these tests were marked Flaky, this will reduce the failure rate

# Additional Notes

### DIFF
--- a/dd-java-agent/instrumentation/graphql-java-14.0/src/test/groovy/GraphQLTest.groovy
+++ b/dd-java-agent/instrumentation/graphql-java-14.0/src/test/groovy/GraphQLTest.groovy
@@ -234,7 +234,7 @@ abstract class GraphQLTest extends VersionedNamingTestBase {
       trace(6) {
         span {
           operationName operation()
-          resourceName operation()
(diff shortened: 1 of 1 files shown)


smola:
Unless the flaky test is addressed, I would rather not remove the `@Flaky` annotations. We might, however, consider making the flaky annotations here more narrow and mark only some individual tests instead of the whole suite.

theJC:
Ok.  I updated this PR, and removed the commit that adjusted the @Flaky annotations and left them alone for now.

PR #5015, opened 2023-04-03
Decouple time in queue from legacy tracing

amarziali:
# What Does This Do

Decouple time in queue tracking from legacy tracing.
The time in queue can be driven independently with the parameter `dd.<instrumentation>.time-in-queue.enabled`
By default it will be enabled if legacy tracing is disabled (as today)

Quick note: for the v1 naming schema, time in queue is disabled by default because synthetic spans should not be present in the service map

# Motivation

# Additional Notes

### DIFF
--- a/dd-java-agent/agent-bootstrap/src/main/java/datadog/trace/bootstrap/instrumentation/jms/SessionState.java
+++ b/dd-java-agent/agent-bootstrap/src/main/java/datadog/trace/bootstrap/instrumentation/jms/SessionState.java
@@ -62,17 +62,17 @@ public final class SessionState {
   @SuppressFBWarnings("IS2_INCONSISTENT_SYNC")
   private boolean capturingFlipped = false;
 
-  public SessionState(int ackMode, boolean legacyTracing) {
+  public SessionState(int ackMode, boolean timeInQueueEnabled) {
     this.ackMode = ackMode;
     if (isAutoAcknowledge()) {
       this.capturedSpans = null; // unused in auto-ack
     } else {
       this.capturedSpans = new ArrayDeque<>();
     }
-    if (legacyTracing) {
-      this.timeInQueueSpans = Collections.emptyMap();
-    } else {
+    if (timeInQueueEnabled) {
       this.timeInQueueSpans = new ConcurrentHashMap<>();
+    } else {
+      this.timeInQueueSpans = Collections.emptyMap();
     }
   }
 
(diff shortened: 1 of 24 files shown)


dd-java-agent/agent-bootstrap/src/main/java/datadog/trace/bootstrap/instrumentation/jms/SessionState.java
```diff
@@ -62,17 +62,17 @@ public int compare(Map.Entry<Thread, TimeInQueue> o1, Map.Entry<Thread, TimeInQu
   @SuppressFBWarnings("IS2_INCONSISTENT_SYNC")
   private boolean capturingFlipped = false;
 
-  public SessionState(int ackMode, boolean legacyTracing) {
+  public SessionState(int ackMode, boolean timeInQueueEnabled) {
```
ygree, line 65: Nice renaming for better clarity

dd-java-agent/instrumentation/rabbitmq-amqp-2.7/src/main/java/datadog/trace/instrumentation/rabbitmq/amqp/RabbitDecorator.java
```diff
@@ -49,6 +49,12 @@ public class RabbitDecorator extends MessagingClientDecorator {
   public static final boolean RABBITMQ_LEGACY_TRACING =
       Config.get()
           .isLegacyTracingEnabled(SpanNaming.instance().version() == 0, "rabbit", "rabbitmq");
+
+  public static final boolean TIME_IN_QUEUE_ENABLED =
+      Config.get()
+          .isTimeInQueueEnabled(
+              !RABBITMQ_LEGACY_TRACING && SpanNaming.instance().version() == 0, "rabbitmq");
```
ygree, line 56: I wonder why it only checks "rabbitmq" but not "rabbit"?

ygree (approved):
Looks good to me!

dd-java-agent/instrumentation/rabbitmq-amqp-2.7/src/main/java/datadog/trace/instrumentation/rabbitmq/amqp/RabbitDecorator.java
amarziali: oh nice spot it should be both. Thanks Yury

PR #5023, opened 2023-04-04
Add IAST support to hibernate queries

manuel-alvarez-alvarez:
# What Does This Do
Add IAST support to hibernate queries for SQL injection

# Motivation

# Additional Notes

### DIFF
--- a/dd-java-agent/instrumentation/hibernate/core-3.3/src/main/java/datadog/trace/instrumentation/hibernate/core/v3_3/IastQueryInstrumentation.java
+++ b/dd-java-agent/instrumentation/hibernate/core-3.3/src/main/java/datadog/trace/instrumentation/hibernate/core/v3_3/IastQueryInstrumentation.java
@@ -1,0 +1,59 @@
+package datadog.trace.instrumentation.hibernate.core.v3_3;
+
+import static datadog.trace.agent.tooling.bytebuddy.matcher.NameMatchers.named;
+import static net.bytebuddy.matcher.ElementMatchers.isMethod;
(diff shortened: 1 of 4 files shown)


dd-java-agent/instrumentation/hibernate/core-3.3/src/main/java/datadog/trace/instrumentation/hibernate/core/v3_3/IastQueryInstrumentation.java
```diff
@@ -0,0 +1,59 @@
...
+        IastQueryInstrumentation.class.getName() + "$QueryMethodAdvice");
+  }
+
+  public static class QueryMethodAdvice {
+
+    @Advice.OnMethodEnter(suppress = Throwable.class)
+    public static void beforeMethod(@Advice.This final Query query) {
+      final SqlInjectionModule module = InstrumentationBridge.SQL_INJECTION;
+      if (module != null) {
+        module.onJdbcQuery(query.getQueryString());
```
manuel-alvarez-alvarez, line 42: Not sure if we want to differentiate SQL and HQL at the moment

smola: I think it's fair not to, at the moment. We might revisit this if we need to differentiate for evidence post-processing purposes (e,g, sensitive data) or something like that.

smola (approved):

PR #5028, opened 2023-04-04
[6to7] Add malformed_tid and inconsistent_tid decoding error support

PerfectSlayer:
# What Does This Do

Add `malformed_tid` and `inconsistent_tid` decoding error support from propagation tags.

# Motivation

# Additional Notes


dd-trace-core/src/main/java/datadog/trace/core/propagation/ptags/W3CPTagsCodec.java
```diff
@@ -664,24 +672,45 @@ private static class W3CPTags extends PTags {
...
+      this.error = null;
+    }
+
+    @Override
+    String getError() {
+      if (this.error != null) {
+        return this.error;
+      }
+      return super.getError();
+    }
```
PerfectSlayer, line 704: @bantonsson I wonder if `error` field would not be better stored in `PTags` class now errors can be handled from elsewhere than `PTagsWithError`.
The `PtagsWithError` would become a static factory method from `PTags` instead: `PTags.withError(String)`.
WDYT?

bantonsson: That sounds reasonable.

bantonsson (approved):
👍🏻 on moving the `error` field

PR #5030, opened 2023-04-04
Remove SummaryBuilder interface

jpbempel:
# What Does This Do
Changing LogTemplateBuilder api
use directly LogTemplateBuilder into LogProbe during evaluation

# Motivation
Summary was removed no need to keep the SummaryBuilder interface

# Additional Notes


shatzi (approved):
This is great. thank you!

PR #5035, opened 2023-04-05
lettuce: surface remote host and db index information

amarziali:
# What Does This Do

This PR wants to surface information we have at connection time on all lettuce spans.
Those tags are:
* peer.hostname
* peer.port
* db.redis.dbIndex

This works for all types of lettuce clients: sync, async and reactive

# Motivation

# Additional Notes


ygree (approved):
LGTM

PR #5041, opened 2023-04-07
vertx-redis: implement v1 naming and surface remote connection info

amarziali:
# What Does This Do

Implements opt-in v1 naming convestions:
* operation: redis.command
* service: DD_SERVICE

Surface remote host connection tags:
* peer.hostname
* peer.port

# Motivation

# Additional Notes

### DIFF
--- a/dd-java-agent/instrumentation/vertx-redis-client-3.9/src/main/java/datadog/trace/instrumentation/vertx_redis_client/RedisAPICallAdvice.java
+++ b/dd-java-agent/instrumentation/vertx-redis-client-3.9/src/main/java/datadog/trace/instrumentation/vertx_redis_client/RedisAPICallAdvice.java
@@ -134,7 +134,6 @@ public class RedisAPICallAdvice {
         break;
       default:
     }
-
     /*
     Store the response handler in the context so that it can be retrieved in RedisAPIImplSendAdvice
     */
--- a/dd-java-agent/instrumentation/vertx-redis-client-3.9/src/main/java/datadog/trace/instrumentation/vertx_redis_client/RedisAPIImplSendAdvice.java
+++ b/dd-java-agent/instrumentation/vertx-redis-client-3.9/src/main/java/datadog/trace/instrumentation/vertx_redis_client/RedisAPIImplSendAdvice.java
@@ -2,19 +2,25 @@ package datadog.trace.instrumentation.vertx_redis_client;
 
 import static datadog.trace.bootstrap.instrumentation.api.AgentTracer.activateSpan;
 import static datadog.trace.bootstrap.instrumentation.api.AgentTracer.noopSpan;
+import static datadog.trace.instrumentation.vertx_redis_client.VertxRedisClientDecorator.DECORATE;
 
 import datadog.trace.bootstrap.InstrumentationContext;
 import datadog.trace.bootstrap.instrumentation.api.AgentScope;
 import io.vertx.core.Future;
+import io.vertx.core.net.SocketAddress;
 import io.vertx.redis.RedisClient;
 import io.vertx.redis.client.Redis;
 import io.vertx.redis.client.RedisAPI;
+import io.vertx.redis.client.RedisConnection;
 import io.vertx.redis.client.Response;
 import net.bytebuddy.asm.Advice;
 
 public class RedisAPIImplSendAdvice {
   @Advice.OnMethodExit(suppress = Throwable.class)
-  public static void afterSend(@Advice.This RedisAPI self, @Advice.Return Future<Response> future) {
+  public static void afterSend(
+      @Advice.This RedisAPI self,
+      @Advice.FieldValue("connection") final RedisConnection connection,
+      @Advice.Return Future<Response> future) {
     /*
     Here we can safely set the handler for a command related Future instance.
     We need to take some precautions due to a non-existent operation which would allow setting the handler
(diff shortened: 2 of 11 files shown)


dd-java-agent/instrumentation/vertx-redis-client-3.9/src/main/java/datadog/trace/instrumentation/vertx_redis_client/RedisAPIImplSendAdvice.java
```diff
@@ -26,15 +32,25 @@ public static void afterSend(@Advice.This RedisAPI self, @Advice.Return Future<R
...
+        if (handler.clientSpan != null && connection != null) {
+          final SocketAddress socketAddress =
+              InstrumentationContext.get(RedisConnection.class, SocketAddress.class)
+                  .get(connection);
+          if (socketAddress != null) {
+            DECORATE.onConnection(handler.clientSpan, socketAddress);
+            DECORATE.setPeerPort(handler.clientSpan, socketAddress.port());
+          }
+        }
+        if (!future.isComplete()) {
```
ygree, line 45: NIce!

dd-java-agent/instrumentation/vertx-redis-client-3.9/src/main/java/datadog/trace/instrumentation/vertx_redis_client/ConnectionAsyncResultWrapper.java
```diff
@@ -0,0 +1,34 @@
...
+    this.contextStore = contextStore;
+    this.redisInstance = redisInstance;
+  }
+
+  @Override
+  public void handle(AsyncResult<RedisConnection> event) {
+    delegate.handle(
+        event.map(
+            connection -> {
+              contextStore.put(redisInstance, connection);
```
ygree, line 30: Here we associate one connection with a Redis client instance. Could there be more than one active connections for the same Redis client instance?

dd-java-agent/instrumentation/vertx-redis-client-3.9/src/main/java/datadog/trace/instrumentation/vertx_redis_client/RedisAPICallAdvice.java
```diff
@@ -134,7 +137,23 @@ public static boolean beforeCall(
         break;
       default:
     }
-
+    if (self instanceof RedisAPIImpl) {
+      Object connection =
+          InstrumentationContext.get(RedisAPIImpl.class, Object.class).get((RedisAPIImpl) self);
+      if (connection instanceof Redis) {
```
ygree, line 143: This two if checks are read as that the connection object associated with RedisAPIImpl can be either Redis or RedisConnection which is a bit confusing. If it's never the case and the object associated with RedisAPIImpl is always Redis then moving the next if-check inside of the first if-check scope would be less confusing.

ygree: The same applies to [RedisSendAdvice.java](https://github.com/DataDog/dd-trace-java/pull/5041/files#diff-8981ca5bfd751f6700ee29de7d46bce2444dd47b3cc34333da14bb796adb7b2b)

ygree (changes requested):
Looks good! Just got a couple of questions

dd-java-agent/instrumentation/vertx-redis-client-3.9/src/main/java/datadog/trace/instrumentation/vertx_redis_client/RedisAPIImplSendAdvice.java
amarziali: Thanks

dd-java-agent/instrumentation/vertx-redis-client-3.9/src/main/java/datadog/trace/instrumentation/vertx_redis_client/RedisAPICallAdvice.java
amarziali: Finally I found another approach so this is now unnecessary

dd-java-agent/instrumentation/vertx-redis-client-3.9/src/main/java/datadog/trace/instrumentation/vertx_redis_client/ConnectionAsyncResultWrapper.java
amarziali: Yes completely. You are right. I completely changed the approach in order to have something only depending to the connection (without relating to other objects) 
Now it looks to my eyes way better

dd-java-agent/instrumentation/vertx-redis-client-3.9/src/main/java/datadog/trace/instrumentation/vertx_redis_client/RedisConnectAdvice.java
```diff
@@ -0,0 +1,26 @@
...
+import datadog.trace.bootstrap.instrumentation.api.AgentScope;
+import datadog.trace.bootstrap.instrumentation.api.AgentSpan;
+import net.bytebuddy.asm.Advice;
+
+public class RedisConnectAdvice {
+  @Advice.OnMethodEnter
+  public static AgentScope before() {
+    final AgentSpan span = activeSpan();
+    if (span != null && VertxRedisClientDecorator.REDIS_COMMAND.equals(span.getOperationName())) {
+      return activateSpan(span, true);
```
ygree, line 15: I'm not sure why do we need this advice? RedisAPICallAdvice should already activate the command span scope.

ygree (approved):
Nice fix! The only thing I'm not sure is still needed is RedisConnectAdvice

dd-java-agent/instrumentation/vertx-redis-client-3.9/src/main/java/datadog/trace/instrumentation/vertx_redis_client/RedisConnectAdvice.java
amarziali: Actually when used in pooled mode `Redis#send` is calling itself `connect` in a lambda. Without activating in this way I do not have the span in the other advices. The solution I found is to activate here

PR #5043, opened 2023-04-10
Implement automatic CI Visibility configuration for Maven projects

nikita-tkachenko-datadog:
# What Does This Do
Updates Maven instrumentation so that CI Visibility configuration is done automatically:
- when parent build process is run with the tracer attached, forked Surefire/Failsafe processes are automatically configured to run with the same tracer
- when parent build process is run with the tracer attached, forked Javac processes are automatically configured to run with the DD Javac Plugin

# Motivation
Simplifying onboarding for CI Visibility customers


dd-java-agent/instrumentation/maven-3.2.1/src/main/java/datadog/trace/instrumentation/maven3/MavenPluginVersion.java
```diff
@@ -0,0 +1,40 @@
...
+      }
+      if (otherToken > thisToken) {
+        return false;
+      }
+    }
+    return tokens.length >= other.tokens.length;
+  }
+
+  static MavenPluginVersion from(String s) {
+    String[] stringTokens = DOT.split(s);
```
albertvaka, line 33: What's the difference between this and `s.split('.')`? Sorry if it's a newbie question but this seems more complicated for nothing.

dd-java-agent/instrumentation/maven-3.2.1/src/main/java/datadog/trace/instrumentation/maven3/MavenProjectConfigurator.java
```diff
@@ -0,0 +1,231 @@
...
+        projectDependencies.add(javacPluginClientDependency);
+      }
+
+      MavenPluginVersion mavenPluginVersion =
+          compilerPlugin.getVersion() != null
+              ? MavenPluginVersion.from(compilerPlugin.getVersion())
+              : MavenPluginVersion.UNKNOWN;
+
+      if (mavenPluginVersion.isLaterThanOrEqualTo(ANNOTATION_PROCESSOR_PATHS_SUPPORTED_VERSION)) {
+        String lombokVersion = getLombokVersion(projectDependencies);
```
albertvaka, line 126: Is the need for lombok on older Maven versions documented anywhere?

dd-java-agent/instrumentation/maven-3.2.1/src/main/java/datadog/trace/instrumentation/maven3/MavenProjectConfigurator.java
```diff
@@ -0,0 +1,231 @@
...
+      Properties systemProperties = System.getProperties();
+      for (Map.Entry<Object, Object> e : systemProperties.entrySet()) {
+        String propertyName = (String) e.getKey();
+        if (propertyName.startsWith(Config.PREFIX)) {
+          modifiedArgLine
+              .append("-D")
+              .append(propertyName)
+              .append('=')
+              .append(e.getValue())
+              .append(System.lineSeparator());
```
albertvaka, line 61: Can we split the arguments by newlines? Is that supported in all OSes? Shouldn't they be just split by spaces?

dd-java-agent/instrumentation/maven-3.2.1/src/main/java/datadog/trace/instrumentation/maven3/MavenProjectConfigurator.java
```diff
@@ -0,0 +1,231 @@
...
+      Integer ciVisibilityDebugPort = Config.get().getCiVisibilityDebugPort();
+      if (ciVisibilityDebugPort != null) {
+        modifiedArgLine
+            .append("-agentlib:jdwp=transport=dt_socket,server=y,suspend=y,address=")
+            .append(ciVisibilityDebugPort)
+            .append(System.lineSeparator());
+      }
+
+      File agentJar = AgentUtils.getAgentJar();
+      modifiedArgLine.append("-javaagent:").append(agentJar.toPath());
```
albertvaka, line 74: What happens if you have already configured this manually & we try to auto-configure it on top of that? You might end with two `-javaagent` arguments. Did you test that scenario? Same for the compiler plugin.

dd-java-agent/instrumentation/maven-3.2.1/src/main/java/datadog/trace/instrumentation/maven3/MavenProjectConfigurator.java
```diff
@@ -0,0 +1,231 @@
...
+    if (pluginCompilerId != null && !JAVAC_COMPILER_ID.equals(pluginCompilerId)) {
+      return;
+    }
+
+    for (PluginExecution execution : compilerPlugin.getExecutions()) {
+      Xpp3Dom configuration = (Xpp3Dom) execution.getConfiguration();
+
+      String compilerId = MavenUtils.getConfigurationValue(configuration, "compilerId");
+      if (compilerId != null && !JAVAC_COMPILER_ID.equals(compilerId)) {
+        // Javac plugin does not work with other compilers
```
albertvaka, line 106: Maybe we could keep a `javacFound` boolean and at the end of the loop if we didn't find any instance of it we could print an info/debug message saying the compiler plugin couldn't be added. Right now we continue silently and maybe the user is expecting the plugin to be added when it's not, and there's no way to know.

dd-java-agent/instrumentation/maven-3.2.1/src/main/java/datadog/trace/instrumentation/maven3/MavenProjectConfigurator.java
```diff
@@ -0,0 +1,231 @@
...
+      execution.setConfiguration(configuration);
+    }
+  }
+
+  // Dependency class does not override equals, so we have to do it like this
+  private static boolean contains(List<Dependency> projectDependencies, Dependency dependency) {
+    for (Dependency projectDependency : projectDependencies) {
+      if (projectDependency.getGroupId().equals(dependency.getGroupId())
+          && projectDependency.getArtifactId().equals(dependency.getArtifactId())
+          && projectDependency.getVersion().equals(dependency.getVersion())) {
```
albertvaka, line 166: Do we only want to consider it already has a dependency if it has the exact same version we want? What if it has a newer version?

albertvaka (commented):
Is there some automated testing we can do here?

internal-api/src/main/java/datadog/trace/util/AgentUtils.java
```diff
@@ -0,0 +1,25 @@
...
+import datadog.trace.api.Config;
+import java.io.File;
+import java.net.URI;
+import java.net.URISyntaxException;
+
+public abstract class AgentUtils {
+
+  private AgentUtils() {}
+
+  public static File getAgentJar() {
```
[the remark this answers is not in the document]
nikita-tkachenko-datadog: > Is there some automated testing we can do here?

Unit testing seems kind of pointless to me, as verifying that a number of methods was executed on a Maven domain object does not give much confidence in the end result.
Integration tests we cannot use, as autoconfig requires an actual agent distribution JAR to be present on the classpath.
It has to be either tracer smoke tests or our test environment.
I chose the latter since it was easier to implement.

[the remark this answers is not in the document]
nikita-tkachenko-datadog: True, moved the method to config

dd-java-agent/instrumentation/maven-3.2.1/src/main/java/datadog/trace/instrumentation/maven3/MavenPluginVersion.java
nikita-tkachenko-datadog: Avoiding `String.split()` is a habit from the old times: it accepts a regex, and providing it like this would cause it to compile on every call, which imposed significant performance overhead.
The issue was "fixed" a while ago - there's no a "shortcut" to avoid compiling the regex if it's amount to a literal.

Changed to `s.split("\\.")`

dd-java-agent/instrumentation/maven-3.2.1/src/main/java/datadog/trace/instrumentation/maven3/MavenProjectConfigurator.java
nikita-tkachenko-datadog: It's not like we're adding Lombok here, this is about ensuring Lombok does not break if the project is already using it.

In _newer_ Maven compiler plugin versions - the ones that support annotation processor paths - whenever we're adding DD Javac Plugin as annotation processor, Lombok (if it happens to be used in the project) stops working, unless we add it as an annotation processor as well.

This is an edge case that is documented in [DD Javac Plugin documentation](https://github.com/DataDog/dd-javac-plugin#using-the-plugin-with-lombok)

dd-java-agent/instrumentation/maven-3.2.1/src/main/java/datadog/trace/instrumentation/maven3/MavenProjectConfigurator.java
nikita-tkachenko-datadog: Now I have doubts :)

I always used to use newlines when configuring Maven manually, and the surefire plugin code contains logic that replaced all the whitespace character with the literal space.

But now I'm not sure about the older plugin versions, and the documentation says nothing about it.

Replaced it with space to be on the safe side.

dd-java-agent/instrumentation/maven-3.2.1/src/main/java/datadog/trace/instrumentation/maven3/MavenProjectConfigurator.java
nikita-tkachenko-datadog: Yes, I have verified it, since I assumed there can be customers who might just update to the new tracer versions without reading the docs.

The tracer contains a safeguard that protects it from initialising twice, and the compiler plugin does not break as well.

dd-java-agent/instrumentation/maven-3.2.1/src/main/java/datadog/trace/instrumentation/maven3/MavenProjectConfigurator.java
nikita-tkachenko-datadog: I am a bit hesitant here, not sure if it's appropriate to ask customers to change the compiler that they use. And if we don't ask, then the warning makes little sense

dd-java-agent/instrumentation/maven-3.2.1/src/main/java/datadog/trace/instrumentation/maven3/MavenProjectConfigurator.java
nikita-tkachenko-datadog: Good point, I have updated the code so that the project has the latest of the two versions.

dd-java-agent/instrumentation/maven-3.2.1/src/main/java/datadog/trace/instrumentation/maven3/MavenProjectConfigurator.java
albertvaka: There are a few more places where `System.lineSeparator()` is used. If it works we can leave it, but let's be consistent and use the same thing everywhere.

dd-java-agent/instrumentation/maven-3.2.1/src/main/java/datadog/trace/instrumentation/maven3/MavenProjectConfigurator.java
albertvaka: Ok

albertvaka (approved):
Left one last comment about `System.lineSeparator()`, other than that LGTM 🚀

dd-java-agent/instrumentation/maven-3.2.1/src/main/java/datadog/trace/instrumentation/maven3/MavenProjectConfigurator.java
nikita-tkachenko-datadog: Good catch, I missed those. Changed everything to using spaces now

PR #5061, opened 2023-04-13
[data streams] Default message timestamp to Kafka message timestamp

piochelepiotr:
# What Does This Do

# Motivation

# Additional Notes


dd-java-agent/agent-bootstrap/src/main/java/datadog/trace/bootstrap/instrumentation/decorator/HttpServerDecorator.java
```diff
@@ -131,7 +131,7 @@ public AgentSpan startSpan(REQUEST_CARRIER carrier, AgentSpan.Context.Extracted
     if (null != carrier && null != getter) {
       PathwayContext pathwayContext = propagate().extractPathwayContext(carrier, getter);
       span.mergePathwayContext(pathwayContext);
-      tracer().setDataStreamCheckpoint(span, SERVER_PATHWAY_EDGE_TAGS);
+      tracer().setDataStreamCheckpoint(span, SERVER_PATHWAY_EDGE_TAGS, null);
```
dougqh, line 134: Why did you decide to pass null instead of overload the method signature.
If you want to be explicit about emptiness, you might consider OptionalLong rather than Long.

Typically, I'd avoid Optional when it is boxing an object, but since Long is already a box around a long it might make sense here.

PR #5063, opened 2023-04-13
Introduce InstrumentationContext class

jpbempel:
# What Does This Do

# Motivation
https://github.com/DataDog/dd-trace-java/pull/5048#discussion_r1163283367

# Additional Notes

### DIFF
--- a/dd-java-agent/agent-debugger/src/main/java/com/datadog/debugger/agent/DebuggerTransformer.java
+++ b/dd-java-agent/agent-debugger/src/main/java/com/datadog/debugger/agent/DebuggerTransformer.java
@@ -3,6 +3,7 @@ package com.datadog.debugger.agent;
 import static java.util.Collections.singletonList;
 import static java.util.stream.Collectors.toList;
 
+import com.datadog.debugger.instrumentation.InstrumentationContext;
 import com.datadog.debugger.instrumentation.InstrumentationResult;
 import com.datadog.debugger.probe.LogProbe;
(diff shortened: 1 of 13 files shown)


dd-java-agent/agent-debugger/src/main/java/com/datadog/debugger/agent/DebuggerTransformer.java
```diff
@@ -419,18 +420,20 @@ private InstrumentationResult applyInstrumentation(
     InstrumentationResult.Status status =
         preCheckInstrumentation(diagnostics, classLoader, methodNode);
```
shatzi, line 421: Maybe InstrumentationContext can also be merged InstrumentationResult ? it only need Status.

Moreover, maybe we should put status inside InstrumentationContext and utility functions to addDiagnostic(DiagnosticMessage.Kind.ERROR, t), maybe that can also set the status to ERROR (but that might feel like dark magic)

which then we can also move preCheckInstrumentation into InsturmentationContext as well.

WDYT?

shatzi (approved):
it looks better IMHO

richardstartin:
I just noticed this, there is already an `InstrumentationContext` class and overloading its name like this might create confusion for people talking about the architecture, people will have to say "Instrumentation context as in the debugger" or "Instrumentation context as in field injection" or actually say the package names when talking about the software.

jpbempel:
will revisit this later

PR #5065, opened 2023-04-13
Avoid duplicate elements in TaintedMap

smola:
# What Does This Do
In TaintedMap, insert at the end of the chain, instead of at the head. This allows us to test for duplicates. This prevents the pathological case where the same object is tainted many times, making the linked list of a bucket have an unbound length and distorting the size estimate.

This should be, in theory, slower. But microbenchmarks do not show any noticeable difference. Perhaps because given our parameters, less than 0.2% of the chains are expect to be longer than 2 elements.

# Motivation

# Additional Notes


dd-java-agent/agent-iast/src/main/java/com/datadog/iast/taint/DefaultTaintedMap.java
```diff
@@ -107,30 +107,50 @@ public TaintedObject get(final @Nonnull Object key) {
...
-      table[index] = entry;
-      if ((entry.positiveHashCode & PURGE_MASK) == 0) {
-        // To mitigate the cost of maintaining an atomic counter, we only update the size every
-        // <PURGE_COUNT> puts. This is just an approximation, we rely on key's identity hash code as
-        // if it was a random number generator, and we assume duplicate keys are rarely inserted.
-        estimatedSize.addAndGet(PURGE_COUNT);
-        purge();
+      while (cur.next != null) {
+        if (cur.positiveHashCode == entry.positiveHashCode && cur.get() == entry.get()) {
+          // Duplicate, exit early.
```
manuel-alvarez-alvarez, line 141: I tried to think of a fair case where we might need to update the tainted vale present in the map, but couldn't figure out one so it's OK to me.

manuel-alvarez-alvarez (approved):

PR #5073, opened 2023-04-17
Add flag DD_TELEMETRY_METRICS_ENABLED disabled by default

manuel-alvarez-alvarez:
# What Does This Do
Adds a new flag to disable telemetry metrics by default.

# Motivation
Disables the telemetry payloads until the telemetry-intake is ready.

# Additional Notes

### DIFF
--- a/internal-api/src/main/java/datadog/trace/api/Config.java
+++ b/internal-api/src/main/java/datadog/trace/api/Config.java
@@ -65,6 +65,7 @@ import static datadog.trace.api.ConfigDefaults.DEFAULT_SERVLET_ROOT_CONTEXT_SERV
 import static datadog.trace.api.ConfigDefaults.DEFAULT_SITE;
 import static datadog.trace.api.ConfigDefaults.DEFAULT_TELEMETRY_DEPENDENCY_COLLECTION_ENABLED;
 import static datadog.trace.api.ConfigDefaults.DEFAULT_TELEMETRY_HEARTBEAT_INTERVAL;
+import static datadog.trace.api.ConfigDefaults.DEFAULT_TELEMETRY_METRICS_ENABLED;
 import static datadog.trace.api.ConfigDefaults.DEFAULT_TELEMETRY_METRICS_INTERVAL;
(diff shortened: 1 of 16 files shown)


internal-api/src/main/java/datadog/trace/api/MetricCollector.java
```diff
@@ -31,29 +31,47 @@ public static MetricCollector get() {
   private static final AtomicRequestCounter wafBlockedRequestCounter = new AtomicRequestCounter();
 
   public boolean wafInit(String wafVersion, String rulesVersion) {
+    if (!Config.get().isTelemetryMetricsEnabled()) {
+      return true;
+    }
```
PerfectSlayer, line 36: It feels like it would be better to create a noop collector if metrics are disabled rather than checking the config state for each call.

internal-api/src/main/java/datadog/trace/api/Config.java
```diff
@@ -1220,6 +1223,9 @@ private Config(final ConfigProvider configProvider, final InstrumenterConfig ins
             TELEMETRY_DEPENDENCY_COLLECTION_ENABLED,
             DEFAULT_TELEMETRY_DEPENDENCY_COLLECTION_ENABLED);
 
+    telemetryMetricsEnabled =
+        configProvider.getBoolean(TELEMETRY_METRICS_ENABLED, DEFAULT_TELEMETRY_METRICS_ENABLED);
```
PerfectSlayer, line 1227: Does it take care if the whole telemetry thing is enabled?

PerfectSlayer (changes requested):

internal-api/src/main/java/datadog/trace/api/MetricCollector.java
manuel-alvarez-alvarez: Hello! sorry for the late reply to the feedback :cry: , I should have closed this PR long ago as the requirement regarding telemetry is not here anymore.

Thanks for the feedback!

PR #5080, opened 2023-04-18
Exclude spotbugsLatestDepTest

smola:
# What Does This Do

# Motivation
This is almost redundant with spotbugsTest and takes a long time to run.
# Additional Notes


.circleci/config.yml
```diff
@@ -368,6 +368,7 @@ jobs:
             GRADLE_OPTS="-Dorg.gradle.jvmargs='-Xmx2G -Xms2G -XX:ErrorFile=/tmp/hs_err_pid%p.log -XX:+HeapDumpOnOutOfMemoryError -XX:HeapDumpPath=/tmp'"
             ./gradlew
             << parameters.gradleTarget >>
+            -x spotbugsLatestDepTest
```
bantonsson, line 371: I thought that the `spotbugs*` tasks were disabled for everything but `Main` and `Main_java11` in the `spotbugs.gradle` file. Maybe that configuration is broken?

smola: @bantonsson I still have to verify the fix, which might be incorrect. But something is pulling all compile tasks for latestDepTest. It might be `codenarcLatestDepTest` or it might be `spotbugsLatestDepTest` pulling required tasks even if it's disabled?

PR #5086, opened 2023-04-19
Add back version in snapshot for probe details

jpbempel:
# What Does This Do

# Motivation
was removed with recent refactorings

# Additional Notes


ojung (commented):
Ah just noticed: what about probe statuses? There we also need the `probeVersion` field. Was that removed in the refactoring as well?

jpbempel:
> Ah just noticed: what about probe statuses? There we also need the `probeVersion` field. Was that removed in the refactoring as well?

Did not touch that, here: https://github.com/DataDog/dd-trace-java/blob/master/dd-java-agent/agent-debugger/src/main/java/com/datadog/debugger/agent/ProbeStatus.java#L111

PR #5089, opened 2023-04-19
Support AWS Lambda parent inheritance

DarcyRaynerDD:
# What Does This Do

Fixes a bug with Lambda span generation, when using an abstract base class. 
Also fixes a bug where com.amazonaws.services.lambda.runtime.RequestStreamHandler was not being wrapped correctly, because it has a void return type, which the ExtensionAdvice class wasn't matching properly. I've added a test case for this.

Before:
Screenshot 2023-04-19 at 5 04 24 PM
After:
Screenshot 2023-04-19 at 4 52 58 PM

# Motivation

If a customer creates an Abstract Base Class, which implements the `com.amazonaws.services.lambda.runtime.RequestHandler`, currently our instrumentation won't correctly wrap the `handleRequest` method, leading to broken traces. This changes the instrumentation matcher to find all classes in the hierachy which implement the RequestHandler interface.

# Additional Notes

### DIFF
--- a/dd-java-agent/instrumentation/aws-lambda-handler/src/main/java/datadog/trace/instrumentation/aws/v1/lambda/LambdaHandlerInstrumentation.java
+++ b/dd-java-agent/instrumentation/aws-lambda-handler/src/main/java/datadog/trace/instrumentation/aws/v1/lambda/LambdaHandlerInstrumentation.java
@@ -10,6 +10,7 @@ import static net.bytebuddy.asm.Advice.OnMethodEnter;
 import static net.bytebuddy.asm.Advice.OnMethodExit;
 import static net.bytebuddy.asm.Advice.Origin;
 import static net.bytebuddy.asm.Advice.This;
+import static net.bytebuddy.implementation.bytecode.assign.Assigner.Typing.DYNAMIC;
 import static net.bytebuddy.matcher.ElementMatchers.isMethod;
 import static net.bytebuddy.matcher.ElementMatchers.takesArgument;
 
@@ -36,14 +37,14 @@ public class LambdaHandlerInstrumentation extends Instrumenter.Tracing
 
   @Override
   public String hierarchyMarkerType() {
-    return "com.amazonaws.services.lambda.runtime.RequestHandler";
+    return "com.amazonaws.services.lambda.runtime.RequestStreamHandler";
   }
 
   @Override
   public ElementMatcher<TypeDescription> hierarchyMatcher() {
     return implementsInterface(
         named(hierarchyMarkerType())
-            .or(named("com.amazonaws.services.lambda.runtime.RequestStreamHandler")));
+            .or(named("com.amazonaws.services.lambda.runtime.RequestHandler")));
   }
 
   @Override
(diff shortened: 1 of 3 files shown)


dd-java-agent/instrumentation/aws-lambda-handler/src/main/java/datadog/trace/instrumentation/aws/v1/lambda/LambdaHandlerInstrumentation.java
```diff
@@ -17,33 +18,30 @@
...
-      } else {
-        this.instrumentedType = handler.substring(0, split);
-        this.methodName = handler.substring(split + HANDLER_SEPARATOR.length());
-      }
-    }
+  }
+
+  @Override
+  public String hierarchyMarkerType() {
+    return "com.amazonaws.services.lambda.runtime.RequestHandler";
```
maxday, line 39: Is this breaking streaming support? Do we need to add
`com.amazonaws.services.lambda.runtime.RequestStreamHandler`, see Handler interfaces paragraph here: https://docs.aws.amazon.com/lambda/latest/dg/java-handler.html

DarcyRaynerDD: Good point, should be easy enough to fix. I'll add a test for it as well.

maxday (commented):
LGTM!

dd-java-agent/instrumentation/aws-lambda-handler/src/main/java/datadog/trace/instrumentation/aws/v1/lambda/LambdaHandlerInstrumentation.java
```diff
@@ -108,8 +100,9 @@ static AgentScope enter(
     static void exit(
         @Origin String method,
         @Enter final AgentScope scope,
-        @Advice.Return final Object result,
+        @Advice.Return(typing = DYNAMIC) final Object result,
```
DarcyRaynerDD, line 103: https://javadoc.io/static/net.bytebuddy/byte-buddy/1.10.5/net/bytebuddy/asm/Advice.Return.html


> Indicates that the annotated parameter should be mapped to the return value of the instrumented method. If the instrumented method terminates exceptionally, the type's default value is assigned to the parameter, i.e. 0 for numeric types and null for reference types. If the return type is void, the annotated value is null if and only if typing() is set to Assigner.Typing#DYNAMIC.

The "handleRequest" method on RequestStreamHandler has a void return type, which was causing this to silently fail to apply the advice for that type.

bantonsson: What am I missing here? Is there one method that returns something and one that doesn't? If there isn't then shouldn't the `@Advice.Return ...` just be removed?

DarcyRaynerDD: There are two interfaces, `RequestStreamHandler` and `RequestHandler`. On `RequestHandler` the handleRequest method returns a value, (the interface is generic), and `RequestStreamHandler` returns void. We want to capture the return value for the  RequestHandler case

bantonsson: Thanks @DarcyRaynerDD, makes perfect sense then.

bantonsson (approved):

bantonsson:
@DarcyRaynerDD could you please squash the commits when you merge the PR?

PR #5096, opened 2023-04-21
Improve OpenTelemetry instrumentation

PerfectSlayer:
# What Does This Do

Add empty string value support for span attribute,
Fix span context sampling flag,
Fix tests after RFC change.

# Motivation

Empty string value is a meaningful attribute value for OTel span.
So it should not be ignored nor cleared from span tags.

# Additional Notes

### DIFF
--- a/dd-trace-core/src/main/java/datadog/trace/core/DDSpan.java
+++ b/dd-trace-core/src/main/java/datadog/trace/core/DDSpan.java
@@ -327,13 +327,27 @@ public class DDSpan
 
   @Override
   public final DDSpan setTag(final String tag, final String value) {
-    context.setTag(tag, value);
+    if (value == null || value.isEmpty()) {
+      // Remove the tag
+      context.setTag(tag, null);
+    } else {
+      context.setTag(tag, value);
+    }
     return this;
   }
 
   @Override
   public final DDSpan setTag(final String tag, final boolean value) {
     context.setTag(tag, value);
+    return this;
+  }
+
+  @Override
+  public AgentSpan setAttribute(String key, Object value) {
+    if (key == null || key.isEmpty() || value == null) {
+      return this;
+    }
+    this.context.setTag(key, value);
     return this;
   }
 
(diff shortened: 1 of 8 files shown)


dd-trace-core/src/main/java/datadog/trace/core/DDSpan.java
```diff
@@ -327,7 +327,12 @@ public DDSpan addThrowable(final Throwable error) {
 
   @Override
   public final DDSpan setTag(final String tag, final String value) {
-    context.setTag(tag, value);
+    if (value == null || value.isEmpty()) {
+      // Remove the tag
```
bantonsson, line 331: I think that this does not cover all the cases. You could still in the old code call `setTag("foo", (Number)null)`, which would now have a different behavior. Can't all the `setTag` variants in `DDSpan`, call the `setTag(String, Object)` variant that does the checks and then calls `context.SetTag`?

bantonsson: Ignore me completely. It has the same behavior 🤦🏼‍♂️

dd-trace-core/src/main/java/datadog/trace/core/DDSpanContext.java
```diff
@@ -631,13 +631,18 @@ public void setMetric(final CharSequence key, final Number value) {
...
 
   /**
-   * Add a tag to the span. Tags are not propagated to the children
+   * Set a tag to the span. Tags are not propagated to the children.
    *
-   * @param tag the tag-name
-   * @param value the value of the tag. tags with null values are ignored.
+   * <ul>
+   *   <li>Setting a tag with a {@code null} value will remove the tag from the span.
+   *   <li>Setting a tag with a {@code null} value will remove the tag from the span.
```
bantonsson, line 638: Duplicate lines.

bantonsson (approved):
👍🏻 Only nitpicks.

bantonsson (approved):

PR #5098, opened 2023-04-21
Refactor Metric instrumentation

jpbempel:
# What Does This Do
use the expression visitor to emit byte code instructions and replace the valueRefResolver that was used before but was limiting the scope of values.

# Motivation
It will help to support for feature form the Expression Language like count/len on map/list/set, or array indexing

# Additional Notes

### DIFF
--- a/dd-java-agent/agent-debugger/src/main/java/com/datadog/debugger/instrumentation/MetricInstrumentor.java
+++ b/dd-java-agent/agent-debugger/src/main/java/com/datadog/debugger/instrumentation/MetricInstrumentor.java
@@ -3,15 +3,39 @@ package com.datadog.debugger.instrumentation;
 import static com.datadog.debugger.instrumentation.Types.*;
 
 import com.datadog.debugger.el.InvalidValueException;
-import com.datadog.debugger.el.Literal;
-import com.datadog.debugger.el.Value;
+import com.datadog.debugger.el.Visitor;
+import com.datadog.debugger.el.expressions.BinaryExpression;
+import com.datadog.debugger.el.expressions.BinaryOperator;
+import com.datadog.debugger.el.expressions.BooleanExpression;
+import com.datadog.debugger.el.expressions.ComparisonExpression;
+import com.datadog.debugger.el.expressions.ComparisonOperator;
+import com.datadog.debugger.el.expressions.ContainsExpression;
+import com.datadog.debugger.el.expressions.EndsWithExpression;
+import com.datadog.debugger.el.expressions.FilterCollectionExpression;
+import com.datadog.debugger.el.expressions.GetMemberExpression;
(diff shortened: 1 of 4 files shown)


dd-java-agent/agent-debugger/src/main/java/com/datadog/debugger/instrumentation/MetricInstrumentor.java
```diff
@@ -235,112 +237,223 @@ private boolean isCompatible(Type argType, Type expectedType) {
...
-      if (target == Values.THIS_OBJECT) {
-        insnList = new InsnList();
-        currentType = tryRetrieveField(name, insnList);
-        convertIfRequired(currentType, expectedType, insnList);
-      } else if (target instanceof ResolverResult) {
-        ResolverResult result = (ResolverResult) target;
-        insnList = result.insnList;
-        currentType = followReferences(result.type, name, insnList);
+    public VisitorResult visit(GetMemberExpression getMemberExpression) {
+      VisitorResult result = getMemberExpression.getTarget().accept(this);
```
shatzi, line 370: Results can be null - should we throw unsupported expression instead of nulls? that would make better error message and also avoid null checks. WDYT?

dd-java-agent/agent-debugger/src/main/java/com/datadog/debugger/instrumentation/MetricInstrumentor.java
```diff
@@ -235,112 +237,223 @@ private boolean isCompatible(Type argType, Type expectedType) {
...
+    final Type type;
+    final InsnList insnList;
+
+    public VisitorResult(Type type, InsnList insnList) {
+      this.type = type;
+      this.insnList = insnList;
+    }
+  }
+
+  private static class MetricValueVisitor implements Visitor<VisitorResult> {
```
shatzi, line 250: Lovely.

Not sure if that will be nicer. 
We can created DefaultVisitor with a visitDefault method that can be overwritten.

Then this call can override the default with throw unsupported exception +  override only supported nodes

That can make the code import less classes and make it more readable.

shatzi (approved):
So cool

dd-java-agent/agent-debugger/src/main/java/com/datadog/debugger/instrumentation/MetricInstrumentor.java
jpbempel: long term we will fill probably all the methods when we will support the full EL.

dd-java-agent/agent-debugger/src/main/java/com/datadog/debugger/instrumentation/MetricInstrumentor.java
jpbempel: done

PR #5106, opened 2023-04-24
Add support for OpenTelemetry programmatic context propagation

PerfectSlayer:
# What Does This Do

This PR brings the support of OpenTelemetry ContextPropagator API.
It will allow to inject and extract context using OpenTelemetry API.

# Motivation

Improve the support of OpenTelemetry Tracing APIs.

# Additional Notes

### DIFF
--- a/dd-java-agent/instrumentation/opentelemetry/opentelemetry-1.4/src/main/java/datadog/trace/instrumentation/opentelemetry14/OtelExtractedContext.java
+++ b/dd-java-agent/instrumentation/opentelemetry/opentelemetry-1.4/src/main/java/datadog/trace/instrumentation/opentelemetry14/OtelExtractedContext.java
@@ -50,7 +50,7 @@ class OtelExtractedContext implements AgentSpan.Context {
 
   @Override
   public AgentTrace getTrace() {
-    return null;
+    return AgentTracer.NoopAgentTrace.INSTANCE;
   }
 
   @Override
--- a/dd-java-agent/instrumentation/opentelemetry/opentelemetry-1.4/src/main/java/datadog/trace/instrumentation/opentelemetry14/OtelExtractedContext.java
+++ b/dd-java-agent/instrumentation/opentelemetry/opentelemetry-1.4/src/main/java/datadog/trace/instrumentation/opentelemetry14/OtelExtractedContext.java
@@ -1,0 +1,70 @@
+package datadog.trace.instrumentation.opentelemetry14;
+
+import datadog.trace.api.DDSpanId;
+import datadog.trace.api.DDTraceId;
+import datadog.trace.api.sampling.PrioritySampling;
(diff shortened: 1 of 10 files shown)


dd-java-agent/instrumentation/opentelemetry/opentelemetry-1.4/src/main/java/datadog/trace/instrumentation/opentelemetry14/OtelExtractedContext.java
```diff
@@ -0,0 +1,70 @@
...
+class OtelExtractedContext implements AgentSpan.Context {
+  private final DDTraceId traceId;
+  private final long spanId;
+  private final int prioritySampling;
+
+  private OtelExtractedContext(SpanContext context) {
+    this.traceId = DDTraceId.fromHex(context.getTraceId());
+    this.spanId = DDSpanId.fromHex(context.getSpanId());
+    this.prioritySampling =
+        context.isSampled() ? PrioritySampling.SAMPLER_KEEP : PrioritySampling.UNSET;
```
bantonsson, line 24: Is it always ok to interpret not sampling as unset? I mean can't OTel communicate _drop_ in some way?

PerfectSlayer: Not that I am aware of. Spans have a `sampled` flag, accessible from their [SpanContext](https://opentelemetry.io/docs/reference/specification/trace/api/#spancontext):

> TraceFlags contain details about the trace. Unlike TraceState values, TraceFlags are present in all traces. The current version of the specification only supports a single flag called [sampled](https://www.w3.org/TR/trace-context/#sampled-flag).

And the [OTel Sampling documentation](https://opentelemetry.io/docs/concepts/sampling/#terminology) only lists two states without distinctions for either _being dropped_ nor _not defined_:

> It’s important to use consistent terminology when discussing sampling. A trace or span is considered “sampled” or “not sampled”:
> * Sampled: A trace or span is processed and exported. Because it is chosen by the sampler as a representive of the population, it is considered “sampled”.
> * Not sampled: A trace or span is not processed or exported. Because it is not chosen by the sampler, it is considered “not sampled”.

About other context extractors, `UNSET` seems to be the default value for not defined sampling priority (except for `Haystack` that uses `KEEP`).

bantonsson: Yeah, it's a deficiency of the OTel hard head sampling decisions. Let's just interpret it as `UNSET` for now.

bantonsson (approved):

PR #5113, opened 2023-04-25
Improvements for WAF telemetry metrics

smola:
# What Does This Do

Improvements for WAF telemetry metrics

* Prefix `MetricCollector` and related classes with `Waf`. These are currently strongly coupled to WAF metrics.
* Add missing tags for waf.updates and waf.requests.
* Whenever there is a rules update, flush waf.requests, since further requests need a new event_rules_version tag.
* Do not double count blocked requests.

# Motivation

Get our WAF telemetry up to date with the spec.

# Additional Notes

It might be a good idea to unify some of the code to collect IAST and WAF telemetry metrics, but that's out of scope for this PR.


PerfectSlayer (approved):
Weird that git fails to detect file rename 🤷

ValentinZakharov (approved):

PR #5115, opened 2023-04-25
Switch reliability env deploys to be scheduled

randomanderson:
# What Does This Do
Changes reliability env deployments to only happen when scheduled or manually triggered

# Motivation
Deploy on master commit is too noisy. Switching to a scheduled deploy should help


smola (approved):
@randomanderson What will be the initial schedule? Nightly?

PR #5125, opened 2023-04-26
Fix timestamp units in WAF telemetry metrics

smola:
# What Does This Do
Convert timestamps in WAF telemetry metric points from milliseconds to seconds.

# Motivation
Timestamps for telemetry metrics are defined in seconds (as a double). We can just encode it as an integer, since the sub-second resolution seems to be pointless here.

# Additional Notes

### DIFF
--- a/internal-api/src/main/java/datadog/trace/api/WafMetricCollector.java
+++ b/internal-api/src/main/java/datadog/trace/api/WafMetricCollector.java
@@ -129,7 +129,7 @@ public class WafMetricCollector {
 
     public WafMetric(String metricName, long counter, final String... tags) {
       this.metricName = metricName;
-      this.timestamp = System.currentTimeMillis();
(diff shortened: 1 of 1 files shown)


internal-api/src/main/java/datadog/trace/api/WafMetricCollector.java
```diff
@@ -129,7 +129,7 @@ public abstract static class WafMetric {
 
     public WafMetric(String metricName, long counter, final String... tags) {
       this.metricName = metricName;
-      this.timestamp = System.currentTimeMillis();
+      this.timestamp = System.currentTimeMillis() / 1000;
```
manuel-alvarez-alvarez, line 132: Same thing happened to me while building IAST metrics :facepalm:

manuel-alvarez-alvarez (approved):

PR #5128, opened 2023-04-27
Safely parse URL and respect http server status code error rule

amarziali:
# What Does This Do

This PR is doing a couple of things:
* Safely parse the URL. If the URL is un parsable, when possible, tag the span with the raw uri (and not the parsed one since it will be null or broken)
* For tomcat, make sure that when an exception is captured, the span is marked as errored only if the http response code matches the related rule

# Motivation

tldr; on recent tomcat versions backslashes are not allowed (it's configurable with `org.apache.catalina.connector.CoyoteAdapter.ALLOW_BACKSLASH=true`) while before it was.

This parsing error is captured because reported among the servlet response parameters. While we should log the stacktrace on the span, we do not should (unless configured) mark the span as error because we have a 4XX error

# Additional Notes

### DIFF
--- a/dd-java-agent/instrumentation/apache-httpasyncclient-4/src/main/java/datadog/trace/instrumentation/apachehttpasyncclient/ApacheHttpAsyncClientDecorator.java
+++ b/dd-java-agent/instrumentation/apache-httpasyncclient-4/src/main/java/datadog/trace/instrumentation/apachehttpasyncclient/ApacheHttpAsyncClientDecorator.java
@@ -1,5 +1,6 @@
 package datadog.trace.instrumentation.apachehttpasyncclient;
 
+import datadog.trace.bootstrap.instrumentation.api.URIUtils;
 import datadog.trace.bootstrap.instrumentation.api.UTF8BytesString;
 import datadog.trace.bootstrap.instrumentation.decorator.HttpClientDecorator;
 import java.net.URI;
@@ -52,7 +53,7 @@ public class ApacheHttpAsyncClientDecorator extends HttpClientDecorator<HttpRequ
       return ((HttpUriRequest) request).getURI();
     } else {
       final RequestLine requestLine = request.getRequestLine();
-      return requestLine == null ? null : new URI(requestLine.getUri());
+      return requestLine == null ? null : URIUtils.safeParse(requestLine.getUri());
     }
   }
 
--- a/dd-java-agent/instrumentation/netty-4.0/src/main/java/datadog/trace/instrumentation/netty40/server/NettyHttpServerDecorator.java
+++ b/dd-java-agent/instrumentation/netty-4.0/src/main/java/datadog/trace/instrumentation/netty40/server/NettyHttpServerDecorator.java
@@ -7,6 +7,7 @@ import datadog.trace.api.gateway.BlockResponseFunction;
 import datadog.trace.bootstrap.instrumentation.api.AgentPropagation;
 import datadog.trace.bootstrap.instrumentation.api.ContextVisitors;
 import datadog.trace.bootstrap.instrumentation.api.URIDataAdapter;
+import datadog.trace.bootstrap.instrumentation.api.URIDataAdapterBase;
 import datadog.trace.bootstrap.instrumentation.api.URIDefaultDataAdapter;
 import datadog.trace.bootstrap.instrumentation.api.UTF8BytesString;
 import datadog.trace.bootstrap.instrumentation.decorator.HttpServerDecorator;
@@ -20,7 +21,6 @@ import io.netty.handler.codec.http.HttpRequest;
 import io.netty.handler.codec.http.HttpResponse;
 import java.net.InetSocketAddress;
 import java.net.SocketAddress;
-import java.net.URI;
 import java.util.Map;
 import org.slf4j.Logger;
 import org.slf4j.LoggerFactory;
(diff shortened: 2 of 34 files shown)


dd-java-agent/instrumentation/apache-httpasyncclient-4/src/main/java/datadog/trace/instrumentation/apachehttpasyncclient/ApacheHttpAsyncClientDecorator.java
```diff
@@ -52,7 +53,7 @@ protected URI url(final HttpRequest request) throws URISyntaxException {
       return ((HttpUriRequest) request).getURI();
     } else {
       final RequestLine requestLine = request.getRequestLine();
-      return requestLine == null ? null : new URI(requestLine.getUri());
+      return URIUtils.safeParse(requestLine.getUri());
```
ygree, line 56: There used to have a null check for requestLine but not anymore. I wonder if requestLine can be ever null?

dd-java-agent/instrumentation/netty-4.0/src/main/java/datadog/trace/instrumentation/netty40/server/NettyHttpServerDecorator.java
```diff
@@ -66,13 +66,17 @@ protected String method(final HttpRequest httpRequest) {
 
   @Override
   protected URIDataAdapter url(final HttpRequest request) {
-    final URI uri = URI.create(request.getUri());
-    if ((uri.getHost() == null || uri.getHost().equals("")) && request.headers().contains(HOST)) {
-      return new URIDefaultDataAdapter(
-          URI.create("http://" + request.headers().get(HOST) + request.getUri()));
-    } else {
-      return new URIDefaultDataAdapter(uri);
-    }
```
ygree, line 75: It's a minor, but for curiosity sake I wonder if it's really worth using lambdas here when the difference is at most just one line of code. Without passing a lambda it would be almost the same from the reading complexity perspective.

```java
    final String uri = request.getUri();
    final URI parsed = URIUtils.safeParse(uri);
    if (parsed != null) {
          if ((uri.getHost() == null || uri.getHost().equals(""))
              && request.headers().contains(HOST)) {
            return URIDataAdapterBase.fromURI(
                "http://" + request.headers().get(HOST) + request.getUri(),
                URIDefaultDataAdapter::new);
          }
          return new URIDefaultDataAdapter(uri);
    }
    return new UnparseableURIDataAdapter(uri);
```

ygree: It would open opportunity to remove the mapper parameter completely and just use URIDefaultDataAdapter's constructor within URIDataAdapterBase.fromURI for simple use-cases.

dd-java-agent/testing/src/main/groovy/datadog/trace/agent/test/base/HttpServerTest.groovy
```diff
@@ -1493,6 +1497,23 @@ abstract class HttpServerTest<SERVER> extends WithHttpServer<SERVER> {
     }
   }
 
+  def "test bad url not cause span marked as error"() {
```
ygree, line 1500: Nice to have it covered with a test.

internal-api/src/main/java/datadog/trace/bootstrap/instrumentation/api/URIDataAdapterBase.java
```diff
@@ -32,4 +34,17 @@ public String raw() {
     }
     return raw;
   }
+
+  @Override
+  public boolean isValid() {
+    return true;
+  }
+
+  public static URIDataAdapter fromURI(String uri, Function<URI, URIDataAdapter> mapper) {
```
ygree, line 43: As I mentioned before, I'm not completely sure if the mapper parameter is worth it. With or without it it's the same amount of boilerplate for the custom use-cases, and is more lightweight for simple cases.

internal-api/src/test/groovy/datadog/trace/bootstrap/instrumentation/api/URIDataAdapterTest.groovy
```diff
@@ -156,4 +156,25 @@ class URINoRawDataAdapterTest extends URIDataAdapterTest {
       return null
     }
   }
+
+  static class UnparseableURIAdapterTest extends DDSpecification {
```
ygree, line 160: Nice

ygree (approved):
Looks good to me. Just not sure about worthiness of the mapper parameter that takes a lambda.

dd-java-agent/instrumentation/apache-httpasyncclient-4/src/main/java/datadog/trace/instrumentation/apachehttpasyncclient/ApacheHttpAsyncClientDecorator.java
amarziali: good observation. I'm protecting with a null check thanks

internal-api/src/main/java/datadog/trace/bootstrap/instrumentation/api/URIDataAdapterBase.java
amarziali: I agree with you. I leverage the mapping in the netty decorator, where the parsed uri is used, eventually, with host coming from the http headers to fix the url. So having a mapping function saves me some if and some other boilerplate to parse the URI. I may think about a better way to do it without having a mapper fuction and improve in a next PR

PR #5130, opened 2023-04-27
Implement automatic CI Visibility configuration for Gradle projects

nikita-tkachenko-datadog:
# What Does This Do
Updates Gradle instrumentation so that CI Visibility configuration is done automatically:
- when parent build process is run with the tracer attached, forked Test processes are automatically configured to run with the same tracer
- when parent build process is run with the tracer attached, forked Javac processes are automatically configured to run with the DD Javac Plugin

# Motivation
Simplifying onboarding for CI Visibility customers


dd-java-agent/instrumentation/gradle/src/main/groovy/datadog/trace/instrumentation/gradle/GradleProjectConfigurator.groovy
```diff
@@ -0,0 +1,140 @@
...
+      for (def e : systemProperties.entrySet()) {
+        if (e.key.startsWith(Config.PREFIX)) {
+          jvmArgs.add("-D" + e.key + '=' + e.value)
+        }
+      }
+
+      def ciVisibilityDebugPort = Config.get().ciVisibilityDebugPort
+      if (ciVisibilityDebugPort != null) {
+        jvmArgs.add(
+          "-agentlib:jdwp=transport=dt_socket,server=y,suspend=y,address="
```
smola, line 57: I wonder how this interacts when gradle runs with `--debug-jvm` or anything that adds a debugger. Is it worth it to detect it? Or CI Visibility is not expected to run in this kind of scenario?

nikita-tkachenko-datadog: If the two are used together, Gradle will ignore `--debug-jvm` saying `Debug configuration ignored in favor of the supplied JVM arguments`.

I added this property mostly for internal use. It is a bit more convenient than the standard one, since I can run `gradle build` and any task that extends from `Test` will be debugged. Whereas with `--debug-jvm` I need to explicitly list the names of the tasks

smola (approved):

PR #5131, opened 2023-04-28
create span for http 400 at parsing time for undertow >=2.2

amarziali:
# What Does This Do

On undertow 2.2+ we do not create any span if the request parsing fails. This because the instrumentation happens at dispatch time and the parsing is before. In case of failure there is no dispatch hence no span created.

This did not happen on undertow 2.0

In order to close the gap, this PR is instrumenting the parsing handler in order to catch as much info as we can at that time before the response is written on the network socket

# Motivation

# Additional Notes


ygree (approved):
Good observation! Thanks for the fix!

PR #5132, opened 2023-04-28
Fix Git info extraction logic to fall back to alternative providers when encountering empty or blank strings

nikita-tkachenko-datadog:
# What Does This Do
There is a component in the tracer that handles Git metadata (repository URL, commit sha, commit author email, etc).
This component uses a variety of sources (env variables set by CI provider, local `.git` folder, user-supplied values, etc).
If a particular value is not available in one source, the component falls back to alternative sources.
This PR changes the notion of "not available": not only `null` strings, but also empty or blank strings will now result in fallback.

# Motivation
Some CI providers (e.g. Azure Pipelines) sometimes set Git-related environment variables to empty strings.

### DIFF
--- a/internal-api/src/main/java/datadog/trace/util/Strings.java
+++ b/internal-api/src/main/java/datadog/trace/util/Strings.java
@@ -282,4 +282,21 @@ public final class Strings {
     json.append("]");
     return json.toString();
   }
+
+  public static boolean isNotBlank(String s) {
+    if (s == null || s.isEmpty()) {
+      return false;
+    }
+
+    final int length = s.length();
+    for (int offset = 0; offset < length; ) {
+      final int codepoint = s.codePointAt(offset);
+      if (!Character.isWhitespace(codepoint)) {
+        return true;
+      }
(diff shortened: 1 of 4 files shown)


internal-api/src/main/java/datadog/trace/util/Strings.java
```diff
@@ -282,4 +282,21 @@ public static String toJson(final Iterable<String> items) {
     json.append("]");
     return json.toString();
   }
+
+  public static boolean isNotBlank(String s) {
+    if (s == null || s.isEmpty()) {
+      return false;
+    }
+
+    final int length = s.length();
```
ygree, line 291: A comment explaining what this block does would be helpful. I suspect it's something that is not covered by `isEmpty` but not sure what exactly it is.

nikita-tkachenko-datadog: The code checks that a string is not blank, i.e. has at least one non-whitespace character. Added a comment with a brief explanation

smola (approved):

PR #5133, opened 2023-04-28
Migrate native-image smoke tests from 2.7 beta to Spring Boot 3

PerfectSlayer:
# What Does This Do

Migrate `native-image` smoke tests from 2.7 beta to Spring Boot 3

# Motivation

Spring Native project is no more support and never went GA.
It was release as part of Spring Boot 3.
The `native-image` smoke tests will now be based on Spring Boot 3 `native-image` support.

# Additional Notes

cc @mcculls 
I discovered you added `@Trace` and `DD_TRACE_METHOD` tests. Nice addition 👍


dd-smoke-tests/spring-boot-3.0-native/application/build.gradle
```diff
@@ -0,0 +1,42 @@
...
+apply from: "$sharedConfigDirectory/repositories.gradle"
+apply from: "$sharedConfigDirectory/spotless.gradle"
+
+if (hasProperty('appBuildDir')) {
+  buildDir = property('appBuildDir')
+}
+
+// Required for 'nativeCompile' goal
+group = 'smoke-tests'
+group = 'spring-boot-native'
```
ygree, line 22: Interesting, group defined twice. How does it work?

dd-smoke-tests/spring-boot-3.0-native/application/.gitignore
```diff
@@ -0,0 +1,6 @@
+# Ignore all project specific gradle directories/files
+.gradle
```
ygree, line 2: Could these ignores be defined for all projects inside `dd-smoke-tests`?

ygree (approved):
LGTM! Just a couple of questions.

dd-smoke-tests/spring-boot-3.0-native/application/.gitignore
PerfectSlayer: I guess each _smoke test applications_ tends to be the closest to any regular Spring app. It tries to not depend of the global project.
The only changes from the default String starter is the `build.gradle` to add instrumentation. 
I would be in favor of leaving as it is if it is okay with you?

dd-smoke-tests/spring-boot-3.0-native/application/build.gradle
PerfectSlayer: I got some errors during the development process but it can't trigger them anymore since everything is in place.
I just removed this part then. Good catch 👍

PR #5134, opened 2023-05-01
Fix automatic demo application deploy on release

shurivich:
# What Does This Do
Currently, the demo applications deployment to the backend is not triggered when a new version is released. 
This is because the deployment is only triggered for the default branch, and the new version is committed to the release branch. 
The current PR fix this issue by adding a trigger from the release branch to ensure that the deployment is also triggered when a new version is released.


ojung (approved):
Thanks 👍

PR #5145, opened 2023-05-03
[Spark] Use the name of the last stage as the job name

paul-laffon-dd:
# What Does This Do

Currently, the name of **first** stage is used as the job name. This PR changes that to use the name of the last stage instead, to match how it is currently displayed in the spark UI


raphaelgavache (approved):

dd-java-agent/instrumentation/spark/src/main/java/datadog/trace/instrumentation/spark/DatadogSparkListener.java
```diff
@@ -140,6 +139,11 @@ public void onJobStart(SparkListenerJobStart jobStart) {
     AgentSpan jobSpan = jobSpanBuilder.start();
     jobSpan.setMeasured(true);
 
+    if (jobStart.stageInfos().nonEmpty()) {
```
bantonsson, line 142: If you move this up before `jobSpanBuilder.start()`, you can apply it to the `jobSpanBuilder` instead, or better yet move it up to the original place of setting the `RESOURCE_NAME`.

paul-laffon-dd: Moved it up, to apply it to `jobSpanBuilder`

dd-java-agent/instrumentation/spark/src/main/java/datadog/trace/instrumentation/spark/DatadogSparkListener.java
```diff
@@ -109,7 +109,6 @@ public void onJobStart(SparkListenerJobStart jobStart) {
             .withStartTimestamp(jobStart.time() * 1000)
             .withTag("job_id", jobStart.jobId())
             .withTag("stage_count", jobStart.stageInfos().size())
-            .withTag(DDTags.RESOURCE_NAME, jobStart.stageInfos().apply(0).name())
```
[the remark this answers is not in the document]
paul-laffon-dd: In fact this code might node be 100% safe, in some rare case it seems possible that a spark job has no stages, so it should be safer to have a condition checking it

bantonsson (approved):

bantonsson:
@paul-laffon-dd could you please squash this PR when merging?

PR #5146, opened 2023-05-03
Per test code coverage in CI Visibility

nikita-tkachenko-datadog:
# What Does This Do
Adds per-test code coverage support for CI Visibility product: code coverage info is collected separately for every test case.

# Motivation
Per-test code coverage data is needed to support CI Visibility Intelligent Test Runner.

# Additional Notes
Code coverage is implemented on top of Jacoco, by instrumenting Jacoco classes and injecting bytecode that inserts additional test probes.


internal-api/src/main/java/datadog/trace/api/civisibility/coverage/TestReportFileEntry.java
```diff
@@ -0,0 +1,99 @@
...
+    public int getEndColumn() {
+      return endColumn;
+    }
+
+    public int getNumberOfExecutions() {
+      return numberOfExecutions;
+    }
+
+    @Override
+    public boolean equals(Object obj) {
```
smola, line 71: Is this `equals` implementation really needed?

smola (approved):

PR #5147, opened 2023-05-03
Update ASM blocking templates

smola:
# What Does This Do
* Latest JSON version: now it is minified.
* Latest HTML version: now it is fully ASCII and minified.

# Motivation

# Additional Notes
Depends on https://github.com/DataDog/system-tests/pull/1142


cataphract (approved):

dd-java-agent/agent-bootstrap/src/main/java/datadog/trace/bootstrap/blocking/BlockingActionHelper.java
```diff
@@ -26,7 +26,7 @@ public class BlockingActionHelper {
   private static volatile byte[] TEMPLATE_HTML;
   private static volatile byte[] TEMPLATE_JSON;
 
-  public static final String CONTENT_TYPE_HTML = "text/html;charset=utf-8";
+  public static final String CONTENT_TYPE_HTML = "text/html; charset=utf-8";
```
cataphract, line 29: This form is not the preferred one, see https://www.rfc-editor.org/rfc/rfc9110#media.type (earlier specs seem to have disallowed the space). Also note that some servers reformat the content-type header, which is why the tests for blocking match against a regular expression: `/(?i)\Aapplication\/json(?:;\s?charset=utf-8)?\z/`

PR #5149, opened 2023-05-03
integrate APM Test Agent within CircleCI instrumentation tests

wconti27:
**What does this PR do?**
This PR adds the APM Test Agent to run alongside integration/plugin tests within CI testing, handling all produced traces during a test run and asserting with any enabled checks. This change was made by modifying the AgentTestRunner to have to contain an additional DDAgentWriter instance, only if the `CI_USE_TEST_AGENT` env variable is set. After test assertions are complete, traces are written from the original ListWriter used for testing to the DDAgentWriter, then flushed to the Test-Agent. 

NOTE: Traces are written and flushed to the DDAgentWriter after all test assertions, as opposed to using the MultiWriter class as it would result in trace / span processing, causing instrumentation tests to fail.

 The APM Test Agent is being added to every DD APM Tracer with the primary purpose of providing a centralized location for writing tests which will subsequently be used for asserting on handled APM traces, APM stats, and/or APM Telemetry. This strategy will help bring consistency to Datadog APM Tagging/ integration support.  [More information on the APM Test Agent is available here.](https://github.com/DataDog/dd-apm-test-agent)

**Motivation**
This testing strategy will be used by Service Naming, Unified Naming Conventions tagging, etc as a centralized location for asserting on integration traces. It allows us to leverage traces produced from integrations during testing, and assert consistently across APM tracers.

**How to test the change?**
Verify via logs that the Test-Agent is receiving traces from instrumentation tests [here](https://app.circleci.com/pipelines/github/DataDog/dd-trace-java/25235/workflows/62d91df2-7241-4366-be47-bc52093fbf82/jobs/778542?invite=true#step-2-974).
Verify via logs that the Test-Agent is asserting on traces with Trace Checks [here](https://app.circleci.com/pipelines/github/DataDog/dd-trace-java/25235/workflows/62d91df2-7241-4366-be47-bc52093fbf82/jobs/778542?invite=true#step-118-26).


dd-java-agent/testing/src/main/groovy/datadog/trace/agent/test/AgentTestRunner.groovy
```diff
@@ -188,6 +214,13 @@ abstract class AgentTestRunner extends DDSpecification implements AgentBuilder.L
       TEST_DATA_STREAMS_MONITORING = new DefaultDataStreamsMonitoring(sink, features, SystemTimeSource.INSTANCE, wellKnownTags, TEST_DATA_STREAMS_WRITER, bucketDuration)
     }
     TEST_WRITER = new ListWriter()
+
+    USE_TEST_AGENT_WRITER = System.getenv("CI_USE_TEST_AGENT").equals("true") && !(System.getProperty("CI_USE_TEST_AGENT").equals("false"))
```
amarziali, line 218: checking == true implies checking != false -> only one is needed

dd-java-agent/testing/src/main/groovy/datadog/trace/agent/test/AgentTestRunner.groovy
```diff
@@ -82,6 +84,22 @@ abstract class AgentTestRunner extends DDSpecification implements AgentBuilder.L
     configureLoggingLevels()
   }
 
+  static void addEnvironmentVariablesToHeaders(DDAgentWriter agentWriter) {
```
amarziali, line 87: in `DDSpecification` there is a check to fail the test if any of the system environment variable starts with `DD_`
So I think that you do not have any chance to find such env variable hence this method is not required

dd-trace-core/src/main/java/datadog/trace/common/writer/ddagent/DDAgentApi.java
```diff
@@ -155,4 +155,8 @@ private void handleAgentChange(String state) {
   protected Logger getLogger() {
     return log;
   }
+
+  public void setHeader(String k, String v) {
```
amarziali, line 159: if you remove the need to pass env vars this is not required as well

dd-java-agent/testing/src/main/groovy/datadog/trace/agent/test/AgentTestRunner.groovy
```diff
@@ -91,6 +109,14 @@ abstract class AgentTestRunner extends DDSpecification implements AgentBuilder.L
   @Shared
   ListWriter TEST_WRITER
 
+  @SuppressWarnings('PropertyName')
+  @Shared
+  DDAgentWriter TEST_AGENT_WRITER
+
+  @SuppressWarnings('PropertyName')
+  @Shared
+  Boolean USE_TEST_AGENT_WRITER
```
amarziali, line 118: you can use `boolean` primitive also if you set always

dd-java-agent/instrumentation/okhttp-3/src/test/groovy/OkHttp3Test.groovy
```diff
@@ -15,6 +15,7 @@ abstract class OkHttp3Test extends HttpClientTest {
     // disable tracer metrics because it uses OkHttp and class loading is
     // not isolated in tests
     injectSysConfig("dd.trace.tracer.metrics.enabled", "false")
+    System.setProperty("CI_USE_TEST_AGENT", "false")
```
amarziali, line 18: I think you want rather to set an env variable and not a system property

dd-java-agent/instrumentation/okhttp-3/src/test/groovy/IastOkHttp3InstrumentationTest.groovy
```diff
@@ -17,6 +17,7 @@ class IastOkHttp3InstrumentationTest extends AgentTestRunner {
     // HttpUrl gets loaded early so we have to disable the advice transformer
     IastHttpUrlInstrumentation.DISABLE_ADVICE_TRANSFORMER = true
     injectSysConfig('dd.iast.enabled', 'true')
+    System.setProperty("CI_USE_TEST_AGENT", "false")
```
amarziali, line 20: same you should set a env variable and not a system property.

Instead of doing it in this way you can create something like this in `AgentTestRunner`

```
boolean shouldRunWithTestAgent() {
return Boolean.parseBoolean(System.getEnv("CI_USE_TEST_AGENT"));
}
```

Then you can override the method on `OkHttp3Test` and in this test to return directly false

dd-java-agent/testing/src/main/groovy/datadog/trace/agent/test/AgentTestRunner.groovy
```diff
@@ -247,7 +290,20 @@ abstract class AgentTestRunner extends DDSpecification implements AgentBuilder.L
   }
 
   void cleanup() {
+    if (USE_TEST_AGENT_WRITER) {
+      // save Datadog environment to DDAgentWriter header
+      addEnvironmentVariablesToHeaders(TEST_AGENT_WRITER)
+
+      // write ListWriter traces to the AgentWriter at cleanup so trace-processing changes occur after span assertions
+      def array = TEST_WRITER.toArray()
+      def traces = (Arrays.asList(array) as List<List<DDSpan>>)
```
amarziali, line 299: why converting an array to a list is needed? You can always iterate with an array

dd-java-agent/testing/src/main/groovy/datadog/trace/agent/test/AgentTestRunner.groovy
```diff
@@ -188,6 +218,13 @@ abstract class AgentTestRunner extends DDSpecification implements AgentBuilder.L
       TEST_DATA_STREAMS_MONITORING = new DefaultDataStreamsMonitoring(sink, features, SystemTimeSource.INSTANCE, wellKnownTags, TEST_DATA_STREAMS_WRITER, bucketDuration)
     }
     TEST_WRITER = new ListWriter()
+
+    USE_TEST_AGENT_WRITER = this.isTestAgentEnabled()
+    if (USE_TEST_AGENT_WRITER) {
```
amarziali, line 223: you can just do `if (isTestAgentEnabled())` and get rid of that field `USE_TEST_AGENT_WRITER`

amarziali (approved):
Globally ok. I left some nits for readability

dd-java-agent/testing/src/main/groovy/datadog/trace/agent/test/AgentTestRunner.groovy
```diff
@@ -82,6 +92,21 @@ abstract class AgentTestRunner extends DDSpecification implements AgentBuilder.L
...
 
+  static void addEnvironmentVariablesToHeaders(DDAgentApi agentapi) {
+    StringBuilder ddEnvVars = new StringBuilder()
+    for (Map.Entry<Object, Object> entry : System.getProperties().entrySet()) {
+      if (entry.getKey().toString().startsWith("dd.")) {
+        ddEnvVars.append(Strings.systemPropertyNameToEnvironmentVariableName(entry.getKey().toString()))
+          .append("=").append(entry.getValue()).append(",")
+      }
+    }
+    ddEnvVars.append("DD_SERVICE").append("=").append(Config.get().getServiceName())
```
nayeem-kamal, line 103: Looks like this chain of `.append()`s is causing a codenarc violation for you in CI. Multiple appends for string literals is redundant so I think you can make this statement into: `ddEnvVars.append("DD_SERVICE=").append(Config.get().getServiceName())`

nayeem-kamal (approved):
LGTM

dd-java-agent/testing/src/main/groovy/datadog/trace/agent/test/AgentTestRunner.groovy
wconti27: should be all set

PR #5156, opened 2023-05-04
Add probe id and eval errors as _dd tags

jpbempel:
# What Does This Do
Add probe id as `_dd.di.dynamictag.probe_id` tag
Add evaluation error as `_dd.di.dynamictag.evaluation_error` tag send snapshot containing evaluation errors when decoration condition evaluation has errors

# Motivation
correlation between probe and span tags
report evaluation errors
# Additional Notes

### DIFF
--- a/dd-java-agent/agent-debugger/src/main/java/com/datadog/debugger/probe/SpanDecorationProbe.java
+++ b/dd-java-agent/agent-debugger/src/main/java/com/datadog/debugger/probe/SpanDecorationProbe.java
@@ -19,6 +19,7 @@ import datadog.trace.bootstrap.debugger.ProbeId;
 import datadog.trace.bootstrap.debugger.ProbeImplementation;
 import datadog.trace.bootstrap.instrumentation.api.AgentSpan;
 import datadog.trace.bootstrap.instrumentation.api.AgentTracer;
+import datadog.trace.util.TagsHelper;
 import java.util.ArrayList;
 import java.util.Arrays;
 import java.util.Collections;
@@ -31,6 +32,8 @@ import org.slf4j.LoggerFactory;
 
 public class SpanDecorationProbe extends ProbeDefinition {
   private static final Logger LOGGER = LoggerFactory.getLogger(SpanDecorationProbe.class);
+  private static final String PROBEID_DD_TAGS_FORMAT = "_dd.%s.probe_id";
+  private static final String EVALERROR_DD_TAGS_FORMAT = "_dd.%s.evaluation_error";
 
   public enum TargetSpan {
     ACTIVE,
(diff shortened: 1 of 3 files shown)


dd-java-agent/agent-debugger/src/main/java/com/datadog/debugger/probe/SpanDecorationProbe.java
```diff
@@ -146,9 +149,11 @@ public void evaluate(CapturedContext context, CapturedContext.Status status) {
             serializeValue(sb, tag.value.getDsl(), tagValue.getValue(), status);
           }
           spanStatus.addTag(tag.name, sb.toString());
+          spanStatus.addTag("_dd." + tag.name + ".probe_id", getProbeId().getId());
```
shatzi, line 152: Do we sanitize tag.name some where ? maybe we should validate it lengths and that it doens't end with ".probe_id" and ".evaluation_error" - maybe that something that should happen on the back end as well.

Also asked the team if we should have longer prefix like "_dd.dynamic." to ensure we don't pollute "_dd" prefix with user generated data

shatzi (approved):
Awesome

dd-java-agent/agent-debugger/src/main/java/com/datadog/debugger/probe/SpanDecorationProbe.java
jpbempel: > Do we sanitize tag.name some where ? maybe we should validate it lengths and that it doens't end with ".probe_id" and ".evaluation_error" - maybe that something that should happen on the back end as well.
I have added call to `TagsHelper.sanitize` for standard sanitization of tags.
Not sure about preventing ".probe_id" or ".evaluation_error": those terms could be used by the user in their tag names

jpbempel:
still need to discuss where to put evaluation errors and the forat of probe id and prefixes

dd-java-agent/agent-debugger/src/main/java/com/datadog/debugger/probe/SpanDecorationProbe.java
```diff
@@ -145,15 +152,21 @@ public void evaluate(CapturedContext context, CapturedContext.Status status) {
           } else {
             serializeValue(sb, tag.value.getDsl(), tagValue.getValue(), status);
           }
-          spanStatus.addTag(tag.name, sb.toString());
+          spanStatus.addTag(tagName, sb.toString());
+          spanStatus.addTag(String.format(PROBEID_DD_TAGS_FORMAT, tagName), getProbeId().getId());
         } catch (EvaluationException ex) {
           LOGGER.debug("Evaluation error: ", ex);
           status.addError(new EvaluationError(ex.getExpr(), ex.getMessage()));
+          spanStatus.addTag(String.format(EVALERROR_DD_TAGS_FORMAT, tagName), ex.toString());
```
shatzi, line 160: I don't think we should pass ExceptionType, but just send the message: ex.getMessage()
Looks like leaking internal information as it will always be EvaluationException

shatzi (approved):

dd-java-agent/agent-debugger/src/main/java/com/datadog/debugger/probe/SpanDecorationProbe.java
jpbempel: done

PR #5157, opened 2023-05-04
Add basic Spock Framework support to CI Visibility

nikita-tkachenko-datadog:
# What Does This Do
Updates CI Visibility to better integrate with Spock Framework.
The following issues are addressed:
- determining proper test case name for parameterized tests
- determining test case methods
- determining correct test framework name and version

# Motivation
There are Datadog customers who use CI Visibility to monitor Spock Framework tests

# Additional Notes
Recent versions of Spock Framework run on top of JUnit 5, so most of the changes are done in JUnit 5 integration code.


dd-java-agent/instrumentation/junit-5.3/src/main/java/datadog/trace/instrumentation/junit5/SpockUtils.java
richardstartin, line 1: Fewer helper classes is always better for an instrumentation, so if you only have static utilities (i.e. no state to encapsulate) I would opt for generalising the name of `JUnit5Utils` and add util methods there.

richardstartin (approved):

PR #5167, opened 2023-05-09
Aggregate tests in check, allTests and allLatestDepTests targets

cataphract:
# What Does This Do

* Add all test targets to the `check` target.
* Create the targets `allTests` and `allLatestDepTests`, aggregating forking and non-forking tests depending on whether they include `test` in the name
* Create new task on the root project named `baseLatestDepTest`, which is partition-aware. Run the latest dep tests on CI using that instead on relying on disabling test tasks, in order to avoid the dependencies of the test tasks from being executed.

### DIFF
--- a/.circleci/config.yml
+++ b/.circleci/config.yml
@@ -808,7 +808,7 @@ build_test_jobs: &build_test_jobs
       cacheType: inst
   - build:
       name: build_latestdep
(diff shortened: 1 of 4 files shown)


cataphract:
Will merge through #5175

PR #5169, opened 2023-05-10
uniform v1 database naming to otel

amarziali:
# What Does This Do

Basically align v1 database naming to OTEL standards (table [here](https://opentelemetry.io/docs/specification/otel/trace/semantic_conventions/database/))

Notable changes:
* postgres -> postgresql
* mongo -> mongodb
* sqlserver -> mssql

This normalized name is used as part of operation (e.g. `postgresql.query`) and also to set the `db.type` tag.

Changes are only for v1 (v0 is not impacted)

The PR takes also the occasion to do a quick refactor  to avoid erasing the method `postProcessServiceAndOperationName`  in each decorator since it's only used on jdbc and vertx-jdbc. Instead, it's only declared in those 2 decorators.

# Motivation

# Additional Notes

### DIFF
--- a/dd-java-agent/instrumentation/aerospike-4/src/main/java/datadog/trace/instrumentation/aerospike4/AerospikeClientDecorator.java
+++ b/dd-java-agent/instrumentation/aerospike-4/src/main/java/datadog/trace/instrumentation/aerospike4/AerospikeClientDecorator.java
@@ -109,7 +109,4 @@ public class AerospikeClientDecorator extends DBTypeProcessingDatabaseClientDeco
     beforeFinish(span);
     span.finish();
   }
-
-  @Override
-  protected void postProcessServiceAndOperationName(AgentSpan span, String dbType) {}
 }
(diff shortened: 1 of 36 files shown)


dd-java-agent/instrumentation/aerospike-4/src/main/java/datadog/trace/instrumentation/aerospike4/AerospikeClientDecorator.java
```diff
@@ -109,7 +109,4 @@ public void finishAerospikeSpan(final AgentSpan span, final Throwable error) {
     beforeFinish(span);
     span.finish();
   }
-
-  @Override
-  protected void postProcessServiceAndOperationName(AgentSpan span, String dbType) {}
```
ygree, line 114: Nice cleanup removing all these empty methods!

ygree (approved):
LGTM

dd-java-agent/instrumentation/aerospike-4/src/main/java/datadog/trace/instrumentation/aerospike4/AerospikeClientDecorator.java
amarziali: thank you

PR #5170, opened 2023-05-10
[Spark] Add appId and sparkVersion tags to all spark spans

paul-laffon-dd:
# What Does This Do

Adding the tags `app_id` and `spark_version` to all spark spans. Before this PR `app_id` was only present on the application span, adding it to all the spark spans will make it easier to get all the spans of a given execution and `spark_version` was not present

### DIFF
--- a/dd-java-agent/instrumentation/spark/src/main/java/datadog/trace/instrumentation/spark/DatadogSparkListener.java
+++ b/dd-java-agent/instrumentation/spark/src/main/java/datadog/trace/instrumentation/spark/DatadogSparkListener.java
@@ -351,12 +351,7 @@ public class DatadogSparkListener extends SparkListener {
   }
 
   private AgentTracer.SpanBuilder buildSparkSpan(String spanName) {
-    return tracer
-        .buildSpan(spanName)
-        .withSpanType("spark")
-        // The volume is quite low and dropping a trace can will to incomplete data
-        .withTag(DDTags.MANUAL_KEEP, true)
-        .withTag("app_id", appId);
+    return tracer.buildSpan(spanName).withSpanType("spark").withTag("app_id", appId);
   }
 
   private long stageSpanKey(int stageId, int attemptId) {
--- a/dd-java-agent/instrumentation/spark/src/main/java/datadog/trace/instrumentation/spark/DatadogSparkListener.java
+++ b/dd-java-agent/instrumentation/spark/src/main/java/datadog/trace/instrumentation/spark/DatadogSparkListener.java
@@ -8,6 +8,7 @@ import java.util.HashMap;
 import java.util.Map;
 import java.util.Properties;
 import org.apache.spark.SparkConf;
+import org.apache.spark.SparkContext;
 import org.apache.spark.TaskFailedReason;
 import org.apache.spark.scheduler.*;
 import scala.Tuple2;
@@ -21,6 +22,8 @@ public class DatadogSparkListener extends SparkListener {
   private final int MAX_COLLECTION_SIZE = 1000;
 
   private final SparkConf sparkConf;
+  private final String sparkVersion;
+  private final String appId;
 
   private final AgentTracer.TracerAPI tracer;
 
@@ -44,9 +47,11 @@ public class DatadogSparkListener extends SparkListener {
   private int maxExecutorCount = 0;
   private long availableExecutorTime = 0;
 
-  public DatadogSparkListener(SparkConf _sparkConf) {
+  public DatadogSparkListener(SparkContext sparkContext) {
     tracer = AgentTracer.get();
-    sparkConf = _sparkConf;
+    sparkConf = sparkContext.getConf();
+    sparkVersion = sparkContext.version();
+    appId = sparkContext.applicationId();
 
     isRunningOnDatabricks = sparkConf.contains("spark.databricks.sparkContextId");
   }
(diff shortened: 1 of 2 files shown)


dd-java-agent/instrumentation/spark/src/main/java/datadog/trace/instrumentation/spark/DatadogSparkListener.java
```diff
@@ -353,6 +348,16 @@ public void onExecutorRemoved(SparkListenerExecutorRemoved executorRemoved) {
     }
   }
 
+  private AgentTracer.SpanBuilder buildSparkSpan(String spanName) {
+    return tracer
+        .buildSpan(spanName)
+        .withSpanType("spark")
+        // The volume is quite low and dropping a trace can will to incomplete data
+        .withTag(DDTags.MANUAL_KEEP, true)
```
raphaelgavache, line 356: nit: priority can be set only once on the first span, for example with span.SetTag() called where the config is set

dd-java-agent/instrumentation/spark/src/main/java/datadog/trace/instrumentation/spark/DatadogSparkListener.java
```diff
@@ -353,6 +348,16 @@ public void onExecutorRemoved(SparkListenerExecutorRemoved executorRemoved) {
...
+        .withTag("app_id", appId)
+        .withTag("spark_version", sparkVersion);
+  }
```
raphaelgavache, line 359: you can also add the Measured tag here and remove the setMeasured on all spans above. The tag is handled by the [interceptor](https://github.com/DataDog/dd-trace-java/blob/master/dd-trace-core/src/main/java/datadog/trace/core/taginterceptor/TagInterceptor.java#L103-L104))

dd-java-agent/instrumentation/spark/src/main/java/datadog/trace/instrumentation/spark/DatadogSparkListener.java
```diff
@@ -353,6 +348,16 @@ public void onExecutorRemoved(SparkListenerExecutorRemoved executorRemoved) {
...
   }
 
+  private AgentTracer.SpanBuilder buildSparkSpan(String spanName) {
+    return tracer
+        .buildSpan(spanName)
+        .withSpanType("spark")
+        // The volume is quite low and dropping a trace can will to incomplete data
+        .withTag(DDTags.MANUAL_KEEP, true)
+        .withTag("app_id", appId)
+        .withTag("spark_version", sparkVersion);
```
raphaelgavache, line 358: why set this on all spans and not on the span containing the config?

paul-laffon-dd: Not particular reason, moved it in the config only

raphaelgavache (approved):

dd-java-agent/instrumentation/spark/src/main/java/datadog/trace/instrumentation/spark/DatadogSparkListener.java
amarziali: did you test that it works? As far as I know you cannot set the sampling priority with setTag. Instead there is a specialized method `setSamplingPriority` on the DDSpan. This method is not today visible to the `AgentSpan` so that you have to expose it

paul-laffon-dd: I didn't test it thoroughly because generating 10 traces per second should be unlikely for spark. I was following this documentation for the force keep https://docs.datadoghq.com/tracing/trace_pipeline/ingestion_mechanisms/?tab=java#force-keep-and-drop

dd-java-agent/instrumentation/spark/src/main/java/datadog/trace/instrumentation/spark/DatadogSparkListener.java
amarziali: oh today I learn something. there is an interceptor doing manual keep so it will be ok. But I think for measured using the API will be a better choice

dd-java-agent/instrumentation/spark/src/main/java/datadog/trace/instrumentation/spark/DatadogSparkListener.java
amarziali: oh today I learn something. there is an interceptor doing manual keep so it will be ok.

bantonsson: Yes, but that is probably wrong for billing audit purposes since it's the instrumentation that is keeping the trace and not the end user, which is what the interceptor will mark it as.

raphaelgavache: We havn't done it yet but we'll set a special ingestion reason tag at the ingestion layer that will be used to bill/audit ingested bytes.

bantonsson: Still, you shouldn't use the tag setting approach to keep it since it will mark the sampling mechanism as `MANUAL`, when it really should be something else.

amarziali:
I think you should rebase instead of merging master into this branch

dd-java-agent/instrumentation/spark/src/main/java/datadog/trace/instrumentation/spark/DatadogSparkListener.java
paul-laffon-dd: I removed the `DDTags.MANUAL_KEEP` and will handle the sampling priority in another PR to keep this one simple

bantonsson (approved):

PR #5173, opened 2023-05-10
[Spark] Link databricks spark jobs to databricks workflow tasks

paul-laffon-dd:
# What Does This Do

Databricks spark spans are generated from this spark instrumentation, while databricks workflow task spans are generated from the databricks API. The traceId/spanId is computed using the same hash on both sides to link everything in the same trace

Those hash are computed using a combinaison of the `job_id`, `job_run_id` and `task_run_id`

### DIFF
--- a/dd-java-agent/instrumentation/spark/src/main/java/datadog/trace/instrumentation/spark/DatadogSparkListener.java
+++ b/dd-java-agent/instrumentation/spark/src/main/java/datadog/trace/instrumentation/spark/DatadogSparkListener.java
@@ -121,15 +121,21 @@ public class DatadogSparkListener extends SparkListener {
       }
 
       if (jobStart.properties() != null) {
-        // ids to link those spans to databricks job/task traces
-        jobSpanBuilder.withTag(
-            "databricks_job_id", jobStart.properties().get("spark.databricks.job.id"));
-        jobSpanBuilder.withTag(
-            "databricks_job_run_id", getDatabricksJobRunId(jobStart.properties()));
+        String databricksJobId = (String) jobStart.properties().get("spark.databricks.job.id");
+        String databricksJobRunId = getDatabricksJobRunId(jobStart.properties());
 
         // spark.databricks.job.runId is the runId of the task, not of the Job
-        jobSpanBuilder.withTag(
-            "databricks_task_run_id", jobStart.properties().get("spark.databricks.job.runId"));
+        String databricksTaskRunId =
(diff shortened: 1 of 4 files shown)


dd-java-agent/instrumentation/spark/src/main/java/datadog/trace/instrumentation/spark/DatadogSparkListener.java
```diff
@@ -121,15 +121,21 @@ public void onJobStart(SparkListenerJobStart jobStart) {
...
+            (String) jobStart.properties().get("spark.databricks.job.runId");
+
+        // ids to link those spans to databricks job/task traces
+        jobSpanBuilder.withTag("databricks_job_id", databricksJobId);
+        jobSpanBuilder.withTag("databricks_job_run_id", databricksJobRunId);
+        jobSpanBuilder.withTag("databricks_task_run_id", databricksTaskRunId);
+
+        AgentSpan.Context parentContext =
+            new DatabricksParentContext(databricksJobId, databricksJobRunId, databricksTaskRunId);
+        jobSpanBuilder.asChildOf(parentContext);
```
raphaelgavache, line 138: When a DatabricksParentContext can't be computed the fallback should be the default random traceID/spanID generation instead of setting traceID and spanID to 0

paul-laffon-dd: From my understanding, if the traceId is 0, it will be randomly [generated here](https://github.com/DataDog/dd-trace-java/blob/master/dd-trace-core/src/main/java/datadog/trace/core/CoreTracer.java#L1456) and the default behaviour in the tracer to indicate that there is no parentSpanId is to use 0 ([from here](https://github.com/DataDog/dd-trace-java/blob/master/dd-trace-core/src/main/java/datadog/trace/core/CoreTracer.java#L1466))

PerfectSlayer: Trace id = 0 is consider as invalid trace id.
I would also recommend not calling `asChildOf()` for better clarity.

paul-laffon-dd: Added a condition to only call it when the traceId is valid

PerfectSlayer (approved):

PR #5182, opened 2023-05-11
Fix encoding of remote config capabilities

smola:
# What Does This Do

# Motivation
Only the first remote config capabilities were sent, others were truncated.

# Additional Notes

### DIFF
--- a/remote-config/src/main/java/datadog/remoteconfig/tuf/RemoteConfigRequest.java
+++ b/remote-config/src/main/java/datadog/remoteconfig/tuf/RemoteConfigRequest.java
@@ -93,22 +93,12 @@ public class RemoteConfigRequest {
       this.products = productNames;
       this.tracerInfo = tracerInfo;
 
-      // Little-endian encoding of the `long` capabilities, stripping any trailing zero bytes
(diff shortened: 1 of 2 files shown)


remote-config/src/main/java/datadog/remoteconfig/tuf/RemoteConfigRequest.java
```diff
@@ -87,12 +87,29 @@ public ClientInfo(
...
+      // (except the first one)
+      int size = 1;
+      long accum = capabilities;
+      for (int i = 0; i < 8; i++) {
+        accum = accum >>> 8;
+        if (accum == 0) {
+          size = i + 1;
+          break;
+        }
+      }
```
cataphract, line 106: You could use Long.numberOfLeadingZeros() and avoid the loop, as in `Math.max(1, Long.BYTES - Long.numberOfLeadingZeros(capabilities) / 8)`.

Or replace everything with `ByteBuffer.allocate(Long.BYTES).putLong(capabilities).array()`. I guess the extra 0s won't hurt.

cataphract (approved):

PR #5187, opened 2023-05-11
Detect unvalidated redirect vulnerabilities in servlets

jandro996:
# What Does This Do
Add instrumentation to javax.servlet.http.HttpServletResponse:
- addHeader/setHeader for location headers
- sendRedirect

Add tainting support to javax.servlet.http.HttpServletResponse:
- encodeURL
- encodeRedirectURL

Unify Header instrumentation callbacks, we need to analice the headers with multiple sink modules at the same time

# Motivation
- Support Unvalidated redirect vulnerability

# Additional Notes
This is a first iteration, only supports servlet unvalidated redirect vulnerability

### DIFF
--- a/dd-java-agent/instrumentation/servlet-common/src/main/java/datadog/trace/instrumentation/servlet/HttpServletResponseInstrumentation.java
+++ b/dd-java-agent/instrumentation/servlet-common/src/main/java/datadog/trace/instrumentation/servlet/HttpServletResponseInstrumentation.java
@@ -87,7 +87,7 @@ public final class HttpServletResponseInstrumentation extends Instrumenter.Iast
   }
 
   public static class EncodeURLAdvice {
-    @Advice.OnMethodExit(onThrowable = Throwable.class, suppress = Throwable.class)
+    @Advice.OnMethodExit(suppress = Throwable.class)
     public static void onExit(@Advice.Argument(0) final String url, @Advice.Return String encoded) {
       final PropagationModule module = InstrumentationBridge.PROPAGATION;
       if (module != null) {
--- a/dd-java-agent/instrumentation/servlet-common/src/main/java/datadog/trace/instrumentation/servlet/HttpServletResponseInstrumentation.java
+++ b/dd-java-agent/instrumentation/servlet-common/src/main/java/datadog/trace/instrumentation/servlet/HttpServletResponseInstrumentation.java
@@ -68,16 +68,8 @@ public final class HttpServletResponseInstrumentation extends Instrumenter.Iast
     @Advice.OnMethodEnter(suppress = Throwable.class)
     public static void onEnter(
         @Advice.Argument(0) final String name, @Advice.Argument(1) String value) {
-      if ("Set-Cookie".equalsIgnoreCase(name) && null != value && value.length() > 0) {
-        final InsecureCookieModule module = InstrumentationBridge.INSECURE_COOKIE;
-        if (module != null) {
-          module.onCookieHeader(value);
-        }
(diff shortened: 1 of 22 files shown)


dd-java-agent/instrumentation/servlet-common/src/main/java/datadog/trace/instrumentation/servlet/HttpServletResponseInstrumentation.java
```diff
@@ -63,11 +68,40 @@ public static class AddHeaderAdvice {
...
-        if ("Set-Cookie".equalsIgnoreCase(name) && null != value && value.length() > 0) {
+      if ("Set-Cookie".equalsIgnoreCase(name) && null != value && value.length() > 0) {
+        final InsecureCookieModule module = InstrumentationBridge.INSECURE_COOKIE;
+        if (module != null) {
           module.onCookieHeader(value);
         }
+      } else if ("Location".equals(name) && null != value && value.length() > 0) {
+        final UnvalidatedRedirectModule module = InstrumentationBridge.UNVALIDATED_REDIRECT;
+        if (module != null) {
+          module.onRedirect(value);
```
smola, line 79: Maybe we should have a single callback for `onHeader(name, value)`, and detect both insecure cookie, unvalidated redirect, and other header-related vulnerabilities there.

jandro996: I've made some changes to solve this problem

dd-smoke-tests/springboot/src/main/java/datadog/smoketest/springboot/controller/IastWebController.java
```diff
@@ -73,6 +74,22 @@ public String insecureCookieFromHeader(HttpServletResponse response) {
...
+    response.addHeader("Location", param);
+    response.setStatus(HttpStatus.FOUND.value());
+    return "Unvalidated redirect";
+  }
+
+  @GetMapping("/unvalidated_redirect_from_send_redirect")
+  public String unvalidatedRedirectFromSendRedirect(
+      @RequestParam String param, HttpServletResponse response) throws IOException {
+    response.sendRedirect(param);
+    response.setStatus(HttpStatus.FOUND.value());
```
manuel-alvarez-alvarez, line 89: The method `sendRedirect` shod set the header automatically, no need to do it explictly

anderruiz:
I would say that you are not ignore path inputs in the getTainted checks as implemented here:

https://bitbucket.org/hdivsecurity/hdiv-enterprise/src/31eba5058d78b0a66e2c6519bfc88b528134d9a5/hdiv-ee-agent-core/src/main/java/org/hdiv/ee/agent/core/analyzers/http/UnvalidatedRedirectAnalyzer.java#lines-78:~:text=79-,78,-77

At the same time, I assume that we are aware that it will not work (I suspect, with Spring)

smola (approved):

jandro996:
> I would say that you are not ignore path inputs in the getTainted checks as implemented here:
> 
> https://bitbucket.org/hdivsecurity/hdiv-enterprise/src/31eba5058d78b0a66e2c6519bfc88b528134d9a5/hdiv-ee-agent-core/src/main/java/org/hdiv/ee/agent/core/analyzers/http/UnvalidatedRedirectAnalyzer.java#lines-78:~:text=79-,78,-77
> 
> At the same time, I assume that we are aware that it will not work (I suspect, with Spring)

We will solve the path variable case in another PR

Thanks for the comment!!

PR #5188, opened 2023-05-11
Allow DD Intake Writer to use multiple dispatchers

nikita-tkachenko-datadog:
# What Does This Do
Allows `DDIntakeWriter` to write traces to multiple `PayloadDispatcher` instances. 
Different dispatchers can use different `RemoteMapper` and `RemoteApi` instances (i.e. each dispatcher has its own serialiser, its own buffer for serialised data and its own endpoint config).

# Motivation
In CI Visibility there is a need to write data from each trace to two different backend endpoints:
* one endpoint for spans data
* the other endpoint for code coverage data that is associated with the spans

Splitting data flow at payload dispatcher allows to reuse logic that asynchronously queues traces for serializing.

# Additional Notes
This PR only adds the mechanism that allows to have multiple dispatchers. Actual dispatcher config/creation will be done in a subsequent PR with CI Vis code coverage changes.

`PayloadDispatcher` concrete class was turned into interface. Implementation logic was pushed to `PayloadDispatcherImpl`.

Some redundant constructors and fields were removed from `DDIntakeWriter`, `DDAgentWriter` and `RemoteWriter`.


dd-trace-core/src/main/java/datadog/trace/common/writer/CompositePayloadDispatcher.java
```diff
@@ -0,0 +1,45 @@
...
+  @Override
+  public void flush() {
+    for (PayloadDispatcher delegate : delegates) {
+      delegate.flush();
+    }
+  }
+
+  @Override
+  public Collection<Class<? extends RemoteApi>> getApis() {
+    Collection<Class<? extends RemoteApi>> apis = new ArrayList<>();
```
dougqh, line 39: This should be sized appropriately -- it looks like the number of elements is known

nikita-tkachenko-datadog: Good catch, fixed

bantonsson (approved):

PR #5189, opened 2023-05-11
record time spent in TPE work queue

richardstartin:
# What Does This Do

Records time spent in the `ThreadPoolExecutor` work queue and emits an event before the scheduled task runs if a threshold of 100ms is breached.

This is the first step in better accounting for time spent off-CPU. Currently we have some profiling for off-CPU time in our wallclock profiler, but this only can only records samples when a thread is blocked. Reactive frameworks tend to avoid blocking, so instead we emit latency thresholded events. The timer itself currently emits an OpenJDK JFR event for ease of development, but in the future this will be moved into our native profiler to better support non-OpenJDK JVMs such as J9. 

The timer state is propagated as a field in `State`. This was chosen over `Continuation` to have more control over which context switches are timed, since there can be multiple individual `Continutation`s per context switch (e.g. `CompletableFuture` can create a `FutureTask`s, creating stacked `Continuation`s) we only want to time a subset of `Continuation`s. 

We record various fields on the events:
* the class name of the queue implementation
* the class name of the task (which is intended to help users to figure out which task was waiting in the queue
* the name of the scheduler (in this case, `ThreadPoolExecutor`, but this will encompass at least the FJP and netty in the future). 

Getting access to the task name is tricky because there is generally a lot of wrapping involved in asynchronous task submission, and we want to record the name of the user's `Runnable` or `Callable`, not some `FutureTask` or some other adapter. To achieve this, certain known wrapper classes (currently there are two: `FutureTask` and `Executors$RunnableAdapter`) are modified to implement `TaskWrapper`, which exposes the field we want to unwrap. The interface has static functionality to recursively unwrap the type. Care needs to be taken here to avoid cycles or very deep delegation chains.

The new instrumentation is disabled unless profiling is enabled, and the events are disabled unless `-Ddd.profiling.experimental.queueing.time.enabled=true` is set.

# Motivation

# Additional Notes

I have some concerns about the unwrapping approach in light of [JDK-8180450](https://bugs.openjdk.org/browse/JDK-8180450). Successful interface checks on a type lead to updating `secondary_super_cache`, and we currently do at least 2 interface checks on certain types used in context propagation: against `RunnableFuture` in various places and against `FieldBackedContextAccessor` before accessing the context fields. This PR adds another interface check which will not make the status quo any worse, but will make it harder to get down to one (I think most `RunnableFuture` checks could be refactored away).

### DIFF
--- a/dd-java-agent/instrumentation/java-concurrent/src/main/java/datadog/trace/instrumentation/java/concurrent/TaskWrapperInstrumentation.java
+++ b/dd-java-agent/instrumentation/java-concurrent/src/main/java/datadog/trace/instrumentation/java/concurrent/TaskWrapperInstrumentation.java
@@ -1,0 +1,33 @@
+package datadog.trace.instrumentation.java.concurrent;
+
+import com.google.auto.service.AutoService;
+import datadog.trace.agent.tooling.Instrumenter;
+import datadog.trace.agent.tooling.bytebuddy.profiling.UnwrappingVisitor;
+
+@AutoService(Instrumenter.class)
+public class TaskWrapperInstrumentation extends Instrumenter.Profiling
+    implements Instrumenter.ForKnownTypes {
+  public TaskWrapperInstrumentation() {
+    super("java_concurrent", "wrapper-task");
+  }
+
+  @Override
+  public AdviceTransformer transformer() {
+    return new VisitingTransformer(
+        new UnwrappingVisitor(
+            "java.util.concurrent.FutureTask",
+            "callable",
+            "java.util.concurrent.Executors$RunnableAdapter",
+            "task"));
+  }
+
+  @Override
+  public void adviceTransformations(AdviceTransformation transformation) {}
+
+  @Override
+  public String[] knownMatchingTypes() {
+    return new String[] {
+      "java.util.concurrent.FutureTask", "java.util.concurrent.Executors$RunnableAdapter"
+    };
+  }
+}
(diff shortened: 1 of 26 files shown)


dd-java-agent/instrumentation/java-concurrent/src/main/java/datadog/trace/instrumentation/java/concurrent/TaskWrapperInstrumentation.java
```diff
@@ -0,0 +1,33 @@
...
+  }
+
+  @Override
+  public AdviceTransformer transformer() {
+    return new VisitingTransformer(
+        new UnwrappingVisitor(
+            "java.util.concurrent.FutureTask",
+            "callable",
+            "java.util.concurrent.Executors$RunnableAdapter",
+            "task"));
```
richardstartin, line 21: this will be expanded later to include various FJP tasks and adapters

dd-java-agent/instrumentation/java-concurrent/src/test/groovy/ExecutorInstrumentationTest.groovy
```diff
@@ -65,6 +66,9 @@ abstract class ExecutorInstrumentationTest extends AgentTestRunner {
 
     injectSysConfig("dd.trace.executors", "CustomThreadPoolExecutor")
     injectSysConfig("trace.thread-pool-executors.exclude", "ExecutorInstrumentationTest\$ToBeIgnoredExecutor")
+    // FIXME settings not being applied, which means the task unwrapping is not being tested
+    injectSysConfig("dd.profiling.enabled", "true")
+    injectSysConfig("dd.profiling.experimental.queueing.time.enabled", "true")
```
richardstartin, line 71: I haven't been able to figure out why this configuration is not respected, but I'm testing end to end execution of the instrumentation and the event emission via `JFRBasedProfilingIntegrationTest` for now

jbachorik (approved):
👍

PR #5194, opened 2023-05-12
Cache TestEventsHandler instance

nikita-tkachenko-datadog:
# What Does This Do
Wraps `TestEventsHandler.Factory` used in CI Visibility with a caching decorator.

# Motivation
[PR 5005](https://github.com/DataDog/dd-trace-java/pull/5005) changed the way we hook into JUnit 4.
As the result, multiple `TracerListener` instances can now be created when attaching to JUnit 4 that is run with Gradle.
Each tracer listener obtains a `TestEventsHandler`, and if each handler is different, it will cause errors and performance degradation (as the handler is stateful - it stores executed suites and test cases, as well as indirectly the results of some expensive operations, e.g. repo index build).

# Additional Notes
This fixes performance degradation observed when tracing Gradle/JUnit4 without using DD Javac Plugin.


dd-trace-api/src/main/java/datadog/trace/api/ConfigDefaults.java
```diff
@@ -111,6 +111,7 @@ public final class ConfigDefaults {
   static final boolean DEFAULT_CIVISIBILITY_AUTO_CONFIGURATION_ENABLED = true;
   static final boolean DEFAULT_CIVISIBILITY_COMPILER_PLUGIN_AUTO_CONFIGURATION_ENABLED = true;
   static final String DEFAULT_CIVISIBILITY_COMPILER_PLUGIN_VERSION = "0.1.6";
+  static final int DEFAULT_CIVISIBILITY_TEST_EVENTS_HANDLER_CACHE_SIZE = 1;
```
smola, line 114: Is this really a good default? I guess it's because they are expected to be run sequentially, but what about parallel tests?

nikita-tkachenko-datadog: We would need multiple instances only if the same JVM was used to run tests for multiple repositories or test frameworks. The only case when we have a process that spans multiple repos is when a Gradle daemon is instrumented, and Gradle always forks a new process to run tests in a module, so multiple repo case looks impossible.
Multiple frameworks is a rare, but possible scenario. Currently asking for a new handler instance in each test case only happens for JUnit 4, but it is possible that other frameworks we will support in the future will have the same problem.

Considering that the price we pay for increasing the cache size is quite small, I increased the default size to 4 just to be on the safe side.

Issue #5191, opened 2023-05-11
Trace and Span IDs with logback-access in Spring Boot with Tomcat 9


internetstaff:
We're using Spring Boot 2.7.11 and Tomcat 9.0.74. We recently added `logback-access` for improved access logging and realized we're not getting trace or span IDs.

The root cause appears to be:

* Datadog agent injects into `org.apache.coyote.Request` using `setAttribute()`
* `org.apache.coyote.Request` is encapsulated by `org.apache.catalina.connector.Request`
* `logback-access` sees a `org.apache.catalina.connector.Request`
* Logback `AccessEvent` copies all the attributes to avoid mutation.
* Logback calls `getAttributeNames()` on `org.apache.catalina.connector.Request`, which does not return names from the encapsulated `org.apache.coyote.Request` despite those attributes being accessible via `getAttribute()`.
* Logback then does not have the trace_id and span_id available because they are not in the `AccessEvent`.

If Datadog agent could call `setAttribute()` on the `org.apache.catalina.connector.Request` this should be fixed, but I'm not clear on the downsides of this.

In the mean time, I simply worked around this with a custom Valve that effectively copies the two attributes out of the encapsulated coyote Request into the connector Request:

```java
      request.setAttribute("dd.trace_id", request.getAttribute("dd.trace_id"));
      request.setAttribute("dd.span_id", request.getAttribute("dd.span_id"));
```

Notes:
https://github.com/qos-ch/logback/blob/master/logback-access/src/main/java/ch/qos/logback/access/spi/AccessEvent.java#L407
https://github.com/apache/tomcat/blob/main/java/org/apache/catalina/connector/Request.java#L902

amarziali:
👋 The issue should be fixed by the PR just merged. It should be included in the next release

PR #5195, opened 2023-05-12
set traceid and spanid on catalina request

amarziali:
# What Does This Do

Propose a fix for #5191 

# Motivation

# Additional Notes

### DIFF
--- a/dd-java-agent/instrumentation/tomcat-5.5/src/main/java/datadog/trace/instrumentation/tomcat/TomcatServerInstrumentation.java
+++ b/dd-java-agent/instrumentation/tomcat-5.5/src/main/java/datadog/trace/instrumentation/tomcat/TomcatServerInstrumentation.java
@@ -20,6 +20,7 @@ import datadog.trace.api.GlobalTracer;
 import datadog.trace.api.gateway.Flow;
 import datadog.trace.bootstrap.instrumentation.api.AgentScope;
 import datadog.trace.bootstrap.instrumentation.api.AgentSpan;
+import datadog.trace.bootstrap.instrumentation.api.AgentTracer;
 import datadog.trace.bootstrap.instrumentation.java.concurrent.ExcludeFilter;
 import java.util.Arrays;
 import java.util.Collection;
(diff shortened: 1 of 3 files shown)


dd-java-agent/instrumentation/tomcat-5.5/src/main/java/datadog/trace/instrumentation/tomcat/TomcatServerInstrumentation.java
```diff
@@ -161,16 +161,20 @@ public static void afterParse(
...
-        AgentSpan.Context.Extracted ctx =
-            ctxObj instanceof AgentSpan.Context.Extracted
-                ? (AgentSpan.Context.Extracted) ctxObj
-                : null;
-        DECORATE.onRequest(span, req, req, ctx);
-        Flow.Action.RequestBlockingAction rba = span.getRequestBlockingAction();
-        if (rba != null) {
-          TomcatBlockingHelper.commitBlockingResponse(req, resp, rba);
-          ret = false; // skip pipeline
+        try (final AgentScope scope = activateSpan(span)) {
```
bantonsson, line 164: I don't think that you need to activate any span to read the `traceId` and `spanId`.

amarziali: CoreTracer.getTraceId is using the activeSpan so I wanted to make sure it's active.  Directly using the traceId / spanId seems not OK seems that method is checking when using hex encoding for 128bit etc etc

bantonsson: Then why not break out that code into `getLoggingTraceId(AgentSpan span)` and `getLoggingSpanId(AgentSpan span)` and use that?

amarziali: now should be in a better shape. Could you please review again?

bantonsson (approved):

PerfectSlayer (approved):

PR #5197, opened 2023-05-12
Add runtimeId to probe statuses

ojung:
# What Does This Do

Adds the runtimeId to probe statuses.

# Motivation

Probe statuses can be different for a single probe in a service fleet. Probe status were originally introduced when only one instance per service was applying instrumentation. With DI logs, metrics and spans probe statuses are emitted from more instances of a service.

# Additional Notes


dd-java-agent/agent-debugger/src/main/java/com/datadog/debugger/sink/Snapshot.java
```diff
@@ -16,7 +16,6 @@
 public class Snapshot {
   private static final String LANGUAGE = "java";
   private static final int VERSION = 2;
-
```
jpbempel, line 19: I prefer to keep separated static and instance fields here :)

ojung: Oops, that wasn't supposed to make it into the PR :)

PR #5201, opened 2023-05-12
Add support for private fields for metric values

jpbempel:
# What Does This Do
Call helper methods when trying to access a field that is not accessible directly by GETFIELD bytecode. Those helpers will try by reflection to set accessible to true and get he value of the field For primitive access need special call for each type of primitive

# Motivation

# Additional Notes

### DIFF
--- a/dd-java-agent/agent-debugger/debugger-bootstrap/src/main/java/datadog/trace/bootstrap/debugger/el/ReflectiveFieldValueResolver.java
+++ b/dd-java-agent/agent-debugger/debugger-bootstrap/src/main/java/datadog/trace/bootstrap/debugger/el/ReflectiveFieldValueResolver.java
@@ -4,7 +4,7 @@ import java.lang.reflect.Field;
 import java.lang.reflect.Modifier;
 
 /** A helper class to resolve a reference path using reflection. */
-public final class ReflectiveFieldValueResolver {
+public class ReflectiveFieldValueResolver {
   public static Object resolve(Object target, Class<?> targetType, String fldName) {
     Field fld = getField(targetType, fldName);
     if (fld == null) {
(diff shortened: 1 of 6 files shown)


dd-java-agent/agent-debugger/debugger-bootstrap/src/main/java/datadog/trace/bootstrap/debugger/el/ReflectiveFieldValueResolver.java
```diff
@@ -17,6 +17,62 @@ public static Object resolve(Object target, Class<?> targetType, String fldName)
     }
   }
 
+  public static Object getFieldValue(Object target, String fieldName)
+      throws NoSuchFieldException, IllegalAccessException {
+    return getField(target, fieldName).get(target);
+  }
+
+  public static long getFieldValueAsLong(Object target, String fieldName)
+      throws NoSuchFieldException, IllegalAccessException {
```
shatzi, line 26: https://docs.oracle.com/en/java/javase/11/docs/api/java.base/java/lang/reflect/Field.html#getFloat(java.lang.Object)

Can't this also throws IllegalArgumentException? do we need to specify this? or we always ensure we don't access fields we can convert?

shatzi (approved):
Nice work

dd-java-agent/agent-debugger/debugger-bootstrap/src/main/java/datadog/trace/bootstrap/debugger/el/ReflectiveFieldValueResolver.java
jpbempel: `IllegalArgumentException` is a RuntimeException unlike `NoSuchFieldException` and `IllegalAccessException` which are checked exception and so nned to be explictly put in `throws` clause it no catch in this method

PR #5206, opened 2023-05-15
fix http resource mapping

amarziali:
# What Does This Do

Fixes the urlAsAResource rule on the tag interceptor. A couple of things:

1.  Uses the path and not the full url to match against the http resource mapping ant patterns
2. Uses the client settings for http client spans (currently uses always the server patterns which is wrong)

# Motivation

# Additional Notes


ygree (approved):
LGTM

PR #5208, opened 2023-05-15
[hackaton] use muzzle to document supported instrumentation versions

amarziali:
# What Does This Do

Adds new muzzle tasks to gather, for each instrumentation, the jar name (group:module) we instrument and its supported version ranges

It produces an aggregated csv at the end of each builds and store it on circleCI

# Motivation

# Additional Notes


buildSrc/src/main/groovy/MuzzlePlugin.groovy
```diff
@@ -164,6 +190,14 @@ class MuzzlePlugin implements Plugin<Project> {
     }
   }
 
+  private static void dumpVersionRanges(Project project, final SortedSet<TestedArtifact> versions) {
+    def filename = project.path.replaceFirst('^:', '').replace(':', '_')
+    def dir = project.file("${project.rootProject.buildDir}/muzzle-deps-results")
+    dir.mkdirs()
+    def file = project.file("${dir}/${filename}.csv")
```
bm1549, line 197: [optional] A header at the top of the file would be nice since it's easier to auto-import into spreadsheets

buildSrc/src/main/groovy/MuzzlePlugin.groovy
```diff
@@ -164,6 +190,14 @@ class MuzzlePlugin implements Plugin<Project> {
...
+    file.text = versions.collect {[it.name, it.low.toString(), it.high.toString()].join(",")}.join("\n")
```
bm1549, line 198: In this code, if a dependency shows up multiple times (for example `com.couchbase.client:java-client`), it will show multiple pairs of `low` and `high` versions

Can we get the min and max for each dependency instead?

amarziali: Yes we should but we also should take into account possibly not supported versions in between

bm1549: Is that the case for any dependency today, where we support a range of dependencies with the exception of a middle version, or is that speculative?

amarziali: tbh we only put the excludes for versions that have broken artifacts so I also think that in real life we can just report a range of versions without holes.

bm1549 (approved):

PR #5214, opened 2023-05-16
[netty] don't create two root spans if 100 continue is expected

amarziali:
# What Does This Do

When the client request a 100 continue to early check the validity of the request without sending the full body, the netty server can send downstream its pipeline a http response with the 100 continue status.

However, this is not the final http response status since there will be another upon termination.
The issue with our instrumentation is that we set the status once and we finish the span. This will cause to create two root spans with wrong information

Example: 

image

# Motivation

# Additional Notes


PerfectSlayer (commented):
Looks good.

Can you set the _imports_ settings according the [contribution guidelines](https://github.com/DataDog/dd-trace-java/blob/master/CONTRIBUTING.md#intellij-idea) so it should revert all the irrelevant changes from the import? Thanks!

PerfectSlayer (approved):

ygree (approved):
LGTM

smola (approved):

PR #5223, opened 2023-05-17
Improve support for SQL escaped literals in redaction

manuel-alvarez-alvarez:
# What Does This Do
Fix issue with SQL tokenizer being too greedy when dealing with oracle and postgresql escaped strings

# Motivation

# Additional Notes

### DIFF
--- a/dd-java-agent/agent-iast/src/test/resources/redaction/evidence-redaction-suite.yml
+++ b/dd-java-agent/agent-iast/src/test/resources/redaction/evidence-redaction-suite.yml
@@ -369,7 +369,54 @@ suite:
         ]
       }
   - type: 'VULNERABILITIES'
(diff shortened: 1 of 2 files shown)


dd-java-agent/agent-iast/src/test/resources/redaction/evidence-redaction-suite.yml
```diff
@@ -866,6 +921,50 @@ suite:
           }
         ]
       }
+
+  - type: 'VULNERABILITIES'
```
manuel-alvarez-alvarez, line 925: This one was introduced by @uurien, I can remove it from this PR and include it in another one since it's not related to the current change.

smola (approved):

PR #5213, opened 2023-05-16
Fix job partitioning in CI

smola:
# What Does This Do
Fix test partitioning, which used project path hashCode without considering it can be negative.

# Motivation

# Additional Notes

## Disabled tests

Tests that are disabled by this PR:
* `PowerWAFModuleSpecification.rule toggling data given through configuration` which was disabled until recently anyway.

## Related PRs

PRs fixing issues that were previously masked by this:

* https://github.com/DataDog/dd-trace-java/pull/5212
* https://github.com/DataDog/dd-trace-java/pull/5216
* https://github.com/DataDog/dd-trace-java/pull/5217
* https://github.com/DataDog/dd-trace-java/pull/5218
* https://github.com/DataDog/dd-trace-java/pull/5225
* https://github.com/DataDog/dd-trace-java/pull/5229
* https://github.com/DataDog/dd-trace-java/pull/5230

## Affected modules

Effectively, some removed from this list since they run in non-parallelized jobs):

```
:dd-java-agent:agent-bootstrap
:dd-java-agent:agent-iast
:dd-java-agent:agent-jmxfetch
:dd-java-agent:agent-tooling
:dd-java-agent:appsec
:dd-java-agent:appsec:weblog
:dd-java-agent:benchmark-integration:play-perftest
:dd-java-agent:instrumentation
:dd-java-agent:instrumentation:akka-concurrent
:dd-java-agent:instrumentation:akka-http-10.0
:dd-java-agent:instrumentation:apache-httpclient-5
:dd-java-agent:instrumentation:caffeine
:dd-java-agent:instrumentation:cdi-1.2
:dd-java-agent:instrumentation:datanucleus-4
:dd-java-agent:instrumentation:dropwizard:dropwizard-views
:dd-java-agent:instrumentation:elasticsearch:rest-6.4
:dd-java-agent:instrumentation:google-http-client
:dd-java-agent:instrumentation:graal:native-image
:dd-java-agent:instrumentation:grizzly-2
:dd-java-agent:instrumentation:grizzly-http-2.3.20
:dd-java-agent:instrumentation:grpc-1.5
:dd-java-agent:instrumentation:hibernate
:dd-java-agent:instrumentation:ignite-2.0
:dd-java-agent:instrumentation:jackson-core
:dd-java-agent:instrumentation:java-concurrent
:dd-java-agent:instrumentation:java-lang:java-lang-9
:dd-java-agent:instrumentation:java-security
:dd-java-agent:instrumentation:jax-rs-annotations-1
[...]

PR #5229, opened 2023-05-17
Fix SessionStateTest

smola:
# What Does This Do

# Motivation
In #5015 the meaning of the boolean flag to SessionState constructor was inverted, but it was not updated in tests.

# Additional Notes
Note that this test is currently not running in CI. See https://github.com/DataDog/dd-trace-java/pull/5213


bantonsson (approved):
You beat me to it. Was just about to open a PR.

amarziali (approved):
Thank you

PR #5231, opened 2023-05-17
Fix java .deb/.rpm versioning

randomanderson:
# What Does This Do
Previously, the version came from `upstream.env` in all cases. The problem was that `UPSTREAM_TRACER_VERSION` also contains the commit hash. The version should be the version of the tag.

Additionally, this moves the package building into its own script because it was getting a bit long to be directly in `.gitlab-ci.yml`.


smola (approved):
@randomanderson Sounds good, but the commit hash could still be included. I think in deb versions this would go in a build suffix (e.g. `+build<commit>`).

PR #5249, opened 2023-05-23
Change span decoration config for interpolation

jpbempel:
Tag values now are using string interpolation like log template

# What Does This Do

# Motivation

# Additional Notes

### DIFF
--- a/dd-java-agent/agent-debugger/src/main/java/com/datadog/debugger/probe/SpanDecorationProbe.java
+++ b/dd-java-agent/agent-debugger/src/main/java/com/datadog/debugger/probe/SpanDecorationProbe.java
@@ -1,13 +1,10 @@
 package com.datadog.debugger.probe;
 
-import static com.datadog.debugger.util.ValueScriptHelper.serializeValue;
-
 import com.datadog.debugger.agent.DebuggerAgent;
 import com.datadog.debugger.agent.Generated;
+import com.datadog.debugger.agent.LogMessageTemplateBuilder;
 import com.datadog.debugger.el.EvaluationException;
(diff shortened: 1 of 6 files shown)


dd-java-agent/agent-debugger/src/main/java/com/datadog/debugger/probe/SpanDecorationProbe.java
```diff
@@ -142,23 +162,20 @@ public void evaluate(CapturedContext context, CapturedContext.Status status) {
...
-        try {
-          Value<?> tagValue = tag.value.execute(context);
-          StringBuilder sb = new StringBuilder();
-          if (tagValue.isUndefined()) {
-            sb.append(tagValue.getValue());
-          } else if (tagValue.isNull()) {
-            sb.append("null");
-          } else {
-            serializeValue(sb, tag.value.getDsl(), tagValue.getValue(), status);
+        LogMessageTemplateBuilder builder = new LogMessageTemplateBuilder(tag.value.getSegments());
```
shatzi, line 165: nit - Suggest rename LogMessageTemplateBuilder to and LogProbe.LogStatus into something more generic that both LogProbe and SpanProbe can use? StringTemplateBuilder & StringTemplateBuilder.Status

shatzi (approved):
Looks good.

dd-java-agent/agent-debugger/src/main/java/com/datadog/debugger/probe/SpanDecorationProbe.java
jpbempel: will rename in a next PR

PR #5252, opened 2023-05-24
[Spark] Compute skew time for spark application / job / stage

paul-laffon-dd:
# What Does This Do

Skew is a common issue in apache spark when the data are not balanced between workers. It is defined as `max - median` tasks duration per stage.

The skew of an application / job is defined as the sum of the skew of its stages. This is an approximation since multiple stages can be running at the same time, but it gives an upper bound of the total skew

## Implementation details

Computing the median is an expensive operation, which can be done
- naively, by keeping all the durations in a list and sorting it
- approximately, like with the https://github.com/DataDog/sketches-java library

Since the number of tasks is generally moderate per stage (default in spark is 200), the overhead of injecting another library + creating the buckets for the approximate algorithm with 1% accuracy outweigh the benefits for 200 tasks, and becomes even around ~10k-20k tasks

### DIFF
--- a/dd-java-agent/instrumentation/spark/src/main/java/datadog/trace/instrumentation/spark/SparkAggregatedTaskMetrics.java
+++ b/dd-java-agent/instrumentation/spark/src/main/java/datadog/trace/instrumentation/spark/SparkAggregatedTaskMetrics.java
@@ -1,11 +1,17 @@
 package datadog.trace.instrumentation.spark;
 
 import datadog.trace.bootstrap.instrumentation.api.AgentSpan;
+import java.util.ArrayList;
+import java.util.Collections;
+import java.util.List;
 import org.apache.spark.TaskFailedReason;
 import org.apache.spark.executor.TaskMetrics;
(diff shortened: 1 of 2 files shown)


dd-java-agent/instrumentation/spark/src/main/java/datadog/trace/instrumentation/spark/SparkAggregatedTaskMetrics.java
```diff
@@ -82,6 +91,23 @@ public void addTaskMetrics(SparkListenerTaskEnd taskEnd) {
...
+    if (taskDurations.size() > 0) {
+      Collections.sort(taskDurations);
+
+      int medianIndex = taskDurations.size() / 2;
+      long median = taskDurations.get(medianIndex);
+      long max = taskDurations.get(taskDurations.size() - 1);
+      skewTime = max - median;
+
+      taskDurations.clear();
     }
```
PerfectSlayer, line 111: Can this part be call concurrently? I fear that `add()`/`clear()` might trigger concurrent issues while running `sort()`.

PR #5256, opened 2023-05-24
[Spark] Separate the error stack trace from the error message

paul-laffon-dd:
# What Does This Do

Before this PR, the full error message from spark (which also contained the stack trace) was added to the `span.setErrorMessage`. Changing that to separate the message from the stack trace in the span so it can be better represented in the UI

Another change is to only send task errors that `countTowardsTaskFailures` as the ones that don't count as a failure tend to be very noisy (`FetchFailed`, `Resubmitted`, ...)

### DIFF
--- a/dd-java-agent/instrumentation/spark/src/main/java/datadog/trace/instrumentation/spark/DatadogSparkListener.java
+++ b/dd-java-agent/instrumentation/spark/src/main/java/datadog/trace/instrumentation/spark/DatadogSparkListener.java
@@ -440,6 +440,10 @@ public class DatadogSparkListener extends SparkListener {
   }
 
   private String getErrorMessageWithoutStackTrace(String errorMessage) {
+    if (errorMessage == null) {
+      return null;
+    }
+
     int stackTraceIndex = errorMessage.indexOf("\tat ");
     if (stackTraceIndex != -1) {
       return errorMessage.substring(0, stackTraceIndex);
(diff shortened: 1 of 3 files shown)


raphaelgavache (approved):
very nice

dd-java-agent/instrumentation/spark/src/main/java/datadog/trace/instrumentation/spark/DatadogSparkListener.java
```diff
@@ -384,4 +412,19 @@ private static String getDatabricksJobRunId(Properties jobProperties) {
...
     return null;
   }
+
+  private String stackTraceToString(Throwable e) {
+    StringWriter stringWriter = new StringWriter();
+    e.printStackTrace(new PrintWriter(stringWriter));
+    return stringWriter.toString();
+  }
+
+  private String getErrorMessageWithoutStackTrace(String errorMessage) {
```
PerfectSlayer, line 422: `errorMessage` can be `null`

PerfectSlayer (changes requested):
Hi Paul,
Does those changes should come with the related tests too?

dd-java-agent/instrumentation/spark/src/main/java/datadog/trace/instrumentation/spark/DatadogSparkListener.java
paul-laffon-dd: Added a safe guard in case `errorMessage` is null

paul-laffon-dd:
Added a test to check the shape of the error type / message / stacktrace

PerfectSlayer:
> Added a test to check the shape of the error type / message / stacktrace

That’s better, thanks!
But the new test case seem to fail in CI:

```
assert span.operationName.toString() == name
       |    |             |          |  |
       |    'spark.job'   'spark.job'|  'spark.application'
       |                             false
       DDSpan [ t_id=7, s_id=8, p_id=6 ] 
```

paul-laffon-dd:
> But the new test case seem to fail in CI:

Yes, I fixed the test in the new version

bantonsson (approved):

PR #5260, opened 2023-05-24
Improve telemetry intervals behavior

smola:
# What Does This Do
Functional changes:
* Telemetry messages should be sent only at the heartbeat interval.
* The metrics interval is used to accumulate metrics, not sending messages.
* Remove all retry logic, no backoff.

Refactor:
* Separate concerns: `TelemetryRunnable` knows about timing, not HTTP requests. `TelemetryService` knows about HTTP requests, not about timing.

# Motivation

# Additional Notes

### DIFF
--- a/telemetry/src/main/java/datadog/telemetry/TelemetryService.java
+++ b/telemetry/src/main/java/datadog/telemetry/TelemetryService.java
@@ -19,8 +19,6 @@ import java.util.ArrayList;
 import java.util.Collections;
 import java.util.List;
 import java.util.Map;
-import java.util.Queue;
-import java.util.concurrent.ArrayBlockingQueue;
 import java.util.concurrent.BlockingQueue;
 import java.util.concurrent.LinkedBlockingQueue;
 import java.util.function.Supplier;
@@ -55,8 +53,6 @@ public class TelemetryService {
 
   private final BlockingQueue<DistributionSeries> distributionSeries =
       new LinkedBlockingQueue<>(1024);
-
-  private final Queue<Request> queue = new ArrayBlockingQueue<>(16);
 
   private boolean sentAppStarted;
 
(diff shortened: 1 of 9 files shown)


telemetry/src/main/java/datadog/telemetry/TelemetryService.java
```diff
@@ -147,87 +119,148 @@ public boolean addDistributionSeries(DistributionSeries series) {
...
+        state.integrations.rollback();
+        break;
+      }
     }
 
-    // New Distributions
-    while (!distributionSeries.isEmpty()) {
-      Payload payload =
-          new Distributions()
+    while (!state.metrics.isEmpty()) {
```
manuel-alvarez-alvarez, line 185: Looks like some method can be extracted from this 5 loops (at least for dealing with the result)

manuel-alvarez-alvarez (approved):

bantonsson (approved):

telemetry/src/main/java/datadog/telemetry/TelemetryService.java
```diff
@@ -147,87 +119,148 @@ public boolean addDistributionSeries(DistributionSeries series) {
...
-              .get()
-              .build(
-                  RequestType.GENERATE_METRICS, payload.requestType(RequestType.GENERATE_METRICS));
-      queue.offer(request);
+    while (!state.integrations.isEmpty()) {
+      final Payload payload =
+          new AppIntegrationsChange()
+              .integrations(state.integrations.get(maxElementsPerReq))
+              .requestType(RequestType.APP_INTEGRATIONS_CHANGE);
+      final SendResult result = sendRequest(RequestType.APP_INTEGRATIONS_CHANGE, payload);
```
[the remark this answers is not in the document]
smola: Right. There is a message-batch request type in the v2 version of the protocol. The idea with the new version is that most of the time, there will be a single message (batch) per interval.

telemetry/src/main/java/datadog/telemetry/TelemetryService.java
smola: Yeah, although I'm reluctant to spend much on abstracting this right now since the right abstraction might change as soon as we get into implementing message-batch.

PR #5267, opened 2023-05-25
[Spark] Instrument YARN application end to not drop traces

paul-laffon-dd:
# What Does This Do

Instrument the spark [yarn.ApplicationMaster.finish()](https://github.com/apache/spark/blob/master/resource-managers/yarn/src/main/scala/org/apache/spark/deploy/yarn/ApplicationMaster.scala#L378) function in order to
- catch the exit code and exit message of the application
- make sure the `spark.application` span is finished before the trace processor is exited

### Capturing exit code

An application can fail during spark computations (when performing operations on the data) or in the user code (like trying to select a column that does not exists)

Failures during spark computations are caught by the SparkListener, but not the errors in the user code so instrumenting the finish function allows to also capture those

### Finish the `spark.application` span

The spark.application span is the local root span and is finished when the event `onApplicationEnd` is received. However if the sparkContext is not stopped explicitly by the user (using `spark.stop()`), spark is stopped in a shutdown hook. In this case, the `onApplicationEnd` event can be received after the trace processor was exited, leading to the trace being dropped, with the following logs

```
23/05/17 13:11:23 INFO ApplicationMaster: Final app status: SUCCEEDED, exitCode: 0
23/05/17 13:11:23 INFO SparkContext: Invoking stop() from shutdown hook
[dd.trace 2023-05-17 13:11:23:205 +0000] [dd-trace-processor] DEBUG datadog.trace.agent.common.writer.TraceProcessingWorker - Datadog trace processor exited. Publishing traces stopped
[dd.trace 2023-05-17 13:11:23:212 +0000] [spark-listener-group-shared] DEBUG datadog.trace.agent.common.writer.RemoteWriter - Trace written after shutdown.. Counted but dropping trace
```

By manually calling the function behind `onApplicationEnd`, we can make sure that the trace is always finished before the tracer is exited

### DIFF
--- a/dd-java-agent/instrumentation/spark/src/main/java/datadog/trace/instrumentation/spark/SparkInstrumentation.java
+++ b/dd-java-agent/instrumentation/spark/src/main/java/datadog/trace/instrumentation/spark/SparkInstrumentation.java
@@ -67,11 +67,6 @@ public class SparkInstrumentation extends Instrumenter.Tracing
     public static void enter(@Advice.Argument(1) int exitCode, @Advice.Argument(2) String msg) {
       if (DatadogSparkListener.listener != null) {
         DatadogSparkListener.listener.finishApplication(System.currentTimeMillis(), exitCode, msg);
-        try {
-          // So that traces have the time to be submitted to the agent before the JVM gets shutdown
-          Thread.sleep(150);
-        } catch (InterruptedException ignored) {
-        }
       }
     }
   }
--- a/dd-java-agent/instrumentation/spark/src/main/java/datadog/trace/instrumentation/spark/SparkInstrumentation.java
+++ b/dd-java-agent/instrumentation/spark/src/main/java/datadog/trace/instrumentation/spark/SparkInstrumentation.java
@@ -10,7 +10,7 @@ import org.apache.spark.SparkContext;
 
 @AutoService(Instrumenter.class)
 public class SparkInstrumentation extends Instrumenter.Tracing
-    implements Instrumenter.ForSingleType {
+    implements Instrumenter.ForKnownTypes {
 
   public SparkInstrumentation() {
     super("spark", "apache-spark");
@@ -22,8 +22,10 @@ public class SparkInstrumentation extends Instrumenter.Tracing
   }
 
   @Override
-  public String instrumentedType() {
-    return "org.apache.spark.SparkContext";
+  public String[] knownMatchingTypes() {
+    return new String[] {
+      "org.apache.spark.SparkContext", "org.apache.spark.deploy.yarn.ApplicationMaster"
+    };
   }
 
   @Override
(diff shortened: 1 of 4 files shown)


dd-java-agent/instrumentation/spark/src/main/java/datadog/trace/instrumentation/spark/SparkInstrumentation.java
```diff
@@ -39,15 +41,38 @@ public String[] helperClassNames() {
...
+      DatadogSparkListener.listener = new DatadogSparkListener(sparkContext);
+      sparkContext.listenerBus().addToSharedQueue(DatadogSparkListener.listener);
+    }
+  }
+
+  public static class YarnFinishAdvice {
+    @Advice.OnMethodEnter(suppress = Throwable.class)
+    public static void enter(@Advice.Argument(1) int exitCode, @Advice.Argument(2) String msg) {
+      if (DatadogSparkListener.listener != null) {
+        DatadogSparkListener.listener.finishApplication(System.currentTimeMillis(), exitCode, msg);
```
raphaelgavache, line 69: Is YarnFinishAdvice run by a separate thread then the SparkListener? If so finishApplication needs to be thread safe

paul-laffon-dd: Yes, it will run in the main thread of the application while the sparkListener will run on another thread so this method has to be thread safe

dd-java-agent/instrumentation/spark/src/main/java/datadog/trace/instrumentation/spark/SparkInstrumentation.java
```diff
@@ -39,15 +41,38 @@ public String[] helperClassNames() {
...
+        try {
+          // So that traces have the time to be submitted to the agent before the JVM gets shutdown
+          Thread.sleep(150);
+        } catch (InterruptedException ignored) {
+        }
+      }
```
PerfectSlayer, line 75: I don’t think this will do what you expect:

* We do already flush the traces in the shutdown hook
* You don’t know if your hook will run after or before the tracer one. So it might be already dead when you call `sleep`.

Moreover, please avoid using `thread.sleep()`. It should be added to _forbidden API_ by the way.

PerfectSlayer (changes requested):

dd-java-agent/instrumentation/spark/src/main/java/datadog/trace/instrumentation/spark/SparkInstrumentation.java
paul-laffon-dd: I removed the sleep since as you said it was not needed. 

The issue with the dropped trace was because the `onApplicationEnd` could be triggered after the processor was exited, with the log `Datadog trace processor exited. Publishing traces stopped`. 

By manually triggering the event before the `finish` function, we can already be sure that the trace is written before the tracer is exited

PerfectSlayer: > I removed the sleep since as you said it was not needed.

I did not say it was not needed, rather than it might not do what you expect.
Hence, it implies we would like to have some tests to back the feature you introduce.
It would ensure it works properly and we can maintain it in the future.

dd-java-agent/instrumentation/spark/src/main/java/datadog/trace/instrumentation/spark/SparkInstrumentation.java
PerfectSlayer: All methods are now synchronized that could lead to performance issue.
What is the thing you are protecting against in the first place?
If only `finishApplication()` can be called from another thread, it might worth it to handle the thread safety at this level only, like using volatile or atomic boolean for `applicationEnded`. `AgentSpan` instance should be fine.

raphaelgavache (approved):

dd-java-agent/instrumentation/spark/src/main/java/datadog/trace/instrumentation/spark/SparkInstrumentation.java
bantonsson: The _biased_ locking (which I think you are alluding to) have been removed from newer versions of the JDK, since the maintenance overhead outweighed the performance gains. That being said, from what I gather, these `synchronized` blocks are not in a performance critical path, so I think it should be fine for the case of simplicity.

dd-java-agent/instrumentation/spark/src/main/java/datadog/trace/instrumentation/spark/SparkInstrumentation.java
bantonsson: I agree with @PerfectSlayer that it would be good to at least have one test. I know that orchestrating a spark test with yarn is out of the question, but a test that checks that the method gets instrumented and the tags added could be possible anyway. I quickly looked at the `ApplicationMaster` and I think it would be possible, even if it's a horrible hack, to _force_ use it without having it orchestrate the whole spark application. Basically just create a `SparkSession` and then create a `ApplicationMaster` and call `finish(...)` on it. The spark/yarn conf supplied to it would need to be faked to make sure it doesn't try to do anything.

bantonsson (approved):

PR #5280, opened 2023-05-26
Changed OBFUSCATION_QUERY_STRING_REGEXP config variable to match the public documentation

nayeem-kamal:
# What Does This Do
Changes the "obfuscation.query.string.regexp" to "trace.obfuscation.query.string.regexp" which is the publicly documented name.
# Motivation
This is to match the config functionality to the public docs
# Additional Notes

### DIFF
--- a/dd-trace-api/src/main/java/datadog/trace/api/config/TraceInstrumentationConfig.java
+++ b/dd-trace-api/src/main/java/datadog/trace/api/config/TraceInstrumentationConfig.java
@@ -90,7 +90,7 @@ public final class TraceInstrumentationConfig {
 
   public static final String IGNITE_CACHE_INCLUDE_KEYS = "ignite.cache.include_keys";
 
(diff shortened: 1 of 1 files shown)


mcculls (approved):
+1

We should probably also change [this line](https://github.com/DataDog/dd-trace-java/blob/master/internal-api/src/main/java/datadog/trace/api/Config.java#L1466) in `Config.java` from:
```
configProvider.getString(OBFUSCATION_QUERY_STRING_REGEXP);
```
to:
```
configProvider.getString(OBFUSCATION_QUERY_STRING_REGEXP, null, "obfuscation.query.string.regexp");
```
so any people using the previous key are not affected by this change

If we decide not to add this alias then we should call out this change in the release notes.

nayeem-kamal:
I missed Stuart's comment above but I have added an alias to prevent a breaking change: https://github.com/DataDog/dd-trace-java/pull/5288

PR #5117, opened 2023-04-26
Insecure cookies detection for servlet

DDJavierSantos:
# What Does This Do
Instrument HttpServletRequest so we can examine cookies and make sure they have the Secure flag enabled.

# Motivation
This enables IAST to report insecure cookie vulnerabilities

# Additional Notes

PR #5285, opened 2023-05-30
Fix addCookie instrumentation in Jetty

DDJavierSantos:
# What Does This Do
Fix a problem in cookie instrumentation where it would not instrument jetty response object because addCookie signature takes a different cookie class.

# Motivation
To make sure Response is always instrumented

# Additional Notes
See https://github.com/DataDog/dd-trace-java/pull/5117

### DIFF
--- a/dd-java-agent/instrumentation/servlet-common/src/main/java/datadog/trace/instrumentation/servlet/HttpServletResponseInstrumentation.java
+++ b/dd-java-agent/instrumentation/servlet-common/src/main/java/datadog/trace/instrumentation/servlet/HttpServletResponseInstrumentation.java
@@ -1,23 +1,22 @@
 package datadog.trace.instrumentation.servlet;
 
-import com.google.auto.service.AutoService;
-import datadog.trace.agent.tooling.Instrumenter;
-import datadog.trace.api.iast.InstrumentationBridge;
-import datadog.trace.api.iast.propagation.PropagationModule;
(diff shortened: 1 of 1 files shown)


smola (approved):

dd-java-agent/instrumentation/servlet-common/src/main/java/datadog/trace/instrumentation/servlet/HttpServletResponseInstrumentation.java
```diff
@@ -54,11 +54,14 @@ public void adviceTransformations(AdviceTransformation transformation) {
...
   public static class AddCookieAdvice {
     @Advice.OnMethodEnter(suppress = Throwable.class)
-    public static void onEnter(@Advice.Argument(0) final Cookie cookie) {
+    public static void onEnter(@Advice.Argument(0) final Object o) {
       final InsecureCookieModule module = InstrumentationBridge.INSECURE_COOKIE;
       if (module != null) {
-        if (null != cookie) {
-          module.onCookie(cookie.getName(), cookie.getSecure());
+        if (null != o) {
+          if (o instanceof Cookie) {
```
smola, line 61: It'd be better to just instrument the methods that take `Cookie` as a parameter. There's name matcher helpers to pass types as strings rather than `Class` objects.

smola (approved):

PR #5289, opened 2023-05-30
Add exploration tests for debugger

jpbempel:
# What Does This Do
Special mode for dd-java-agent instrumenting every methods in loaded class. We run tests from jackson library with dd-trace-java and this mode enabled and expect no error.
Running those tests with Gitlab. Each job generate an agent.log file containing debug logs of instrumentation

# Motivation

# Additional Notes

### DIFF
--- a/.gitlab-ci.yml
+++ b/.gitlab-ci.yml
@@ -2,12 +2,14 @@ include:
   - remote: https://gitlab-templates.ddbuild.io/libdatadog/include/ci_authenticated_job.yml
   - remote: https://gitlab-templates.ddbuild.io/apm/packaging.yml
   - local: ".gitlab/benchmarks.yml"
+  - local: ".gitlab/exploration-tests.yml"
 
 stages:
   - build
   - package
   - deploy
   - benchmarks
+  - exploration-tests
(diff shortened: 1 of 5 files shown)


shatzi (approved):
I have tested and it seem to run the the tests.

However, there are few things that I find missing here:
1. in the output of gitlab job there is no mention that instrumentation happened, so while I assume it did - there is no way to know.  Can we add confirmation and stats about how many probes were triggered ?
2. For benchmark there is a message added to PR with results. Maybe we want to add message that exploration tests also ran and passed? not sure if that is needed.

jpbempel:
> 1. in the output of gitlab job there is no mention that instrumentation happened, so while I assume it did - there is no way to know.  Can we add confirmation and stats about how many probes were triggered ?

Something I missed in the description: there is an artifact associated to the run `agent.log` that includes all debug logs of instrumentation

> 2. For benchmark there is a message added to PR with results. Maybe we want to add message that exploration tests also ran and passed? not sure if that is needed.

Right now exploration-tests are run manually

smola (approved):

PR #5302, opened 2023-06-01
allow removal of fake services in v0

amarziali:
# What Does This Do

This PR allow to flatten every service name to `DD_SERVICE` also in v0 if the env `DD_TRACE_REMOVE_CLIENT_SERVICE_NAMES_ENABLED` is set to true

Please note that having this activated (like happening in v1) implies:
* legacy tracing disabled (otherwise we'll have fake services like `kafka`)
*  time in queue disabled (same reasons)

I took the time hence to refactor the way we determined if legacy tracing was enabled in the decorators by getting rid of checks like `schemaVersion == 0` that can cause nasty coupling

# Motivation

# Additional Notes


ygree (approved):
LGTM

PR #5306, opened 2023-06-01
Fix handling of local variable table

jpbempel:
# What Does This Do
Add protection about array out of bounds

# Motivation
in some cases there is no match between method arguments and local variable tables (hidden arg, inherited method with no info in local variable table)

# Additional Notes

### DIFF
--- a/dd-java-agent/agent-debugger/src/main/java/com/datadog/debugger/instrumentation/Instrumentor.java
+++ b/dd-java-agent/agent-debugger/src/main/java/com/datadog/debugger/instrumentation/Instrumentor.java
@@ -87,14 +87,16 @@ public abstract class Instrumentor {
     // as stated into the JVM spec:
     // https://docs.oracle.com/javase/specs/jvms/se8/html/jvms-2.html#jvms-2.6.1
     // so we reassigned local var in arg slots if they are empty
-    int slot = isStatic ? 0 : 1;
-    int localVarTableIdx = slot;
-    for (Type t : argTypes) {
-      if (localVars[slot] == null) {
-        localVars[slot] = sortedLocalVars.get(localVarTableIdx);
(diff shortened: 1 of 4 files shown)


dd-java-agent/agent-debugger/src/main/java/com/datadog/debugger/instrumentation/Instrumentor.java
```diff
@@ -87,14 +87,16 @@ private LocalVariableNode[] extractLocalVariables(Type[] argTypes) {
     // as stated into the JVM spec:
     // https://docs.oracle.com/javase/specs/jvms/se8/html/jvms-2.html#jvms-2.6.1
     // so we reassigned local var in arg slots if they are empty
-    int slot = isStatic ? 0 : 1;
-    int localVarTableIdx = slot;
-    for (Type t : argTypes) {
-      if (localVars[slot] == null) {
-        localVars[slot] = sortedLocalVars.get(localVarTableIdx);
+    if (argTypes.length < localVars.length) {
```
shatzi, line 90: not sure what this does... let assume argTypes lengths is == localVars.length ? is that ok? or only if iStatic true?

if argTypes is bigger than localVars - should still try to find those args for the first few? what the implications of not running this logic at all?

shatzi (approved):
not sure I follow the logic here.. but I see what is expected from the unit-test.

dd-java-agent/agent-debugger/src/main/java/com/datadog/debugger/instrumentation/Instrumentor.java
jpbempel: we try best effort to map method args with local variable tables, however there are cases where args are not declared in the local var table because like for enums we are inheriting from a class where there is no local var table, hence no mapping

PR #5308, opened 2023-06-02
Spring mvc unvalidated redirect instrumentation

jandro996:
work in progress


internal-api/src/main/java/datadog/trace/api/iast/sink/UnvalidatedRedirectModule.java
```diff
@@ -7,5 +7,7 @@ public interface UnvalidatedRedirectModule extends HttpHeaderModule {
 
   void onRedirect(@Nullable String value);
 
+  void onRedirect(@Nullable String value, @Nullable String clazz, @Nullable String method);
```
manuel-alvarez-alvarez, line 10: Can `value` really be null?, what happens if `clazz` and `method` are null? should we traverse the stack as by default?

dd-java-agent/agent-iast/src/main/java/com/datadog/iast/model/Location.java
```diff
@@ -21,6 +21,11 @@ public static Location forSpanAndStack(final long spanId, final StackTraceElemen
     return new Location(spanId, stack.getClassName(), stack.getLineNumber(), stack.getMethodName());
   }
 
+  public static Location forSpanAndClassAndMethod(
```
manuel-alvarez-alvarez, line 24: We have to review the hashing code in 'com.datadog.iast.model.Vulnerability' to include the method if available.

dd-smoke-tests/springboot/src/test/groovy/datadog/smoketest/IastSpringBootRedirectSmokeTest.groovy
```diff
@@ -73,4 +73,82 @@ class IastSpringBootRedirectSmokeTest extends AbstractIastServerSmokeTest {
     then:
     hasVulnerabilityInLogs { vul -> vul.type == 'UNVALIDATED_REDIRECT' && vul.location.method == 'unvalidatedRedirectFromForward' }
   }
+
+  def "unvalidated  redirect from RedirectView is present"() {
```
manuel-alvarez-alvarez, line 77: Can we try to work together and merge these tests in 'datadog.smoketest.AbstractIastServerSmokeTest'? this way we will test redirections in all spring versions.

internal-api/src/main/java/datadog/trace/api/iast/sink/UnvalidatedRedirectModule.java
jandro996: You are right! I've checked it and they can't be null 👍

dd-smoke-tests/springboot/src/test/groovy/datadog/smoketest/IastSpringBootRedirectSmokeTest.groovy
jandro996: Do you mean `datadog.smoketest.AbstractIastSpringBootTest`? I put the redirection tests here and they are running ok for  spring-boot.2.6-mvc too

Do you want to extend it to all spring-boot version? maybe this could be done in a different PR, I think it is a bit far from the current PR scope (in `datadog.smoketest.AbstractIastSpringBootTest` there are tests for almost all implemented vulnerabilities)

dd-java-agent/agent-iast/src/main/java/com/datadog/iast/model/Location.java
jandro996: I've changed the `com.datadog.iast.mode.VulnerabilityType.calculateHash(Vulnerability vulnerability)` including the method if it's available. Is enough with that?

smola (approved):

manuel-alvarez-alvarez (approved):

PR #5309, opened 2023-06-02
Fix NullPointerException in map/list serialization

jpbempel:
# What Does This Do

# Motivation
if null elements are present in map or collections, serialization was throwing NullPointerException

# Additional Notes


ojung (approved):
👍

PR #5310, opened 2023-06-02
[Spark] Compute available executor time per stage

paul-laffon-dd:
# What Does This Do

The available executor time is defined as `sum(executor uptime * number of cores)` it is already computed for the whole spark application

This PR introduce the computation to attribute it for each stage
- when only one stage is running, the executor time of the full executor time is attributed to the stage
- when multiple stages are running at the same time, the executor time attribution is weighted by the run time of each stage 

executor_time_attribution

### DIFF
--- a/dd-java-agent/instrumentation/spark/src/main/java/datadog/trace/instrumentation/spark/DatadogSparkListener.java
+++ b/dd-java-agent/instrumentation/spark/src/main/java/datadog/trace/instrumentation/spark/DatadogSparkListener.java
@@ -338,9 +338,9 @@ public class DatadogSparkListener extends SparkListener {
 
     if (taskEnd.taskMetrics() != null) {
       // Record the runtime in each active stage in order to allocate the available executor time
-      long executorRunTime = taskEnd.taskMetrics().executorRunTime();
+      long taskRunTime = SparkAggregatedTaskMetrics.computeTaskRunTime(taskEnd.taskMetrics());
       for (SparkAggregatedTaskMetrics aggMetrics : stageMetrics.values()) {
-        aggMetrics.recordExecutorRunTime(executorRunTime);
+        aggMetrics.recordTotalTaskRunTime(taskRunTime);
       }
     }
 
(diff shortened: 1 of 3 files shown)


dd-java-agent/instrumentation/spark/src/main/java/datadog/trace/instrumentation/spark/DatadogSparkListener.java
```diff
@@ -308,29 +322,27 @@ public synchronized void onStageCompleted(SparkListenerStageCompleted stageCompl
...
-        .computeIfAbsent(stageSpanKey, k -> new SparkAggregatedTaskMetrics())
-        .addTaskMetrics(taskEnd);
+    SparkAggregatedTaskMetrics stageMetric = stageMetrics.get(stageSpanKey);
+    if (stageMetric != null) {
+      stageMetric.addTaskMetrics(taskEnd);
+    }
+
+    if (taskEnd.taskMetrics() != null) {
+      // Record the runtime in each active stage in order to allocate the available executor time
+      long executorRunTime = taskEnd.taskMetrics().executorRunTime();
```
raphaelgavache, line 341: this doesn't take in account resultSerializationTime and executorDeserialize time

raphaelgavache (approved):

dd-java-agent/instrumentation/spark/src/main/java/datadog/trace/instrumentation/spark/DatadogSparkListener.java
paul-laffon-dd: You are right, it does not take the `resultSerializationTime` and `executorDeserialize`. But since the total task duration is also including those, I think we should take them into account

Added a new commit to use the `executorRunTime + resultSerializationTime + executorDeserialize` to compute the allocation

utils/test-utils/src/main/groovy/datadog/trace/test/util/DDSpecification.groovy
```diff
@@ -90,7 +90,7 @@ abstract class DDSpecification extends Specification {
 
   void setupSpec() {
     assert !configModificationFailed: "Config class modification failed.  Ensure all test classes extend DDSpecification"
-    assert System.getenv().findAll { it.key.startsWith("DD_") }.isEmpty()
+    //assert System.getenv().findAll { it.key.startsWith("DD_") }.isEmpty()
```
bantonsson, line 93: This should not be commented out, and the same goes for the three other instances below.

bantonsson (changes requested):

utils/test-utils/src/main/groovy/datadog/trace/test/util/DDSpecification.groovy
paul-laffon-dd: Yes, my bad, I didn't realize that I pushed that 🤦 I removed it

bantonsson (approved):

PR #5318, opened 2023-06-05
Fix Collection/Map serialization

jpbempel:
# What Does This Do
Catch exception and report it safely into the snapshot

# Motivation
Some implementation of collections or maps may throw exceptions in somce situations which can happen during our serialization process 

# Additional Notes

### DIFF
--- a/dd-java-agent/agent-debugger/src/main/java/com/datadog/debugger/util/SerializerWithLimits.java
+++ b/dd-java-agent/agent-debugger/src/main/java/com/datadog/debugger/util/SerializerWithLimits.java
@@ -7,6 +7,7 @@ import java.lang.reflect.Array;
 import java.lang.reflect.Field;
 import java.lang.reflect.Modifier;
 import java.util.Collection;
+import java.util.Collections;
 import java.util.Iterator;
 import java.util.Map;
 import java.util.Set;
@@ -75,7 +76,7 @@ public class SerializerWithLimits {
 
     void mapEntryEpilogue(Map.Entry<?, ?> entry) throws Exception;
 
-    void mapEpilogue(Map<?, ?> map, boolean isComplete) throws Exception;
+    void mapEpilogue(boolean isComplete, int size) throws Exception;
 
     void objectPrologue(Object value) throws Exception;
 
@@ -98,6 +99,8 @@ public class SerializerWithLimits {
     void objectEpilogue(Object value) throws Exception;
 
     void notCaptured(NotCapturedReason reason) throws Exception;
+
+    void notCaptured(String reason) throws Exception;
   }
 
   private final TokenWriter tokenWriter;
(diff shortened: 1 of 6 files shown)


dd-java-agent/agent-debugger/src/main/java/com/datadog/debugger/util/SerializerWithLimits.java
```diff
@@ -166,15 +169,33 @@ public void serialize(Object value, String type, Limits limits) throws Exception
...
       tokenWriter.collectionPrologue(value);
-      Collection<?> col = (Collection<?>) value;
-      boolean isComplete = serializeCollection(col, limits);
-      tokenWriter.collectionEpilogue(value, isComplete, col.size());
+      Collection<?> col;
+      boolean isComplete = true;
+      int size = 0;
+      try {
+        col = (Collection<?>) value;
+        size = col.size(); // /!\ alien call /!\
```
ojung, line 177: Where is the synchronization that makes this an alien call or do you mean something else than that? Btw, how come you're adding the comments now? Did we run into any deadlock?

dd-java-agent/agent-debugger/src/test/java/com/datadog/debugger/agent/SnapshotSerializationTest.java
```diff
@@ -717,6 +718,56 @@ public void capturedValueAdapterNull() {
     Assertions.assertEquals("null", capturedValueAdapter.toJson(null));
   }
 
+  @Test
+  public void collectionValueThrows() throws IOException {
+    JsonAdapter<Snapshot> adapter = createSnapshotAdapter();
+    Snapshot snapshot = createSnapshot();
+    CapturedContext context = new CapturedContext();
+    CapturedContext.CapturedValue mapField =
```
ojung, line 726: Var should probably be called `listField`.

dd-java-agent/agent-debugger/src/test/java/com/datadog/debugger/agent/SnapshotSerializationTest.java
```diff
@@ -717,6 +718,56 @@ public void capturedValueAdapterNull() {
...
+            new ArrayList<String>() {
+              @Override
+              public int size() {
+                throw new UnsupportedOperationException();
+              }
+            });
+    context.addFields(new CapturedContext.CapturedValue[] {mapField});
+    snapshot.setExit(context);
+    String buffer = adapter.toJson(snapshot);
+    System.out.println(buffer);
```
ojung, line 739: Left-over or intended?

ojung (approved):
Thanks, just a couple of questions / minor things.

dd-java-agent/agent-debugger/src/main/java/com/datadog/debugger/util/SerializerWithLimits.java
jpbempel: "Alien call" means we are calling an external code that we do not control and can be throwing whatever exception. see unit tests.
There is no synchronization issue per se, but I am reusing the same terminology as this is similar issue here.

dd-java-agent/agent-debugger/src/test/java/com/datadog/debugger/agent/SnapshotSerializationTest.java
jpbempel: intended, for serialization tests I am keeping the println for troubleshooting purpose

dd-java-agent/agent-debugger/src/test/java/com/datadog/debugger/agent/SnapshotSerializationTest.java
jpbempel: ah yes, good catch

ojung (approved):

PR #5320, opened 2023-06-05
Fix TestNG onBeforeClass/onAfterClass notifications for cases when only a portion of class' test methods is executed

nikita-tkachenko-datadog:
# What Does This Do
Fixes TestNG instrumentation for scenarios when not all of the test case methods in a class are executed.

# Motivation
It is possible to run Test NG with an explicit XML configuration, specifying which exact test cases need to be executed.
And it is possible to specify only a portion of the test cases in any given class, leaving out some test cases.
In these scenarios current instrumentation will not work correctly: `onAfterClass` callback will not be executed because it waits for all test cases in the class to run.

# Additional Notes
The fix is to get the list of executed cases from TestNG's `ISuite`, and use that to determine when a class has finished executing.


smola (approved):
The branch is fairly stale, I'd suggest a rebase before merging.

nikita-tkachenko-datadog:
> The branch is fairly stale, I'd suggest a rebase before merging.

Good point, rebased

PR #5323, opened 2023-06-06
merge Release/v1.15.x into master

amarziali:
# What Does This Do

# Motivation

# Additional Notes


smola:
@amarziali The NPE and TestNG commits are already in master and were cherry-picked to the release branch, so only the disable v05 one needs to be applied.

PR #5339, opened 2023-06-07
[serverless] json serializer skips all unsupported types

joeyzhao2018:
# What Does This Do
Make the JsonAdapter used in serverless cases skip all unsupported types

# Background
### How we do distributed tracing in Serverless for Java
In aws lambda case, we run the serverless version of the agent using lambda extension. 
And at the beginning of the customer's lambda function, it need to notify the agent with the event payload via the [notifyStartInvocation() method](https://github.com/DataDog/dd-trace-java/blob/d9f0e171e1de47588af8ee4bb8e360930355e081/dd-trace-core/src/main/java/datadog/trace/lambda/LambdaHandler.java#L64).
When the customer's lambda function finish running, it need to notify the agent with the result via the [notifyEndInvocation() method.](https://github.com/DataDog/dd-trace-java/blob/d9f0e171e1de47588af8ee4bb8e360930355e081/dd-trace-core/src/main/java/datadog/trace/lambda/LambdaHandler.java#L106)
In both cases, we need a Moshi json adapter to serialize the payload.
# Motivation

- A customer reported that their customized request payload has been breaking the tracer in aws lambda environment.
- Every now and then there's a "new" unsupported types' serialization problem.

# Implementation explanation
- In both serialization cases mentioned above, we don't want to fail the request and we just want to skip anything we don't know how to serialize and write `{}`.
- Currently, `SkipAbstractTypeJsonSerializer` is inserted at the beginning of the adapter factories list. So it tries to catch unsupported types before trying it with the Moshi's built-in factories. But when it fails to recognize a new unsupported type, it would end up throwing an `IllegalArgumentException` and break the tracer.
- Ideally, we just want to append our JsonAdapter's factory after the Moshi built-in factories instead of[ before them](https://github.com/square/moshi/blob/master/moshi/src/main/java/com/squareup/moshi/Moshi.kt#L41-L42). But there's no easy way to do it as the type and methods are all `final` in java. So we just use`moshi.nextAdapter(this, requestedType, annotations);` to look through those build-in adapter factories and if none found it would throw `IllegalArgumentException` and then we use the `SkipUnsupportedTypeJsonAdapter` to handle.
- Renamed the `SkipAbstractTypeJsonSerializer` class to `SkipUnsupportedTypeJsonAdapter` since now it skips all unsupported types.


bantonsson (approved):

mcculls (commented):
LGTM

PR #5343, opened 2023-06-08
DRAFT

jpbempel:
# What Does This Do

# Motivation

# Additional Notes


dd-trace-api/src/main/java/datadog/trace/api/config/DebuggerConfig.java
```diff
@@ -21,6 +21,7 @@ public final class DebuggerConfig {
   public static final String DEBUGGER_INSTRUMENT_THE_WORLD =
       "dynamic.instrumentation.instrument.the.world";
   public static final String DEBUGGER_EXCLUDE_FILE = "dynamic.instrumentation.exclude.file";
+  public static final String DEBUGGER_CAPTURE_TIMEOUT = "dynamic.instrumentation.capture.timeout";
```
shatzi, line 24: is that already defined in debugger spec? suggest sync with team so everyone have the same config.

dd-java-agent/agent-debugger/src/main/java/com/datadog/debugger/util/ValueScriptHelper.java
```diff
@@ -1,20 +1,22 @@
...
 import datadog.trace.bootstrap.debugger.util.TimeoutChecker;
+import java.time.Duration;
+import java.time.temporal.ChronoUnit;
 
 public class ValueScriptHelper {
   private static final Limits LIMITS = new Limits(1, 3, 255, 5);
 
   public static void serializeValue(
       StringBuilder sb, String expr, Object value, CapturedContext.Status status) {
+    Duration timeout = Duration.of(Config.get().getDebuggerCaptureTimeout(), ChronoUnit.MILLIS);
```
shatzi, line 16: suggest create Duration only once, when the config is set. Maybe getDebuggerCaptureTimeout should return a Duration object and not a number?

shatzi (approved):

dd-java-agent/agent-debugger/src/main/java/com/datadog/debugger/util/ValueScriptHelper.java
jpbempel: I am a bit reluctant to introduce `Duration` in the shared class `Config` as it is not used for now. I think this is a legacy because dd-trace-java was supporting JDK 7 and `Duration` was introduced in JDK8.
Let's move it if we measure a serious impact on this

PR #5345, opened 2023-06-08
IAST no HttpOnly cookie vulnerability detection

DDJavierSantos:
# What Does This Do
Implements detection of the NoHttpOnlyCookie vulnerability (a cookie should have the HttpOnly flag set).

# Motivation
We need to report all of the http header vulnerabilities in IAST.

# Additional Notes

### DIFF
--- a/internal-api/src/main/java/datadog/trace/api/iast/InstrumentationBridge.java
+++ b/internal-api/src/main/java/datadog/trace/api/iast/InstrumentationBridge.java
@@ -4,8 +4,10 @@ import datadog.trace.api.iast.propagation.CodecModule;
 import datadog.trace.api.iast.propagation.PropagationModule;
 import datadog.trace.api.iast.propagation.StringModule;
 import datadog.trace.api.iast.sink.CommandInjectionModule;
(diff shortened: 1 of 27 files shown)


internal-api/src/main/java/datadog/trace/api/iast/InstrumentationBridge.java
```diff
@@ -29,9 +31,12 @@ public abstract class InstrumentationBridge {
   public static volatile LdapInjectionModule LDAP_INJECTION;
   public static volatile PropagationModule PROPAGATION;
   public static volatile InsecureCookieModule INSECURE_COOKIE;
+  public static volatile NoHttpOnlyCookieModule NO_HTTP_ONLY_COOKIE;
```
smola, line 34: This vulnerability name is landing as `NO_HTTPONLY_COOKIE` in the spec.

manuel-alvarez-alvarez (approved):

internal-api/src/main/java/datadog/trace/api/iast/InstrumentationBridge.java
DDJavierSantos: Applied

PR #5352, opened 2023-06-09
Unvalidated redirect vertx-3.4

jandro996:
# What Does This Do

- Add  new advice to`io.vertx.ext.web.impl.RoutingContextImpl.reroute(HttpMethod, path)`
- Modify HttpServerRequestInstrumentation
- Rework instrumentation unit test, we decide to start up a vertx server instead of mocking 

# Motivation

Add unvalidated redirect support for vertx 3.4 framework

# Additional Notes

Future PRs will increase coverage for unvalidated redirect in other vertx versions


dd-smoke-tests/vertx-3.9/application/src/main/java/datadog/smoketest/vertx_3_9/IastHandler.java
jandro996, line 1: We need to update the Handler in 3.9 to not break the tests (new tests cases were added to dd-smoke-tests/iast-util/src/testFixtures/groovy/datadog/smoketest/AbstractIastVertxSmokeTest.groovy)

smola (approved):

dd-java-agent/instrumentation/vertx-web-3.4/src/test/groovy/server/IastSourceTest.groovy
```diff
@@ -0,0 +1,74 @@
+package server
```
manuel-alvarez-alvarez, line 1: It would be super nice if you could remove as well `HttpServerRequestInstrumentationTest` and add the sources to this test.

manuel-alvarez-alvarez (approved):

jandro996 (commented):
Thanks!! @manuel-alvarez-alvarez

PR #5355, opened 2023-06-09
Less verbose AppSec start log

smola:
# What Does This Do
By default (AppSec inactive), log to debug instead of info. This prevents getting an uninteresting (by default) early log line with:

>  AppSec is ENABLED_INACTIVE with powerwaf(libddwaf: 1.10.0)

# Motivation

# Additional Notes


ivoanjo:
Thank you for the nice tiny fix! :)

cataphract (approved):

PR #5359, opened 2023-06-09
Add priorities for server set errors

amarziali:
# What Does This Do

Similarly to resource names, it adds priority based error flag settings.

The http servers will set error using a lower priority than the user one. The errors set elsewhere will stay untouched (but it should be changed in the future as well).

It solves a particular two use cases: 
* The error set by the user should have higher prio than the one set according to a http response status code
* The application server throws an error but, despite logged, the error flag should be set according to the http status code

# Motivation

# Additional Notes


dd-java-agent/agent-bootstrap/src/main/java/datadog/trace/bootstrap/instrumentation/decorator/HttpServerDecorator.java
```diff
@@ -363,6 +365,16 @@ private AgentSpan.Context.Extracted callIGCallbackStart(AgentSpan.Context.Extrac
     return context;
   }
 
+  @Override
+  public AgentSpan onError(final AgentSpan span, final Throwable throwable) {
```
ygree, line 369: I can't see any usages of this method

amarziali: it's used by Subclasses (e.g. `datadog.trace.instrumentation.tomcat.TomcatDecorator#onResponse` calls `onError`).
Being overridded now http server based decorators are inheriting this behavior

PR #5362, opened 2023-06-09
get stack depth from Throwable stackTrace to avoid copy

richardstartin:
# What Does This Do

This addresses a fairly large source of overhead in exception profiling when applications throw lots of exceptions: 
Screenshot 2023-06-09 at 18 59 55

The instrumentation will only activate if the `Throwable` class has a field called `stackTrace` (rather than crash) which means exception profiling would be unavailable, but this is unlikely ever to happen.

# Motivation

# Additional Notes


jbachorik (commented):
Considering that the field name hasn't changed for many years and it is in the Java standard libraries which are shared amongst various JVM implementations I would be less pessimistic about this breaking in the foreseeable future.

jbachorik (approved):

PR #5375, opened 2023-06-13
expose cassandra contact points and use for peer.service

amarziali:
# What Does This Do

Surface cassandra contact points on a tag `db.cassandra.contact.points` and use it as a privileged precursor for cassandra peer.service

# Motivation

# Additional Notes


settings.gradle
```diff
@@ -176,6 +176,7 @@ include ':dd-java-agent:instrumentation:couchbase:couchbase-3.1'
 include ':dd-java-agent:instrumentation:couchbase:couchbase-3.2'
 include ':dd-java-agent:instrumentation:datanucleus-4'
 include ':dd-java-agent:instrumentation:datastax-cassandra-3'
+include ':dd-java-agent:instrumentation:datastax-cassandra-3_8'
```
PerfectSlayer, line 179: Can you rename it with `3.8` rather than `3_8`?

dd-java-agent/instrumentation/datastax-cassandra-4/src/main/java/datadog/trace/instrumentation/datastax/cassandra4/ContactPointsUtil.java
```diff
@@ -0,0 +1,36 @@
...
+              .map(EndPoint::resolve)
+              .map(InetSocketAddress.class::cast)
+              .map(
+                  inetSocketAddress -> {
+                    if (inetSocketAddress.getPort() > 0) {
+                      return inetSocketAddress.getHostString() + ":" + inetSocketAddress.getPort();
+                    }
+                    return inetSocketAddress.getHostString();
+                  })
+              .collect(Collectors.joining(","));
```
PerfectSlayer, line 28: Are they implicitly distinct?

dd-java-agent/instrumentation/datastax-cassandra-4/src/main/java/datadog/trace/instrumentation/datastax/cassandra4/ContactPointsUtil.java
```diff
@@ -0,0 +1,36 @@
...
+
+public class ContactPointsUtil {
+  private static final DDCache<Set<EndPoint>, String> CONTACT_POINT_CACHE =
+      DDCaches.newFixedSizeCache(8);
+
+  private static final Function<Set<EndPoint>, String> ADDER =
+      endPoints ->
+          endPoints.stream()
+              .map(EndPoint::resolve)
+              .map(InetSocketAddress.class::cast)
```
PerfectSlayer, line 20: What about using `filter()` to check it implements `InetSocketAddress` before casting it?

PerfectSlayer (commented):
Few inline questions.

About the tests, there is some `System.setProperty("java.net.preferIPv4Stack", "true")` but nothing related to IPv6. Should it be tested too?

PerfectSlayer (changes requested):
Actually, I request to change a package name 😓

dd-java-agent/instrumentation/datastax-cassandra-4/src/main/java/datadog/trace/instrumentation/datastax/cassandra4/ContactPointsUtil.java
amarziali: ok better safe than sorry :)

PerfectSlayer (approved):

PR #5377, opened 2023-06-13
Wait for traces to be published before exiting if CI Visibility is enabled

nikita-tkachenko-datadog:
# What Does This Do
This PR changes the logic that publishes traces: if CI Visibility is enabled, the thread that does the publishing is not marked as daemon, and upon receiving an interrupt request the thread will wait up to 5 seconds to ensure all pending traces were published.

# Motivation
CI Visibility differs from production APM: when a customer shuts down a production service, it is acceptable for some operations that happened right before the shutdown to be missing in Datadog. When running tests, however, customers expect every executed test to be visible in Datadog. That means there's a need to wait until publishing is finished, even when all the tests have executed and JVM is ready to shut down.


ygree (approved):
LGTM

PR #5379, opened 2023-06-14
Fix `this` duplicate serialization in snapshot

jpbempel:
# What Does This Do
Remove duplicate serialization of `this`

# Motivation
`this` instance was serialized twice by instrumentation adding this as
 argument and by explicit serialization in json adapters.
This duplicate was deduplicated automatically when parsed on the backend side

# Additional Notes
Fix unit tests that were incorrect and cannot use directly fields anymore

### DIFF
--- a/dd-java-agent/agent-debugger/debugger-bootstrap/src/main/java/datadog/trace/bootstrap/debugger/ProbeRateLimiter.java
+++ b/dd-java-agent/agent-debugger/debugger-bootstrap/src/main/java/datadog/trace/bootstrap/debugger/ProbeRateLimiter.java
@@ -52,6 +52,11 @@ public class ProbeRateLimiter {
     setGlobalSnapshotRate(DEFAULT_GLOBAL_LOG_RATE);
   }
 
+  public static void resetAll() {
+    PROBE_SAMPLERS.clear();
+    resetGlobalRate();
+  }
+
   private static AdaptiveSampler createSampler(double rate) {
     if (rate < 1) {
       int intRate = (int) Math.round(rate * 10);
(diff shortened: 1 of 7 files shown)


dd-java-agent/agent-debugger/debugger-bootstrap/src/main/java/datadog/trace/bootstrap/debugger/ProbeRateLimiter.java
```diff
@@ -52,6 +52,11 @@ public static void resetGlobalRate() {
     setGlobalSnapshotRate(DEFAULT_GLOBAL_LOG_RATE);
   }
 
+  public static void resetAll() {
```
ojung, line 55: Is this related to the PR?

dd-java-agent/agent-debugger/src/test/java/com/datadog/debugger/agent/SnapshotSerializationTest.java
```diff
@@ -96,18 +97,18 @@ public void roundTripCapturedValue() throws IOException, URISyntaxException {
     JsonAdapter<Snapshot> adapter = createSnapshotAdapter();
     Snapshot snapshot = createSnapshot();
     CapturedContext context = new CapturedContext();
-    CapturedContext.CapturedValue normalValuedField =
-        CapturedContext.CapturedValue.of("normalValuedField", String.class.getTypeName(), "foobar");
-    CapturedContext.CapturedValue normalNullField =
-        CapturedContext.CapturedValue.of("normalNullField", String.class.getTypeName(), null);
+    CapturedContext.CapturedValue normalValuedLocal =
```
ojung, line 100: Why are you changing fields to locals here and below?

ojung (approved):
Nice, good job fixing this 😂

dd-java-agent/agent-debugger/src/test/java/com/datadog/debugger/agent/SnapshotSerializationTest.java
jpbempel: because I cannot put directly fields like before need to generate a this argument to put fields, which more cumbersome, does not change the point of the test in this case

PR #5382, opened 2023-06-14
WIP: Span metrics

PerfectSlayer:
# What Does This Do

1. Introduce a metrics API for core tracer,
2. Use it for span creation and finition telemetry.

# Motivation

A generic metrics API allow gathering metrics from core and dispatch them later to either health metrics or telemetry (or both).
It should also allow deviated metrics in a longer term. Current metric implementations are either for client or telemetric but isn't generic.

# Additional Notes

This PR is still in WIP. I will split it in two distinct PRs later: one for the metric API, another for span metrics.
For now, the new metric API is only directly linked to telemetry using the recently supported periodic action for telemetry metric payload.


PerfectSlayer:
The whole feature was merged to the linked PR.
This PR still proposes a new metric API for history/reference.

PR #5383, opened 2023-06-14
Added NPE checkers in Instrumentation Gateway

ValentinZakharov:
# What Does This Do
The temporary solution. Added Null checker to avoid NPE.

# Motivation
Reported NPE:
```
java.lang.NullPointerException: Cannot invoke "com.datadog.appsec.event.EventDispatcher$DataSubscriberInfoImpl.isEventDispatcher(com.datadog.appsec.event.EventDispatcher)" 
because "subscribers" is null. 
	at com.datadog.appsec.event.EventDispatcher.publishDataEvent(EventDispatcher.java:185)
	at com.datadog.appsec.event.ReplaceableEventProducerService.publishDataEvent(ReplaceableEventProducerService.java:33)
	at com.datadog.appsec.gateway.GatewayBridge.lambda$init$5(GatewayBridge.java:224)
	at com.datadog.appsec.gateway.GatewayBridge$$Lambda$151 (Source)
	at datadog.trace.api.gateway.InstrumentationGateway$6.apply(InstrumentationGateway.java:255)
	at datadog.trace.api.gateway.InstrumentationGateway$6.apply(InstrumentationGateway.java:250)
	at org.glassfish.jersey.server.internal.routing.UriRoutingContext.getPathParameters(UriRoutingContext.java:319)
	at org.glassfish.jersey.server.internal.routing.UriRoutingContext.getPathParameters(UriRoutingContext.java:297)
	at com.fsmatic.launcher.JettyEnvironment.decorateRequest(JettyEnvironment.java:284)
	at com.fsmatic.launcher.JettyEnvironment$2.decorateRequest(JettyEnvironment.java:341)
	at io.opentracing.contrib.jaxrs2.server.ServerTracingFilter.filter(ServerTracingFilter.java:78)
	at org.glassfish.jersey.server.ContainerFilteringStage.apply(ContainerFilteringStage.java:132)
	at org.glassfish.jersey.server.ContainerFilteringStage.apply(ContainerFilteringStage.java:68)
	at org.glassfish.jersey.process.internal.Stages.process(Stages.java:197)
	at org.glassfish.jersey.server.ServerRuntime$1run(ServerRuntime.java:269)
```

# Additional Notes
This is a temporary fix, while we are investigating the root of issue. It seems, there are something is wrong with the jersey instrumentation, that cause the problem.

### DIFF
--- a/dd-java-agent/appsec/src/main/java/com/datadog/appsec/gateway/GatewayBridge.java
+++ b/dd-java-agent/appsec/src/main/java/com/datadog/appsec/gateway/GatewayBridge.java
@@ -218,6 +218,13 @@ public class GatewayBridge {
                 pathParamsSubInfo =
                     producerService.getDataSubscribers(KnownAddresses.REQUEST_PATH_PARAMS);
               }
+              if (pathParamsSubInfo == null) {
+                log.warn("Subscriptions is null for REQUEST_PATH_PARAMS event");
+                return NoopFlow.INSTANCE;
+              }
+              if (pathParamsSubInfo.isEmpty()) {
+                return NoopFlow.INSTANCE;
+              }
               DataBundle bundle =
                   new SingletonDataBundle<>(KnownAddresses.REQUEST_PATH_PARAMS, data);
               try {
@@ -245,6 +252,10 @@ public class GatewayBridge {
               if (rawRequestBodySubInfo == null) {
                 rawRequestBodySubInfo =
                     producerService.getDataSubscribers(KnownAddresses.REQUEST_BODY_RAW);
+              }
+              if (pathParamsSubInfo == null) {
+                log.warn("Subscriptions is null for REQUEST_BODY_DONE event");
+                return NoopFlow.INSTANCE;
               }
               if (rawRequestBodySubInfo.isEmpty()) {
                 return NoopFlow.INSTANCE;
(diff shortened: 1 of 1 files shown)


dd-java-agent/appsec/src/main/java/com/datadog/appsec/gateway/GatewayBridge.java
```diff
@@ -246,6 +253,10 @@ public void init() {
                 rawRequestBodySubInfo =
                     producerService.getDataSubscribers(KnownAddresses.REQUEST_BODY_RAW);
               }
+              if (pathParamsSubInfo == null) {
+                log.warn("Subscriptions is null for REQUEST_BODY_DONE event");
```
smola, line 257: We might want to leave this at debug level.

cataphract (changes requested):
I've looked at the code, at there seems to be a race condition indeed. Code follows this pattern:

```java
    while (true) {
      if (this.respDataSubInfo == null) {
        this.respDataSubInfo =
            producerService.getDataSubscribers(
                KnownAddresses.RESPONSE_STATUS, KnownAddresses.RESPONSE_HEADERS_NO_COOKIES);
      }

      try {
        return producerService.publishDataEvent(this.respDataSubInfo, ctx, bundle, false);
      } catch (ExpiredSubscriberInfoException e) {
        this.respDataSubInfo = null;
      }
    }
```

where `respDataSubInfo` is a volatile field. Problem is some thread may have set the variable to null after the null check on another thread. The field needs to be copied to a local variable before the null check.

dd-java-agent/appsec/src/main/java/com/datadog/appsec/gateway/GatewayBridge.java

smola (approved):

ValentinZakharov:
I'll try to fix the race condition issue in next PR

cataphract (approved):

PR #5385, opened 2023-06-14
Stop using LZ4 via JNI, remove lz4-java native libraries

richardstartin:
# What Does This Do

Stops using the JNI implementations in lz4-java, and then removes the bundled native libraries. I have extensive experience with this library and have found that compressing large binaries via JNI is bad for time-to-safepoint, and that on more recent JDKs the Java code is very competitive with native code (LZ4 and XXHash32 are extremely simple algorithms).

# Motivation

Slightly reduces the agent jar size: 

before:
```
27180	dd-java-agent-1.16.0-SNAPSHOT.jar
inflated top level breakdown of contents:
45776	inst
19556	shared
12400	appsec
7016	profiling
3336	datadog
3224	metrics
1296	debugger
540	iast
452	ci-visibility
```

after:
```
26664	dd-java-agent-1.16.0-SNAPSHOT.jar
inflated top level breakdown of contents:
45776	inst
19556	shared
12400	appsec
5708	profiling
3340	datadog
3224	metrics
1304	debugger
540	iast
452	ci-visibility
```

It's likely that similar economies can be found elsewhere and part of the motivation is to encourage drive by reductions in jar size elsewhere.

# Additional Notes

### DIFF
--- a/dd-java-agent/agent-profiling/profiling-uploader/src/main/java/com/datadog/profiling/uploader/CompressingRequestBody.java
+++ b/dd-java-agent/agent-profiling/profiling-uploader/src/main/java/com/datadog/profiling/uploader/CompressingRequestBody.java
@@ -8,7 +8,9 @@ import java.io.OutputStream;
 import java.util.zip.GZIPOutputStream;
 import javax.annotation.Nonnull;
 import javax.annotation.Nullable;
+import net.jpountz.lz4.LZ4Factory;
 import net.jpountz.lz4.LZ4FrameOutputStream;
+import net.jpountz.xxhash.XXHashFactory;
 import okhttp3.MediaType;
 import okhttp3.RequestBody;
 import okio.BufferedSink;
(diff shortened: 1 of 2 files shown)


jbachorik:
Is the description correct? 
The before size is `27180 ` and the after size is `26960` which does not seem to correlate to the uncompressed profiling changes.

richardstartin:
> Is the description correct? The before size is `27180 ` and the after size is `26960` which does not seem to correlate to the uncompressed profiling changes.

Well, I didn't cook the numbers if that's what you're asking. I can build again and double check, I'm not sure where the discrepancy would come from but I'm trusting `jar -xf` and `du -s` to do the right things.

jbachorik:
> Well, I didn't cook the numbers if that's what you're asking

No, not really. It just that the change seems to actually *increase* the jar size :(

mcculls (approved):

jbachorik (approved):

smola (approved):

PR #5387, opened 2023-06-14
omit dylibs unless building to run on a MacOS laptop

richardstartin:
# What Does This Do

Unless a developer builds with the property `targetMacOs`, native dylibs are excluded from the build, which reduces deployment size by about 1MiB. This assumes that customers do not use appsec on laptops.

before:
```
26312	dd-java-agent-1.16.0-SNAPSHOT.jar
45776	inst
19556	shared
12400	appsec
7016	profiling
3340	datadog
3224	metrics
1304	debugger
540	iast
452	ci-visibility
```

after:
```
25500	dd-java-agent-1.16.0-SNAPSHOT.jar
45776	inst
19556	shared
9620	appsec
6860	profiling   <-- double counts the dylib excluded by the profiling-specific PR
3340	datadog
3224	metrics
1304	debugger
540	iast
452	ci-visibility
```

# Motivation

# Additional Notes

### DIFF
--- a/dd-java-agent/build.gradle
+++ b/dd-java-agent/build.gradle
@@ -35,6 +35,10 @@ ext.generalShadowJarConfig = {
   exclude '**/META-INF/*.kotlin_module'
   exclude '**/module-info.class'
 
+  if (!project.hasProperty("targetMacOs")) {
+    exclude '**/*.dylib'
+  }
+
   // Prevents conflict with other SLF4J instances. Important for premain.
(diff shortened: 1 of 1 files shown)


jbachorik (approved):

smola (changes requested):
As discussed offline, given our own extensive (development) usage in macOS, and the few hundred organizations that are running dd-trace-java on macOS, I don't think we should be doing this unless/until we do provide some alternative artifacts that do not involve building from source.

PR #5391, opened 2023-06-14
Hokitam/dropped payloads java

hokitam:
# What Does This Do

# Motivation

# Additional Notes


dd-trace-core/src/main/java/datadog/trace/core/datastreams/DefaultDataStreamsMonitoring.java
```diff
@@ -123,6 +123,13 @@ public void start() {
     thread.start();
   }
 
+  public PathwayContext setCheckpoint(PathwayContext pathwayContext, LinkedHashMap<String, String> sortedTags) {
+    if (pathwayContext == null) {
```
dougqh, line 127: A null object idiom might be a better solution that defensive null checking everywhere.

dd-trace-core/src/main/java/datadog/trace/core/datastreams/DataStreamsMonitoring.java
```diff
@@ -29,6 +30,8 @@ public interface DataStreamsMonitoring extends AgentDataStreamsMonitoring, AutoC
    */
   void mergePathwayContextIntoSpan(AgentSpan span, DataStreamsContextCarrier carrier);
 
+  PathwayContext setCheckpoint(PathwayContext pathwayContext, LinkedHashMap<String, String> sortedTags);
```
dougqh, line 33: Several us have commented before that we would like to see the LinkedHashMap replaced with a more semantically rich object.  LinkedHashMap doesn't solve the sorting problem and has unnecessary overhead.

Rather than extending the problematic design, I would like to see this code cleaned up.

PR #5395, opened 2023-06-15
add tag to local root span to indicate whether profiling is enabled

richardstartin:
# What Does This Do

Tags the local root span to indicate whether profiling is enabled or not.

# Motivation

# Additional Notes


jbachorik (approved):

bantonsson (approved):

richardstartin:
I've simply removed the OpenLiberty snapshot test because it's not clear it was really testing anything not already tested, and was just preventing me from making a very simple change.

smola (approved):

PR #5398, opened 2023-06-15
Implement auto-configuration for code-coverage in Maven

nikita-tkachenko-datadog:
# What Does This Do
Adds automatic configuration of per-test code coverage for Maven projects.
If `dd.civisibility.jacoco.plugin.version` property is provided when running a Maven build with the tracer attached, the tracer will automatically configure Jacoco plugin of the given version to be used with the build.

# Motivation
In order for the tracer's per-test code coverage feature to work, Jacoco needs to be used in the project.
The usual way of enabling Jacoco would be to update a POM file by adding Jacoco Maven plugin config.
The clients often have troubles doing this, and are in general reluctant to modify their build files.


dd-smoke-tests/maven/src/test/groovy/datadog/smoketest/MavenSmokeTest.groovy
```diff
@@ -0,0 +1,393 @@
...
+      files          : [
+        [
+          filename: "src/test/java/datadog/smoke/TestSucceed.java",
+          segments: [[7, -1, 7, -1, -1], [11, -1, 12, -1, -1]]
+        ]
+      ]
+    ]
+
+    where:
+    mavenVersion << ["3.2.1", "3.2.5", "3.3.9", "3.5.4", "3.6.3", "3.8.8", "3.9.2", "4.0.0-alpha-5"]
```
bantonsson, line 188: If I understand the code correctly, these will be re-downloaded every time that the smoke-test is run. Would it be possible to have them _cached_ somewhere?

bantonsson (approved):
I just have to trust you on the inner workings of maven and the DOM manipulation, and the tests look good.

dd-smoke-tests/maven/src/test/groovy/datadog/smoketest/MavenSmokeTest.groovy
nikita-tkachenko-datadog: By default the distributions are downloaded to `./m2/wrapper/dists`, which (depending on how our CI is set up of course) should survive between test runs.
I changed that to download them to a temporary folder that is removed once the tests are executed - my motivation was to avoid polluting the environment.
It takes ~8-10 seconds to download a distribution on my machine, which is significant given that there are 8 of them in total.
If storing them in the default location (`./m2/wrapper/dists`) is fine, then I am reverting to that.

bantonsson: I think that it's fine to store them in the default location and they will be cached by CI there. The `master` cache is rebuilt on a weekly basis.

nikita-tkachenko-datadog: Done, reverted location to default

PR #5401, opened 2023-06-15
Break potentially circular dependency between Config and TPEHelper

richardstartin:
# What Does This Do

This resolves a circular dependency where `TPEHelper` depends on `Config` for a feature flag check, but `Config.<clinit>` depends on `Runtime.exec` to get the hostname, which depends on `ThreadPoolExecutor` which depends on `TPEHelper` once instrumented. All of this has been set up by the time `InstrumentationBasedProfiling`'s flag is flipped.

# Motivation

# Additional Notes


smola (approved):

mcculls (approved):
+1

Just a note that `Config.<clinit>` only triggers fetching the hostname when debug logging is enabled - otherwise fetching the hostname is deferred until a service requires it (such as remote-config) - but regardless, reversing the check here is definitely a good idea

richardstartin:
I've completely removed the dependency of TPEHelper on Config now, just in case

bantonsson (approved):

PR #5405, opened 2023-06-15
Do not send empty test suite spans

nikita-tkachenko-datadog:
# What Does This Do
Adds a check to avoid reporting empty test suite spans.
A test suite span is considered empty if it has no children (i.e. suite does not include test cases) and it was not explicitly reported by the test framework as skipped or failed.

# Motivation
In certain edge cases a test framework might report a test suite as started, and then all of suite's children might be skipped by Intelligent Test Runner (this happens with parameterized tests in JUnit 5 - the list of test cases is not known in advance, so when suite is started there is no way to know if it will be skipped entirely or not).
These empty suites are confusing for the product's users and must not be reported.


PerfectSlayer (approved):
Thanks for the detailed context.

Quick question: any reason the context status is a bunch of cached String in an interface rather than an enum?
It always feel bad to me to have null check everywhere in its codebase. I would have even define an "unknown" status if needed to prevent it.

Otherwise, it's good to me.

PR #5407, opened 2023-06-15
Refactor CI Tags provider

nikita-tkachenko-datadog:
# What Does This Do
Refactors one of CI Visibility's internal classes: `CiTagsProvider`.
The interface and implementation are united in a single concrete class, and some of the class' internal logic is moved outside.

# Motivation
Historically the class needed to be exposed to instrumentation code, so an interface (`CiTagsProvider`) and its implementation (`CiTagsProviderImpl`) existed.
The class is no longer used by code outside of CI Visibility, so there is no longer a need for the interface.
The class also included logic for obtaining CI Info, that was duplicated in several other places.
That logic was moved outside of the class, as a preparatory step for centralising it (since executing it multiple times results in some overhead).


dd-java-agent/agent-ci-visibility/src/main/java/datadog/trace/civisibility/ci/CITagsProvider.java
```diff
@@ -1,8 +1,177 @@
...
+    public CITagsBuilder withGitCommitMessage(final GitInfo gitInfo) {
+      return putTagValue(Tags.GIT_COMMIT_MESSAGE, gitInfo.getCommit().getFullMessage());
+    }
+
+    public Map<String, String> build() {
+      return ciTags;
+    }
+
+    private CITagsBuilder putTagValue(
+        final String tagKey, final String tagValue, final String... fallbackValues) {
```
PerfectSlayer, line 163: I can't find usage of the varags. Did I miss them?
If not, can we remove it to prevent String[] allocation?

PerfectSlayer (approved):
Only one minor comment. LGTM otherwise.
Thanks for splitting those changes in a dedicated PR.

dd-java-agent/agent-ci-visibility/src/main/java/datadog/trace/civisibility/ci/CITagsProvider.java
nikita-tkachenko-datadog: Good catch! I missed the fact that this parameter is not used any more. Removed

PR #5412, opened 2023-06-15
Implement remote settings fetching for CI Visibility

nikita-tkachenko-datadog:
# What Does This Do
Implements fetching remote tracer settings from CI Visibility backend:
- code coverage enabled/disabled
- intelligent test runner enabled/disabled
- list of skippable tests for intelligent test runner

# Motivation
These settings are stored in the backend and need to be fetched by the tracer at runtime.

# Additional Notes
Settings and skippable tests requests include a "configuration" block, that contains among other things info about the JVM that is used to execute the tests (backend needs those data in order to provide correct response).
Responses to those requests need to be known before forking the JVMs that run the tests (in order to properly set up build tasks that do the forking).
Therefore instrumentation code hooks into build systems (Gradle and Maven) to determine the path to the forked JVM executable and use it to obtain the required details.
That involves launching the JVM, as certain info can only be obtained from a running Java Machine.

This PR only contains component that fetch the settings. To keep the review scope small, actual usage of those components will be included in a separate PR.


internal-api/src/main/java/datadog/trace/api/Config.java
```diff
@@ -2348,8 +2348,8 @@ public boolean isCiVisibilityCompilerPluginAutoConfigurationEnabled() {
     return ciVisibilityCompilerPluginAutoConfigurationEnabled;
   }
 
-  public boolean isCiVisibilityPerTestCodeCoverageEnabled() {
-    return ciVisibilityPerTestCodeCoverageEnabled;
+  public Boolean getCiVisibilityCodeCoverageEnabled() {
```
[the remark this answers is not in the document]
nikita-tkachenko-datadog: In parent process this flag is supposed to be enabled by default (unless the client explicitly disables it), and in children processes it's disabled by default (unless the parent process explicitly informs the child to enable it).
I agree that it is somewhat confusing.
My line of thinking was that having two different flags would be even more confusing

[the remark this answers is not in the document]
nikita-tkachenko-datadog: This is something I forgot to rollback, good catch! I though I'd need the code to behave differently based on whether the flag is specified or omitted, but in the end I could do with just specifying the default value.
Rolled back this change.

dd-java-agent/agent-ci-visibility/src/main/java/datadog/trace/civisibility/config/JvmInfo.java
```diff
@@ -0,0 +1,47 @@
+package datadog.trace.civisibility.config;
+
+public class JvmInfo {
+
+  public static final JvmInfo CURRENT_JVM =
```
bantonsson, line 5: It feels like this is the same thing as the data in `datadog.trace.api.Platform`

dd-java-agent/agent-ci-visibility/src/main/java/datadog/trace/civisibility/config/ModuleExecutionSettingsFactory.java
```diff
@@ -0,0 +1,174 @@
...
+        .env(config.getEnv())
+        .repositoryUrl(gitInfo.getRepositoryURL())
+        .branch(gitInfo.getBranch())
+        .sha(gitInfo.getCommit().getSha())
+        .osPlatform(System.getProperty("os.name"))
+        .osArchitecture(System.getProperty("os.arch"))
+        .osVersion(System.getProperty("os.version"))
+        .runtimeName(jvmInfo.getName())
+        .runtimeVersion(jvmInfo.getVersion())
+        .runtimeVendor(jvmInfo.getVendor())
```
bantonsson, line 87: Why is this code not using the `datadog.telemetry.HostInfo`?

nikita-tkachenko-datadog: Similarly to `Platform`, I missed this class.
And just like with the platform, there's a gotcha: I need the values to fully correspond to the tags that are applied to the test spans (the tags are set in `datadog.trace.civisibility.decorator.TestDecoratorImpl`).
The distinction in this case is in the OS name, which is always taken from the system property for the test spans - in `HostInfo` it is calculated slightly differently.
I could add a new field to the host info (something like `osNameProperty`, I suppose), not sure if this would be better than what I have now.
What do you think?

nikita-tkachenko-datadog:
It would be ideal of course to just use `Platform` and `HostInfo` as the sources for test span tags, the problem here is all the historical data that already exists in the backend: a history of executions is stored for each test case, and the tags in question are part of the composite test case "ID", so changing the logic that populates the tags will break the history for existing customers :(

dd-java-agent/agent-ci-visibility/src/main/java/datadog/trace/civisibility/config/JvmInfo.java
bantonsson: As far as I know, the string in `java.vendor` and `java.vm.vendor` are identical (at least on the handful of JVMs that I just tested with). I hear what you are saying, and maybe consolidating things are not worth it since you don't try to parse out the version components or anything.

dd-java-agent/agent-ci-visibility/src/main/java/datadog/trace/civisibility/config/ModuleExecutionSettingsFactory.java
nikita-tkachenko-datadog: That's a good point, making the dependency more obvious is a good idea. Added the comment.

internal-api/internal-api-9/src/main/java/datadog/trace/util/JDK9ExecutableSupplier.java
```diff
@@ -0,0 +1,19 @@
...
+
+public final class JDK9ExecutableSupplier implements Supplier<String> {
+  private static final Logger log = LoggerFactory.getLogger(JDK9ExecutableSupplier.class);
+
+  @Override
+  public String get() {
+    try {
+      return ProcessHandle.current().info().command().orElse(null);
+    } catch (Throwable e) {
+      log.debug("Cannot get PID through JVM API", e);
```
bantonsson, line 15: This is not the PID.

internal-api/src/main/java/datadog/trace/util/ProcessUtils.java
```diff
@@ -0,0 +1,33 @@
...
+import java.util.function.Supplier;
+import org.slf4j.Logger;
+import org.slf4j.LoggerFactory;
+
+public abstract class ProcessUtils {
+
+  private static final Logger log = LoggerFactory.getLogger(ProcessUtils.class);
+
+  @SuppressForbidden
+  public static String getCurrentExecutablePath() {
```
bantonsson, line 14: Please document/test the different _suprising_ behaviors, not to mention that the `JDK9ExecutableSupplier` will return `null` and not throw an exception. This `null` value would be propagated to the code using this, resulting in a `null` pointer.

bantonsson (changes requested):

internal-api/internal-api-9/src/main/java/datadog/trace/util/JDK9ExecutableSupplier.java
nikita-tkachenko-datadog: My bad, poorly done copy-paste.

internal-api/src/main/java/datadog/trace/util/ProcessUtils.java
nikita-tkachenko-datadog: Updated `JDK9ExecutableSupplier` to throw exception, and documented the fallback strategy which is to return home folder instead of the path.
Now the method will only return `null` if `java.home` System property is not available (which is possible in theory, but I don't think there's a JVM that does not set it).

bantonsson (approved):

PR #5420, opened 2023-06-16
Add support for snapshot pruning

jpbempel:
# What Does This Do
When snapshots are serialized if they are greater than 1MB we are pruning them from the deepest levels first and going up to remove them until the limit is not breached anymore

# Motivation
Snapshot cannot be larger than 1MB to be processed on the backend.
# Additional Notes


dd-java-agent/agent-debugger/src/main/java/com/datadog/debugger/util/SnapshotSlicer.java
```diff
@@ -0,0 +1,149 @@
...
+          jsonWriter.beginObject();
+        }
+        break;
+      case NAME:
+        String name = jsonReader.nextName();
+        if (FIELDS.equals(name)) {
+          fieldsLevels.addLast(tokenLevels.size());
+          if (copy && fieldsLevels.size() > maxDepth) {
+            jsonWriter.name(NOT_CAPTURED_REASON);
+            jsonWriter.value(DEPTH_REASON);
```
shatzi, line 73: should we have a different reason? snapshotToBig?

dd-java-agent/agent-debugger/src/main/java/com/datadog/debugger/util/SnapshotSlicer.java
```diff
@@ -0,0 +1,149 @@
...
+        jsonReader.beginObject();
+        if (copy) {
+          jsonWriter.beginObject();
+        }
+        break;
+      case NAME:
+        String name = jsonReader.nextName();
+        if (FIELDS.equals(name)) {
+          fieldsLevels.addLast(tokenLevels.size());
+          if (copy && fieldsLevels.size() > maxDepth) {
```
shatzi, line 71: this is only place with check maxDepth, aka, 'fields'.
Why we don't check depth for elements or entires?

dd-java-agent/agent-debugger/src/test/resources/com/datadog/debugger/util/largeSnapshot_1.json
shatzi, line 1: allot of '{"notCapturedReason":"depth","type":"org.springframework.samples.petclinic.vet.VetController$LargeObject"}'...

Maybe we should trim arrays at maxLevel-1 ? no need to pass 100 item that are not captured.

dd-java-agent/agent-debugger/src/test/java/com/datadog/debugger/util/SnapshotSlicerTest.java
shatzi, line 1: I think we have a small bug, but I'm not sure it wroth fixing. what happen if an object have fields named: 'fields','elements','entires' -> with those be considered as depth+1? and if that happen - does the snapshot generated still valid? I think it still valid so I don't think that worth fixing. However, if the snapshot trimmed has invalid structure it worth at least make it valid.

shatzi (approved):
LGTM.

dd-java-agent/agent-debugger/src/main/java/com/datadog/debugger/util/SnapshotSlicer.java
jpbempel: not sure worth to do it, what's the added value?

dd-java-agent/agent-debugger/src/main/java/com/datadog/debugger/util/SnapshotSlicer.java
jpbempel: the depth notCpauredReson only appear in a regular snapshot at that level, to keep coherency we put it only for fields

dd-java-agent/agent-debugger/src/test/resources/com/datadog/debugger/util/largeSnapshot_1.json
jpbempel: it's to keep coherency with the way the snapshot is generated in the first place

PR #5422, opened 2023-06-16
Ensure lambda instrumentation only notifies extension once.

purple4reina:
# What Does This Do

Prevent calling `start-invocation` and `end-invocation` on the aws lambda extension multiple times.

The lambda event needs to be serialized to json and shared with the extension. This event was properly serialized for the second wrapping, but not for the first. This means that once the extra handler wrapping is ignored, only one `aws.lambda` span is being created, but the lack of event json means we won't be able to create inferred spans or properly handle trace context.

The solution for this second problem is to add a new json adapter to serialize these types.

# Motivation

In the most recent version of the java tracing layer for aws lambda, the request handler is being wrapped at least twice.

```log
[dd.trace 2023-05-18 21:41:20:530 +0000] [main] DEBUG datadog.trace.agent.tooling.InstrumenterState - Instrumentation applied - instrumentation.names=[aws-lambda] instrumentation.class=datadog.trace.instrumentation.aws.v1.lambda.LambdaHandlerInstrumentation instrumentation.target.classloader=lambdainternal.CustomerClassLoader@4d2a1da3

[dd.trace 2023-05-18 21:41:21:649 +0000] [main] DEBUG datadog.trace.agent.tooling.InstrumenterState - Instrumentation applied - instrumentation.names=[aws-lambda] instrumentation.class=datadog.trace.instrumentation.aws.v1.lambda.LambdaHandlerInstrumentation instrumentation.target.classloader=jdk.internal.loader.ClassLoaders$AppClassLoader@2626b418
```

This double wrapping is causing the tracer to make calls to `start-invocation` and `end-invocation` multiple times. A `aws.lambda` root span is created in the extension for every call to these endpoints.

# Additional Notes

### DIFF
--- a/dd-trace-core/src/main/java/datadog/trace/lambda/ReadFromInputStreamJsonAdapter.java
+++ b/dd-trace-core/src/main/java/datadog/trace/lambda/ReadFromInputStreamJsonAdapter.java
@@ -1,0 +1,51 @@
+package datadog.trace.lambda;
+
+import com.squareup.moshi.JsonAdapter;
+import com.squareup.moshi.JsonReader;
+import com.squareup.moshi.JsonWriter;
+import com.squareup.moshi.Moshi;
+import java.io.ByteArrayInputStream;
+import java.io.ByteArrayOutputStream;
+import java.io.IOException;
+import java.lang.annotation.Annotation;
+import java.lang.reflect.Type;
+import java.util.Set;
+import okio.BufferedSink;
+
+public final class ReadFromInputStreamJsonAdapter<T> extends JsonAdapter<T> {
+
+  @Override
+  public T fromJson(JsonReader reader) throws IOException {
+    throw new UnsupportedOperationException();
+  }
+
+  @Override
+  public void toJson(JsonWriter writer, T value) throws IOException {
+    BufferedSink sink = writer.valueSink();
+    byte[] bytes = getInputBytes(value);
+    sink.write(bytes);
+    sink.flush();
+  }
+
+  private byte[] getInputBytes(T value) throws IOException {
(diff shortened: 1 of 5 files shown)


dd-trace-core/src/main/java/datadog/trace/lambda/ReadFromInputStreamJsonAdapter.java
```diff
@@ -0,0 +1,50 @@
...
+import com.squareup.moshi.JsonWriter;
+import com.squareup.moshi.Moshi;
+import java.io.ByteArrayInputStream;
+import java.io.IOException;
+import java.lang.annotation.Annotation;
+import java.lang.reflect.Type;
+import java.util.Set;
+import okio.BufferedSink;
+
+public final class ReadFromInputStreamJsonAdapter<T> extends JsonAdapter<T> {
```
bantonsson, line 14: Can't `T` be `ByteArrayInputStream` since you know the type?

bantonsson (approved):
Looks good.

The codenarc failure seems to be an unnecessary import in the test 🤷🏼‍♂️

https://output.circle-artifacts.com/output/job/046ad3e3-a643-4ba2-8ba3-14c0c9f1fa2a/artifacts/0/check_reports/dd-trace-core/codenarc/test.html

dd-trace-core/src/main/java/datadog/trace/lambda/ReadFromInputStreamJsonAdapter.java
joeyzhao2018: good point. updated.

PR #5424, opened 2023-06-19
Unvalidated redirect not reported if Referer header is the source

jandro996:
# What Does This Do
Not report unvalidated redirect vulnerability if all tainted object ranges have header Referer as source

# Motivation
Avoid unvalidated redirect false positives

### DIFF
--- a/dd-java-agent/agent-iast/src/test/groovy/com/datadog/iast/sink/UnvalidatedRedirectModuleTest.groovy
+++ b/dd-java-agent/agent-iast/src/test/groovy/com/datadog/iast/sink/UnvalidatedRedirectModuleTest.groovy
@@ -135,22 +135,9 @@ class UnvalidatedRedirectModuleTest extends IastModuleImplTestBase {
     0 * reporter.report(_, _)
   }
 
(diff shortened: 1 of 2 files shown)


smola (approved):

smola (approved):

dd-java-agent/agent-iast/src/test/groovy/com/datadog/iast/sink/UnvalidatedRedirectModuleTest.groovy
```diff
@@ -120,6 +121,51 @@ class UnvalidatedRedirectModuleTest extends IastModuleImplTestBase {
...
+    module.onRedirect(value)
+
+    then:
+    1 * reporter.report(_, _)
+
+    when:
+    module.onRedirect(value2)
+
+    then:
+    1 * reporter.report(_, _)
```
manuel-alvarez-alvarez, line 165: You could easily make it a parameterized test with a `where` since both when/then clauses are the same and only the input parameters (ranges) change

manuel-alvarez-alvarez (approved):

PR #5426, opened 2023-06-19
Add telemetry to IAST advices

manuel-alvarez-alvarez:
# What Does This Do
Adds a `net.bytebuddy.asm.Advice.PostProcessorFactory` to generate the required code to provide IAST related metrics in bytebuddy advices. 

The post processor does two things:

- Calls `IastMetricCollector.add(IastMetric.INSTRUMENTED_$TYPE, $TAG, 1)` whenever a new instrumentation point is detected.
- Injects the required bytecode to call `IastMetricCollector.add(IastMetric.EXECUTED_$TYPE, $TAG, 1)` when the actual code of the advice is executed at runtime.

# Motivation
From the IAST part we need very concrete metrics to understand if the vulnerability detection engine is working correctly. Having these kind of metrics allows us to check if customer spans that have certain operations (for instance a SQL call) are handled correctly by the engine.

# Additional Notes

### DIFF
--- a/dd-java-agent/agent-tooling/src/main/java/datadog/trace/agent/tooling/iast/IastPostProcessorFactory.java
+++ b/dd-java-agent/agent-tooling/src/main/java/datadog/trace/agent/tooling/iast/IastPostProcessorFactory.java
@@ -1,0 +1,176 @@
+package datadog.trace.agent.tooling.iast;
+
+import static datadog.trace.api.iast.telemetry.IastMetric.EXECUTED_PROPAGATION;
+import static datadog.trace.api.iast.telemetry.IastMetric.EXECUTED_SINK;
+import static datadog.trace.api.iast.telemetry.IastMetric.EXECUTED_SOURCE;
+import static datadog.trace.api.iast.telemetry.IastMetric.INSTRUMENTED_PROPAGATION;
+import static datadog.trace.api.iast.telemetry.IastMetric.INSTRUMENTED_SINK;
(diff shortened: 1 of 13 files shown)


dd-java-agent/agent-tooling/src/main/java/datadog/trace/agent/tooling/iast/IastPostProcessorFactory.java
```diff
@@ -0,0 +1,166 @@
...
+      runtime = null;
+    }
+    return new PostProcessor(instrumentation, runtime, tagValue);
+  }
+
+  private AnnotationDescription findAdviceAnnotation(
+      final MethodDescription.InDefinedShape advice) {
+    for (final AnnotationDescription annotation : advice.getDeclaredAnnotations()) {
+      final TypeDescription type = annotation.getAnnotationType();
+      if (type.represents(Propagation.class)) {
```
jandro996, line 94: This could be done in one if sentence :)

jandro996 (approved):

dd-java-agent/agent-tooling/src/main/java/datadog/trace/agent/tooling/iast/IastPostProcessorFactory.java
manuel-alvarez-alvarez: Changed! thanks for the input

smola (approved):
Looks good to me if it does not affect performance of the base case (IAST off).

PR #5427, opened 2023-06-19
Implement a mechanism to report data from children processes to parent process

nikita-tkachenko-datadog:
# What Does This Do
Implements a mechanism that would allow children processes to send data to parent process.
The mechanism is a simple TCP server and a client for it.

# Motivation
When running a service with CI Visibility, we need to instrument the following processes:
- parent (the build system - Gradle or Maven)
- children (JVMs forked to run tests for individual modules in the project)

There is a need to send data from children processes to the parent process: number of tests skipped by Intelligent Test Runner, code coverage stats, etc.

# Additional Notes
This PR only contains code for the server and the client. To keep the review scope small, actual usage of that code will be included in a separate PR.


dd-java-agent/agent-ci-visibility/src/main/java/datadog/trace/civisibility/ipc/ByteUtils.java
```diff
@@ -0,0 +1,32 @@
+package datadog.trace.civisibility.ipc;
+
+public abstract class ByteUtils {
```
bantonsson, line 3: Is there a particular reason to not use `ByteBuffer` for this?

dd-java-agent/agent-ci-visibility/src/main/java/datadog/trace/civisibility/ipc/SignalServer.java
```diff
@@ -0,0 +1,132 @@
...
+    this.port = port;
+  }
+
+  public synchronized void start() {
+    if (serverSocketChannel == null) {
+      try {
+        serverSocketChannel = ServerSocketChannel.open();
+        serverSocketChannel.configureBlocking(false);
+
+        serverSocketChannel.bind(port != -1 ? new InetSocketAddress(port) : null);
```
bantonsson, line 52: Do we expect that customers would want to configure the address that this binds to in some way? I assume that this is only enabled under certain circumstances (and opt in), but could opening a port on the JVM that potentially runs the build server be considered a security risk?

dd-java-agent/agent-ci-visibility/src/main/java/datadog/trace/civisibility/ipc/ByteUtils.java
nikita-tkachenko-datadog: Initial intention was to avoid allocating buffers, but I looked at the final code once again, and the message is wrapped in a buffer anyway before being written to the channel, so that doesn't make much sense in the end.
Changed the code to use `ByteBuffer`.

dd-java-agent/agent-ci-visibility/src/main/java/datadog/trace/civisibility/ipc/SignalServer.java
bantonsson: Ok, I accept that opening the port is probably fine (and no worse than what's there already), but how do you configure which interface/address to bind to?

nikita-tkachenko-datadog: Good point, it used to be `InetAddress.anyLocalAddress()`.
Added a config property so that customers could change this.
Default value is `127.0.0.1` to only allow local connection (which should cover Maven and most Gradle setups).

bantonsson (approved):

PR #5428, opened 2023-06-19
Implement passing module execution results from children processes to parent process

nikita-tkachenko-datadog:
# What Does This Do
Implements passing tests execution results from children processes (JVMs forked to run tests) to the parent process (Gradle or Maven).

# Motivation
The overall execution result (success or failure) is already available in the parent process, but for CI Visibility there's a need to pass more data (such as whether the tests were executed with code coverage, or if any of the test cases were skipped by Intelligent Test Runner).

# Additional Notes
The data is passed via TCP using `SignalServer` and`SignalClient` introduced for this purpose previously (in a separate PR).


tonyredondo (approved):
LGTM

PR #5434, opened 2023-06-20
Implement Git repo unshallowing

nikita-tkachenko-datadog:
# What Does This Do
Implements logic for "unshallowing" a shallow Git repository.
The code checks whether current repo is a shallow one and if so, attempts to refetch the commits data from the server for the past month.

# Motivation
CI Visibility's Intelligent Test Runner requires Git data in order to work: the tracer has to upload Git pack files to a backend endpoint.
Many CI providers clone their repositories with a shallow clone, which means there is no commit history. In those cases, sending a packfile to the backend would not be the correct choice since we would generate a packfile with a single commit instead of the necessary payload, therefore polluting the state of the backend.


nikita-tkachenko-datadog:
> Javadoc would be nice...

Added Javadoc

PR #5435, opened 2023-06-20
Allow trace flush interval to be configurable

randomanderson:
# What Does This Do
Adds the ability to configure the flush interval used to flush traces to the backend.

# Motivation
Logs injection smoke tests take too long with dynamic config enabled.

# Additional Notes
* The unit for frequency would be _flushes/per second_ . "Interval" is the correct when talking about durations. I updated where applicable.

* Almost all of the configurations for intervals use seconds without specifying "seconds" in the config name. I followed that pattern. Internally, milliseconds is used because passing the timeunit through all of those layers would be annoying.

* Lots of tests and smoke tests can be sped up by configuring a shorter flush interval. I did not make any of those changes in this PR.


ygree (approved):
LGTM

dd-trace-core/src/main/java/datadog/trace/common/writer/DDAgentWriter.java
```diff
@@ -35,7 +35,7 @@ public static class DDAgentWriterBuilder {
     long timeoutMillis = TimeUnit.SECONDS.toMillis(DEFAULT_AGENT_TIMEOUT);
     int traceBufferSize = BUFFER_SIZE;
     HealthMetrics healthMetrics = HealthMetrics.NO_OP;
-    int flushFrequencySeconds = 1;
+    int flushIntervalMilliseconds = 1000;
```
bantonsson, line 38: Why not `TimeUnit.SECONDS.toMillis(DEFAULT_TRACE_FLUSH_INTERVAL);`?

PR #5436, opened 2023-06-21
Ensure safe class loading in instrumentation

jpbempel:
# What Does This Do
We introduced a helper method that require to provide the name of the class currently being transformed to ensure it is not the same. We also catch Throwable in transform method to ensure whatever happens during transformation we do not bubble up to the user code and load the original class

# Motivation
Loading a class during transformation/instrumentation MUST not be the current class being transformed otherwise a LinkageError happen with message "attempted duplicate class definition" because we have loaded the class before it was transformed, so when transformation finish the original load of the class is happening and boom! 
# Additional Notes

### DIFF
--- a/dd-java-agent/agent-debugger/src/main/java/com/datadog/debugger/agent/DebuggerTransformer.java
+++ b/dd-java-agent/agent-debugger/src/main/java/com/datadog/debugger/agent/DebuggerTransformer.java
@@ -154,7 +154,7 @@ public class DebuggerTransformer implements ClassFileTransformer {
       // e.g. Main.java contains Main & TopLevel class, line numbers are in TopLevel class
       log.info(
           "type {} matched but no transformation for definitions: {}", classFilePath, definitions);
-    } catch (Exception ex) {
+    } catch (Throwable ex) {
       log.warn("Cannot transform: ", ex);
     }
     return null;
--- a/dd-java-agent/agent-debugger/src/main/java/com/datadog/debugger/instrumentation/ASMHelper.java
+++ b/dd-java-agent/agent-debugger/src/main/java/com/datadog/debugger/instrumentation/ASMHelper.java
@@ -103,6 +103,37 @@ public class ASMHelper {
     return new Type(mainType, genericTypes);
   }
 
+  /**
+   * Makes sure that the class we want to load is not the one that we are currently transforming
+   * otherwise it will lead into a LinkageError
+   *
+   * @param className class name to load in '.' format (Java)
+   * @param currentClassTransformed in '.' format (Java)
+   * @param classLoader use for loading the class
+   * @return the loaded class
+   */
+  public static Class<?> ensureSafeClassLoad(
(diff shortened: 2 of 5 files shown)


dd-java-agent/agent-debugger/src/main/java/com/datadog/debugger/agent/DebuggerTransformer.java
```diff
@@ -154,7 +154,7 @@ public byte[] transform(
       // e.g. Main.java contains Main & TopLevel class, line numbers are in TopLevel class
       log.info(
           "type {} matched but no transformation for definitions: {}", classFilePath, definitions);
-    } catch (Exception ex) {
+    } catch (Throwable ex) {
```
ojung, line 157: Nice, an occasion where this is actually required.

dd-java-agent/agent-debugger/src/main/java/com/datadog/debugger/instrumentation/ASMHelper.java
```diff
@@ -103,6 +103,37 @@ public static Type decodeSignature(String signature) {
     return new Type(mainType, genericTypes);
   }
 
+  /**
+   * Makes sure that the class we want to load is not the one that we are currently transforming
+   * otherwise it will lead into a LinkageError
+   *
+   * @param className class name to load in '.' format (Java)
+   * @param currentClassTransformed in '.' format (Java)
```
ojung, line 111: What's `'.' format (java)`? The canonical name? So `com.foo.bar.Class.InnerClass` instead of `com.foo.bar.Class$InnerClass`?

dd-java-agent/agent-debugger/src/main/java/com/datadog/debugger/instrumentation/ASMHelper.java
```diff
@@ -103,6 +103,37 @@ public static Type decodeSignature(String signature) {
...
+   * @param classLoader use for loading the class
+   * @return the loaded class
+   */
+  public static Class<?> ensureSafeClassLoad(
+      String className, String currentClassTransformed, ClassLoader classLoader) {
+    if (currentClassTransformed == null) {
+      // This is requires to make sure we are not loading class being transformed during
```
ojung, line 118:
```suggestion
      // This is required to make sure we are not loading the class being transformed during
```

ojung (approved):
Nice, good to have this fixed 👍

dd-java-agent/agent-debugger/src/main/java/com/datadog/debugger/instrumentation/ASMHelper.java
jpbempel: no in fact it's `TypeName` and not the internal JVM name (with the slahes '/') or the canonical
so ti would be `com.foo.bar.Class$InnerClass` 
but I would precise it

PR #5441, opened 2023-06-21
Add special support for enum values

jpbempel:
# What Does This Do
We are serializing then the enum constant name in the snapshot instead of handling enum as regular complex class

# Motivation
Enum in java are in fact regular classes with a special treatment by the java compiler. from the POV of the JVM there is no difference. From user perspective however it would be nice to have constant name as enum values.

# Additional Notes


ojung (approved):
👍

PR #5443, opened 2023-06-21
Fold stacktrace for exception

jpbempel:
when trying access fields during capture/serialization and debug logs
# What Does This Do

# Motivation

# Additional Notes


ojung (approved):
Thanks 👍

PR #5450, opened 2023-06-22
Fix inserting line probe before for-loops

jpbempel:
# What Does This Do
for loops generate 2 entries in LineNumberTable, 1 before the loop, 1 for the incrementation part. the second part was only kept into the line map (overwritten). We now always keep the first mapping.

# Motivation

# Additional Notes

### DIFF
--- a/dd-java-agent/agent-debugger/src/main/java/com/datadog/debugger/instrumentation/LineMap.java
+++ b/dd-java-agent/agent-debugger/src/main/java/com/datadog/debugger/instrumentation/LineMap.java
@@ -13,7 +13,11 @@ class LineMap {
   private final TreeMap<Integer, LabelNode> lineLabels = new TreeMap<>();
 
   void addLine(LineNumberNode lineNode) {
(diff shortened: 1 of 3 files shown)


ojung (approved):
Yeah, I saw that when investigating for SymDB as well.. The increment happens after the body of the loop, right?

jpbempel:
> Yeah, I saw that when investigating for SymDB as well.. The increment happens after the body of the loop, right?
hum no, it's generated as a while with inc inside the loop. so if we anchor on the incrementation we are inside the loop.
```
for (int i = 0; i < 10; i++) {
}
```
is translated as:
```
int i = 0;
while (i <10) {
  i++;
}
```

ojung:
But when the for has a body, the increment happens last in the body, right?
```
for (int i = 0; i < 10; i++) {
    System.out.println("hello");
}
```
V
```
int i = 0;
while (i <10) {
    System.out.println("hello");
    i++;
}
```

jpbempel:
yes

PR #5457, opened 2023-06-23
Add support for double values for metric probes

jpbempel:
for gauge and histogram metric types

# What Does This Do

# Motivation

# Additional Notes

### DIFF
--- a/dd-java-agent/agent-debugger/src/main/java/com/datadog/debugger/instrumentation/MetricInstrumentor.java
+++ b/dd-java-agent/agent-debugger/src/main/java/com/datadog/debugger/instrumentation/MetricInstrumentor.java
@@ -48,6 +48,7 @@ import datadog.trace.bootstrap.debugger.el.ValueReferences;
 import java.lang.reflect.Field;
 import java.util.ArrayList;
 import java.util.List;
+import java.util.stream.Collectors;
 import org.objectweb.asm.Opcodes;
 import org.objectweb.asm.Type;
 import org.objectweb.asm.tree.AbstractInsnNode;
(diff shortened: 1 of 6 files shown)


dd-java-agent/agent-debugger/src/main/java/com/datadog/debugger/instrumentation/MetricInstrumentor.java
```diff
@@ -207,55 +208,66 @@ private InsnList internalCallMetric(String metricMethodName, MetricProbe metricP
...
         Types.asArray(STRING_TYPE, 1));
     // stack []
     insnList.add(nullBranch);
     return insnList;
   }
 
-  private void convertIfRequired(Type currentType, Type expectedType, InsnList insnList) {
-    if (expectedType == Type.LONG_TYPE && currentType == Type.INT_TYPE) {
+  private Type convertIfRequired(Type currentType, InsnList insnList) {
+    if (currentType == Type.INT_TYPE) {
```
shatzi, line 248: nit, if someone define a byte variable - does the byte code store those as int? I wonder if byte type are supported by default.

shatzi (approved):
Looks good.

dd-java-agent/agent-debugger/src/main/java/com/datadog/debugger/instrumentation/MetricInstrumentor.java
jpbempel: That's relevant question. The thing is, the JVM on the operand stack treats `byte` `short` & `char` as `int`, see https://docs.oracle.com/javase/specs/jvms/se20/html/jvms-2.html#jvms-2.11.1

> Note that most instructions in [Table 2.11.1-A](https://docs.oracle.com/javase/specs/jvms/se20/html/jvms-2.html#jvms-2.11.1-220) do not have forms for the integral types byte, char, and short. None have forms for the boolean type. A compiler encodes loads of literal values of types byte and short using Java Virtual Machine instructions that sign-extend those values to values of type int at compile-time or run-time. Loads of literal values of types boolean and char are encoded using instructions that zero-extend the literal to a value of type int at compile-time or run-time. Likewise, loads from arrays of values of type boolean, byte, short, and char are encoded using Java Virtual Machine instructions that sign-extend or zero-extend the values to values of type int. Thus, most operations on values of actual types boolean, byte, char, and short are correctly performed by instructions operating on values of computational type int.

jpbempel: Though, will add in a next PR some unit tests with those type because the internal representation with EL may not match this.

PR #5467, opened 2023-06-26
Response header substitution in jetty

cataphract:
Adds the capability of blocking requests based on response headers.

### DIFF
--- a/dd-java-agent/instrumentation/jetty-7.6/main/java/datadog/trace/instrumentation/jetty70/JettyCommitResponseInstrumentation.java
+++ b/dd-java-agent/instrumentation/jetty-7.6/main/java/datadog/trace/instrumentation/jetty70/JettyCommitResponseInstrumentation.java
@@ -1,0 +1,123 @@
+package datadog.trace.instrumentation.jetty70;
+
+import static datadog.trace.agent.tooling.bytebuddy.matcher.NameMatchers.named;
+import static datadog.trace.bootstrap.instrumentation.api.AgentTracer.activateSpan;
+import static datadog.trace.bootstrap.instrumentation.decorator.HttpServerDecorator.DD_IGNORE_COMMIT_ATTRIBUTE;
+import static datadog.trace.bootstrap.instrumentation.decorator.HttpServerDecorator.DD_SPAN_ATTRIBUTE;
+import static datadog.trace.instrumentation.jetty70.JettyDecorator.DECORATE;
+import static java.util.Collections.singletonMap;
+import static net.bytebuddy.matcher.ElementMatchers.isConstructor;
+import static net.bytebuddy.matcher.ElementMatchers.takesArgument;
+import static net.bytebuddy.matcher.ElementMatchers.takesArguments;
+import static net.bytebuddy.matcher.ElementMatchers.takesNoArguments;
+
+import com.google.auto.service.AutoService;
+import datadog.trace.agent.tooling.Instrumenter;
+import datadog.trace.api.Config;
+import datadog.trace.api.CorrelationIdentifier;
+import datadog.trace.api.GlobalTracer;
+import datadog.trace.api.ProductActivation;
+import datadog.trace.api.gateway.BlockResponseFunction;
+import datadog.trace.api.gateway.Flow;
+import datadog.trace.api.gateway.RequestContext;
+import datadog.trace.bootstrap.InstrumentationContext;
+import datadog.trace.bootstrap.instrumentation.api.AgentScope;
+import datadog.trace.bootstrap.instrumentation.api.AgentSpan;
+import datadog.trace.instrumentation.jetty.ConnectionHandleRequestVisitor;
+import java.util.Map;
+import net.bytebuddy.asm.Advice;
+import net.bytebuddy.asm.AsmVisitorWrapper;
+import net.bytebuddy.description.field.FieldDescription;
+import net.bytebuddy.description.field.FieldList;
+import net.bytebuddy.description.method.MethodList;
+import net.bytebuddy.description.type.TypeDescription;
+import net.bytebuddy.implementation.Implementation;
+import net.bytebuddy.jar.asm.ClassVisitor;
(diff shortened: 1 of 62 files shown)


smola:
@cataphract This one needs a rebase.

dd-java-agent/instrumentation/jetty-7.6/main/java/datadog/trace/instrumentation/jetty70/JettyCommitResponseInstrumentation.java
```diff
@@ -0,0 +1,123 @@
...
+        named("commitResponse").and(takesArguments(1)).and(takesArgument(0, boolean.class)).or(
+            named("completeResponse").and(takesArguments(0))
+        ),
+        JettyCommitResponseInstrumentation.class.getName() + "$CommitResponseAdvice"
+    );
+  }
+
+  static class CommitResponseAdvice {
+    @Advice.OnMethodEnter(suppress = Throwable.class, skipOn = Advice.OnNonDefaultValue.class)
+    static boolean /* skip */ before(@Advice.This HttpConnection connection) {
```
ValentinZakharov, line 82: I can see return value from `OnMethodEnter`, but there is no `OnMethodExit`, where this value can be used. What is the purpose of the return value?

ValentinZakharov:
Looks like there is something is wrong with structure:
I can see exactly the same files in different places.
```
dd-java-agent/instrumentation/jetty-7.6/java/datadog/trace/instrumentation/jetty76/JettyCommitResponseInstrumentation.java
dd-java-agent/instrumentation/jetty-7.6/main/java/datadog/trace/instrumentation/jetty76/JettyCommitResponseInstrumentation.java
dd-java-agent/instrumentation/jetty-7.6/src/main/java/datadog/trace/instrumentation/jetty76/JettyCommitResponseInstrumentation.java
```
I suppose first and second one are been accidentally copied.

dd-java-agent/instrumentation/jetty-9/src/main/java/datadog/trace/instrumentation/jetty10/JettyServerInstrumentation.java
```diff
@@ -0,0 +1,148 @@
...
+import net.bytebuddy.description.method.MethodList;
+import net.bytebuddy.description.type.TypeDescription;
+import net.bytebuddy.implementation.Implementation;
+import net.bytebuddy.jar.asm.ClassVisitor;
+import net.bytebuddy.jar.asm.ClassWriter;
+import net.bytebuddy.jar.asm.Opcodes;
+import net.bytebuddy.pool.TypePool;
+
+@AutoService(Instrumenter.class)
+public final class JettyServerInstrumentation extends Instrumenter.Tracing
```
ValentinZakharov, line 34: Shouldn't we extends on `Instrumenter.AppSec` instead?
The blocking functionality it seems AppSec feature.

dd-java-agent/instrumentation/jetty-9/src/main/java_jetty10/datadog/trace/instrumentation/jetty10/JettyDecorator.java
```diff
@@ -0,0 +1,104 @@
...
+  protected BlockResponseFunction createBlockResponseFunction(Request request, Request connection) {
+    return new JettyBlockResponseFunction(request);
+  }
+
+  public static final class OnResponse {
+    public static AgentSpan onResponse(AgentSpan span, HttpChannel channel) {
+      Request request = channel.getRequest();
+      Response response = channel.getResponse();
+      if (Config.get().isServletPrincipalEnabled() && request.getUserPrincipal() != null) {
+        span.setTag(DDTags.USER_NAME, request.getUserPrincipal().getName());
```
ValentinZakharov, line 91: Why do we need to tag user name?

dd-java-agent/instrumentation/jetty-7.6/main/java/datadog/trace/instrumentation/jetty70/JettyCommitResponseInstrumentation.java
cataphract: The purpose is to determine whether the body of the instrumented method is skipped. We have `skipOn = Advice.OnNonDefaultValue.class`, with the return value being a boolean the method is skipped if the return value is not the default value for booleans, i.e, the method is skipped if the return value is `true`.

dd-java-agent/instrumentation/jetty-9/src/main/java/datadog/trace/instrumentation/jetty10/JettyServerInstrumentation.java
cataphract: It adds three advice:

```java
    transformation.applyAdvice(
        takesNoArguments().and(named("handle")), packageName + ".HandleAdvice");
    transformation.applyAdvice(
        named("recycle").and(takesNoArguments()), packageName + ".ResetAdvice");

    if (appSecNotFullyDisabled) {
      transformation.applyAdvice(
          named("handleException").and(takesArguments(1)).and(takesArgument(0, Throwable.class)),
          packageName + ".HandleExceptionAdvice");
    }
```

Only one of them is about appsec. This very similar to the JettyServerInstrumentation for the other versions. It was added here because there is currently no support for jetty 10. We jump from jetty 9 to 11. In order for appsec to work, there needs to be tracing support as well.

dd-java-agent/instrumentation/jetty-9/src/main/java_jetty10/datadog/trace/instrumentation/jetty10/JettyDecorator.java
cataphract: We set this tag on tomcat, servlet, jetty and openliberty. This file is an adaptation of the one for jetty 9. It's not shared because jetty 10 supports only java 11, and so is compiled separately with jdk 11.

ValentinZakharov (approved):

PR #5474, opened 2023-06-27
Fix scanDependencies CLI

smola:
# What Does This Do

# Motivation
Broken by previous class move.
# Additional Notes


bantonsson (approved):
Could we add a red-face test to the `test-published-depenencies`?

anderruiz (approved):

PR #5476, opened 2023-06-27
Mitigate telemetry interval drift

smola:
# What Does This Do
* Compute the deadline to next telemetry action based on incrementing the previous deadline (rather than based on the current time). This should ensure that telemetry intervals do not drift as long as they do not take longer than the interval itself.
* If heartbeat interval is exceeded (can happen with a faulty network), start the next interval immediately, and reset the interval deadlines.
* If metrics interval is exceed (should not happen except, perhaps, under extreme load), skip one interval.

# Motivation

Mitigate telemetry interval drift.

# Additional Notes

### DIFF
--- a/telemetry/src/main/java/datadog/telemetry/TelemetryRunnable.java
+++ b/telemetry/src/main/java/datadog/telemetry/TelemetryRunnable.java
@@ -18,13 +18,7 @@ public class TelemetryRunnable implements Runnable {
   private final TelemetryService telemetryService;
   private final List<TelemetryPeriodicAction> actions;
   private final List<MetricPeriodicAction> actionsAtMetricsInterval;
-  private final ThreadSleeper sleeper;
-
-  private final TimeSource timeSource;
-  private final long metricsIntervalMs;
-  private final long heartbeatIntervalMs;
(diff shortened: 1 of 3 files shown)


telemetry/src/main/java/datadog/telemetry/TelemetryRunnable.java
```diff
@@ -81,38 +77,25 @@ private void mainLoopIteration() throws InterruptedException {
...
       }
     }
 
-    if (currentTime - lastHeartbeatIntervalMs >= heartbeatIntervalMs) {
-      lastHeartbeatIntervalMs = currentTime;
+    if (scheduler.shouldRunHeartbeat()) {
       for (final TelemetryPeriodicAction action : this.actions) {
         action.doIteration(this.telemetryService);
       }
       telemetryService.sendIntervalRequests();
```
manuel-alvarez-alvarez, line 90: Not related to this PR, but I wonder if it makes sense to uncouple the heartbeat message from the rest, so we could disable other payloads in periods of high load.

manuel-alvarez-alvarez (approved):

PR #5478, opened 2023-06-27
Add support for distribution metrics

jpbempel:
add a new kind of metric: Distribution that is already supported by DogStatsD, but was not exposed before to the tracer metric interface

# What Does This Do

# Motivation

# Additional Notes


shatzi (approved):
LGTM

PR #5480, opened 2023-06-27
Add span metrics

PerfectSlayer:
# What Does This Do

This PR introduces span metrics.
The first to metrics available are:
* Span created
* Span finish

Those metrics are connected to telemetry by a new core metrics collector and period action.

# Motivation

# Additional Notes

This PR is the second part of the Span Metrics PR https://github.com/DataDog/dd-trace-java/pull/5382

### DIFF
--- a/internal-api/src/main/java/datadog/trace/api/metrics/SpanMetricsImpl.java
+++ b/internal-api/src/main/java/datadog/trace/api/metrics/SpanMetricsImpl.java
@@ -1,0 +1,63 @@
+package datadog.trace.api.metrics;
+
+import java.util.ArrayList;
+import java.util.Collection;
+import java.util.Collections;
+import java.util.List;
+import java.util.concurrent.atomic.AtomicLong;
+
+/** The default implementation of {@link SpanMetrics} based on atomic counters. */
+public class SpanMetricsImpl implements SpanMetrics {
+  private final String instrumentationName;
+  private final AtomicLong spanCreatedCounter;
+  private final AtomicLong spanFinishedCounter;
+  private final Collection<CoreCounter> coreCounters;
+
+  public SpanMetricsImpl(String instrumentationName) {
+    this.instrumentationName = instrumentationName;
+    this.spanCreatedCounter = new AtomicLong(0);
+    this.spanFinishedCounter = new AtomicLong(0);
+    List<CoreCounter> coreCounters = new ArrayList<>(2);
+    coreCounters.add(new SpanCounter("span_created", this.spanCreatedCounter));
+    coreCounters.add(new SpanCounter("span_finished", this.spanFinishedCounter));
+    this.coreCounters = Collections.unmodifiableList(coreCounters);
+  }
+
+  @Override
+  public void onSpanCreated() {
+    this.spanCreatedCounter.incrementAndGet();
+  }
+
+  @Override
+  public void onSpanFinished() {
+    this.spanFinishedCounter.incrementAndGet();
+  }
+
+  public String getInstrumentationName() {
+    return this.instrumentationName;
+  }
+
+  public Collection<CoreCounter> getCounters() {
+    return this.coreCounters;
+  }
+
+  private static class SpanCounter implements CoreCounter {
+    private final String name;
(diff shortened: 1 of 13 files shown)


internal-api/src/main/java/datadog/trace/api/metrics/SpanMetricsImpl.java
```diff
@@ -0,0 +1,39 @@
...
+  public void onSpanFinished() {
+    this.spanFinishedCounter.incrementAndGet();
+  }
+
+  public String getInstrumentationName() {
+    return this.instrumentationName;
+  }
+
+  public Map<String, Long> getAndResetCounters() {
+    Map<String, Long> counters = new HashMap<>();
```
bantonsson, line 34: I know that this part is not performance critical, but I would rather have this be an `Iterable` of some more concrete type like a `Counter`, since this is ever only iterated over to pull out the values.

internal-api/src/main/java/datadog/trace/api/telemetry/CoreMetricCollector.java
```diff
@@ -0,0 +1,72 @@
...
+      for (Map.Entry<String, Long> entry : spanMetricsImpl.getAndResetCounters().entrySet()) {
+        String name = entry.getKey();
+        Long value = entry.getValue();
+        if (value == 0) {
+          // Skip not updated counters
+          continue;
+        }
+        CoreMetric metric = new CoreMetric(METRIC_NAMESPACE, true, name, "count", value, tag);
+        if (!this.metricsQueue.offer(metric)) {
+          // Stop adding metrics if the queue is full
```
bantonsson, line 44: Won't we always skip the same metrics if all metrics are updated and there are more metrics than queue slots? Not to mention that we have already _lost_ the metric in question since we have reset the counters.

PerfectSlayer: Oh sure! It is the beauty of the current design ❤️ 
It is exactly the same behavior in IAST and WAF collector. I kept it for consistency.
Except I stop reading at counter as soon as it is full where this other collectors don't care about empty queue and reset everything 🤷 

I don't think this should be fixed in this PR but should be part on another one to improve the whole telemetry thing.
What do you think about it?

internal-api/src/main/java/datadog/trace/api/metrics/SpanMetricsImpl.java
bantonsson: The new code looks better, but this method is still here 🤷🏼‍♂️

internal-api/src/main/java/datadog/trace/api/telemetry/CoreMetricCollector.java
PerfectSlayer: Done. I added it at telemetry level as it applies to all collectors.

internal-api/src/main/java/datadog/trace/api/metrics/SpanMetricsImpl.java
PerfectSlayer: Whoopsie... Pushed the wrong commit 😞 
Thanks for the catch!

internal-api/src/main/java/datadog/trace/api/telemetry/MetricCollector.java
```diff
@@ -13,6 +13,9 @@ public interface MetricCollector<M extends Metric> {
 
   int RAW_QUEUE_SIZE = 1024;
 
+  // TODO All implements are based on queueing metrics and never care if there queue is full or not.
+  // TODO This leads to always resetting counters even if the related metric could not be enqueue.
+  // TODO So we may loose information during this call.
```
bantonsson, line 18:
```suggestion
  // TODO All implementations are based on queueing metrics and never care if the queue is full or not.
  // TODO This leads to always resetting counters even if the related metric could not be enqueued.
  // TODO So we may loose information during this call.
```

bantonsson (approved):

PR #5482, opened 2023-06-27
Change traceDebug with dynamic config

randomanderson:
# What Does This Do
This PR allows the changing of `traceDebug` with dynamic config. Since `traceDebug` only controls the log level of the tracer, the majority of change are related to logging.

There were a few changes made for this to work:

**Simplify the log level switcher**:
I removed `SwitchableLogLevelFactory` and merged it into `DDLoggerFactory` to remove 1 level of indirection and confusion. (The current code is still a little confusing but less so)

**Move classes into `internal-api`**: 
`GlobalLogLevelSwitcher`, `LogLevelSwitcher`, and `LogLevel` needed to be in internal-api.

**Hook log level switching into `TracingConfigPoller`**
Change the log level when we receive a true or false

## Testing
Manually tested using the [test agent](https://github.com/DataDog/dd-apm-test-agent)


dd-java-agent/agent-logging/src/test/groovy/datadog/trace/logging/ddlogger/DDLoggerTest.groovy
```diff
@@ -36,23 +36,23 @@ class DDLoggerTest extends LogValidatingSpecification {
...
     then:
     validateEnabled(logger, trace, debug, info, warn, error)
     factory.switchLevel(LogLevel.TRACE)
     validateEnabled(logger, true, true, true, true, true)
     factory.switchLevel(LogLevel.DEBUG)
-    validateEnabled(logger, trace, true, true, true, true)
+    validateEnabled(logger, false, true, true, true, true)
     factory.switchLevel(LogLevel.INFO)
-    validateEnabled(logger, trace, debug, true, true, true)
+    validateEnabled(logger, false, false, true, true, true)
```
bantonsson, line 49: So according to this test, you can now switch to any log level, not only more verbose ones, which was the case before, but the doc in `LogLevelSwitcher` still says:
> Temporarily switch the current LogLevel to a new LogLevel. Will only change LogLevel to a more verbose level.

dd-trace-core/src/main/java/datadog/trace/core/TracingConfigPoller.java
```diff
@@ -97,6 +99,14 @@ void applyConfigOverrides(LibConfig libConfig) {
     DynamicConfig<?>.Builder builder = dynamicConfig.initial();
 
     maybeOverride(builder::setDebugEnabled, libConfig.debugEnabled);
+    if (libConfig.debugEnabled != null) {
+      if (Boolean.TRUE.equals(libConfig.debugEnabled)) {
+        GlobalLogLevelSwitcher.get().switchLevel(LogLevel.DEBUG);
+      } else {
+        GlobalLogLevelSwitcher.get().switchLevel(LogLevel.INFO);
```
bantonsson, line 106: Shouldn't this be `GlobalLogLevelSwitcher.get().restore()` as well? I mean if the startup logs have been disabled, the the log level should be `WARN`.

randomanderson: It shouldn't be `restore()` but the logic is more complicated than I thought.

The `false` case is to turn off debug mode if it was enabled at startup. The default log level for the tracer without debug mode is `INFO` _unless_ startup logs is disabled in which case the default level is `WARN`.

I'll update the code

dd-java-agent/agent-logging/src/test/groovy/datadog/trace/logging/ddlogger/DDLoggerTest.groovy
randomanderson: I updated the docs in LogLevelSwitcher

dd-trace-core/src/main/java/datadog/trace/core/TracingConfigPoller.java
randomanderson: Code updated

bantonsson (approved):

PR #5484, opened 2023-06-28
fix: Fix release creation

PerfectSlayer:
# What Does This Do

This should fix the release creation. It will also brings the capability of rerunning the action on an existing tag.
This step was not tested on repository to prevent creating test releases...

# Motivation

# Additional Notes


richardstartin (approved):
🤞🏻

PR #5488, opened 2023-06-28
Add GitHub actions documentation

PerfectSlayer:
# What Does This Do

This PR adds a README about installed GitHub actions and their usage.

# Motivation

# Additional Notes

`create-next-milestone` action looks suspicious.
Should we clean the `gradle-wrapper-validation` action?

### DIFF
--- a/.github/workflows/README.md
+++ b/.github/workflows/README.md
@@ -1,0 +1,107 @@
+# GitHub Actions Documentation
+
+This lists and describes the repository GitHub actions.
+
+## Release Management
+
+### add-assets-to-release
+
+_Trigger:_ When a release is published
+
+_Actions:_
+* Ensure the release name is properly formatted (using `x.y.z` format),
+* Download `dd-java-agent`, `dd-trace-api` and `dd-trace-ot` artifacts from Sonatype (aka _Maven Central_ and upload them to the release (`dd-java-agent` will also be uploaded without version number).
+
+_Recovery:_ Download artifacts and upload them manually to the release.
+
+### add-milestone-to-pull-requests
+
+_Trigger:_ When a PR to `master` is closed
+
+_Action:_ Get the last (by name) opened milestone and affect it to the closed pull request.
+
+_Recovery:_ Attach the milestone by hand to the PR.
+
+### create-next-milestone
+
+_Trigger:_ When closing a milestone
+
+_Action:_ Create a new milestone by incrementing minor version.
+
+_Comment:_ Already done when closing a tag. To delete?
+
+### draft-release-notes-on-tag
+
+_Trigger:_ When creating a tag, or manually (providing a tag)
+
+_Actions:_
+
+* Fetch merged pull requests from the related tag milestone,
+* Generate changelog draft,
+* Create a new draft release for given tag with the generated changelog.
+
+_Recovery:_ Manually trigger the action again on the relevant tag.
+
+## increment-milestone-on-tag
+
+_Trigger:_ When creating a tag
+
+_Actions:_
+* Close the milestone related to the tag,
+* Create a new milestone by incrementing minor version.
+
+_Recovery:_ Manually close the related milestone and create a new one.
+
+_Notes:_ This actions will handle _minor_ releases only.
+As there is no milestone for _patch_ releases, it won't close and create _patch_ releated milestone.
+
+## update-download-releases
+
+_Trigger:_ When a release is published
+
+_Action:_ Update the _download releases_ with the latest release artifact.
+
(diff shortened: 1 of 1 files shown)


.github/workflows/README.md
```diff
@@ -0,0 +1,88 @@
...
+
+**TO DELETE?**
+
+_Trigger:_ When closing a milestone
+
+_Action:_ Create a new milestone by incrementing minor version.
+
+### draft-release-notes-on-tag
+
+_Trigger:_ When creating a tag, or manully (providing a tag)
```
ygree, line 35: A typo in "manully" should be "manually"

ygree (approved):
That's useful! Thanks!

.github/workflows/README.md
PerfectSlayer: Nice catch. Thanks!

.github/workflows/README.md
```diff
@@ -0,0 +1,88 @@
...
+
+_Recovery:_ Download artifacts and upload them manually to the release.
+
+### add-milestone-to-pull-requests
+
+_Trigger:_ When a PR to `master` is closed
+
+_Action:_ Get the last (by name) opened milestone and affect it to the closed pull request.
+
+_Recovery:_ Affect the milestone by hand to the PR.
```
bantonsson, line 23:
```suggestion
_Recovery:_ Attach the milestone by hand to the PR.
```

.github/workflows/README.md
```diff
@@ -0,0 +1,88 @@
...
+* Ensure the release name is properly formatted (using `x.y.z` format),
+* Download `dd-java-agent`, `dd-trace-api` and `dd-trace-ot` artifacts from Sonatype (aka _Maven Central_ and upload them to the release (`dd-java-agent` will also be uploaded without version number).
+
+_Recovery:_ Download artifacts and upload them manually to the release.
+
+### add-milestone-to-pull-requests
+
+_Trigger:_ When a PR to `master` is closed
+
+_Action:_ Get the last (by name) opened milestone and affect it to the closed pull request.
```
bantonsson, line 21: As far as I can see it sorts by the milestone `number`, selecting the most recently created open milestone.

.github/workflows/README.md
```diff
@@ -0,0 +1,88 @@
...
+
+_Recovery:_ Affect the milestone by hand to the PR.
+
+### create-next-milestone
+
+**TO DELETE?**
```
bantonsson, line 27: Not do be fixed in this PR, but I think we should if we create the new milestone on the tag anyway.

.github/workflows/README.md
```diff
@@ -0,0 +1,88 @@
...
+
+_Trigger:_ When creating a tag, or manually (providing a tag)
+
+_Actions:_
+
+* Fetch merged pull requests from the related tag milestone,
+* Generate changelog draft,
+* Create a new draft release for given tag with the generated changelog.
+
+_Recovery:_ Manually trigger the action again.
```
bantonsson, line 43:
```suggestion
_Recovery:_ Manually trigger the action again on the relevant tag.
```

.github/workflows/README.md
```diff
@@ -0,0 +1,88 @@
...
+
+_Trigger:_ When pushing commits to `master` or any pull request to `master`
+
+_Action:_ Run GitHub CodeQL action and upload result to GitHub security tab.
+
+## Disabled
+
+### gradle-wrapper-validation
+
+**TO DELETE?**
```
bantonsson, line 86: Yeah, that one we haven't run in quite a while.

.github/workflows/README.md
PerfectSlayer: You are right. That means we might need to fix this to get them by name using semver order...
What do you thing of it?

bantonsson: Yeah, the _quick_ fix would be to splitting names and sort by the semver order. The longer term fix for supporting multiple releases would be to find the latest release tag from the commit and select the milestone based on that (then we could trigger this on `release/` branches as well.

bantonsson: Ah, the node JS way of writing your own library 😉 Joking aside, it would probably be good to list the different things we want to do and see what is already out there.

PerfectSlayer: Most of the stuff is already done in GitHub actions script and I must have already write the missing parts for the `update-download-releases` action.

> what is already out there

We mainly relies on third-party actions. This cause our workflows to be quite verbose and makes it difficult to set more complicated logic in place (the main reason is when you call another action, you will have to serialize your context to step output and restore it later in another action step, playing with step output and step run condition 😫)

Let's move this discussion for next guild meeting 😉

PR #5491, opened 2023-06-29
Remove warning about unknown grpc.server.request.metadata address

smola:
# What Does This Do

# Motivation
Whenever we load the WAF, we issue a warning `WAF has rule against unknown address grpc.server.request.metadata`. This is not actionable: it is a known address, even if we don't support it yet.

# Additional Notes


dd-java-agent/appsec/src/main/java/com/datadog/appsec/event/data/KnownAddresses.java
```diff
@@ -93,6 +93,9 @@ public interface KnownAddresses {
 
   Address<Object> GRPC_SERVER_REQUEST_MESSAGE = new Address<>("grpc.server.request.message");
 
+  // XXX: Not really used yet, but it's a known address and we should treat it as unknown.
```
bantonsson, line 96:
```suggestion
  // XXX: Not really used yet, but it's a known address and we should not treat it as unknown.
```

PR #5492, opened 2023-06-29
Provide instrumentation name context to span

PerfectSlayer:
# What Does This Do

This PR introduce instrumentation name as context to the span.
The context was added for the following cases:
* opentracing instrumentation
* opentelemetry instrumentation
* annotation instrumentation
* dd-trace-ot module

# Motivation

This is required for span metrics.

# Additional Notes

This PR is the third part of the Span Metrics PR https://github.com/DataDog/dd-trace-java/pull/5382

I might do a pass in most of instrumentations to add this context if needed in another PR.
There might be complication with common decorator though.

### DIFF
--- a/dd-java-agent/agent-ci-visibility/src/main/java/datadog/trace/civisibility/DDTestImpl.java
+++ b/dd-java-agent/agent-ci-visibility/src/main/java/datadog/trace/civisibility/DDTestImpl.java
@@ -49,7 +49,7 @@ public class DDTestImpl implements DDTest {
 
     AgentTracer.SpanBuilder spanBuilder =
         AgentTracer.get()
-            .buildSpan("ci-visibility", testDecorator.component() + ".test")
+            .buildSpan(testDecorator.component() + ".test")
             .ignoreActiveSpan()
             .asChildOf(null)
             .withRequestContextData(
--- a/dd-java-agent/agent-ci-visibility/src/main/java/datadog/trace/civisibility/DDTestImpl.java
+++ b/dd-java-agent/agent-ci-visibility/src/main/java/datadog/trace/civisibility/DDTestImpl.java
@@ -49,7 +49,7 @@ public class DDTestImpl implements DDTest {
 
     AgentTracer.SpanBuilder spanBuilder =
         AgentTracer.get()
-            .buildSpan(testDecorator.component() + ".test")
+            .buildSpan("ci-visibility", testDecorator.component() + ".test")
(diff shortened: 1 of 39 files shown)


richardstartin:
Nice - this would be a good context field to add to profile samples. We would need to dictionarize the instrumentation name and set it through the profiling context integration the way we do the span name currently. I will watch the PR to know when it's available to build against.

dd-java-agent/agent-ci-visibility/src/main/java/datadog/trace/civisibility/DDTestImpl.java
```diff
@@ -49,7 +49,7 @@ public DDTestImpl(
 
     AgentTracer.SpanBuilder spanBuilder =
         AgentTracer.get()
-            .buildSpan(testDecorator.component() + ".test")
+            .buildSpan("ci-visibility", testDecorator.component() + ".test")
```
nikita-tkachenko-datadog, line 52: CI Visibility also starts spans in `DDTestSuiteImpl`, `DDTestModuleImpl` and `DDTestSessionImpl`. I assume those should to be updated as well?

PerfectSlayer: Ideally, they all should get updated at the end.

I reverted this change because I introduced deprecated backward-compatible methods to support _the old way_ of creating span (ie without instrumentation context name) in the meantime.

dd-java-agent/instrumentation/opentelemetry/opentelemetry-1.4/src/main/java/datadog/trace/instrumentation/opentelemetry14/OtelTracer.java
```diff
@@ -7,6 +7,7 @@
 
 @ParametersAreNonnullByDefault
 class OtelTracer implements Tracer {
+  private static final String INSTRUMENTATION_NAME = "opentelemetry";
```
bantonsson, line 10: This should be `otel` as per the RFC

PerfectSlayer: Oh... With all the discussions we had to prevent using _otel_ in our docs, I got wrong this part 😓

bantonsson (approved):

PR #5497, opened 2023-06-29
Add context propagation diagnostic information to debug logs

PerfectSlayer:
# What Does This Do

This PR will add diagnostic information to debug logs for context propagation investigation issues.

# Motivation

I had to address few context propagation issues the past weeks and it feels this kind of information will be helpful.

# Additional Notes

I don't think logging this kind of information will be problematic as they are send/receive from outside.

### DIFF
--- a/dd-trace-core/src/main/java/datadog/trace/core/propagation/ExtractedContext.java
+++ b/dd-trace-core/src/main/java/datadog/trace/core/propagation/ExtractedContext.java
@@ -2,6 +2,7 @@ package datadog.trace.core.propagation;
 
 import datadog.trace.api.DDTraceId;
 import datadog.trace.api.TraceConfig;
+import datadog.trace.api.sampling.PrioritySampling;
 import datadog.trace.bootstrap.instrumentation.api.TagContext;
 import java.util.Map;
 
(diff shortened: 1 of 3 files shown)


dd-trace-core/src/main/java/datadog/trace/core/propagation/ExtractedContext.java
```diff
@@ -58,4 +58,29 @@ public final long getEndToEndStartTime() {
...
+    if (getOrigin() != null) {
+      builder.append("origin=").append(getOrigin()).append(", ");
+    }
+    if (getTags() != null) {
+      builder.append("tags=").append(getTags()).append(", ");
+    }
+    if (getBaggage() != null) {
+      builder.append("baggage=").append(getBaggage()).append(", ");
+    }
+    builder.append("samplingPriority=").append(getSamplingPriority()).append('}');
```
bantonsson, line 83: I think we should skip the `samplingPriority` if it is `PrioritySampling.UNSET`

bantonsson (approved):

dd-trace-core/src/main/java/datadog/trace/core/propagation/ExtractedContext.java
PerfectSlayer: Agreed & done 👌

PR #5501, opened 2023-06-30
Clean up lib-injection docker images

PerfectSlayer:
# What Does This Do

This PR introduces GitHub workflow to clean up GitHub container registry from lib-injection docker images.

# Motivation

`lib-injection` produces nearly 2k docker images per week. This should cut our costs on GH storage.

# Additional Notes

I spent few days to clean up GHCR first and removed two others unused packages (`dd-java-agent` and `dd-trace-java`).

### DIFF
--- a/.github/workflows/lib-injection-prune-registry.yaml
+++ b/.github/workflows/lib-injection-prune-registry.yaml
@@ -1,0 +1,48 @@
+name: Prune Lib Injection images
+on:
+  schedule:
+    - cron: '15 3 * * *'
+  workflow_dispatch:
+
+jobs:
+  prune-init-images:
+    name: Prune dd-lib-java-init docker images
+    runs-on: ubuntu-latest
+    permissions:
+      contents: read
+      packages: write
+    steps:
+    - name: Prune registry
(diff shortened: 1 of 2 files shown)


.github/workflows/lib-injection-prune-registry.yaml
```diff
@@ -0,0 +1,48 @@
...
+      with:
+        token: ${{ secrets.GITHUB_TOKEN }}
+        organization: Datadog
+        container: dd-trace-java/dd-lib-java-init
+        keep-younger-than: 7 # days
+        keep-last: 10
+        keep-tags: |
+          latest_snapshot
+        prune-tags-regexes: |
+          ^[a-z0-9]{40}$
```
bantonsson, line 26: I guess this is only removing images tagged with the hash?

bantonsson (approved):

.github/workflows/lib-injection-prune-registry.yaml
PerfectSlayer: Yes it is. If we come to push tagged images with another kind of label (let's say `1.18.0` for example 😉 ), I don't want it to be removed because we forgot about our CI.

PerfectSlayer:
I just removed `dd-java-agent` and `dd-trace-java` package from our repository.
Our registry must be clean from now on 👍

PR #5505, opened 2023-06-30
[Spark] Instrument SparkSubmit to capture errors outside of SparkListener interface

paul-laffon-dd:
# What Does This Do

Instrument the `SparkSubmit.runMain(...)` function to capture all errors occurring in a spark application

This SparkSubmit class is used by spark to launch the application in non YARN/Mesos environnements. The YARN case is already covered by instrumenting the `ApplicationMaster.finish()` function (from this PR https://github.com/DataDog/dd-trace-java/pull/5267)

# Motivation

We are currently using the interface `SparkListener` to receive events from spark. However, this interface is only capturing errors that occur during the spark computations, but not in arbitrary user code

```scala
def main() = {
  val spark = SparkSession.builder.getOrCreate()
  val df = spark.read.parquet("...")
  
  // Exception will be captured because it's happening during spark computations
  df.map(x => new RuntimeException("Some exception"))

  // Exception currently not captured because it is thrown outside of spark computations
  throw new RuntimeException("Some exception")
}
```

### DIFF
--- a/dd-java-agent/instrumentation/spark/src/test/groovy/SparkTest.groovy
+++ b/dd-java-agent/instrumentation/spark/src/test/groovy/SparkTest.groovy
@@ -6,6 +6,7 @@ import datadog.trace.instrumentation.spark.DatadogSparkListener
 import datadog.trace.test.util.Flaky
 import org.apache.hadoop.yarn.api.records.FinalApplicationStatus
 import org.apache.hadoop.yarn.conf.YarnConfiguration
+import org.apache.spark.deploy.SparkSubmit
 import org.apache.spark.deploy.yarn.ApplicationMaster
 import org.apache.spark.deploy.yarn.ApplicationMasterArguments
 import org.apache.spark.sql.SparkSession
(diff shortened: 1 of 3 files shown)


dd-java-agent/instrumentation/spark/src/test/groovy/SparkTest.groovy
```diff
@@ -172,6 +173,36 @@ class SparkTest extends AgentTestRunner {
...
+
+    expect:
+    assertTraces(1) {
+      trace(1) {
+        span {
+          operationName "spark.application"
+          resourceName "spark.application"
+          spanType "spark"
+          errored true
+          parent()
```
PerfectSlayer, line 197: Should you test span tags like error message and type?

paul-laffon-dd: Yes, those can be tested to validate we are catching the right exception. The small caveat is if spark changes this error type / message in future versions, but this should be minor

PerfectSlayer (approved):
LGTM. Don't forget to squash before merging 👍

PR #5507, opened 2023-06-30
[DSM] - calculate edge latency using message timestamp

kr-igor:
# What Does This Do
DSM will now use message timestamp to measure latency at the beginning on the pathway.

#  Motivation
When the customer has consumer instrumented, but not producer they will see `0` as edge latency.  At the same time, most of the queueing tech have a way of injecting message timestamp, which shows the time when the message was produced. We want to use this timestamp if we can't calculate the latency in any other way.

# Additional Notes
Changes affect DSM only, in particular DSM instrumentation for Kafka and RabbitMQ.

### DIFF
--- a/internal-api/src/main/java/datadog/trace/bootstrap/instrumentation/api/AgentTracer.java
+++ b/internal-api/src/main/java/datadog/trace/bootstrap/instrumentation/api/AgentTracer.java
@@ -199,10 +199,10 @@ public class AgentTracer {
      *
      * @param span active span
      * @param sortedTags alphabetically sorted tags for the checkpoint (direction, queue type etc)
-     * @param defaultTimestamp unix timestamp to use as a start of the pathway if this is the first checkpoint in the chain.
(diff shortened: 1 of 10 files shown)


internal-api/src/main/java/datadog/trace/bootstrap/instrumentation/api/AgentTracer.java
```diff
@@ -194,7 +194,8 @@ public interface TracerAPI
 
     CallbackProvider getUniversalCallbackProvider();
 
-    void setDataStreamCheckpoint(AgentSpan span, LinkedHashMap<String, String> sortedTags);
+    void setDataStreamCheckpoint(
+        AgentSpan span, LinkedHashMap<String, String> sortedTags, long defaultTimestamp);
```
PerfectSlayer, line 198: Can you provide javadoc with the unit of the timestamp and describe the special `0` value case?

PerfectSlayer (approved):

PR #5512, opened 2023-07-03
Implement repository index sharing between processes

nikita-tkachenko-datadog:
# What Does This Do

# Motivation

# Additional Notes


dd-java-agent/agent-ci-visibility/src/main/java/datadog/trace/civisibility/ipc/RepoIndexResponse.java
```diff
@@ -0,0 +1,32 @@
...
+    return SignalType.REPO_INDEX_RESPONSE;
+  }
+
+  @Override
+  public ByteBuffer serialize() {
+    return index.serialize();
+  }
+
+  public static RepoIndexResponse deserialize(ByteBuffer buffer) {
+    RepoIndex repoIndex = RepoIndex.deserialize(buffer);
```
tonyredondo, line 29: How big is the `RepoIndex` object in serialized bytes?
What is the impact of sending multiple times for each module over TCP? I guess much lower than building the index in each module.

dd-java-agent/agent-ci-visibility/src/main/java/datadog/trace/civisibility/ipc/SignalType.java
```diff
@@ -1,9 +1,13 @@
...
 public enum SignalType {
+  // requests
   MODULE_EXECUTION_RESULT((byte) 0),
-  ACK((byte) 1),
-  ERROR((byte) 2);
+  REPO_INDEX_REQUEST((byte) 1),
+  // responses
+  ACK((byte) 2),
+  ERROR((byte) 3),
+  REPO_INDEX_RESPONSE((byte) 4);
```
tonyredondo, line 10: There's no way to mix versions of client and server right? because this is a breaking change and different versions of client and server will be unable to connect.

In this case I kind of prefer to have a bigger separation between requests and responses signals, something like:

[0-99] -> Requests
[100-199] -> Responses

So you will have enough slots to add signals without changing the value of existing ones.

tonyredondo (approved):
LGTM, only a couple of questions...

dd-java-agent/agent-ci-visibility/src/main/java/datadog/trace/civisibility/ipc/RepoIndexResponse.java
nikita-tkachenko-datadog: Yep, in the vast majority of cases the parent and the children will run on the same machine, sending the index over TCP is on average two orders of magnitude faster than building it from scratch

PR #5518, opened 2023-07-03
Add INTEGRATIONS system tests to Java Tracer CI

hokitam:
# What Does This Do
Add the INTEGRATIONS scenario from System Tests to the Java Tracer's CI.

# Motivation
Enabling INTEGRATIONS scenario allow us to catch potential breakages to integrations early for future PRs.

# Additional Notes

### DIFF
--- a/.circleci/config.yml
+++ b/.circleci/config.yml
@@ -804,6 +804,14 @@ jobs:
             DD_SITE=datadoghq.com DD_API_KEY=$SYSTEM_TESTS_E2E_DD_API_KEY DD_APPLICATION_KEY=$SYSTEM_TESTS_E2E_DD_APP_KEY ./run.sh APM_TRACING_E2E_SINGLE_SPAN
 
       - run:
+          name: Run APM Integrations tests
+          # Stop the job after 5m to avoid excessive overhead. Will need adjustment as more tests are added.
+          no_output_timeout: 5m
+          command: |
+            cd system-tests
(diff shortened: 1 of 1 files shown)


.circleci/config.yml
```diff
@@ -803,6 +803,14 @@ jobs:
             cd system-tests
             DD_SITE=datadoghq.com DD_API_KEY=$SYSTEM_TESTS_E2E_DD_API_KEY DD_APPLICATION_KEY=$SYSTEM_TESTS_E2E_DD_APP_KEY ./run.sh APM_TRACING_E2E_SINGLE_SPAN
 
+      - run:
+          name: Run APM Integrations tests
+          # Stop the job after 5m to avoid excessive overhead. Will need adjustment as more tests are added.
+          no_output_timeout: 5m
+          command: |
+            cd system-tests
+            DD_SITE=datadoghq.com DD_API_KEY=$SYSTEM_TESTS_E2E_DD_API_KEY DD_APPLICATION_KEY=$SYSTEM_TESTS_E2E_DD_APP_KEY ./run.sh INTEGRATIONS
```
bantonsson, line 812: I don't know enough about how the `system-tests` are structured when it comes to these tests, but do they need to run for every `weblog-variant`?

hokitam: Right now we only run them for spring-boot, but we're also working on expanding the coverage to cover the other weblog_variants in parallel, so I think it's OK to have them run for every weblog-variant here similar to all the other tests we already have in the CI, wdyt?

bantonsson: Sure, I was just wondering if it made sense from a testing perspective to run the DSM tests on different web frameworks, since it's not so much about testing the web framework as it is about testing a _queue_ like `kafka` or `rabbitmq`.

hokitam: updated the PR to only run this for springboot! ptal!

bantonsson (approved):

PR #5520, opened 2023-07-04
HSTS missing header vulnerability detection

DDJavierSantos:
# What Does This Do
Detect and report HSTS header missing vulnerabilities
# Motivation
To cover all of the header vulnerabilities
# Additional Notes

### DIFF
--- a/dd-java-agent/agent-bootstrap/src/main/java/datadog/trace/bootstrap/instrumentation/decorator/HttpServerDecorator.java
+++ b/dd-java-agent/agent-bootstrap/src/main/java/datadog/trace/bootstrap/instrumentation/decorator/HttpServerDecorator.java
@@ -401,20 +401,33 @@ public abstract class HttpServerDecorator<REQUEST, CONNECTION, RESPONSE, REQUEST
   }
 
   private Flow<Void> callIGCallbackRequestHeaders(AgentSpan span, REQUEST_CARRIER carrier) {
-    CallbackProvider cbp = tracer().getCallbackProvider(RequestContextSlot.APPSEC);
+    CallbackProvider cbpAppsec = tracer().getCallbackProvider(RequestContextSlot.APPSEC);
+    CallbackProvider cbpIast = tracer().getCallbackProvider(RequestContextSlot.IAST);
     RequestContext requestContext = span.getRequestContext();
     AgentPropagation.ContextVisitor<REQUEST_CARRIER> getter = getter();
-    if (requestContext == null || cbp == null || getter == null) {
+    if (requestContext == null || getter == null) {
       return Flow.ResultFlow.empty();
     }
-    IGKeyClassifier igKeyClassifier =
-        IGKeyClassifier.create(
-            requestContext,
-            cbp.getCallback(EVENTS.requestHeader()),
-            cbp.getCallback(EVENTS.requestHeaderDone()));
-    if (null != igKeyClassifier) {
-      getter.forEachKey(carrier, igKeyClassifier);
-      return igKeyClassifier.done();
+    if (cbpIast != null) {
+      IGKeyClassifier igKeyClassifier =
+          IGKeyClassifier.create(
+              requestContext,
+              cbpIast.getCallback(EVENTS.requestHeader()),
+              cbpIast.getCallback(EVENTS.requestHeaderDone()));
+      if (null != igKeyClassifier) {
+        getter.forEachKey(carrier, igKeyClassifier);
+      }
+    }
+    if (cbpAppsec != null) {
+      IGKeyClassifier igKeyClassifier =
+          IGKeyClassifier.create(
+              requestContext,
+              cbpAppsec.getCallback(EVENTS.requestHeader()),
+              cbpAppsec.getCallback(EVENTS.requestHeaderDone()));
+      if (null != igKeyClassifier) {
+        getter.forEachKey(carrier, igKeyClassifier);
+        return igKeyClassifier.done();
+      }
     }
     return Flow.ResultFlow.empty();
   }
--- a/dd-java-agent/agent-iast/src/main/java/com/datadog/iast/IastRequestContext.java
+++ b/dd-java-agent/agent-iast/src/main/java/com/datadog/iast/IastRequestContext.java
@@ -19,6 +19,9 @@ public class IastRequestContext implements HasMetricCollector {
   private final TaintedObjects taintedObjects;
   private final OverheadContext overheadContext;
   private final IastMetricCollector collector;
+  private volatile boolean hstsHeaderIsSet;
+  private volatile boolean xForwardedProtoIsHtttps;
+  private volatile String contentType;
 
   public IastRequestContext() {
     this(TaintedObjects.acquire(), null);
@@ -32,6 +35,9 @@ public class IastRequestContext implements HasMetricCollector {
       final TaintedObjects taintedObjects, final IastMetricCollector collector) {
     this.vulnerabilityBatch = new VulnerabilityBatch();
     this.spanDataIsSet = new AtomicBoolean(false);
+    this.hstsHeaderIsSet = false;
+    this.xForwardedProtoIsHtttps = false;
+    this.contentType = null;
     this.overheadContext = new OverheadContext();
     this.taintedObjects = taintedObjects;
     this.collector = collector;
@@ -39,6 +45,30 @@ public class IastRequestContext implements HasMetricCollector {
 
   public VulnerabilityBatch getVulnerabilityBatch() {
     return vulnerabilityBatch;
+  }
+
+  public void setHstsHeaderIsSet() {
+    hstsHeaderIsSet = true;
+  }
+
+  public boolean getHstsHeaderIsSet() {
+    return hstsHeaderIsSet;
+  }
+
+  public void setXForwardedProtoIsHtttps() {
+    xForwardedProtoIsHtttps = true;
+  }
+
+  public boolean getXForwardedProtoIsHtttps() {
+    return xForwardedProtoIsHtttps;
+  }
+
+  public void setContentType(String contentType) {
+    this.contentType = contentType;
+  }
+
+  public String getContentType() {
+    return contentType;
   }
 
   public boolean getAndSetSpanDataIsSet() {
--- a/dd-java-agent/agent-iast/src/main/java/com/datadog/iast/RequestEndedHandler.java
+++ b/dd-java-agent/agent-iast/src/main/java/com/datadog/iast/RequestEndedHandler.java
@@ -9,6 +9,8 @@ import com.datadog.iast.taint.TaintedObjects;
 import datadog.trace.api.gateway.Flow;
 import datadog.trace.api.gateway.IGSpanInfo;
 import datadog.trace.api.gateway.RequestContext;
+import datadog.trace.api.iast.InstrumentationBridge;
+import datadog.trace.api.iast.sink.HstsMissingHeaderModule;
 import datadog.trace.api.internal.TraceSegment;
 import java.util.function.BiFunction;
 import javax.annotation.Nonnull;
@@ -26,6 +28,10 @@ public class RequestEndedHandler implements BiFunction<RequestContext, IGSpanInf
     final TraceSegment traceSegment = requestContext.getTraceSegment();
     final IastRequestContext iastRequestContext = IastRequestContext.get(requestContext);
     if (iastRequestContext != null) {
+      HstsMissingHeaderModule mod = InstrumentationBridge.HSTS_MISSING_HEADER_MODULE;
+      if (mod != null) {
+        mod.onRequestEnd(iastRequestContext, igSpanInfo);
+      }
       try {
         ANALYZED.setTagTop(traceSegment);
         final TaintedObjects taintedObjects = iastRequestContext.getTaintedObjects();
--- a/dd-java-agent/agent-iast/src/main/java/com/datadog/iast/RequestHeaderHandler.java
+++ b/dd-java-agent/agent-iast/src/main/java/com/datadog/iast/RequestHeaderHandler.java
@@ -1,0 +1,19 @@
+package com.datadog.iast;
+
+import datadog.trace.api.function.TriConsumer;
+import datadog.trace.api.gateway.RequestContext;
+import datadog.trace.api.gateway.RequestContextSlot;
+
+public class RequestHeaderHandler implements TriConsumer<RequestContext, String, String> {
+  private static final String X_FORWARDED_PROTO_HEADER = "X-Forwarded-Proto";
+
+  @Override
+  public void accept(RequestContext requestContext, String key, String value) {
+    final IastRequestContext ctx = requestContext.getData(RequestContextSlot.IAST);
+    if (null != ctx
+        && X_FORWARDED_PROTO_HEADER.equalsIgnoreCase(key)
+        && "https".equalsIgnoreCase(value)) {
+      ctx.setXForwardedProtoIsHtttps();
+    }
+  }
+}
--- a/dd-java-agent/agent-iast/src/main/java/com/datadog/iast/sink/HstsMissingHeaderModuleImpl.java
+++ b/dd-java-agent/agent-iast/src/main/java/com/datadog/iast/sink/HstsMissingHeaderModuleImpl.java
@@ -74,10 +74,11 @@ public class HstsMissingHeaderModuleImpl extends SinkModuleBase implements HstsM
     if (null == value || 0 == value.length()) {
       return false;
     }
-    if (value.toLowerCase().contains("max-age=0")) {
+    String lowercaseValue = value.toLowerCase();
+    if (lowercaseValue.contains("max-age=0")) {
       return true;
     }
-    if (value.toLowerCase().contains("max-age=-1")) {
+    if (lowercaseValue.contains("max-age=-1")) {
       return true;
     }
     return false;
--- a/dd-java-agent/agent-iast/src/main/java/com/datadog/iast/sink/HstsMissingHeaderModuleImpl.java
+++ b/dd-java-agent/agent-iast/src/main/java/com/datadog/iast/sink/HstsMissingHeaderModuleImpl.java
@@ -11,10 +11,15 @@ import datadog.trace.bootstrap.instrumentation.api.AgentTracer;
 import java.net.HttpURLConnection;
 import java.net.URL;
 import java.util.Map;
+import java.util.regex.Matcher;
+import java.util.regex.Pattern;
+import javax.annotation.Nullable;
 import org.slf4j.Logger;
 import org.slf4j.LoggerFactory;
 
 public class HstsMissingHeaderModuleImpl extends SinkModuleBase implements HstsMissingHeaderModule {
+  private static final Pattern MAX_AGE =
+      Pattern.compile("max-age=(\\d+)", Pattern.CASE_INSENSITIVE);
 
   private static final Logger LOGGER = LoggerFactory.getLogger(HstsMissingHeaderModuleImpl.class);
 
@@ -59,7 +64,7 @@ public class HstsMissingHeaderModuleImpl extends SinkModuleBase implements HstsM
 
   @Override
   public void onHstsHeader(String value) {
-    if (!isMaxAgeZero(value)) {
+    if (isValidMaxAge(value)) {
       final AgentSpan span = AgentTracer.activeSpan();
       final IastRequestContext ctx = IastRequestContext.get(span);
       if (ctx == null) {
@@ -70,17 +75,14 @@ public class HstsMissingHeaderModuleImpl extends SinkModuleBase implements HstsM
     }
   }
 
-  public static boolean isMaxAgeZero(String value) {
-    if (null == value || 0 == value.length()) {
+  private static boolean isValidMaxAge(@Nullable final String value) {
+    if (value == null) {
       return false;
     }
-    String lowercaseValue = value.toLowerCase();
-    if (lowercaseValue.contains("max-age=0")) {
-      return true;
+    final Matcher matcher = MAX_AGE.matcher(value);
+    if (!matcher.find()) {
+      return false;
     }
-    if (lowercaseValue.contains("max-age=-1")) {
-      return true;
-    }
-    return false;
+    return Integer.parseInt(matcher.group(1)) > 0;
   }
 }
--- a/dd-java-agent/agent-iast/src/main/java/com/datadog/iast/sink/HstsMissingHeaderModuleImpl.java
+++ b/dd-java-agent/agent-iast/src/main/java/com/datadog/iast/sink/HstsMissingHeaderModuleImpl.java
@@ -52,7 +52,7 @@ public class HstsMissingHeaderModuleImpl extends SinkModuleBase implements HstsM
           }
         }
       } catch (Throwable e) {
-        LOGGER.error("Exception while checking for hsts vulnerability", e);
+        LOGGER.debug("Exception while checking for missing HSTS headers vulnerability", e);
       }
     }
   }
--- a/dd-java-agent/agent-iast/src/main/java/com/datadog/iast/sink/HstsMissingHeaderModuleImpl.java
+++ b/dd-java-agent/agent-iast/src/main/java/com/datadog/iast/sink/HstsMissingHeaderModuleImpl.java
@@ -1,0 +1,85 @@
+package com.datadog.iast.sink;
+
+import com.datadog.iast.IastRequestContext;
+import com.datadog.iast.model.Vulnerability;
+import com.datadog.iast.model.VulnerabilityType;
+import com.datadog.iast.overhead.Operations;
+import datadog.trace.api.gateway.IGSpanInfo;
+import datadog.trace.api.iast.sink.HstsMissingHeaderModule;
+import datadog.trace.bootstrap.instrumentation.api.AgentSpan;
+import datadog.trace.bootstrap.instrumentation.api.AgentTracer;
+import java.net.HttpURLConnection;
+import java.net.URL;
+import java.util.Map;
+import org.slf4j.Logger;
+import org.slf4j.LoggerFactory;
+
+public class HstsMissingHeaderModuleImpl extends SinkModuleBase implements HstsMissingHeaderModule {
+
+  private static final Logger LOGGER = LoggerFactory.getLogger(HstsMissingHeaderModuleImpl.class);
+
+  @Override
+  public void onRequestEnd(final Object iastRequestContextObject, final IGSpanInfo igSpanInfo) {
+
+    final IastRequestContext iastRequestContext = (IastRequestContext) iastRequestContextObject;
+
+    if (!iastRequestContext.getHstsHeaderIsSet()) {
+      try {
+        Map<String, Object> tags = igSpanInfo.getTags();
+        String urlString = (String) tags.get("http.url");
+        if (null != tags.get("http.status_code")) {
+          Integer httpStatus = (Integer) tags.get("http.status_code");
+          if (httpStatus == HttpURLConnection.HTTP_MOVED_PERM
+              || httpStatus == HttpURLConnection.HTTP_MOVED_TEMP
+              || httpStatus == HttpURLConnection.HTTP_NOT_MODIFIED
+              || httpStatus == HttpURLConnection.HTTP_NOT_FOUND
+              || httpStatus == HttpURLConnection.HTTP_GONE
+              || httpStatus == HttpURLConnection.HTTP_INTERNAL_ERROR
+              || httpStatus == 307) {
+            return;
+          }
+        }
+        URL url = new URL(urlString);
+        if (null != iastRequestContext.getContentType()
+            && (iastRequestContext.getContentType().contains("text/html")
+                || iastRequestContext.getContentType().contains("application/xhtml+xml"))
+            && (url.getProtocol().equals("https")
+                || iastRequestContext.getXForwardedProtoIsHtttps())) {
+          final AgentSpan span = AgentTracer.activeSpan();
+          if (overheadController.consumeQuota(Operations.REPORT_VULNERABILITY, span)) {
+            reporter.report(
+                span, new Vulnerability(VulnerabilityType.HSTS_HEADER_MISSING, null, null));
+          }
+        }
+      } catch (Throwable e) {
+        LOGGER.error("Exception while checking for hsts vulnerability", e);
+      }
+    }
+  }
+
+  @Override
+  public void onHstsHeader(String value) {
+    if (!isMaxAgeZero(value)) {
+      final AgentSpan span = AgentTracer.activeSpan();
+      final IastRequestContext ctx = IastRequestContext.get(span);
+      if (ctx == null) {
+        return;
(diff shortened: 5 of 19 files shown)


dd-java-agent/agent-iast/src/main/java/com/datadog/iast/RequestEndedHandler.java
```diff
@@ -4,28 +4,57 @@
...
+              final AgentSpan span = AgentTracer.activeSpan();
+              if (overheadController.consumeQuota(Operations.REPORT_VULNERABILITY, span)) {
+                reporter.report(
+                    span, new Vulnerability(VulnerabilityType.HSTS_HEADER_MISSING, null, null));
+              }
+            }
+          }
+        } catch (Throwable e) {
+          // suppress
+          System.out.println("!!! Exception " + e);
```
smola, line 55: No println.

dd-java-agent/agent-iast/src/main/java/com/datadog/iast/RequestEndedHandler.java
```diff
@@ -4,28 +4,57 @@
...
+    this.reporter = dependencies.getReporter();
   }
 
   @Override
   public Flow<Void> apply(final RequestContext requestContext, final IGSpanInfo igSpanInfo) {
     final TraceSegment traceSegment = requestContext.getTraceSegment();
     final IastRequestContext iastRequestContext = IastRequestContext.get(requestContext);
+    System.out.println("!!! requestEndedHandler");
     if (iastRequestContext != null) {
+      if (!iastRequestContext.getHstsHeaderIsSet()) {
```
smola, line 38: Let's move this logic to a separate method or class.

dd-java-agent/agent-iast/src/main/java/com/datadog/iast/RequestEndedHandler.java
```diff
@@ -4,28 +4,57 @@
...
   public RequestEndedHandler(@Nonnull final Dependencies dependencies) {
     this.overheadController = dependencies.getOverheadController();
+    this.reporter = dependencies.getReporter();
   }
 
   @Override
   public Flow<Void> apply(final RequestContext requestContext, final IGSpanInfo igSpanInfo) {
     final TraceSegment traceSegment = requestContext.getTraceSegment();
     final IastRequestContext iastRequestContext = IastRequestContext.get(requestContext);
+    System.out.println("!!! requestEndedHandler");
```
smola, line 36: No println.

dd-java-agent/agent-iast/src/main/java/com/datadog/iast/sink/HttpResponseHeaderModuleImpl.java
```diff
@@ -18,9 +19,44 @@
...
+      }
+    }
+    if (CONTENT_TYPE_HEADER.equalsIgnoreCase(name)) {
+      if ("text/html".equalsIgnoreCase(value) || "application/xhtml+xml".equalsIgnoreCase(value)) {
+        final AgentSpan span = AgentTracer.activeSpan();
+        final IastRequestContext ctx = IastRequestContext.get(span);
+        if (ctx == null) {
+          return;
+        } else {
+          ctx.setHstsContentTypeIsHtml();
```
smola, line 45:
```suggestion
          ctx.setContentTypeIsHtml();
```

dd-java-agent/agent-iast/src/main/java/com/datadog/iast/sink/HttpResponseHeaderModuleImpl.java
```diff
@@ -18,9 +19,44 @@
...
+    if (HSTS_HEADER.equalsIgnoreCase(name)) {
+      final AgentSpan span = AgentTracer.activeSpan();
+      final IastRequestContext ctx = IastRequestContext.get(span);
+      if (ctx == null) {
+        return;
+      } else {
+        ctx.setHstsHeaderIsSet();
+      }
+    }
+    if (CONTENT_TYPE_HEADER.equalsIgnoreCase(name)) {
```
smola, line 38: Use `if else`, since only one header can match.

dd-java-agent/agent-iast/src/main/java/com/datadog/iast/sink/HttpResponseHeaderModuleImpl.java
```diff
@@ -18,9 +19,44 @@
...
+      }
+    }
+    if (X_FORWARDED_PROTO_HEADER.equalsIgnoreCase(name)) {
+      if ("https".equalsIgnoreCase(value)) {
+        final AgentSpan span = AgentTracer.activeSpan();
+        final IastRequestContext ctx = IastRequestContext.get(span);
+        if (ctx == null) {
+          return;
+        } else {
+          ctx.setHstsXForwardedProtoIsHtttps();
```
smola, line 56:
```suggestion
          ctx.setXForwardedProtoIsHtttps();
```

dd-java-agent/agent-iast/src/main/java/com/datadog/iast/RequestEndedHandler.java
DDJavierSantos: removed

dd-java-agent/agent-iast/src/main/java/com/datadog/iast/RequestEndedHandler.java
DDJavierSantos: done

dd-java-agent/agent-iast/src/main/java/com/datadog/iast/sink/HttpResponseHeaderModuleImpl.java
DDJavierSantos: done

dd-java-agent/agent-iast/src/main/java/com/datadog/iast/sink/HttpResponseHeaderModuleImpl.java
DDJavierSantos: done

dd-java-agent/agent-iast/src/main/java/com/datadog/iast/RequestEndedHandler.java
```diff
@@ -40,4 +51,39 @@ public Flow<Void> apply(final RequestContext requestContext, final IGSpanInfo ig
     }
     return Flow.ResultFlow.empty();
   }
+
+  static void checkForHstsMissingVulnerability(
```
manuel-alvarez-alvarez, line 55: I don't think this is the place for checking vulnerabilities, instead we could provide something to let our modules listen to the end request event and do the check in the `HttpResponseHeaderModuleImpl`

dd-java-agent/agent-iast/src/main/java/com/datadog/iast/RequestEndedHandler.java
```diff
@@ -40,4 +51,39 @@ public Flow<Void> apply(final RequestContext requestContext, final IGSpanInfo ig
...
+              || iastRequestContext.getXForwardedProtoIsHtttps()) {
+            final AgentSpan span = AgentTracer.activeSpan();
+            if (overheadController.consumeQuota(Operations.REPORT_VULNERABILITY, span)) {
+              reporter.report(
+                  span, new Vulnerability(VulnerabilityType.HSTS_HEADER_MISSING, null, null));
+            }
+          }
+        }
+      } catch (Throwable e) {
+        // do nothing
```
manuel-alvarez-alvarez, line 85: Don't do that please, at least log it as `debug`/`warn` so we can troubleshoot if something goes wrong.

dd-java-agent/agent-iast/src/main/java/com/datadog/iast/sink/HttpResponseHeaderModuleImpl.java
```diff
@@ -18,10 +19,42 @@
...
+    if (HSTS_HEADER.equalsIgnoreCase(name) && !isMaxAgeZero(value)) {
+      final AgentSpan span = AgentTracer.activeSpan();
+      final IastRequestContext ctx = IastRequestContext.get(span);
+      if (ctx == null) {
+        return;
+      } else {
+        ctx.setHstsHeaderIsSet();
+      }
+    } else if (CONTENT_TYPE_HEADER.equalsIgnoreCase(name)) {
+      if ("text/html".equalsIgnoreCase(value) || "application/xhtml+xml".equalsIgnoreCase(value)) {
```
manuel-alvarez-alvarez, line 38: Did you discuss using the appsec `Address<Map<String, List<String>>> RESPONSE_HEADERS_NO_COOKIES`?, I'm pretty sure that we might lose some headers (specially if they are set by the frameworks themselves)

dd-java-agent/agent-iast/src/main/java/com/datadog/iast/RequestEndedHandler.java
smola: That sounds reasonable, but let's keep it simple. For example, I don't think this needs to go to `internal-api` or `InstrumentationBridge` and it should stay within `agent-iast` module boundaries.

manuel-alvarez-alvarez: Yep, I was thinking about something living inside our `IastSystem`

DDJavierSantos: Moved this code to HttpResponseHeaderModuleImpl

dd-java-agent/agent-iast/src/main/java/com/datadog/iast/RequestEndedHandler.java
DDJavierSantos: exception logged now

dd-java-agent/agent-iast/src/main/java/com/datadog/iast/sink/HttpResponseHeaderModuleImpl.java
DDJavierSantos: After some conversation we think the current design is simpler and will keep it

dd-java-agent/agent-iast/src/main/java/com/datadog/iast/sink/HttpResponseHeaderModuleImpl.java
```diff
@@ -1,27 +1,70 @@
...
   private static final String SET_COOKIE_HEADER = "Set-Cookie";
+  private static final String HSTS_HEADER = "Strict-Transport-Security";
+
+  private static final String CONTENT_TYPE_HEADER = "Content-Type";
+  private static final String X_FORWARDED_PROTO_HEADER = "X-Forwarded-Proto";
 
   @Override
   public void onHeader(@Nonnull final String name, final String value) {
-    if (SET_COOKIE_HEADER.equalsIgnoreCase(name)) {
+    if (HSTS_HEADER.equalsIgnoreCase(name) && !isMaxAgeZero(value)) {
```
smola, line 39: I think `max-age` is mandatory? This only checks that it's positive when it's present?

dd-java-agent/agent-iast/src/main/java/com/datadog/iast/RequestEndedHandler.java
```diff
@@ -16,16 +17,20 @@
...
     this.overheadController = dependencies.getOverheadController();
+    this.reporter = dependencies.getReporter();
   }
 
   @Override
   public Flow<Void> apply(final RequestContext requestContext, final IGSpanInfo igSpanInfo) {
     final TraceSegment traceSegment = requestContext.getTraceSegment();
     final IastRequestContext iastRequestContext = IastRequestContext.get(requestContext);
     if (iastRequestContext != null) {
+      HttpResponseHeaderModuleImpl.checkForHstsMissingVulnerability(
```
smola, line 32: We could probably just call this `checkForHeaderVulnerabilities`, since we need to add at least one more here.

dd-java-agent/agent-iast/src/main/java/com/datadog/iast/sink/HttpResponseHeaderModuleImpl.java
DDJavierSantos: Yes, it is mandatory:

dd-java-agent/agent-jmxfetch/integrations-core
manuel-alvarez-alvarez, line 1: Be careful with this or you might break the build :cry:

dd-java-agent/agent-iast/src/main/java/com/datadog/iast/sink/HttpResponseHeaderModuleImpl.java
DDJavierSantos: But we don't need to validate the header, only detect if it is there, so after some conversation with Santi we are keeping this implementation.

dd-java-agent/agent-iast/src/main/java/com/datadog/iast/RequestEndedHandler.java
DDJavierSantos: Done

dd-java-agent/agent-iast/src/main/java/com/datadog/iast/IastRequestContext.java
```diff
@@ -32,6 +35,9 @@ public IastRequestContext(
       final TaintedObjects taintedObjects, final IastMetricCollector collector) {
     this.vulnerabilityBatch = new VulnerabilityBatch();
     this.spanDataIsSet = new AtomicBoolean(false);
+    this.hstsHeaderIsSet = false;
+    this.xForwardedProtoIsHtttps = false;
+    this.contentTypeIsHtml = false;
```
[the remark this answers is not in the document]
DDJavierSantos: I have changed it so we store the content type string now

dd-java-agent/agent-jmxfetch/integrations-core
smola, line 1: Please, see https://github.com/DataDog/dd-trace-java/pull/5522#issue-1787790804

You can prevent this in the future setting this:
```
git config --local submodule.recurse true
```

To fix it in this PR: `git checkout -f origin/master -- dd-java-agent/agent-jmxfetch/integrations-core`

dd-smoke-tests/spring-boot-2.6-webmvc/src/main/java/datadog/smoketest/springboot/controller/IastWebController.java
```diff
@@ -264,6 +264,13 @@ public String xpathInjectionEvaluate(final HttpServletRequest request)
     return "XPath Injection page";
   }
 
+  @GetMapping(value = "/hstsmissing", produces = "text/html")
+  public String hstsHeaderMissing(HttpServletResponse response) {
+    response.addHeader("X-Forwarded-Proto", "https");
```
smola, line 269: I'd swear `X-Forwarded-Proto` needs to be set by the caller in the request, not by the app in the response.

smola (changes requested):

dd-java-agent/agent-iast/src/main/java/com/datadog/iast/sink/HttpResponseHeaderModuleImpl.java
```diff
@@ -82,4 +123,55 @@ public void onCookie(
...
+    if (value.toLowerCase().contains("max-age=0")) {
+      return true;
+    }
+    if (value.toLowerCase().contains("max-age=-1")) {
+      return true;
+    }
+    return false;
+  }
+
+  public static void checkForHeaderVulnerabilities(
```
manuel-alvarez-alvarez, line 140: You don't need a static method with all the dependencies, you can create an instance one like:

```java
@Override
public void onRequestEnded(final IastRequestContext iastRequestContext, final AgentSpan span) {
   ...
}
```

And call it in the `RequestEndedHandler` via the bridge as usual:

```java
try {
  HttpResponseHeaderModule module = InstrumentationBridge.RESPONSE_HEADER_MODULE;
  if (module != null) {
    module.onRequestEnded(iastRequestContext, (AgentSpan) igSpanInfo);
  }
} catch (Throwable e) {
   LOGGER.error("Failed to handle request ended in headers module", e);
}
```

dd-smoke-tests/spring-boot-2.6-webmvc/src/main/java/datadog/smoketest/springboot/controller/IastWebController.java
DDJavierSantos: Design changed, please check

dd-java-agent/agent-iast/src/main/java/com/datadog/iast/sink/HttpResponseHeaderModuleImpl.java
DDJavierSantos: Design changed, please check

dd-java-agent/agent-iast/src/main/java/com/datadog/iast/RequestEndedHandler.java
```diff
@@ -9,23 +9,32 @@
...
 
   @Override
   public Flow<Void> apply(final RequestContext requestContext, final IGSpanInfo igSpanInfo) {
     final TraceSegment traceSegment = requestContext.getTraceSegment();
     final IastRequestContext iastRequestContext = IastRequestContext.get(requestContext);
     if (iastRequestContext != null) {
+      HttpResponseHeaderModule mod = InstrumentationBridge.RESPONSE_HEADER_MODULE;
+      if (mod != null) {
+        mod.checkForHeaderVulnerabilities(
+            iastRequestContext, igSpanInfo, overheadController, reporter);
```
manuel-alvarez-alvarez, line 36: The `HttpResponseHeaderModule` already has the overhead controller and reported dependencies injected, no need to pass them around.

dd-java-agent/instrumentation/servlet/request-2/src/main/java/datadog/trace/instrumentation/servlet2/Servlet2Advice.java
```diff
@@ -61,6 +63,12 @@ public static boolean onEnter(
     httpServletRequest.setAttribute(
         CorrelationIdentifier.getSpanIdKey(), GlobalTracer.get().getSpanId());
 
+    HttpResponseHeaderModule mod = InstrumentationBridge.RESPONSE_HEADER_MODULE;
```
manuel-alvarez-alvarez, line 66: This approach introduces IAST related logic into the tracing advice that doesn't feel right, maybe an API where the request start callbacks can specify headers that they are interested in?

internal-api/src/main/java/datadog/trace/api/iast/sink/HttpResponseHeaderModule.java
```diff
@@ -9,4 +10,9 @@ public interface HttpResponseHeaderModule extends IastModule {
   void onHeader(@Nonnull String name, String value);
 
   void onCookie(@Nonnull Cookie cookie);
+
+  void onXForwardedProtoHeaderValue(String value);
```
manuel-alvarez-alvarez, line 14: Can we try to find a different place for this?, the original idea of the module is to deal with response headers.

dd-java-agent/instrumentation/servlet/request-2/src/main/java/datadog/trace/instrumentation/servlet2/Servlet2Advice.java
DDJavierSantos: design changed, please check

internal-api/src/main/java/datadog/trace/api/iast/sink/HttpResponseHeaderModule.java
DDJavierSantos: design changed, please check

dd-java-agent/agent-bootstrap/src/main/java/datadog/trace/bootstrap/instrumentation/decorator/HttpServerDecorator.java
```diff
@@ -553,6 +566,13 @@ private IGKeyClassifier(
 
     @Override
     public boolean accept(String key, String value) {
+      System.out.println(
```
manuel-alvarez-alvarez, line 569: This should be removed

dd-java-agent/agent-iast/src/main/java/com/datadog/iast/RequestEndedHandler.java
```diff
@@ -9,23 +9,31 @@
...
+import datadog.trace.api.iast.InstrumentationBridge;
+import datadog.trace.api.iast.sink.HttpResponseHeaderModule;
 import datadog.trace.api.internal.TraceSegment;
 import java.util.function.BiFunction;
 import javax.annotation.Nonnull;
 
 public class RequestEndedHandler implements BiFunction<RequestContext, IGSpanInfo, Flow<Void>> {
 
   private final OverheadController overheadController;
+  private final Reporter reporter;
```
manuel-alvarez-alvarez, line 21: I think this is not needed

dd-java-agent/agent-iast/src/main/java/com/datadog/iast/RequestHeaderHandler.java
```diff
@@ -0,0 +1,19 @@
+package com.datadog.iast;
+
+import datadog.trace.api.function.TriConsumer;
+import datadog.trace.api.gateway.RequestContext;
+import datadog.trace.api.gateway.RequestContextSlot;
+
+public class RequestHeaderHandler implements TriConsumer<RequestContext, String, String> {
+  private static final String X_FORWARDED_PROTO_HEADER = "X-Forwarded-Proto";
```
manuel-alvarez-alvarez, line 8: Maybe not for this PR, but since we already have a lot of places where we have header names, maybe a headers class that groups all the logic makes sense

dd-java-agent/appsec/src/main/java/com/datadog/appsec/gateway/GatewayBridge.java
```diff
@@ -421,6 +421,8 @@ private static class NewRequestHeaderCallback
       implements TriConsumer<RequestContext, String, String> {
     @Override
     public void accept(RequestContext ctx_, String name, String value) {
+      new Throwable().printStackTrace();
```
manuel-alvarez-alvarez, line 424: This should be gone

dd-java-agent/agent-iast/src/main/java/com/datadog/iast/sink/HttpResponseHeaderModuleImpl.java
```diff
@@ -88,4 +116,55 @@ private static List<HttpCookieModule<VulnerabilityType>> httpCookieModules() {
...
+    if (value.toLowerCase().contains("max-age=-1")) {
+      return true;
+    }
+    return false;
+  }
+
+  public void checkForHeaderVulnerabilities(
+      final Object iastRequestContextObject, final IGSpanInfo igSpanInfo) {
+
+    final IastRequestContext iastRequestContext = (IastRequestContext) iastRequestContextObject;
```
manuel-alvarez-alvarez, line 136: Doesn't this make more sense to have a `HstsHeaderModule`, we forced small modules for insecure cookies et all to be able to switch them on/off so it looks weird to me to embed this one inside the response headers module.

dd-java-agent/agent-bootstrap/src/main/java/datadog/trace/bootstrap/instrumentation/decorator/HttpServerDecorator.java
DDJavierSantos: removed

dd-java-agent/agent-iast/src/main/java/com/datadog/iast/RequestEndedHandler.java
DDJavierSantos: removed

dd-java-agent/appsec/src/main/java/com/datadog/appsec/gateway/GatewayBridge.java
DDJavierSantos: removed

dd-java-agent/agent-iast/src/main/java/com/datadog/iast/sink/HttpResponseHeaderModuleImpl.java
DDJavierSantos: I have added a module named HstsMissingHeaderModule

dd-java-agent/agent-iast/build.gradle
```diff
@@ -52,8 +52,14 @@ ext {
     'com.datadog.iast.model.json.SourceTypeAdapter',
     'com.datadog.iast.model.json.VulnerabilityTypeAdapter',
   ]
-  excludedClassesBranchCoverage = []
-  excludedClassesInstructionCoverage = []
+  excludedClassesBranchCoverage = [
```
manuel-alvarez-alvarez, line 55: I think we should not ignore coverage for those classes, can you try to test all branches?

dd-java-agent/agent-iast/src/main/java/com/datadog/iast/sink/HstsMissingHeaderModuleImpl.java
```diff
@@ -0,0 +1,57 @@
...
+  public void onRequestEnd(final Object iastRequestContextObject, final IGSpanInfo igSpanInfo) {
+
+    final IastRequestContext iastRequestContext = (IastRequestContext) iastRequestContextObject;
+
+    if (!iastRequestContext.getHstsHeaderIsSet()) {
+      try {
+        Map<String, Object> tags = igSpanInfo.getTags();
+        String urlString = (String) tags.get("http.url");
+        Integer httpStatus = (Integer) tags.get("http.status_code");
+        if (httpStatus == HttpURLConnection.HTTP_MOVED_PERM
```
manuel-alvarez-alvarez, line 31: This will throw a NPE if there is no status due to unboxing, if the status code is null we should consider as a reportable status code

dd-java-agent/agent-iast/src/main/java/com/datadog/iast/sink/HttpResponseHeaderModuleImpl.java
```diff
@@ -88,4 +113,17 @@ private static List<HttpCookieModule<VulnerabilityType>> httpCookieModules() {
...
   }
+
+  static boolean isMaxAgeZero(String value) {
+    if (null == value || 0 == value.length()) {
+      return false;
+    }
+    if (value.toLowerCase().contains("max-age=0")) {
+      return true;
+    }
+    if (value.toLowerCase().contains("max-age=-1")) {
```
manuel-alvarez-alvarez, line 124: Why only `0` and `-1`? Maybe we can cover all cases with a regexp (like `max-age`, `max-age=`, `max-age=-2`):

```java
  private static final Pattern MAX_AGE = Pattern.compile("max-age=(\\d+)", Pattern.CASE_INSENSITIVE);

  private static boolean isValidMaxAge(@Nullable final String value) {
    if (value == null) {
     return false;
    }
    final Matcher matcher = MAX_AGE.matcher(value);
    if (!matcher.find()) {
      return false;
    }
    return Integer.parseInt(matcher.group(1)) > 0;
  }

```

And probably it will be a better approach to move this code to the HSTS module rather than here.

manuel-alvarez-alvarez (approved):

dd-java-agent/agent-iast/src/main/java/com/datadog/iast/sink/HstsMissingHeaderModuleImpl.java
DDJavierSantos: fixed

dd-java-agent/agent-iast/build.gradle
DDJavierSantos: fixed

dd-java-agent/agent-iast/src/main/java/com/datadog/iast/sink/HttpResponseHeaderModuleImpl.java
DDJavierSantos: I'd rather have a simple comparison rather than a regex for performance reasons. I have moved the function to the Hsts module.

dd-java-agent/agent-iast/src/main/java/com/datadog/iast/sink/HstsMissingHeaderModuleImpl.java
```diff
@@ -0,0 +1,85 @@
...
+        ctx.setHstsHeaderIsSet();
+      }
+    }
+  }
+
+  public static boolean isMaxAgeZero(String value) {
+    if (null == value || 0 == value.length()) {
+      return false;
+    }
+    if (value.toLowerCase().contains("max-age=0")) {
```
manuel-alvarez-alvarez, line 77: Fair enough if you don't want to use regexps but don't duplicate the `toLowerCase`, we are still missing cases like `max-age=012345` (not sure if we will ever find it though)

manuel-alvarez-alvarez (approved):

dd-java-agent/agent-iast/src/main/java/com/datadog/iast/sink/HstsMissingHeaderModuleImpl.java
```diff
@@ -0,0 +1,85 @@
...
+            && (url.getProtocol().equals("https")
+                || iastRequestContext.getXForwardedProtoIsHtttps())) {
+          final AgentSpan span = AgentTracer.activeSpan();
+          if (overheadController.consumeQuota(Operations.REPORT_VULNERABILITY, span)) {
+            reporter.report(
+                span, new Vulnerability(VulnerabilityType.HSTS_HEADER_MISSING, null, null));
+          }
+        }
+      } catch (Throwable e) {
+        LOGGER.error("Exception while checking for hsts vulnerability", e);
```
smola, line 55:
```suggestion
        LOGGER.debug("Exception while checking for missing HSTS headers vulnerability", e);
```

Changed to debug. If this turns out to happen too often, we don't want to spam the logs.

smola (changes requested):
One small change, then good to go.

smola (approved):
Applied the change myself.

dd-java-agent/agent-iast/src/main/java/com/datadog/iast/sink/HstsMissingHeaderModuleImpl.java
DDJavierSantos: Well, max-age=012345 is still not zero so the function will work as intended. About the duplication of toLowerCase, it is fixed now.

dd-java-agent/agent-iast/src/main/java/com/datadog/iast/sink/HstsMissingHeaderModuleImpl.java
DDJavierSantos: applied

PR #5525, opened 2023-07-04
Implement ITR skippable tests request for Gradle

nikita-tkachenko-datadog:
# What Does This Do
Implements Intelligent Test Runner in Gradle builds.
Parent process (Gradle) requests remote settings (code coverage enabled, intelligent test runner enabled), and the list of skippable tests from the backend. Then the settings and the skippable tests are propagated to the children processes (JVMs forked to run tests).

# Motivation
Intelligent Test Runner is a feature of CI Visibility product.

# Additional Notes
Request for settings and skippable tests must include JVM executable details.
It is possible to use multiple JVMs during the build (e.g. some tests are run with Java 8, while others are run with Java 11).
Calculating the list of skippable tests is an expensive operation, therefore in order to make as few requests as possible, first all the Gradle test tasks are grouped by the JVM that they use, then a single settings request and a single skippable tests request is issued for each group.

### DIFF
--- a/dd-java-agent/instrumentation/gradle/src/main/groovy/datadog/trace/instrumentation/gradle/GradleProjectConfigurator.groovy
+++ b/dd-java-agent/instrumentation/gradle/src/main/groovy/datadog/trace/instrumentation/gradle/GradleProjectConfigurator.groovy
@@ -1,10 +1,19 @@
 package datadog.trace.instrumentation.gradle
 
 import datadog.trace.api.Config
+import datadog.trace.api.civisibility.config.SkippableTest
+import datadog.trace.api.civisibility.config.SkippableTestsSerializer
+import datadog.trace.api.config.CiVisibilityConfig
 import datadog.trace.bootstrap.DatadogClassLoader
+import datadog.trace.util.Strings
 import org.gradle.api.Project
+import org.gradle.api.Task
+import org.gradle.internal.jvm.Jvm
+import org.slf4j.Logger
+import org.slf4j.LoggerFactory
 
 import java.nio.file.Files
+import java.nio.file.Path
 import java.nio.file.Paths
 import java.util.regex.Pattern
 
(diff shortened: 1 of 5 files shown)


dd-java-agent/instrumentation/gradle/src/main/groovy/datadog/trace/instrumentation/gradle/GradleProjectConfigurator.groovy
```diff
@@ -29,36 +38,38 @@ import java.util.regex.Pattern
...
-      }
+    // propagate to child process all "dd." system properties available in current process
+    for (def e : propagatedSystemProperties.entrySet()) {
+      jvmArgs.add("-D" + e.key + '=' + e.value)
+    }
 
-      jvmArgs.add("-javaagent:" + Config.get().ciVisibilityAgentJarFile.toPath())
+    if (skippableTests != null && !skippableTests.isEmpty()) {
+      String skippableTestsString = SkippableTestsSerializer.serialize(skippableTests)
+      jvmArgs.add("-D" + Strings.propertyNameToSystemPropertyName(CiVisibilityConfig.CIVISIBILITY_SKIPPABLE_TESTS) + '=' + skippableTestsString)
```
tonyredondo, line 60: Basically the same question I did for the Maven version, is there a size limit to the `jvmArgs`, I'm worried about the length of that `skippableTestsString`.

tonyredondo (approved):
LGTM

dd-java-agent/instrumentation/gradle/src/main/groovy/datadog/trace/instrumentation/gradle/GradleProjectConfigurator.groovy
nikita-tkachenko-datadog: This is a good point. There is a limit on the max length of command line which, if I'm not mistaken, is ~1-2 mb in most Linux distros.
This string contains only the tests for a specific module and not for the entire project, besides I am compressing it with gzip.
Nevertheless, theoretically we could still hit this limit.
I have an alternative mechanism to pass data from parent process to children (via TCP), but it is not merged yet.
Once that is available, I will rewrite this piece to use it instead of command line arguments

PR #5526, opened 2023-07-04
Do tags and baggage processing in ListWriter

bantonsson:
# What Does This Do

Calls `span.processTagsAndBaggage` in the `ListWriter` used in tests.

# Motivation

This forces some lazy evaluation processing to happen, so that we more accurately test what will actually be sent to the agent in the end.

# Additional Notes


dd-java-agent/agent-jmxfetch/integrations-core
smola, line 1: Unintended change?

dd-trace-core/src/main/java/datadog/trace/core/MetadataConsumer.java
```diff
@@ -1,5 +1,8 @@
 package datadog.trace.core;
 
-public abstract class MetadataConsumer {
-  public abstract void accept(Metadata metadata);
+public interface MetadataConsumer {
+
+  MetadataConsumer NO_OP = (md) -> {};
+
+  void accept(Metadata metadata);
```
PerfectSlayer, line 7: What about extending `Consumer<Metadata>`? Could allow `andThen()` operation.
`@FunctionalInterface` could be nice too.

dd-java-agent/instrumentation/jsp-2.3/src/test/groovy/JSPInstrumentationBasicTests.groovy
```diff
@@ -110,7 +110,7 @@ class JSPInstrumentationBasicTests extends JSPTestBase {
             "$Tags.SPAN_KIND" Tags.SPAN_KIND_SERVER
             "$Tags.PEER_HOST_IPV4" "127.0.0.1"
             "$Tags.PEER_PORT" Integer
-            "$Tags.HTTP_URL" "http://localhost:$port/$jspWebappContext/getQuery.jsp"
+            "$Tags.HTTP_URL" "http://localhost:$port/$jspWebappContext/getQuery.jsp?HELLO"
```
PerfectSlayer, line 113: That is neatpicking but can we use a more _self explanatory_ value for when tests will break? Like `QueryParameter` rather than `HELLO`?
It should give more meaningful expected message 🙏

PerfectSlayer (approved):

dd-java-agent/agent-jmxfetch/integrations-core
PerfectSlayer: Might be related to the upgrade from Stuart?

dd-java-agent/instrumentation/jsp-2.3/src/test/groovy/JSPInstrumentationBasicTests.groovy
bantonsson: I agree with you, but I'm not intending to change the tests right now. Only to make them work.

dd-trace-core/src/main/java/datadog/trace/core/MetadataConsumer.java
bantonsson: That is a good idea.

PR #5528, opened 2023-07-04
Update IAST redaction algorithm

manuel-alvarez-alvarez:
# What Does This Do
Updates the redaction algorithm to be more lenient

# Motivation

# Additional Notes

### DIFF
--- a/dd-java-agent/agent-iast/src/main/java/com/datadog/iast/model/json/EvidenceAdapter.java
+++ b/dd-java-agent/agent-iast/src/main/java/com/datadog/iast/model/json/EvidenceAdapter.java
@@ -6,6 +6,7 @@ import com.datadog.iast.model.Source;
 import com.datadog.iast.model.Vulnerability;
 import com.datadog.iast.model.VulnerabilityType;
 import com.datadog.iast.model.json.AdapterFactory.Context;
+import com.datadog.iast.model.json.AdapterFactory.RedactionContext;
 import com.datadog.iast.sensitive.SensitiveHandler;
 import com.datadog.iast.sensitive.SensitiveHandler.Tokenizer;
 import com.datadog.iast.util.Ranged;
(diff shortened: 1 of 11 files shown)


dd-java-agent/agent-iast/src/test/groovy/com/datadog/iast/model/json/VulnerabilityEncodingTest.groovy
```diff
@@ -19,6 +19,22 @@ class VulnerabilityEncodingTest extends DDSpecification {
...
+    final value = new VulnerabilityBatch()
+    value.add(null)
+
+    when:
+    final result = VulnerabilityEncoding.toJson(value)
+
+    then:
+    JSONAssert.assertEquals('''{
+      "vulnerabilities": [
+        null
```
smola, line 33: Do we really want to do this? We should probably skip the element altogether?

manuel-alvarez-alvarez: Yep it's quite weird. It was like this before the PR but jacoco forced me to add the tests due to coverage, while doing the tests I thought about changing it, so I'll skip the vulnerability if it's null.

dd-java-agent/agent-iast/src/main/java/com/datadog/iast/model/json/EvidenceAdapter.java
```diff
@@ -15,6 +16,8 @@
 import com.squareup.moshi.Moshi;
 import datadog.trace.api.Config;
 import java.io.IOException;
+import java.util.*;
```
smola, line 19: We generally avoid wildcard imports: https://github.com/DataDog/dd-trace-java/blob/master/CONTRIBUTING.md#intellij-idea

smola (approved):

dd-java-agent/agent-iast/src/main/java/com/datadog/iast/model/json/EvidenceAdapter.java
manuel-alvarez-alvarez: Yeah, I copied from another project that did not have the settings and IntelliJ did not change it :cry: 

Updated, thanks!

PR #5531, opened 2023-07-05
Clearer API to taint non-Strings

smola:
# What Does This Do

Use `taintObject` (single `Object`) and `taintObjects` (array or collection of `Object`) to taint non-Strings.

# Motivation

* Too many overloads for `taint`, sometimes leading to calls to the incorrect versions.
* It is easy to call a taint method that is meant for non-Strings (and does not handle range sizes for strings appropriately).
* Avoid some unnecessary work at runtime when these are called with single objects (most calls).

# Additional Notes

### DIFF
--- a/dd-java-agent/agent-iast/src/test/groovy/com/datadog/iast/propagation/PropagationModuleTest.groovy
+++ b/dd-java-agent/agent-iast/src/test/groovy/com/datadog/iast/propagation/PropagationModuleTest.groovy
@@ -75,7 +75,7 @@ class PropagationModuleTest extends IastModuleImplTestBase {
     'taint'                    | [SourceTypes.REQUEST_PARAMETER_VALUE, 'name', '']
     'taintObjects'             | [SourceTypes.REQUEST_PARAMETER_VALUE, null as Object[]]
     'taintObjects'             | [SourceTypes.REQUEST_PARAMETER_VALUE, [] as Object[]]
-    'taint'                    | [SourceTypes.REQUEST_PARAMETER_VALUE, null as Collection]
+    'taintObjects'             | [SourceTypes.REQUEST_PARAMETER_VALUE, null as Collection]
   }
 
   void '#method without span'() {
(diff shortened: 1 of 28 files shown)


dd-java-agent/instrumentation/jose-jwt/src/main/java/datadog/trace/instrumentation/josejwt/JSONObjectUtilsInstrumentation.java
```diff
@@ -43,7 +43,9 @@ public static void onEnter(@Advice.Return Map<String, Object> map) {
       if (module != null) {
         for (Map.Entry entry : map.entrySet()) {
           if (entry.getValue() instanceof String) {
-            module.taint(SourceTypes.REQUEST_HEADER_VALUE, (String) entry.getValue());
```
manuel-alvarez-alvarez, line 46: Not for this PR but both instrumentations (`JSONObjectUtilsInstrumentation` and `JWTParserInstrumentation`) are weird, they should check if the string passed to the method is tainted and infer the source type accordingly, doing so it might help as well to fetch the actual header name instead of using `null` as the name.

manuel-alvarez-alvarez (approved):

dd-java-agent/agent-iast/src/test/groovy/com/datadog/iast/propagation/PropagationModuleTest.groovy
```diff
@@ -123,30 +124,126 @@ class PropagationModuleTest extends IastModuleImplTestBase {
...
-    'taintIfAnyInputIsTainted' | [new Object(), ['I am an string'].toArray()]                                                    | { it[0] }          | { it[1][0] }
-    'taintIfAnyInputIsTainted' | [new Object(), [new Object()].toArray()]                                                        | { it[0] }          | { it[1][0] }
-    'taintIfAnyInputIsTainted' | [new Object(), [new MockTaintable()].toArray()]                                                 | { it[0] }          | { it[1][0] }
-    'taintIfAnyInputIsTainted' | ['Hello', ['I am an string'].toArray()]                                                         | { it[0] }          | { it[1][0] }
-    'taintIfAnyInputIsTainted' | ['Hello', [new Object()].toArray()]                                                             | { it[0] }          | { it[1][0] }
-    'taintIfAnyInputIsTainted' | ['Hello', [new MockTaintable()].toArray()]                                                      | { it[0] }          | { it[1][0] }
+    'taintIfInputIsTainted'    | [new Object(), 'I am an string']                                                                | {
```
jandro996, line 127: issues with your IDE formatter?

smola: I guess so. I was trying back and forth with IntelliJ + spotless, and it couldn't keep the format straight...

PR #5532, opened 2023-07-05
disable queue time instrumentations unless feature enabled

richardstartin:
# What Does This Do

This minimises the footprint from queue time instrumentation unless the feature is enabled

# Motivation

# Additional Notes

### DIFF
--- a/dd-java-agent/instrumentation/java-concurrent/src/main/java/datadog/trace/instrumentation/java/concurrent/TaskUnwrappingInstrumentation.java
+++ b/dd-java-agent/instrumentation/java-concurrent/src/main/java/datadog/trace/instrumentation/java/concurrent/TaskUnwrappingInstrumentation.java
@@ -3,12 +3,22 @@ package datadog.trace.instrumentation.java.concurrent;
 import com.google.auto.service.AutoService;
 import datadog.trace.agent.tooling.Instrumenter;
 import datadog.trace.agent.tooling.bytebuddy.profiling.UnwrappingVisitor;
+import datadog.trace.api.config.ProfilingConfig;
+import datadog.trace.bootstrap.config.provider.ConfigProvider;
 
 @AutoService(Instrumenter.class)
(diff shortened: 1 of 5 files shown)


dd-java-agent/instrumentation/java-concurrent/src/main/java/datadog/trace/instrumentation/java/concurrent/TaskUnwrappingInstrumentation.java
```diff
@@ -11,6 +13,14 @@ public TaskUnwrappingInstrumentation() {
     super("java_concurrent", "task-unwrapping");
   }
 
+  @Override
+  protected boolean defaultEnabled() {
+    return ConfigProvider.getInstance()
+        .getBoolean(
+            ProfilingConfig.PROFILING_QUEUEING_TIME_ENABLED,
+            ProfilingConfig.PROFILING_QUEUEING_TIME_ENABLED_DEFAULT);
+  }
```
PerfectSlayer, line 22: I might have missed something here but why not directly set it to `false` and expect the user to set `dd.integrations.task-unwrapping.enabled=true` instead of introducing a new environment variable?

richardstartin: 1. This is not a new environment variable being introduced. 
2. This is a feature aimed at profiling users (which is why it's a profiling instrumentation) who generally set `dd.profiling.*` flags to toggle features, not `dd.integration.*` flags.
3. This is consistent with how other profiling features implemented as instrumentations are toggled

PerfectSlayer: Sounds good to me then!

PerfectSlayer (approved):

PR #5542, opened 2023-07-06
[Spark] Compute distribution of task metrics for each stage

paul-laffon-dd:
# What Does This Do

- Expose a new tracer interface `AgentHistogram` backed by a DDSketch, that can be used inside an instrumentation
- Use a `LogarithmicMapping` since it is the mapping supported by the most languages, with a relative error of 3%
- Serialize the DDSketch and encode it in base64 to add it as an hidden span tag

Using a total of 6 histogram to capture the most important spark metrics. For most of the stages, only 3 histograms will contain non-zero values (stages generally only have either input/output or shuffle read/write)
- Duration
- Input bytes
- Output bytes
- Shuffle Read bytes
- Shuffle Write bytes
- Spill

Added the histogram behind the feature flag `-Ddd.spark.task-histogram.enabled` (or `DD_SPARK_TASK_HISTOGRAM_ENABLED`). It is enabled by default, but can be disabled if histograms are generating a big amount of data

# Motivation

Expose imbalance in the tasks duration (also referred as skew) for each spark stages

### DIFF
--- a/dd-java-agent/instrumentation/spark/src/main/java/datadog/trace/instrumentation/spark/SparkAggregatedTaskMetrics.java
+++ b/dd-java-agent/instrumentation/spark/src/main/java/datadog/trace/instrumentation/spark/SparkAggregatedTaskMetrics.java
@@ -223,10 +223,7 @@ class SparkAggregatedTaskMetrics {
     span.setMetric(prefix + ".skew_time", skewTime);
 
     if (taskRunTimeHistogram.getCount() > 0) {
-      String base64hist = histogramToBase64(taskRunTimeHistogram);
-      span.setTag("_dd.spark.task_run_time", base64hist);
-      span.setMetric(prefix + ".histogram_serialized_length", base64hist.length());
-      span.setMetric(prefix + ".max_task_run_time", taskRunTimeHistogram.getMaxValue());
+      span.setTag("_dd.spark.task_run_time", histogramToBase64(taskRunTimeHistogram));
     }
     if (inputBytesHistogram.getCount() > 0) {
       span.setTag("_dd.spark.input_bytes", histogramToBase64(inputBytesHistogram));
--- a/dd-java-agent/instrumentation/spark/src/main/java/datadog/trace/instrumentation/spark/SparkAggregatedTaskMetrics.java
+++ b/dd-java-agent/instrumentation/spark/src/main/java/datadog/trace/instrumentation/spark/SparkAggregatedTaskMetrics.java
@@ -10,6 +10,9 @@ import org.apache.spark.executor.TaskMetrics;
 import org.apache.spark.scheduler.SparkListenerTaskEnd;
 
 class SparkAggregatedTaskMetrics {
+  private static final double HISTOGRAM_RELATIVE_ACCURACY = 1 / 32.0;
+  private static final int HISTOGRAM_MAX_NUM_BINS = 512;
+
   private long executorDeserializeTime = 0L;
   private long executorDeserializeCpuTime = 0L;
   private long executorRunTime = 0L;
@@ -48,15 +51,12 @@ class SparkAggregatedTaskMetrics {
   private long totalTaskRunTimeSinceLastStage = 0L;
   private long skewTime = 0;
 
-  private final AgentHistogram taskRunTimeHistogram = AgentTracer.get().newHistogram(1 / 32.0, 512);
-  private final AgentHistogram inputBytesHistogram = AgentTracer.get().newHistogram(1 / 32.0, 512);
-  private final AgentHistogram outputBytesHistogram = AgentTracer.get().newHistogram(1 / 32.0, 512);
-  private final AgentHistogram shuffleReadBytesHistogram =
-      AgentTracer.get().newHistogram(1 / 32.0, 512);
-  private final AgentHistogram shuffleWriteBytesHistogram =
-      AgentTracer.get().newHistogram(1 / 32.0, 512);
-  private final AgentHistogram diskBytesSpilledHistogram =
-      AgentTracer.get().newHistogram(1 / 32.0, 512);
+  private AgentHistogram taskRunTimeHistogram;
+  private AgentHistogram inputBytesHistogram;
+  private AgentHistogram outputBytesHistogram;
+  private AgentHistogram shuffleReadBytesHistogram;
+  private AgentHistogram shuffleWriteBytesHistogram;
+  private AgentHistogram diskBytesSpilledHistogram;
 
   public SparkAggregatedTaskMetrics() {}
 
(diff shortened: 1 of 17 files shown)


dd-java-agent/instrumentation/spark/src/main/java/datadog/trace/instrumentation/spark/SparkAggregatedTaskMetrics.java
```diff
@@ -187,11 +220,42 @@ public void setSpanMetrics(AgentSpan span, String prefix) {
     span.setMetric(prefix + ".task_with_output_count", taskWithOutputCount);
 
     span.setMetric(prefix + ".available_executor_time", attributedAvailableExecutorTime);
+    span.setMetric(prefix + ".skew_time", skewTime);
+
+    if (taskRunTimeHistogram.getCount() > 0) {
+      String base64hist = histogramToBase64(taskRunTimeHistogram);
+      span.setTag("_dd.spark.task_run_time", base64hist);
+      span.setMetric(prefix + ".histogram_serialized_length", base64hist.length());
```
bantonsson, line 228: Is this only here for testing purposes?

dd-java-agent/instrumentation/spark/src/main/java/datadog/trace/instrumentation/spark/SparkAggregatedTaskMetrics.java
```diff
@@ -42,6 +46,17 @@ class SparkAggregatedTaskMetrics {
...
+
+  private final AgentHistogram taskRunTimeHistogram = AgentTracer.get().newHistogram(1 / 32.0, 512);
+  private final AgentHistogram inputBytesHistogram = AgentTracer.get().newHistogram(1 / 32.0, 512);
+  private final AgentHistogram outputBytesHistogram = AgentTracer.get().newHistogram(1 / 32.0, 512);
+  private final AgentHistogram shuffleReadBytesHistogram =
+      AgentTracer.get().newHistogram(1 / 32.0, 512);
+  private final AgentHistogram shuffleWriteBytesHistogram =
+      AgentTracer.get().newHistogram(1 / 32.0, 512);
+  private final AgentHistogram diskBytesSpilledHistogram =
+      AgentTracer.get().newHistogram(1 / 32.0, 512);
```
bantonsson, line 59: Since you mention that only about 3 of the histograms will contain non-zero values, would it be possible to make the creation of them _lazy_ so they are only instantiated when needed?

bantonsson (commented):
This looks fine from a code perspective, but I'm worried about the tag size/traffic. I would prefer if there was a feature flag for the library and this was opt-in or could be easily turned off.

dd-java-agent/instrumentation/spark/src/main/java/datadog/trace/instrumentation/spark/SparkAggregatedTaskMetrics.java
paul-laffon-dd: Yes, this was only for testing purpose and get figures about the size of the histograms, removing those

dd-java-agent/instrumentation/spark/src/main/java/datadog/trace/instrumentation/spark/SparkAggregatedTaskMetrics.java
paul-laffon-dd: Yes, this should something possible to do, I will look into that

paul-laffon-dd (commented):
> I would prefer if there was a feature flag for the library and this was opt-in or could be easily turned off.

The spark integration is [disabled by default](https://github.com/DataDog/dd-trace-java/blob/master/dd-java-agent/instrumentation/spark/src/main/java/datadog/trace/instrumentation/spark/SparkInstrumentation.java#L21C1-L21C1) and has to be enabled using `dd.integration.spark.enabled=true` would that be enough or do you mean having a specific feature flag for reporting the histograms ?

dd-java-agent/instrumentation/spark/src/main/java/datadog/trace/instrumentation/spark/SparkAggregatedTaskMetrics.java
paul-laffon-dd: Implemented the lazy instantiation in this commit https://github.com/DataDog/dd-trace-java/pull/5542/commits/9a84e2bf5fa5cf1ae5d1faff8d1a8630fac53fc9

dd-java-agent/instrumentation/spark/src/main/java/datadog/trace/instrumentation/spark/SparkAggregatedTaskMetrics.java
```diff
@@ -187,11 +232,61 @@ public void setSpanMetrics(AgentSpan span, String prefix) {
...
   }
 
   public static long computeTaskRunTime(TaskMetrics metrics) {
     return metrics.executorDeserializeTime()
         + metrics.executorRunTime()
         + metrics.resultSerializationTime();
   }
+
+  private static String histogramToBase64(AgentHistogram hist) {
+    ByteBuffer bb = hist.serialize();
```
dougqh, line 286: I'm assuming this code isn't terribly hot, so this is probably fine.
But if this is hot, this is pretty inefficient.

We're creating a ByteBuffer (presumably backed by a byte[]), then copying to a byte[], and then encoding to a String.  In theory, this could be done in a streaming fashion with much less copying.

paul-laffon-dd: This code is not very hot, but I agree that it could be more efficient. The copy from ByteBuffer to a byte[] can be removed if the underlying byte[] of the buffer is accessible https://github.com/DataDog/dd-trace-java/pull/5542/commits/e8926208dc04c3acdc4b89eef4e51172b56a7cdb

paul-laffon-dd:
Added the histogram behind the feature flag `-Ddd.spark.task-histogram.enabled` (or `DD_SPARK_TASK_HISTOGRAM_ENABLED`). It is enabled by default, but can be disabled if histograms are generating a big amount of data

dd-java-agent/instrumentation/spark/src/main/java/datadog/trace/instrumentation/spark/SparkAggregatedTaskMetrics.java
```diff
@@ -94,7 +110,24 @@ public void addTaskMetrics(SparkListenerTaskEnd taskEnd) {
         taskWithOutputCount += 1;
       }
 
-      taskRunTimeSinceLastStage += computeTaskRunTime(taskMetrics);
+      long taskRunTime = computeTaskRunTime(taskMetrics);
+      taskRunTimeSinceLastStage += taskRunTime;
+
+      if (Config.get().isSparkTaskHistogramEnabled()) {
```
bantonsson, line 116: You could to this check at creation time of the `SparkAggregatedTaskMetrics`

dd-java-agent/instrumentation/spark/src/main/java/datadog/trace/instrumentation/spark/SparkAggregatedTaskMetrics.java
```diff
@@ -187,11 +230,67 @@ public void setSpanMetrics(AgentSpan span, String prefix) {
...
+      span.setTag("_dd.spark.disk_bytes_spilled", histogramToBase64(diskBytesSpilledHistogram));
+    }
+  }
+
+  /**
+   * Lazy creation of histograms, only creating one if there is a non-zero value. Spark stages
+   * usually involve either input/output or shuffle read/write operations, resulting in an average
+   * of 3 histograms having non-zero values
+   */
+  private AgentHistogram lazyHistogramAccept(AgentHistogram hist, double value) {
```
bantonsson, line 260: If this is not touching any instance variables, it could be `static`.

dd-java-agent/instrumentation/spark/src/main/java/datadog/trace/instrumentation/spark/SparkAggregatedTaskMetrics.java
```diff
@@ -187,11 +230,67 @@ public void setSpanMetrics(AgentSpan span, String prefix) {
...
+    if (hist != null) {
+      hist.accept(value);
+    } else {
+      if (value != 0) {
+        hist = AgentTracer.get().newHistogram(HISTOGRAM_RELATIVE_ACCURACY, HISTOGRAM_MAX_NUM_BINS);
```
bantonsson, line 265: If this code is ever called concurrently for the same instance you could loose values. If we know this is single threaded, then there needs to be a comment explaining that.

dd-trace-core/src/main/java/datadog/trace/core/histogram/Histogram.java
```diff
@@ -0,0 +1,59 @@
+package datadog.trace.core.histogram;
+
+import com.datadoghq.sketch.ddsketch.DDSketch;
+import datadog.trace.bootstrap.instrumentation.api.AgentHistogram;
+import java.nio.ByteBuffer;
+
+/** Wrapper around the DDSketch library so that it can be used in an instrumentation */
```
dougqh, line 7: Ah, I missed this explanatory comment the first time around.  That make sense.

dougqh (approved):

PR #5543, opened 2023-07-06
httpasyncclient4: fix url parsing and make host/port extraction happening

amarziali:
# What Does This Do

Http async client 4 may not produce absolute urls in `http.url` tag if the client is configured with HttpHost taget and the request uri is hence relative.

This happens when the client executes the request with host and port explicitly provided with a `HttpHost` argument.

Impacted signatures:

* [execute](https://javadoc.io/static/org.apache.httpcomponents/httpasyncclient/4.1.5/org/apache/http/nio/client/HttpAsyncClient.html#execute(org.apache.http.HttpHost,%20org.apache.http.HttpRequest,%20org.apache.http.concurrent.FutureCallback))(org.apache.http.HttpHost target, org.apache.http.HttpRequest request, org.apache.http.concurrent.FutureCallback<org.apache.http.HttpResponse> callback)
* [execute](https://javadoc.io/static/org.apache.httpcomponents/httpasyncclient/4.1.5/org/apache/http/nio/client/HttpAsyncClient.html#execute(org.apache.http.HttpHost,%20org.apache.http.HttpRequest,%20org.apache.http.protocol.HttpContext,%20org.apache.http.concurrent.FutureCallback))(org.apache.http.HttpHost target, org.apache.http.HttpRequest request, org.apache.http.protocol.HttpContext context, org.apache.http.concurrent.FutureCallback<org.apache.http.HttpResponse> callback)

The side effects are:
* `http.url` tag will miss scheme, host, port
* MIssing `peer.hostname`,`peer.port` and `peer.service` tags
* Split by domain feature will not work (since the captured url is missing the host part)

This PR fixes this buggy behaviour and it's a `breaking change`. It also impact the following instrumentations which are relying on httpasyncclient4:
* elasticsearch rest
* opensearch rest

In case this change is not suitable , the old tracing flavour can be restored by adding the following feature flag:
* Sysprop: `-Ddd.httpasyncclient4.legacy.tracing.enabled=true`
* Env: `DD_HTTPASYNCCLIENT4_LEGACY_TRACING_ENABLED=true`

# Motivation

# Additional Notes


amarziali:
drafting since it needs some adjustement

internal-api/src/main/java/datadog/trace/bootstrap/instrumentation/api/URIUtils.java
```diff
@@ -249,4 +249,30 @@ public String get() {
       return lazy;
     }
   }
+
+  /**
+   * Cconcatenate two URI parts to form the complete one. Mostly used for apache http client
```
PerfectSlayer, line 254: Minor typo: `Cconcatenate`

PerfectSlayer (commented):
I get it is breaking changes for `httpasyncclient-4` integration but it should align its behavior with other client integrations, right?

amarziali:
> I get it is breaking changes for `httpasyncclient-4` integration but it should align its behavior with other client integrations, right?

yes it will made http async client 4 behave like other version of same library on the reported used case. So basically it's a bugfix

PerfectSlayer:
So, what was the outcome of the discussion about the breaking behavior?

amarziali:
> So, what was the outcome of the discussion about the breaking behavior?

@PerfectSlayer normally breaking change by default behind a feature flag (as done in this PR). It's ok to ship?

PerfectSlayer (approved):

PR #5546, opened 2023-07-06
activate spans before calling decorators

richardstartin:
# What Does This Do

This means that any tracing latency overhead can be explained by the profiler, which helps users understand request latency

# Motivation

# Additional Notes


dougqh (commented):
Looks good to me

bantonsson (approved):

PR #5547, opened 2023-07-07
Fix maven dependency version parsing

nikita-tkachenko-datadog:
# What Does This Do
Fixes logic that parses Maven artifact versions to support versions that contain dash, e.g. `4.0.0-alpha-5`, `3.0.0-M6`


dd-java-agent/instrumentation/maven-3.2.1/src/main/java/datadog/trace/instrumentation/maven3/MavenDependencyVersion.java
```diff
@@ -1,19 +1,23 @@
...
+  private final String remainder;
 
-  MavenDependencyVersion(int... tokens) {
+  MavenDependencyVersion(int[] tokens, String remainder) {
     this.tokens = tokens;
+    this.remainder = remainder;
   }
 
-  boolean isLaterThanOrEqualTo(MavenDependencyVersion other) {
+  public boolean isLaterThanOrEqualTo(MavenDependencyVersion other) {
```
PerfectSlayer, line 20: What about implementing `Comparable<MavenDependencyVersion>`?

dd-java-agent/instrumentation/maven-3.2.1/src/main/java/datadog/trace/instrumentation/maven3/MavenDependencyVersion.java
```diff
@@ -1,19 +1,23 @@
...
 import java.util.regex.Pattern;
 
 public class MavenDependencyVersion {
 
-  public static final MavenDependencyVersion UNKNOWN = new MavenDependencyVersion();
+  public static final MavenDependencyVersion UNKNOWN = new MavenDependencyVersion(new int[0], null);
 
   private static final Pattern DOT = Pattern.compile("\\.");
   private final int[] tokens;
+  private final String remainder;
```
PerfectSlayer, line 13: This is also called as _qualifier_ in Maven terminology.

dd-java-agent/instrumentation/maven-3.2.1/src/main/java/datadog/trace/instrumentation/maven3/MavenDependencyVersion.java
```diff
@@ -25,15 +29,35 @@ boolean isLaterThanOrEqualTo(MavenDependencyVersion other) {
...
+    if (tokens.length != other.tokens.length) {
+      return tokens.length >= other.tokens.length;
+    }
+    if (remainder == null && other.remainder != null) {
+      return true;
+    }
+    if (remainder != null && other.remainder == null) {
+      return false;
+    }
+    return Objects.compare(remainder, other.remainder, Comparator.naturalOrder()) > 0;
```
PerfectSlayer, line 41: This can be simplified using `nullsFirst()` or `nullsLast()`: https://docs.oracle.com/javase/8/docs/api/java/util/Comparator.html#nullsFirst-java.util.Comparator-

PerfectSlayer (commented):
Some minor comments

dd-java-agent/instrumentation/maven-3.2.1/src/main/java/datadog/trace/instrumentation/maven3/MavenDependencyVersion.java
smola: +1, also I think this still misses a few cases of Maven's version ordering rules: https://maven.apache.org/ref/3.5.2/maven-artifact/apidocs/org/apache/maven/artifact/versioning/ComparableVersion.html

nikita-tkachenko-datadog:
Oof, disregard all of this: I completely missed Maven's `ComparableVersion`. It takes care of many things and there's no point in replicating its logic given that it's available at runtime. I have removed my ad-hoc class and replaced its usages with the Maven's one.

PerfectSlayer (approved):
That’s a very valid way to fix it 😋

PR #5553, opened 2023-07-07
enable apm test agent service naming checks

wconti27:
Enables APM Test Agent Service Naming and Peer Service checks for all integrations


bantonsson (approved):
Would you mind changing the PR title and remove the WIP and also do a Squash Merge to get a cleaner git history?

PR #5556, opened 2023-07-09
Fix exploration tests

jpbempel:
Introduced multiple exclude files for known issue with some classes one exclude file general, and one specific per project. If an instrumentation error is detected it is reported into instrumentation.log file that is tested at the end of the job for indicating if a failure occurred.

# What Does This Do

# Motivation

# Additional Notes

### DIFF
--- a/dd-java-agent/agent-debugger/src/main/java/com/datadog/debugger/agent/DebuggerTransformer.java
+++ b/dd-java-agent/agent-debugger/src/main/java/com/datadog/debugger/agent/DebuggerTransformer.java
@@ -13,14 +13,17 @@ import datadog.trace.agent.tooling.AgentStrategies;
 import datadog.trace.api.Config;
 import datadog.trace.bootstrap.debugger.ProbeId;
 import datadog.trace.bootstrap.debugger.ProbeImplementation;
+import java.io.FileWriter;
 import java.io.IOException;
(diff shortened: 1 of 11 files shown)


dd-java-agent/agent-debugger/src/main/java/com/datadog/debugger/agent/DebuggerTransformer.java
```diff
@@ -91,27 +96,34 @@ public DebuggerTransformer(Config config, Configuration configuration) {
...
+      }
+      try {
+        Files.lines(excludePath)
+            .forEach(
+                line -> {
+                  if (line.startsWith("#")) {
+                    return;
+                  }
+                  if (line.endsWith("*")) {
+                    excludeTrie.insert(line.substring(0, line.length() - 1));
```
shatzi, line 118: maybe also log.debug adding to excludeTrie?

shatzi (approved):
LGTM

dd-java-agent/agent-debugger/src/main/java/com/datadog/debugger/agent/DebuggerTransformer.java
jpbempel: removing the debug log in fact it was left over!

PR #5563, opened 2023-07-11
Added config option to disable baggage as tag injection

nayeem-kamal:
# What Does This Do
Introduces a new config option (`dd.writer.baggage.inject`) that allows the user to disable injecting baggage as tags when sending traces to the Trace Agent.

# Motivation
When sending to the trace agent, the Java tracer merges baggage & tags into the meta section of the span.  From the user perspective in the UI, this effectively copies any baggage into tags. This behavior is not standard compliant.

# Additional Notes


dd-trace-core/src/main/java/datadog/trace/core/DDSpanContext.java
```diff
@@ -729,8 +732,11 @@ public Map<String, Object> getTags() {
 
   public void processTagsAndBaggage(final MetadataConsumer consumer, int longRunningVersion) {
     synchronized (unsafeTags) {
-      Map<String, String> baggageItemsWithPropagationTags = new HashMap<>(baggageItems);
-      propagationTags.fillTagMap(baggageItemsWithPropagationTags);
+      Map<String, String> baggageItemsWithPropagationTags = EMPTY_BAGGAGE;
+      if (injectBaggageAsTags) {
+        baggageItemsWithPropagationTags = new HashMap<>(baggageItems);
+        propagationTags.fillTagMap(baggageItemsWithPropagationTags);
```
bantonsson, line 738: Shouldn't the propagation tags always be present?

bantonsson: I added some comments to the new code.

dd-trace-core/src/main/java/datadog/trace/core/DDSpanContext.java
```diff
@@ -736,7 +735,7 @@ public void processTagsAndBaggage(final MetadataConsumer consumer, int longRunni
               threadId,
               threadName,
               postProcessor.processTags(unsafeTags),
-              baggageItemsWithPropagationTags,
+              injectBaggageAsTags ? baggageItemsWithPropagationTags : EMPTY_BAGGAGE,
```
bantonsson, line 738: This has the same problem as the previous solution, that you completely loose the propagation tags if the `injectBaggageAsTags` is `false`. You also copy the whole baggage map even if you throw it away. I would prefer if the code above does something like
```
if (injectBaggageAsTags) {
  baggageItemsWithPropagationTags = new HashMap<>(baggageItems);
  propagationTags.fillTagMap(baggageItemsWithPropagationTags);
} else {
  baggageItemsWithPropagationTags = propagationTags.createTagMap();
}
```

bantonsson (commented):
Some additions to the baggage and propagation tags tests with this setting 'off' would be a good addition.

dd-trace-core/src/main/java/datadog/trace/core/DDSpanContext.java
nayeem-kamal: I see, Thanks @bantonsson ! I'll make the fix and add some tests.

dd-trace-core/src/test/groovy/datadog/trace/core/DDSpanSerializationTest.groovy
```diff
@@ -428,6 +428,149 @@ class DDSpanSerializationTest extends DDCoreSpecification {
     cleanup:
     tracer.close()
   }
+  def "serialize trace with tags correctly v0.4 with baggage injection disabled"() {
```
PerfectSlayer, line 431: Instead of duplicating the test case, can you parametrise `"serialize trace with baggage and tags correctly v0.4"` with baggage injection?

dd-trace-core/src/main/java/datadog/trace/core/DDSpanContext.java
```diff
@@ -859,4 +863,8 @@ public void setDataCurrent(String key, Object value) {
     String tagKey = "_dd." + key + ".json";
     this.setTag(tagKey, value);
   }
+  // This method is for testing functionality of dd.writer.baggage.inject
+  public void setInjectBaggageAsTags(boolean injectBaggageAsTags) {
```
PerfectSlayer, line 867: Is this needed? Can you use config injection in groovy test instead?

PerfectSlayer (commented):
Bjorn is off, I will try to it from here. 
I added some comments about testing that might simplify the whole thing.

dd-trace-core/src/main/java/datadog/trace/core/DDSpanContext.java
```diff
@@ -187,7 +187,54 @@ public DDSpanContext(
...
+        tagsSize,
+        trace,
+        requestContextDataAppSec,
+        requestContextDataIast,
+        null,
+        pathwayContext,
+        disableSamplingMechanismValidation,
+        propagationTags,
+        ProfilingContextIntegration.NoOp.INSTANCE,
+        injectBaggageAsTags);
```
PerfectSlayer, line 237: Not sure a 4th constructor is needed 🤷 
I would just add the `injectBaggageAsTags` parameter in the last and most complete constructor to keep it simple but that's nitpicking.

PerfectSlayer (approved):

PR #5564, opened 2023-07-11
Tag test spans with method description

nikita-tkachenko-datadog:
# What Does This Do
Adds `test.source.method` tag to each test span. The tag contains concatenation of test method name and method descriptor.
Adds `test.source.class` tag that contains test class. This tag is only set if the contents of existing `test.suite` tag do not match class name (should be a very rare case when Manual API is used and test suite is set to something other than test class name).

# Motivation
The new tags are needed in order to identify Java method that corresponds to a test case.
More details can be found in the [RFC](https://docs.google.com/document/d/1MACBTUmRfWoFpjYiu9RQ8CgytdU0zFPb0Xdc_9UHBq0/edit).


tonyredondo (approved):
LGTM

PR #5565, opened 2023-07-11
Update TestNG instrumentation to differentiate between identical test cases executed concurrently

nikita-tkachenko-datadog:
# What Does This Do
Updates TestNG instrumentation to correctly handle scenarios where the same test case is executed concurrently from multiple threads.

# Motivation
It is possible to create a TestNG XML configuration that will have the same test case included in multiple "tests".
In its simplest form such configuration will look like this:
```
<suite name="API Test Suite" configfailurepolicy="continue">
    <test name="Test A">
        <classes>
            <class name="org.example.TestSucceed">
                <methods>
                    <include name="test_succeed"/>
                </methods>
            </class>
        </classes>
    </test>

    <test name="Test B">
        <classes>
            <class name="org.example.TestSucceed">
                <methods>
                    <include name="test_succeed"/>
                </methods>
            </class>
        </classes>
    </test>
</suite>
```

It is then possible to set `parallel="tests"` attribute on the suite, which will result in two test cases being executed concurrently in two threads.

When a test case finishes, CI Visibility logic locates corresponding `DDTestImpl` instance in its internal registry.
This lookup is done using a `TestDescriptor` key that is supposed to uniquely identify a test case.
The problem is that two test cases from the suite above will have identical descriptors, so when one of them finishes looking up the right `DDTestImpl` instance will be impossible.

# Additional Notes
The issue is addressed by extending `TestDescriptor` with an opaque "test qualifier" that acts as a tie-breaker in cases when two descriptors are otherwise identical.
In the case of Test NG the role of qualifier is played by `name` attribute of the `<test/>` tag

### DIFF
--- a/dd-java-agent/agent-ci-visibility/src/main/java/datadog/trace/civisibility/events/TestDescriptor.java
+++ b/dd-java-agent/agent-ci-visibility/src/main/java/datadog/trace/civisibility/events/TestDescriptor.java
@@ -1,6 +1,7 @@
 package datadog.trace.civisibility.events;
 
 import java.util.Objects;
+import javax.annotation.Nullable;
 
 final class TestDescriptor {
   private final String testSuiteName;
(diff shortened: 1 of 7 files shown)


tonyredondo (approved):
LGTM

PR #5566, opened 2023-07-11
Add support for tycho-surefire-plugin instrumentation in Maven builds

nikita-tkachenko-datadog:
# What Does This Do
Adds support for instrumenting `tycho-surefire-plugin` in Maven builds.
When a Maven build is run with the tracer attached, `tycho-surefire-plugin` executions are identified and their configuration is updated so that the JVMs that are forked by the plugin also run with the tracer attached (similar to the way this is already done for standard Surefire and Failsafe plugins).

# Motivation
`tycho-surefire-plugin` is an alternative to standard Surefire and Failsafe plugins that is used for testing Eclipse plugins.
There are CI Visibility customers that are using this plugin and would like to have visibility into tests that are run with it.


tonyredondo (approved):
Nice refactor

PR #5571, opened 2023-07-12
Jsantos/content type sniffing vulnerability

DDJavierSantos:
# What Does This Do
Implement  X Content Type missing header vulnerability detection

# Motivation

# Additional Notes


dd-java-agent/agent-iast/src/main/java/com/datadog/iast/sink/XContentTypeOptionsModuleImpl.java
```diff
@@ -0,0 +1,54 @@
...
+  private static final Logger LOGGER = LoggerFactory.getLogger(XContentTypeOptionsModuleImpl.class);
+
+  @Override
+  public void onRequestEnd(final Object iastRequestContextObject, final IGSpanInfo igSpanInfo) {
+    try {
+
+      final IastRequestContext iastRequestContext = (IastRequestContext) iastRequestContextObject;
+
+      if (!iastRequestContext.getContentTypeOptionsIsNoSniff()) {
+        if (null != iastRequestContext.getContentType()
```
manuel-alvarez-alvarez, line 27: This is a verbatim copy of the logic from the HSTS module, can you refactor it a bit?

dd-java-agent/agent-iast/src/main/java/com/datadog/iast/sink/HttpResponseHeaderModuleImpl.java
```diff
@@ -51,6 +52,18 @@ public void onHeader(@Nonnull final String name, final String value) {
     } else if (SET_COOKIE_HEADER.equalsIgnoreCase(name)) {
       CookieSecurityParser cookieSecurityInfo = new CookieSecurityParser();
       onCookies(CookieSecurityParser.parse(value));
+    } else if (X_CONTENT_TYPE_OPTIONS.equalsIgnoreCase(name)) {
+      final AgentSpan span = AgentTracer.activeSpan();
```
manuel-alvarez-alvarez, line 56: Can you do like in the HSTS module?

dd-smoke-tests/springboot/src/main/java/datadog/smoketest/springboot/jwt/JwtSecurityConfigurerAdapter.java
```diff
@@ -38,7 +38,10 @@ protected void configure(HttpSecurity http) throws Exception {
         .accessDeniedHandler(
             (request, response, accessDeniedException) -> {
               response.sendError(403, "Access Denied.");
-            });
+            })
+        .and()
+        .headers()
+        .disable();
```
manuel-alvarez-alvarez, line 44: Nice catch!

dd-java-agent/agent-iast/src/main/java/com/datadog/iast/sink/XContentTypeOptionsModuleImpl.java
DDJavierSantos: Fixed with the refactor

dd-java-agent/agent-iast/src/main/java/com/datadog/iast/sink/HttpResponseHeaderModuleImpl.java
DDJavierSantos: Fixed with the refactor

dd-smoke-tests/springboot/src/main/java/datadog/smoketest/springboot/jwt/JwtSecurityConfigurerAdapter.java
DDJavierSantos: Thanks!!!

dd-java-agent/agent-iast/src/main/java/com/datadog/iast/util/HttpHeader.java
```diff
@@ -0,0 +1,86 @@
...
+
+  public boolean matches(final String name) {
+    return name != null && name.equalsIgnoreCase(this.name);
+  }
+
+  public static HttpHeader from(final String name) {
+    return Values.HEADERS.get(name.toLowerCase(Locale.ROOT));
+  }
+
+  public abstract static class ContextAwareHeader extends HttpHeader {
```
smola, line 26: Do we need this intermediate class? We don't seem to create relevant instances of `HttpHeader` that are not also `ContextAwareHeader`, except for `HttpHeader.from`, which we only use to then check if it's an instance of `ContextAwareHeader`.

It seems it might be better to just have `HttpHeader.from` return null when the header is unknown or irrelevant to us. And then have the same type of callback for unvalidated redirect and set-cookie. Although even if we decide to change this, I'm fine with a separate PR, since this also would touch on set-cookie and redirect logic.

internal-api/src/main/java/datadog/trace/api/iast/VulnerabilityTypes.java
```diff
@@ -14,6 +14,8 @@ private VulnerabilityTypes() {}
   public static final String INSECURE_COOKIE = "INSECURE_COOKIE";
   public static final String NO_HTTPONLY_COOKIE = "NO_HTTPONLY_COOKIE";
   public static final String HSTS_HEADER_MISSING = "HSTS_HEADER_MISSING";
+  public static final String XCONTENTTYPEOPTIONS_HEADER_MISSING =
```
smola, line 17: The official ID in the schema is `XCONTENTTYPE_HEADER_MISSING`.
(which makes me think we're missing a system test for this)

internal-api/src/main/java/datadog/trace/api/iast/InstrumentationBridge.java
```diff
@@ -43,6 +44,9 @@ public abstract class InstrumentationBridge {
   public static volatile WeakRandomnessModule WEAK_RANDOMNESS;
   public static volatile HttpResponseHeaderModule RESPONSE_HEADER_MODULE;
   public static volatile HstsMissingHeaderModule HSTS_MISSING_HEADER_MODULE;
+
+  public static volatile XContentTypeOptionsModule X_CONTENT_TYPE_OPTIONS_HEADER_MODULE;
```
smola, line 48: I don't this is ever used?

internal-api/src/main/java/datadog/trace/api/iast/InstrumentationBridge.java
```diff
@@ -43,6 +44,9 @@ public abstract class InstrumentationBridge {
   public static volatile WeakRandomnessModule WEAK_RANDOMNESS;
   public static volatile HttpResponseHeaderModule RESPONSE_HEADER_MODULE;
   public static volatile HstsMissingHeaderModule HSTS_MISSING_HEADER_MODULE;
```
smola, line 46: Do we still need these modules? I thought we would need to exponse only `RESPONSE_HEADER_MODULE`.

dd-java-agent/agent-iast/src/main/java/com/datadog/iast/util/HttpHeader.java
manuel-alvarez-alvarez: `ContextAwareHeader` is used for those headers that put extra information in the context to be used latter during the detection of vulnerabilities (for instance the `Content-Type`). It's true that we could remove this class and keep the callback in the `HttpHeader` leavING it empty by default (making it a no-op for classes that don't deal with the context)

internal-api/src/main/java/datadog/trace/api/iast/VulnerabilityTypes.java
DDJavierSantos: changed vulnerability and module name to XCONTENTTYPE_HEADER_MISSING

dd-java-agent/agent-iast/src/main/java/com/datadog/iast/util/HttpHeader.java
DDJavierSantos: I agree. I will do the refactor in a separated PR

internal-api/src/main/java/datadog/trace/api/iast/InstrumentationBridge.java
DDJavierSantos: The reason for this modules to exist is so the specific vulnerability can be turned on and off separatedly.

internal-api/src/main/java/datadog/trace/api/iast/InstrumentationBridge.java
DDJavierSantos: It is used in RequestEndModules

smola: Only in:
https://github.com/DataDog/dd-trace-java/pull/5571/files#diff-c291a984ba7378263e7b3fe8b85757e0290ca17c3092484217e4b267d2b36b31R51-R55
which is within the `agent-iast` module. It does not seem this needs to leak to `internal-api`?

DDJavierSantos: With this module iin internal-api the specific X Content Type missing header vulnerability can be enabled and disabled we have done pretty much the same for the header vulnerabilities.

smola: Ok. I think we'll need to rethink that to avoid leaking unnecessary details to `internal-api`, but better to think of it independently of this feature. Marking as resolved.

dd-java-agent/agent-iast/src/main/java/com/datadog/iast/sink/HstsMissingHeaderModuleImpl.java
```diff
@@ -85,4 +64,19 @@ public static boolean isValidMaxAge(@Nullable final String value) {
     }
     return Integer.parseInt(matcher.group(1)) > 0;
   }
+
+  static boolean isHttps(@Nullable final String urlString, @Nullable final String forwardedFor)
+      throws MalformedURLException {
+    if (urlString == null) {
+      return false;
+    }
+    final URL url = new URL(urlString);
```
smola, line 73: * Prefer URI class, because `URL` makes DNS requests on some methods like `equals`.
* Do not let this kind of exception propagate. We don't want to be spamming logs just because requests have invalid headers. This is a controller error.
* In this case, we might want to just check if the URL starts with `https://`?

DDJavierSantos: fixed

smola (approved):

PR #5589, opened 2023-07-17
Add support to IAST XSS vulnerability

jandro996:
# What Does This Do
- New XSS Module
- Add Call Site Instrumentation to `java.io.PrintWriter` methods used in OWASP Benchmark
- Add spring-boot smoke tests

# Motivation
Pass OWASP Benchmark related to XSS vulnerability

# Additional Notes
Full XSS topic is not covered, only the necessary for OWASP Benchmark


dd-smoke-tests/spring-boot-2.6-webmvc/src/main/java/datadog/smoketest/springboot/controller/IastWebController.java
```diff
@@ -264,6 +264,18 @@ public String xpathInjectionEvaluate(final HttpServletRequest request)
     return "XPath Injection page";
   }
 
+  @GetMapping("/xss/write")
+  @SuppressFBWarnings
+  public String xssWrite(final HttpServletRequest request, final HttpServletResponse response)
+      throws XPathExpressionException {
```
manuel-alvarez-alvarez, line 270: The `throws` looks suspicious, and if you are writing to the response directly probably the method should be `void`

manuel-alvarez-alvarez (approved):

PR #5591, opened 2023-07-18
Add OpenTelemetry span name tests

PerfectSlayer:
# What Does This Do

This minor PR updates the existing test cases to ensure the current beharior: spans generated from OpenTelemetry instrumentation should have their operation name and resource name set with the OpenTelemetry span name.

# Motivation

Ensure consistant behavior between tracer implementations.

# Additional Notes


nayeem-kamal (approved):
LGTM

PR #5594, opened 2023-07-18
More accurately report whether the request was blocked

cataphract:
The tag `appsec.blocked` was being set only in code paths where decorator would invoke the WAF and call `datadog.trace.api.gateway.IGSpanInfo#setRequestBlockingAction`. This was both underinclusive (many blocking advice would not call `setRequestBlockingAction`) and overinclusive (there was no guarantee that the advice would call `datadog.trace.api.gateway.IGSpanInfo#getRequestBlockingAction` or that it would be able to block).

So set the tag closer to where the blocking is actually happening so that the value `appsec.blocked` is more accurate.

### DIFF
--- a/dd-java-agent/appsec/src/main/java/com/datadog/appsec/gateway/GatewayBridge.java
+++ b/dd-java-agent/appsec/src/main/java/com/datadog/appsec/gateway/GatewayBridge.java
@@ -118,9 +118,6 @@ public class GatewayBridge {
           if (traceSeg != null) {
             traceSeg.setTagTop("_dd.appsec.enabled", 1);
             traceSeg.setTagTop("_dd.runtime_family", "jvm");
-            if (spanInfo.getRequestBlockingAction() != null) {
-              traceSeg.setTagTop("appsec.blocked", "true");
-            }
 
             Collection<AppSecEvent100> collectedEvents = ctx.transferCollectedEvents();
 
@@ -226,7 +223,8 @@ public class GatewayBridge {
               DataBundle bundle =
                   new SingletonDataBundle<>(KnownAddresses.REQUEST_PATH_PARAMS, data);
               try {
-                return producerService.publishDataEvent(subInfo, ctx, bundle, false);
+                Flow<Void> flow = producerService.publishDataEvent(subInfo, ctx, bundle, false);
+                return flow;
               } catch (ExpiredSubscriberInfoException e) {
                 pathParamsSubInfo = null;
               }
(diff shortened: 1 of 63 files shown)


dd-trace-api/src/main/java/datadog/appsec/api/blocking/Blocking.java
```diff
@@ -70,8 +70,10 @@ public static boolean tryCommitBlockingResponse(
    */
   public static boolean tryCommitBlockingResponse(int statusCode, BlockingContentType contentType) {
     try {
-      return SERVICE.tryCommitBlockingResponse(
-          statusCode, BlockingContentType.NONE, Collections.emptyMap());
+      boolean committedBlockingResponse =
```
jandro996, line 73: I think we don't need this boolean

dd-smoke-tests/appsec/springboot/src/test/groovy/datadog/smoketest/appsec/SpringBootSmokeTest.groovy
```diff
@@ -12,6 +12,7 @@ class SpringBootSmokeTest extends AbstractAppSecServerSmokeTest {
 
     List<String> command = new ArrayList<>()
     command.add(javaPath())
+    //    command.addAll(['-agentlib:jdwp=transport=dt_socket,server=y,suspend=y,address=0.0.0.0:5005'])
```
jandro996, line 15: I don't know if you forget to remove this comment related with remote debugging

dd-java-agent/appsec/src/main/java/com/datadog/appsec/gateway/GatewayBridge.java
```diff
@@ -226,7 +223,8 @@ public void init() {
               DataBundle bundle =
                   new SingletonDataBundle<>(KnownAddresses.REQUEST_PATH_PARAMS, data);
               try {
-                return producerService.publishDataEvent(subInfo, ctx, bundle, false);
+                Flow<Void> flow = producerService.publishDataEvent(subInfo, ctx, bundle, false);
```
jandro996, line 226: This change is needed?

jandro996 (approved):

dd-smoke-tests/appsec/springboot/src/test/groovy/datadog/smoketest/appsec/SpringBootSmokeTest.groovy
cataphract: thanks, I'll remove it

dd-trace-api/src/main/java/datadog/appsec/api/blocking/Blocking.java
cataphract: I prefer it this way because it's better for debugging. I can read the return value value without going up the stack, as well as put a conditional breakpoint.

dd-java-agent/appsec/src/main/java/com/datadog/appsec/gateway/GatewayBridge.java
cataphract: same comment as above

PR #5602, opened 2023-07-19
Implement Intelligent Test Runner metadata tags

nikita-tkachenko-datadog:
# What Does This Do
Adds ITR (Intelligent Test Runner) metadata tags to test session and test module spans:
* `test.itr.tests_skipping.type`: `"test"` - to indicate that Java tracer skips individual test cases when ITR is enabled
* `test.itr.tests_skipping.count`: `number` - to tell how many test cases were skipped in a given session / module

And to test spans:
`test.skipped_by_itr`: `true` - if the test was skipped by ITR

# Motivation
The tracer needs to provide this info so that it could be displayed to the customers in the UI.
More details can be found in the [RFC](https://docs.google.com/document/d/105YXO2r2hQqwkbEgKOoO8MuT2OAawa5Wh8PwujJC9LY/edit).

### DIFF
--- a/dd-java-agent/agent-ci-visibility/src/main/java/datadog/trace/civisibility/DDTestModuleParent.java
+++ b/dd-java-agent/agent-ci-visibility/src/main/java/datadog/trace/civisibility/DDTestModuleParent.java
@@ -129,8 +129,11 @@ public class DDTestModuleParent extends DDTestModuleImpl {
     }
     span.setTag(Tags.TEST_STATUS, context.getStatus());
 
-    if (testsSkipped) {
+    long testsSkippedTotal = testsSkipped.sum();
+    if (testsSkippedTotal > 0) {
       setTag(DDTags.CI_ITR_TESTS_SKIPPED, true);
+      setTag(Tags.TEST_ITR_TESTS_SKIPPING_TYPE, "test");
+      setTag(Tags.TEST_ITR_TESTS_SKIPPING_COUNT, testsSkippedTotal);
     }
 
     Object testFramework = context.getChildTag(Tags.TEST_FRAMEWORK);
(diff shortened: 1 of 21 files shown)


PerfectSlayer (approved):
LGTM.
I was not familiar with what `ITR` stands for so I updated the title and description for others 😉

dd-java-agent/agent-ci-visibility/src/main/java/datadog/trace/civisibility/DDTestModuleParent.java
```diff
@@ -129,8 +129,11 @@ public void end(@Nullable Long endTime) {
     }
     span.setTag(Tags.TEST_STATUS, context.getStatus());
 
-    if (testsSkipped) {
+    long testsSkippedTotal = testsSkipped.sum();
+    if (testsSkippedTotal > 0) {
       setTag(DDTags.CI_ITR_TESTS_SKIPPED, true);
+      setTag(Tags.TEST_ITR_TESTS_SKIPPING_TYPE, "test");
+      setTag(Tags.TEST_ITR_TESTS_SKIPPING_COUNT, testsSkippedTotal);
```
gnufede, line 136: I believe we may send these last two even if no tests were skipped

romainkomorndatadog:
LGTM but I agree with @gnufede that we want the data no matter what.

dd-java-agent/agent-ci-visibility/src/main/java/datadog/trace/civisibility/DDTestModuleParent.java
nikita-tkachenko-datadog: Good catch, Fede! Fixed

PR #5608, opened 2023-07-20
Add Trust Boundary Violation OWASP Benchmark  escape functions taint tracking 

jandro996:
# What Does This Do

Add CSI to `org.springframework.web.util.HtmlUtils#htmlEscape` and `org.apache.commons.lang.StringEscapeUtils#escapeHtml` and `org.owasp.esapi.Encoder#encodeForHTML`
Mark for XSS if it's tainted

# Motivation

Pass trust boundary violation OWASP Benchmark

# Additional Notes

This PR only covers methods that we need to pass the OWASP Benchmark, future PRs will deal with the rest of escape functions


dd-java-agent/agent-iast/src/main/java/com/datadog/iast/propagation/PropagationModuleImpl.java
```diff
@@ -244,6 +246,23 @@ public Taintable.Source firstTaintedSource(@Nullable final Object input) {
     return firstTaintedSource(taintedObjects, input);
   }
 
+  @Override
+  public void taintAndMarkXSSIfInputIsTainted(
```
manuel-alvarez-alvarez, line 250: Can't you create and overload of the current tainted method that takes an int with the marks? something like:

```java
public void taintIfInputIsTainted(String toTaint, Object input, final int marks) {
}
```

dd-java-agent/agent-iast/src/main/java/com/datadog/iast/propagation/PropagationModuleImpl.java
```diff
@@ -269,7 +288,7 @@ private static Source firstTaintedSource(
     } else {
       final TaintedObject tainted = taintedObjects.get(object);
       final Range[] ranges = tainted == null ? null : tainted.getRanges();
-      return ranges != null && ranges.length > 0 ? ranges[0].getSource() : null;
+      return ranges != null && ranges.length > 0 ? highestPriorityRange(ranges).getSource() : null;
```
manuel-alvarez-alvarez, line 291: Can you rename the method? it's not the `firstTaintedSource` anymore

manuel-alvarez-alvarez (approved):

PR #5613, opened 2023-07-20
Report test framework data from child processes instead of parsing project dependencies

nikita-tkachenko-datadog:
# What Does This Do
Update CI Visibility: test framework name and version are now sent from children processes to the parent process using signal server.

# Motivation
Improving the accuracy of test framework name and version detection.

Test session and test modules spans are created by the parent process.
These spans need to be tagged with test framework name and version.
Previous implementation attempted to extract that data by parsing the list of project's Maven or Gradle dependencies.
That approach often gave inaccurate results (for instance, in cases when a project had JUnit dependency but ran another framework such as Spock or Cucumber on top of JUnit).
The new approach is more accurate since the data is obtained from test framework instrumentations.

### DIFF
--- a/dd-java-agent/agent-ci-visibility/src/main/java/datadog/trace/civisibility/context/ParentProcessTestContext.java
+++ b/dd-java-agent/agent-ci-visibility/src/main/java/datadog/trace/civisibility/context/ParentProcessTestContext.java
@@ -38,8 +38,8 @@ public class ParentProcessTestContext extends AbstractTestContext implements Tes
   public void reportChildTag(String key, Object value) {
     // the method leaves room for a
     // proper implementation using a thread-safe map,
-    // but for not it's just this,
-    // in order to save some performance costs
+    // but for now it's just this,
+    // to save some performance costs
     switch (key) {
       case Tags.TEST_FRAMEWORK:
         testFramework = String.valueOf(value);
--- a/dd-java-agent/agent-ci-visibility/src/main/java/datadog/trace/civisibility/context/ParentProcessTestContext.java
+++ b/dd-java-agent/agent-ci-visibility/src/main/java/datadog/trace/civisibility/context/ParentProcessTestContext.java
@@ -1,11 +1,14 @@
 package datadog.trace.civisibility.context;
 
 import datadog.trace.bootstrap.instrumentation.api.AgentSpan;
+import datadog.trace.bootstrap.instrumentation.api.Tags;
 import javax.annotation.Nonnull;
 import javax.annotation.Nullable;
 
 public class ParentProcessTestContext extends AbstractTestContext implements TestContext {
 
+  private volatile String testFramework;
+  private volatile String testFrameworkVersion;
   private final long sessionId;
   private final long moduleId;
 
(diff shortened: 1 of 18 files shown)


dd-java-agent/agent-ci-visibility/src/main/java/datadog/trace/civisibility/context/ParentProcessTestContext.java
```diff
@@ -30,4 +33,34 @@ public Long getParentId() {
   public AgentSpan getSpan() {
     return null;
   }
+
+  @Override
+  public void reportChildTag(String key, Object value) {
+    // the method leaves room for a
+    // proper implementation using a thread-safe map,
+    // but for not it's just this,
+    // in order to save some performance costs
```
juan-fernandez, line 42: nitpick 😄 
```suggestion
    // but for now it's just this,
    // to save some performance costs
```

dd-java-agent/agent-ci-visibility/src/test/groovy/datadog/trace/civisibility/ipc/ModuleExecutionResultTest.groovy
```diff
@@ -16,11 +16,11 @@ class ModuleExecutionResultTest extends Specification {
...
-      new ModuleExecutionResult(12345, 67890, false, false, 0),
-      new ModuleExecutionResult(12345, 67890, true, false, 1),
-      new ModuleExecutionResult(12345, 67890, false, true, 2),
-      new ModuleExecutionResult(12345, 67890, false, false, 3),
-      new ModuleExecutionResult(12345, 67890, true, true, Integer.MAX_VALUE)
+      new ModuleExecutionResult(12345, 67890, false, false, 0, null, null),
+      new ModuleExecutionResult(12345, 67890, true, false, 1, "abc", "def"),
+      new ModuleExecutionResult(12345, 67890, false, true, 2, null, "def"),
+      new ModuleExecutionResult(12345, 67890, false, false, 3, "abc", null),
+      new ModuleExecutionResult(12345, 67890, true, true, Integer.MAX_VALUE, "abc", "def")
```
juan-fernandez, line 23: do the last two values represent `testFramwork` and `testFrameworkVersion`? If that's the case, is there value in setting more real values? Just to make it easier to read

juan-fernandez: if doing this implies doing a bigger refactor feel free to ignore it

dd-java-agent/instrumentation/maven-3.2.1/src/test/groovy/MavenTest.groovy
```diff
@@ -279,12 +279,12 @@ class MavenTest extends CiVisibilityTest {
...
   @Override
   String expectedTestFramework() {
-    return "junit4"
+    return null
   }
 
   @Override
   String expectedTestFrameworkVersion() {
-    return "4.13.2"
+    return null
```
juan-fernandez, line 287: why is this now null?

juan-fernandez (approved):

dd-java-agent/agent-ci-visibility/src/test/groovy/datadog/trace/civisibility/ipc/ModuleExecutionResultTest.groovy
nikita-tkachenko-datadog: I don't think there's much value other than some readability gains. Still, updated this

dd-java-agent/agent-ci-visibility/src/main/java/datadog/trace/civisibility/context/ParentProcessTestContext.java
nikita-tkachenko-datadog: Updated

PR #5617, opened 2023-07-21
Introduce permanent errors

jpbempel:
# What Does This Do
When conditions or value expressions contain permanent errors like referencing a variable not defined, we report the error through probe statuses instead of snapshot and evaluation errors. We keep evaluation errors for transient error like a null pointer.

# Motivation

# Additional Notes

### DIFF
--- a/dd-java-agent/agent-debugger/src/main/java/com/datadog/debugger/agent/LogMessageTemplateBuilder.java
+++ b/dd-java-agent/agent-debugger/src/main/java/com/datadog/debugger/agent/LogMessageTemplateBuilder.java
@@ -1,27 +1,29 @@
 package com.datadog.debugger.agent;
 
 import static com.datadog.debugger.util.ValueScriptHelper.serializeValue;
+import static java.util.Collections.singletonList;
 
 import com.datadog.debugger.el.EvaluationException;
+import com.datadog.debugger.el.PermanentEvaluationException;
 import com.datadog.debugger.el.Value;
 import com.datadog.debugger.el.ValueScript;
+import com.datadog.debugger.instrumentation.DiagnosticMessage;
 import com.datadog.debugger.probe.LogProbe;
 import datadog.trace.bootstrap.debugger.CapturedContext;
 import datadog.trace.bootstrap.debugger.EvaluationError;
+import datadog.trace.bootstrap.debugger.ProbeId;
 import java.util.List;
 import org.slf4j.Logger;
 import org.slf4j.LoggerFactory;
 
 public class LogMessageTemplateBuilder {
   private static final Logger LOGGER = LoggerFactory.getLogger(LogMessageTemplateBuilder.class);
-  /**
-   * Serialization limits for log messages. Most values are lower than snapshot because you can
-   * directly reference values that are in your interest with Expression Language:
-   * obj.field.deepfield or array[1001]
-   */
+
+  private final ProbeId probeId;
   private final List<LogProbe.Segment> segments;
 
-  public LogMessageTemplateBuilder(List<LogProbe.Segment> segments) {
+  public LogMessageTemplateBuilder(ProbeId probeId, List<LogProbe.Segment> segments) {
+    this.probeId = probeId;
     this.segments = segments;
   }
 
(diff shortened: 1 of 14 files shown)


dd-java-agent/agent-debugger/src/main/java/com/datadog/debugger/agent/LogMessageTemplateBuilder.java
```diff
@@ -45,6 +47,10 @@ public String evaluate(CapturedContext context, LogProbe.LogStatus status) {
             } else {
               serializeValue(sb, segment.getParsedExpr().getDsl(), result.getValue(), status);
             }
+          } catch (PermanentEvaluationException ex) {
+            DiagnosticMessage diagnosticMessage =
+                new DiagnosticMessage(DiagnosticMessage.Kind.ERROR, ex.getMessage());
+            DebuggerAgent.getSink().addDiagnostics(probeId, singletonList(diagnosticMessage));
```
shatzi, line 53: what happen if we have multiple evaluation errors? do we generate multiple error status messages?
What happen if we get both permanent and evaluation errors at the same time?

For example if we have a log template with two expressions: one fails permanent and the other does not. I think in that case we should just emit a snapshot with evaluation errors for both.

dd-java-agent/agent-debugger/src/main/java/com/datadog/debugger/probe/SpanDecorationProbe.java
```diff
@@ -150,6 +151,13 @@ public void evaluate(CapturedContext context, CapturedContext.Status status) {
           if (!condition) {
             continue;
           }
+        } catch (PermanentEvaluationException ex) {
+          LOGGER.debug("Permanent evaluation error: ", ex);
```
shatzi, line 155: SpanTag can have multiple conditions and therefore multiple errors. I think in those cases we should collect all errors and eval both diagnostic message + eval errors on the span. currently permanent errors are not reported as errors when looking at the span

dd-java-agent/agent-debugger/src/main/java/com/datadog/debugger/probe/SpanDecorationProbe.java
```diff
@@ -150,6 +151,13 @@ public void evaluate(CapturedContext context, CapturedContext.Status status) {
...
+          DiagnosticMessage diagnosticMessage =
+              new DiagnosticMessage(DiagnosticMessage.Kind.ERROR, ex.getMessage());
+          DebuggerAgent.getSink()
```
shatzi, line 158: does the sink sends a new message every time or there is a logic that if the message is the same it not resend it?
Does that logic also work if we have two failed conditions with the different errors?

dd-java-agent/agent-debugger/src/test/java/com/datadog/debugger/agent/CapturedSnapshotTest.java
```diff
@@ -980,12 +989,12 @@ public void mergedProbesConditionMainErrorAdditionalFalse()
...
-    List<Snapshot> snapshots = doMergedProbeConditions(condition1, condition2, 1);
-    List<EvaluationError> evaluationErrors = snapshots.get(0).getEvaluationErrors();
-    Assertions.assertEquals(1, evaluationErrors.size());
-    Assertions.assertEquals("fld", evaluationErrors.get(0).getExpr());
-    Assertions.assertEquals(
-        "Cannot dereference to field: fld", evaluationErrors.get(0).getMessage());
+    DebuggerTransformerTest.TestSnapshotListener listener =
+        doMergedProbeConditions(condition1, condition2);
+    assertEquals(0, listener.snapshots.size());
+    List<DiagnosticMessage> errors = listener.errors.get(PROBE_ID1.getId());
```
shatzi, line 995: what happen if we run this function 1000 times?

dd-java-agent/agent-debugger/src/main/java/com/datadog/debugger/agent/LogMessageTemplateBuilder.java
```diff
@@ -1,27 +1,29 @@
...
-   * Serialization limits for log messages. Most values are lower than snapshot because you can
-   * directly reference values that are in your interest with Expression Language:
-   * obj.field.deepfield or array[1001]
-   */
+
+  private final ProbeId probeId;
   private final List<LogProbe.Segment> segments;
 
-  public LogMessageTemplateBuilder(List<LogProbe.Segment> segments) {
+  public LogMessageTemplateBuilder(ProbeId probeId, List<LogProbe.Segment> segments) {
```
shatzi, line 25: alternative design would be is to add isPermanent flag to EvaluationError.  
Aka: status.addError(EvaluationError(isPermanent=true))

Then at the commit stage we can collect and see if we have permanent errors and report an Probe Status message with the right list of errors.

shatzi (commented):
Lets chat about it tomorrow, I would love to go over edge cases of multiple errors per probe.

dd-java-agent/agent-debugger/src/main/java/com/datadog/debugger/agent/LogMessageTemplateBuilder.java
jpbempel: we don't support multiple error messages in probe statuses

PR #5620, opened 2023-07-21
Fix log template issue for duplicated line probes

jpbempel:
# What Does This Do
With line probe, the freeze/commit of the data is done just after the evaluation per probe id.
The fix is to perform 2 loops on the probe ids, first only evaluation and then commit.

# Motivation
Having 2 line probes on the same location with same template or accessing the same field, arg or var result in an error evaluating the template.

# Additional Notes


shatzi (approved):
should we assert not call evaluate if the probed was committed already? or the unit test is enough?

jpbempel:
> should we assert not call evaluate if the probed was committed already? or the unit test is enough?

unit test is enough

PR #5624, opened 2023-07-24
Add more org.owasp.esapi.Encoder escape functions taint tracking

jandro996:
# What Does This Do

# Motivation

# Additional Notes


dd-java-agent/instrumentation/owasp-esapi-2/src/main/java/datadog/trace/instrumentation/owasp/esapi/EncoderCallSite.java
```diff
@@ -28,4 +29,111 @@ public static String afterEncodeForHTML(
     }
     return result;
   }
+
+  @CallSite.After("java.lang.String org.owasp.esapi.Encoder.canonicalize(java.lang.String)")
```
manuel-alvarez-alvarez, line 33: I miss some smoke tests to validate that we don't trigger XSS vulnerabilities when the encoder is used

manuel-alvarez-alvarez (approved):

smola (approved):

PR #5627, opened 2023-07-24
Disable code coverage segments data gathering by default

nikita-tkachenko-datadog:
# What Does This Do
Update code coverage logic in CI Visibility: by default when per-test code coverage is enabled, only the list of covered files will be collected for each test case.
In order to collect covered code segment that include line numbers, a separate feature flag will have to be enabled.

# Motivation
Intelligent Test Runner only requires the list of covered files in order to work.
Backend does not use covered segments data at all at the moment, and gathering this data on the tracer side results in significant (up to 30%) overhead.

# Additional Notes
The overhead from collecting segments data mainly comes from turning the list of activated coverage probes into actual source code line numbers: in order to do that, Jacoco has to parse and analyze the class file, which is a computationally-intensive operation.
Because it is not unusual for a class to be covered by multiple test cases, parsing every class happens multiple times.
There is room for optimisation here, but that is left for future PRs.


juan-fernandez (approved):

PerfectSlayer (approved):
LGTM

PR #5629, opened 2023-07-24
[Spark] Add support for spark structured streaming

paul-laffon-dd:
# What Does This Do

[Spark structured streaming](https://spark.apache.org/docs/latest/structured-streaming-programming-guide.html) is a stream processing engine built on the Spark SQL. It process data streams as a series of small batch jobs, where each batch jobs is composed of the regular spark operations (job / stage / task)

A streaming applications is composed of 
- one or more long running queries, identified a `queryId`
- each query continuously executes batches identified by a incrementing `batchId`

Spark is emitting 3 type of events in structured streaming:
- `QueryStartedEvent` 
- `QueryTerminatedEvent` 
- `QueryProgressEvent`

`QueryProgressEvent` is emitted at the end of each micro-batch in the [finishTrigger](https://github.com/apache/spark/blob/master/sql/core/src/main/scala/org/apache/spark/sql/execution/streaming/MicroBatchExecution.scala#L293) method and has a lot of metrics about the executed batch

There is no event emitted at the start of a micro batch (QueryStartedEvent is emitted at the start of the long running query, not for every micro-batch). To circumvent that, the batch span is started at the start of its first job, the caveat is that the first job is started after the `getBatch` and `queryPlanning` computations, leading to the batch span being shorter that the real batch duration

Spark structured streaming is part of the `spark-sql` module, meaning that the muzzle directive has to be changed. This is acceptable as Spark SQL is the recommended way to use spark and vendors like EMR or databricks always provide `spark-sql` in their managed versions

### DIFF
--- a/dd-java-agent/instrumentation/spark/src/main/java/datadog/trace/instrumentation/spark/DatadogSparkListener.java
+++ b/dd-java-agent/instrumentation/spark/src/main/java/datadog/trace/instrumentation/spark/DatadogSparkListener.java
@@ -47,6 +47,7 @@ public class DatadogSparkListener extends SparkListener {
   private final AgentTracer.TracerAPI tracer;
 
   private AgentSpan applicationSpan;
+  private SparkListenerApplicationStart applicationStart;
   private final HashMap<String, AgentSpan> streamingBatchSpans = new HashMap<>();
   private final HashMap<Integer, AgentSpan> jobSpans = new HashMap<>();
   private final HashMap<Long, AgentSpan> stageSpans = new HashMap<>();
@@ -68,6 +69,7 @@ public class DatadogSparkListener extends SparkListener {
   private boolean lastJobFailed = false;
   private String lastJobFailedMessage;
   private String lastJobFailedStackTrace;
+  private int jobCount = 0;
   private int currentExecutorCount = 0;
   private int maxExecutorCount = 0;
   private long availableExecutorTime = 0;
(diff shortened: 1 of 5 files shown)


dd-java-agent/instrumentation/spark/src/main/java/datadog/trace/instrumentation/spark/DatadogSparkListener.java
```diff
@@ -173,7 +202,7 @@ public synchronized void onJobStart(SparkListenerJobStart jobStart) {
         }
       }
     } else {
-      // In non-databricks env, the spark application is the local root spans
+      // In non-databricks, non-streaming env, the spark application is the local root span
```
raphaelgavache, line 205: Spark application spans will still be created and be periodically flushed with the long running mechanism.
Shouldn't applicationSpan.start() be applied instead, and end() SparkApplicationEnd event be ignored if we're not tracing the application

raphaelgavache (approved):
very nice, lgtm!

dd-java-agent/instrumentation/spark/src/main/java/datadog/trace/instrumentation/spark/DatadogSparkListener.java
```diff
@@ -327,6 +380,15 @@ public synchronized void onStageCompleted(SparkListenerStageCompleted stageCompl
       jobMetrics
           .computeIfAbsent(jobId, k -> new SparkAggregatedTaskMetrics())
           .accumulateStageMetrics(stageMetric);
+
+      Properties prop = stageProperties.remove(stageSpanKey);
+      String batchKey = getStreamingBatchKey(prop);
```
bantonsson, line 385: I'm a bit confused about how this works and what metrics we want to accumulate and show. This code combined with the code below creates one new `SparkAggregatedTaskMetrics` per query_id and micro_batch_id combination. I assume that is what we want. How many Histograms will this create on _normal_ workloads?

bantonsson (commented):
Looks good, with only one single question.

dd-java-agent/instrumentation/spark/src/main/java/datadog/trace/instrumentation/spark/DatadogSparkListener.java
paul-laffon-dd: The [accumulateStageMetrics](https://github.com/DataDog/dd-trace-java/blob/master/dd-java-agent/instrumentation/spark/src/main/java/datadog/trace/instrumentation/spark/SparkAggregatedTaskMetrics.java#L154-L189) of `SparkAggregatedTaskMetrics` keeps an aggregated sum of the metrics present in the tasks of this query_id/micro_batch_id (sum of cpu time / gc time / input / output ...). The idea is to have all those values aggregated and available at each spark level (application, batch, job, stage)

The histograms are not accumulated and are kept at the stage level (each stage is performing a different operation, so it does not really make sense to accumulate histograms from different stages), so no new histogram is created from this accumulation per query_id/micro_batch_id

bantonsson (approved):

PR #5649, opened 2023-07-27
Fix async propagation in some versions of undertow

cataphract:
The executor that netty uses in more recent versions is not instrumented.

### DIFF
--- a/dd-java-agent/instrumentation/java-concurrent/src/main/java/datadog/trace/instrumentation/java/concurrent/WrapRunnableAsNewTaskInstrumentation.java
+++ b/dd-java-agent/instrumentation/java-concurrent/src/main/java/datadog/trace/instrumentation/java/concurrent/WrapRunnableAsNewTaskInstrumentation.java
@@ -39,7 +39,8 @@ public final class WrapRunnableAsNewTaskInstrumentation extends Instrumenter.Tra
       "io.netty.util.concurrent.GlobalEventExecutor",
       "io.netty.util.concurrent.SingleThreadEventExecutor",
       "java.util.concurrent.AbstractExecutorService",
-      "org.glassfish.grizzly.threadpool.GrizzlyExecutorService"
+      "org.glassfish.grizzly.threadpool.GrizzlyExecutorService",
+      "org.jboss.threads.EnhancedQueueExecutor",
     };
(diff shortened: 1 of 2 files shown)


mcculls (approved):

richardstartin:
I think this will cause issues: this executor wraps the runnable to preserve the classloader:

```java
final Runnable realRunnable = JBossExecutors.classLoaderPreservingTaskUnchecked(runnable);
```

And this instrumentation will wrap the runnable in a bootstrap type (FutureTask), which breaks this classloader preservation. The wrapper type `ContextClassLoaderSavingRunnable` is already instrumented to activate its embedded continuation, so what about instrumenting its constructor to capture an active scope, or capture on exit from `BossExecutors.classLoaderPreservingTaskUnchecked(Runnable)` instead?

richardstartin:
The title and label are also misleading because this isn't part of netty but belongs to jboss-threads.

cataphract:
> I think this will cause issues: this executor wraps the runnable to preserve the classloader:
> 
> ```java
> final Runnable realRunnable = JBossExecutors.classLoaderPreservingTaskUnchecked(runnable);
> ```
> 
> And this instrumentation will wrap the runnable in a bootstrap type (FutureTask), which breaks this classloader preservation. The wrapper type `ContextClassLoaderSavingRunnable` is already instrumented to activate its embedded continuation, so what about instrumenting its constructor to capture an active scope, or capture on exit from `BossExecutors.classLoaderPreservingTaskUnchecked(Runnable)` instead?

Is the concern here that we might ourselves use or alter the context classloader? Because all I can see it does is save the context classloader at the time the `Runnable` is submitted and restore it for the duration of the execution of the `Runnable`. As long as we don't ourselves modify the context classloader, it shouldn't matter whether we execute our code around the "preserving" runnable (in the wrapper we're adding) or inside it.

richardstartin:
Apologies - the implementation of the wrapping method indeed does not depend on the type of the runnable, so this fix is fine. I'd just make the description and labels refer to quarkus instead of netty though.

cataphract:
> I'd just make the description and labels refer to quarkus instead of netty though.

Right I got confused. This causes problems in undertow, not netty.

PR #5652, opened 2023-07-27
[FRAPMS-3788] Openliberty jakarta integration

nayeem-kamal:
# What Does This Do
Adds instrumentation for openliberty 23+ following the changes from javax to jakarta.
# Motivation
Increases support for new versions of openliberty
# Additional Notes
Async Servlet tests will need to be added to achieve parity with the changes made to liberty-20 instrumentation.
Latest dep tests will be included in a follow up PR.


cataphract:
Please note that I've made changes to the liberty-20 module that you'll want to reflect on this PR.

Also, I found that the openliberty instrumentation has a serious shortcoming on request finish. I added this comment:

```
/**
 * XXX: {@link SRTServletResponse#finish()} is not appropriate method to look at the response
 * headers. By this time, the headers may have been cleared already. They are cleared when the
 * output is closed. There are other problems with this instrumentation. See the comments and NPEs
 * being caught. In short, the method runs too late. Maybe intercept {@link
 * SRTServletResponse#closeResponseOutput()} instead?
 */
@AutoService(Instrumenter.class)
public class ResponseFinishInstrumentation extends Instrumenter.Tracing
```

dd-smoke-tests/springboot-openliberty-23/application/src/main/java/com/openliberty/demo/DemoController.java
```diff
@@ -0,0 +1,54 @@
...
+  @Value("${server.port}")
+  private String httpPort;
+
+  public DemoController() {
+    this.users = new int[100];
+  }
+
+  @RequestMapping("/connect")
+  public String connect() {
+    int userId = new Random().nextInt(users.length);
```
am312, line 27: Do we prefer ThreadLocalRandom.current().nextInt(...) here?

dd-smoke-tests/springboot-openliberty-23/application/src/main/java/com/openliberty/demo/DemoController.java
```diff
@@ -0,0 +1,54 @@
+package com.openliberty.demo;
+
+import java.security.MessageDigest;
+import java.util.Random;
+import lombok.extern.slf4j.Slf4j;
```
am312, line 5: Looks like you're only using lombok's @slf4j in two places. You might remove those to avoid taking the lombok dependency if you want.

dd-java-agent/instrumentation/servlet/request-5/src/main/java/datadog/trace/instrumentation/servlet5/JakartaServletBlockingHelper.java
```diff
@@ -0,0 +1,86 @@
...
+import jakarta.servlet.http.HttpServletResponse;
+import java.io.IOException;
+import java.io.OutputStream;
+import java.util.Map;
+import org.slf4j.Logger;
+import org.slf4j.LoggerFactory;
+
+public class JakartaServletBlockingHelper {
+  private static final byte[] EMPTY_BYTE_ARRAY = new byte[0];
+  private static final Logger log = LoggerFactory.getLogger(ServletBlockingHelper.class);
```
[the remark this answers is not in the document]
nayeem-kamal: Fixed.

dd-smoke-tests/springboot-openliberty-23/application/src/main/java/com/openliberty/demo/DemoController.java
nayeem-kamal: This was copied from the existing openliberty smoke tests for version 20. I tried changing this to threadlocalrandom and it caused the tests to break for some reason. Rather than investigate that now I think it makes sense to leave it as it was.

dd-smoke-tests/springboot-openliberty-23/application/src/main/java/com/openliberty/demo/DemoController.java
nayeem-kamal: This was set up in the openliberty smoke tests previously so I am just using the existing `DemoController` definition. I think since this is a smoke-test it won't make much of a difference in terms of reducing the dependencies.

am312 (approved):

dougqh (commented):
Looks good to me

PR #5653, opened 2023-07-27
Add manual Gitlab job for running CI Visibility test environment

nikita-tkachenko-datadog:
# What Does This Do
Adds a manual Gitlab job that triggers CI Visibility [test-environment](https://github.com/DataDog/test-environment) run with the tracer built in the current branch.

# Motivation
Make it easier to run these external tests before tracer changes are merged.

# Additional Notes
The job is only triggered manually, so it will not interfere with the other products.

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -36,6 +36,7 @@ Thumbs.db
 # Intellij Idea #
 #################
 /.idea
+/.run
 *.iml
 *.ipr
 *.iws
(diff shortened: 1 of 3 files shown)


.gitignore
```diff
@@ -36,6 +36,7 @@ Thumbs.db
 # Intellij Idea #
 #################
 /.idea
+/.run
```
PerfectSlayer, line 39: Is it an `IDEA` config file?

PerfectSlayer (approved):

smola (approved):

.gitignore
nikita-tkachenko-datadog: Yup, this is the folder where IDEA stores custom run configuration by default

PR #5657, opened 2023-07-28
Response blocking in openliberty

cataphract:


dd-java-agent/agent-bootstrap/src/main/java/datadog/trace/bootstrap/instrumentation/decorator/http/LibertyDecoratorHolder.java
```diff
@@ -0,0 +1,9 @@
+package datadog.trace.bootstrap.instrumentation.decorator.http;
+
+import datadog.trace.bootstrap.instrumentation.decorator.HttpServerDecorator;
+
+public class LibertyDecoratorHolder {
```
manuel-alvarez-alvarez, line 5: Is this holder needed?, seems it's not used

dd-java-agent/instrumentation/liberty-20/src/main/java/datadog/trace/instrumentation/liberty20/ArrayOfTypeMatcher.java
```diff
@@ -0,0 +1,26 @@
+package datadog.trace.instrumentation.liberty20;
+
+import static datadog.trace.agent.tooling.bytebuddy.matcher.NameMatchers.named;
+
+import datadog.trace.agent.tooling.bytebuddy.matcher.NameMatchers;
+import net.bytebuddy.description.NamedElement;
+import net.bytebuddy.description.type.TypeDefinition;
+import net.bytebuddy.matcher.ElementMatcher;
+
+public class ArrayOfTypeMatcher<T extends TypeDefinition>
```
manuel-alvarez-alvarez, line 10: Looks to me that this might be useful in other cases and can be moved alongside the other matchers

dd-java-agent/instrumentation/liberty-20/src/main/java/datadog/trace/instrumentation/liberty20/LibertyServerInstrumentation.java
```diff
@@ -98,6 +111,15 @@ public static class HandleRequestAdvice {
       request.setAttribute(CorrelationIdentifier.getTraceIdKey(), GlobalTracer.get().getTraceId());
       request.setAttribute(CorrelationIdentifier.getSpanIdKey(), GlobalTracer.get().getSpanId());
 
+      if (ActiveSubsystems.APPSEC_ACTIVE) {
+        ContextStore store =
```
manuel-alvarez-alvarez, line 115: Out of curiosity, is this is a performance improvement to speed up access to the span using the context store instead of the TL?

dd-java-agent/instrumentation/liberty-20/src/main/java/datadog/trace/instrumentation/liberty20/RequestMessageFromServletRequestHelper.java
```diff
@@ -0,0 +1,41 @@
...
+import com.ibm.websphere.servlet.request.extended.IRequestExtended;
+import com.ibm.ws.webcontainer.srt.SRTServletRequest;
+import com.ibm.wsspi.http.HttpRequest;
+import java.lang.invoke.MethodHandle;
+import java.lang.invoke.MethodHandles;
+import java.lang.reflect.Field;
+import java.util.concurrent.ConcurrentHashMap;
+
+public class RequestMessageFromServletRequestHelper {
+  static ConcurrentHashMap<Class<?>, MethodHandle> FIELD_CACHE = new ConcurrentHashMap<>();
```
manuel-alvarez-alvarez, line 13: Maybe `final`?

dd-java-agent/instrumentation/liberty-20/src/main/java/datadog/trace/instrumentation/liberty20/WebAppHandleExceptionInstrumentation.java
```diff
@@ -0,0 +1,56 @@
...
+            .and(named("handleException"))
+            .and(takesArguments(4))
+            .and(takesArgument(0, Throwable.class))
+            .and(returns(void.class)),
+        WebAppHandleExceptionInstrumentation.class.getName() + "$HandleExceptionAdvice");
+  }
+
+  static class HandleExceptionAdvice {
+    @Advice.OnMethodEnter(suppress = Throwable.class, skipOn = Advice.OnNonDefaultValue.class)
+    static boolean /* skip */ before(@Advice.Argument(0) Throwable throwable_) {
```
manuel-alvarez-alvarez, line 45: Does the skip make sense without an `OnMethodExit`?

manuel-alvarez-alvarez (approved):
Take my review with a grain of salt, I'm no blocking expert and just entered the PR to learn a bit, but looks good to me.

dd-java-agent/agent-bootstrap/src/main/java/datadog/trace/bootstrap/instrumentation/decorator/http/LibertyDecoratorHolder.java
cataphract: Thanks, it's a leftover from an earlier version.

dd-java-agent/instrumentation/liberty-20/src/main/java/datadog/trace/instrumentation/liberty20/WebAppHandleExceptionInstrumentation.java
cataphract: Sure. The return method on the advice serves two purposes. With `skipOn` it can be used to skip the method body and it can be retrieved on `OnMethodExit` with `@Advice.Enter`. In this case I'm only skipping its body; i don't need to do anything on an `OnMethodExit` like setting the return value (this is a `void` method being instrumented).

dd-java-agent/instrumentation/liberty-20/src/main/java/datadog/trace/instrumentation/liberty20/ArrayOfTypeMatcher.java
cataphract: I looked into this but it doesn't seem to fit the matchers in package `package datadog.trace.agent.tooling.bytebuddy.matcher`. They're all rather specialized, including a set of matchers behind an interface there's 3 implementations of.

PR #5669, opened 2023-08-01
Archive parametric tests results before upload

smola:
# What Does This Do
Archive parametric tests results before upload, just like we do with system-tests.

# Motivation

# Additional Notes


bantonsson (approved):
👍🏻

PR #5673, opened 2023-08-02
Add support for OpenTelemetry Context.makeCurrent()

PerfectSlayer:
# What Does This Do

This PR brings a limited support to `Context.makeCurrent()` as a way to activate a span.

There is two ways to make a span active using OTel, explicitly using `Span.makeCurrent()`, or through the `Context` API:
```java
var context = Context.root().with(span);
var scope = context.makeCurrent();
```

This PR will bring support for the second case.

# Motivation

Some library, like clj-otel a Clojure OTel integration is relying on `Context` API to manage active/current spans.
Without such support, our tracer and its auto-instrumentations are not able to find the active spans, creating new traces.

# Additional Notes

Providing support for OTel `Context` can't go so far with our current tracer design.


dd-java-agent/instrumentation/opentelemetry/opentelemetry-1.4/src/test/groovy/OpenTelemetry14Test.groovy
```diff
@@ -1,25 +1,12 @@
...
-import io.opentelemetry.context.propagation.TextMapSetter
 import spock.lang.Subject
 
-import javax.annotation.Nullable
-
-import static datadog.trace.bootstrap.instrumentation.api.ScopeSource.MANUAL
-import static io.opentelemetry.api.trace.StatusCode.ERROR
-import static io.opentelemetry.api.trace.StatusCode.OK
-import static io.opentelemetry.api.trace.StatusCode.UNSET
+import static io.opentelemetry.api.trace.StatusCode.*
```
bantonsson, line 9: Wildcard import...

bantonsson (approved):

PR #5674, opened 2023-08-03
Innovation week - IAST Hadcoded secret detection

jandro996:
# What Does This Do

# Motivation

# Additional Notes

### DIFF
--- a/dd-java-agent/instrumentation/iast-instrumenter/src/main/java/datadog/trace/instrumentation/iastinstrumenter/IastSecretDetectionListener.java
+++ b/dd-java-agent/instrumentation/iast-instrumenter/src/main/java/datadog/trace/instrumentation/iastinstrumenter/IastSecretDetectionListener.java
@@ -7,6 +7,9 @@ import datadog.trace.api.iast.sink.SecretsModule;
 import javax.annotation.Nonnull;
 import net.bytebuddy.description.type.TypeDescription;
 
+import java.util.HashSet;
+import java.util.Set;
+
 public class IastSecretDetectionListener implements Advices.Listener {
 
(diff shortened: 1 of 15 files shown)


dd-java-agent/instrumentation/iast-instrumenter/src/main/java/datadog/trace/instrumentation/iastinstrumenter/IastSecretDetectionListener.java
```diff
@@ -0,0 +1,30 @@
...
+      final @Nonnull TypeDescription type,
+      final @Nonnull ConstantPool pool,
+      final @Nonnull byte[] classFile) {
+    final SecretsModule iastModule = InstrumentationBridge.HARDCODED_SECRET;
+    if (iastModule != null) {
+      for (int index = 1; index < pool.getCount(); index++) {
+        if (pool.getType(index) == ConstantPool.CONSTANT_STRING_TAG) {
+          final int literalIndex = pool.readUnsignedShort(pool.getOffset(index));
+          final String literal = pool.readUTF8(pool.getOffset(literalIndex));
+          iastModule.onStringLiteral(literal, type.getName(), classFile);
```
manuel-alvarez-alvarez, line 25: You should not do this, since the next step is an analysis of the bytecode to match for line numbers ... you should extract all literals from the constant pool and then continue further with a `Set<String>`

jandro996: Thanks for the advice! done

jandro996:
We will resume this task on 2023 Q4

PR #5677, opened 2023-08-03
Experiment with TelemetryService lazy data init

ygree:
# What Does This Do

Try to optimize startup by doing lazy init for TelemetryService

# Motivation

Perf degradation investingation in #5640 confirmed in #5675

# Additional Notes


ygree:
Closed in favor of #5678

PR #5680, opened 2023-08-03
Telemetry v2 and message-batching

ygree:
# What Does This Do

Adjustments for Telemetry V2 protocol.
Add app-started product info (appsec, profiler).
Retry app-started for limited number of times (3), and every heartbeat interval until it's been successfully sent.
Message-batching and message buffering for retries on next heartbeat interval as required by RFC. When sending a request fails it keeps unsent data till the next heartbeat interval for a retry. If it fails again we discard data.

# Motivation

Telemetry V2 implementation.

# Additional Notes

System-tests use this version: https://github.com/DataDog/system-tests/pull/1551

### DIFF
--- a/.circleci/config.continue.yml.j2
+++ b/.circleci/config.continue.yml.j2
@@ -30,7 +30,7 @@ debugger_modules: &debugger_modules "dd-java-agent/agent-debugger|dd-java-agent/
 profiling_modules: &profiling_modules "dd-java-agent/agent-profiling"
 
 # TODO do not merge this line to master
-default_system_tests_commit: &default_system_tests_commit 7655009d9d26d28ff387f009d20c0d41154bde40
+default_system_tests_commit: &default_system_tests_commit ef1b79ee4ee84bb7efb650139157f0dfbf4aa476
 
 parameters:
   nightly:
--- a/.circleci/config.continue.yml.j2
+++ b/.circleci/config.continue.yml.j2
@@ -29,6 +29,7 @@ instrumentation_modules: &instrumentation_modules "dd-java-agent/instrumentation
 debugger_modules: &debugger_modules "dd-java-agent/agent-debugger|dd-java-agent/agent-bootstrap|dd-java-agent/agent-builder|internal-api|communication|dd-trace-core"
 profiling_modules: &profiling_modules "dd-java-agent/agent-profiling"
 
+# TODO do not merge this line to master
 default_system_tests_commit: &default_system_tests_commit 7655009d9d26d28ff387f009d20c0d41154bde40
 
 parameters:
--- a/.circleci/config.continue.yml.j2
+++ b/.circleci/config.continue.yml.j2
@@ -29,7 +29,7 @@ instrumentation_modules: &instrumentation_modules "dd-java-agent/instrumentation
 debugger_modules: &debugger_modules "dd-java-agent/agent-debugger|dd-java-agent/agent-bootstrap|dd-java-agent/agent-builder|internal-api|communication|dd-trace-core"
 profiling_modules: &profiling_modules "dd-java-agent/agent-profiling"
 
-default_system_tests_commit: &default_system_tests_commit 3598f2583cd5c25326cd7a5c01097374d94c1993
+default_system_tests_commit: &default_system_tests_commit 7655009d9d26d28ff387f009d20c0d41154bde40
 
 parameters:
   nightly:
--- a/telemetry/src/main/java/datadog/telemetry/EventSink.java
+++ b/telemetry/src/main/java/datadog/telemetry/EventSink.java
@@ -7,6 +7,11 @@ import datadog.telemetry.api.LogMessage;
 import datadog.telemetry.api.Metric;
 import datadog.telemetry.dependency.Dependency;
 
+/**
+ * A unified interface for telemetry event sink. It is used to buffer events polled from the queues
+ * to reattempt next sending attempt. A NOOP implementation is used to discard event on the next
+ * failing attempt.
+ */
 interface EventSink {
   void addConfigChangeEvent(ConfigChange event);
 
@@ -20,12 +25,10 @@ interface EventSink {
 
   void addLogMessageEvent(LogMessage event);
 
-  static EventSink noop() {
-    return Noop.INSTANCE;
-  }
+  EventSink NOOP = new Noop();
 
-  enum Noop implements EventSink {
-    INSTANCE;
+  class Noop implements EventSink {
+    private Noop() {}
 
     @Override
     public void addConfigChangeEvent(ConfigChange event) {}
--- a/telemetry/src/main/java/datadog/telemetry/EventSink.java
+++ b/telemetry/src/main/java/datadog/telemetry/EventSink.java
@@ -1,0 +1,48 @@
+package datadog.telemetry;
+
+import datadog.telemetry.api.ConfigChange;
+import datadog.telemetry.api.DistributionSeries;
+import datadog.telemetry.api.Integration;
+import datadog.telemetry.api.LogMessage;
+import datadog.telemetry.api.Metric;
+import datadog.telemetry.dependency.Dependency;
+
+interface EventSink {
+  void addConfigChangeEvent(ConfigChange event);
+
+  void addIntegrationEvent(Integration event);
+
+  void addDependencyEvent(Dependency event);
+
+  void addMetricEvent(Metric event);
+
+  void addDistributionSeriesEvent(DistributionSeries event);
+
+  void addLogMessageEvent(LogMessage event);
+
+  static EventSink noop() {
+    return Noop.INSTANCE;
+  }
+
+  enum Noop implements EventSink {
+    INSTANCE;
+
+    @Override
+    public void addConfigChangeEvent(ConfigChange event) {}
+
+    @Override
+    public void addIntegrationEvent(Integration event) {}
+
+    @Override
+    public void addDependencyEvent(Dependency event) {}
+
+    @Override
+    public void addMetricEvent(Metric event) {}
+
+    @Override
+    public void addDistributionSeriesEvent(DistributionSeries event) {}
+
+    @Override
+    public void addLogMessageEvent(LogMessage event) {}
+  }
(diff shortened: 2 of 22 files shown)


telemetry/src/main/java/datadog/telemetry/TelemetryRunnable.java
```diff
@@ -54,6 +56,22 @@ public void run() {
     // Ensure that Config has been initialized, so ConfigCollector can collect all settings first.
     Config.get();
 
+    collectConfigChanges();
+
+    int attempt = 0;
+    while (!Thread.interrupted()
+        && !telemetryService.sendAppStartedEvent()
+        && attempt < MAX_APP_STARTED_RETRIES) {
+      attempt += 1;
```
bantonsson, line 65: Won't this code just try to send the app started event 3 times in quick succession? Wouldn't it be better to put the sending of the app started event as part of the `while` loop below, and retry it before `mainLLoopIteration()` if it has failed before, and skip `mainLoopIteration()` if it fails (for 3 tries of course)?

telemetry/src/main/java/datadog/telemetry/api/RequestType.java
```diff
@@ -2,14 +2,22 @@
...
   APP_STARTED("app-started"),
-  APP_CLIENT_CONFIGURATION_CHANGE("app-client-configuration-change"),
   APP_DEPENDENCIES_LOADED("app-dependencies-loaded"),
   APP_INTEGRATIONS_CHANGE("app-integrations-change"),
+  APP_CLIENT_CONFIGURATION_CHANGE("app-client-configuration-change"),
+  // app-product-change
   APP_HEARTBEAT("app-heartbeat"),
+  // app-extended-heartbeat
+  // message-batch
   APP_CLOSING("app-closing"),
```
bantonsson, line 12: What does the `// message-batch` comment meant in this place?

telemetry/src/main/java/datadog/telemetry/TelemetryRequestState.java
```diff
@@ -0,0 +1,410 @@
...
+  private static final AtomicLong SEQ_ID = new AtomicLong();
+  private static final String TELEMETRY_NAMESPACE_TAG_TRACER = "tracers";
+  private final Buffer body = new Buffer();
+
+  private final JsonWriter bodyWriter = JsonWriter.of(body);
+  private final RequestType requestType;
+  private final Request.Builder requestBuilder;
+  private final boolean debug;
+
+  private enum CommonData {
```
bantonsson, line 46: Another `enum` singleton.

bantonsson (commented):
Looks good overall. Some comments and questions.

telemetry/src/main/java/datadog/telemetry/EventSink.java
```diff
@@ -0,0 +1,48 @@
...
+
+  void addDistributionSeriesEvent(DistributionSeries event);
+
+  void addLogMessageEvent(LogMessage event);
+
+  static EventSink noop() {
+    return Noop.INSTANCE;
+  }
+
+  enum Noop implements EventSink {
```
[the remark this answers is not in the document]
ygree: Good to know!

telemetry/src/main/java/datadog/telemetry/TelemetryRunnable.java
ygree: I think that's a good idea not to send telemetry events after an unsuccessful attempt sending app-started. As well as reattempting app-started next cycle.

telemetry/src/main/java/datadog/telemetry/api/RequestType.java
ygree: Just forgot to remove it

telemetry/src/main/java/datadog/telemetry/EventSource.java
```diff
@@ -0,0 +1,136 @@
...
+      this.integrationQueue = integrationQueue;
+      this.dependencyQueue = dependencyQueue;
+      this.metricQueue = metricQueue;
+      this.distributionSeriesQueue = distributionSeriesQueue;
+      this.logMessageQueue = logMessageQueue;
+    }
+
+    @Override
+    public boolean isEmpty() {
+      return !hasConfigChangeEvent()
```
PerfectSlayer, line 68: Any rational for not checking the `configChangeQueue` directly?
Or if you only use the `hasXXXEvent()`, you can implement `isEmpty()` as default method in the interface itself.

telemetry/src/main/java/datadog/telemetry/TelemetryService.java
```diff
@@ -43,37 +36,37 @@ public class TelemetryService {
...
   private boolean openTelemetryIntegrationEnabled;
 
-  public TelemetryService(final OkHttpClient okHttpClient, final HttpUrl httpUrl) {
-    this(new HttpClient(okHttpClient), httpUrl);
-  }
-
-  public TelemetryService(final HttpClient httpClient, final HttpUrl httpUrl) {
-    this(httpClient, MAX_ELEMENTS_PER_REQUEST, MAX_DEPENDENCIES_PER_REQUEST, httpUrl);
+  public TelemetryService(
+      final OkHttpClient okHttpClient, final HttpUrl httpUrl, final boolean debug) {
```
PerfectSlayer, line 54: Can be useful to document the `debug` is not related to the service itself but for marking `TelemetryRequest(State)` as debug to the backend.

telemetry/src/test/groovy/datadog/telemetry/TestHttpClient.groovy
```diff
@@ -76,40 +73,49 @@ class TestHttpClient extends HttpClient {
...
-      assert request.header('DD-Client-Library-Language') == 'jvm'
-      assert request.header('DD-Client-Library-Version') == TracerVersion.TRACER_VERSION
-      assert request.header('DD-Telemetry-API-Version') == 'v1'
-      assert request.header('DD-Telemetry-Request-Type') == requestType.toString()
+      assert this.request.header('Content-Type') == 'application/json; charset=utf-8'
+      assert this.request.header('DD-Client-Library-Language') == 'jvm'
+      assert this.request.header('DD-Client-Library-Version') == TracerVersion.TRACER_VERSION
+      assert this.request.header('DD-Telemetry-API-Version') == 'v2'
+      assert this.request.header('DD-Telemetry-Request-Type') == requestType.toString()
+      //      assert this.request.header('Content-Length') ==
```
PerfectSlayer, line 90: Is the test missing?

PerfectSlayer (commented):
Nice addition 👍

telemetry/src/main/java/datadog/telemetry/EventSource.java
```diff
@@ -0,0 +1,131 @@
...
+
+import datadog.telemetry.api.ConfigChange;
+import datadog.telemetry.api.DistributionSeries;
+import datadog.telemetry.api.Integration;
+import datadog.telemetry.api.LogMessage;
+import datadog.telemetry.api.Metric;
+import datadog.telemetry.dependency.Dependency;
+import java.util.Queue;
+
+interface EventSource {
```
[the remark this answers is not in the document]
ygree: It depends on the implementation. `BufferedEvents` will through an exception and `EventSource.Queued` will return null. All the `nextXXXXEvent()` methods can only be called if `hasXXXXEvent()` returns `true`, similar to `Iterator`.

telemetry/src/test/groovy/datadog/telemetry/TestHttpClient.groovy
ygree: Content Length depends on the payload which in turn depends on some OS variables. I'll just change this assertion to be greater than zero, so the tests passes in any environment.

telemetry/src/main/java/datadog/telemetry/EventSource.java
ygree: Will move it to EventSource interface as a default method

.circleci/config.continue.yml.j2
```diff
@@ -29,7 +29,7 @@ instrumentation_modules: &instrumentation_modules "dd-java-agent/instrumentation
 debugger_modules: &debugger_modules "dd-java-agent/agent-debugger|dd-java-agent/agent-bootstrap|dd-java-agent/agent-builder|internal-api|communication|dd-trace-core"
 profiling_modules: &profiling_modules "dd-java-agent/agent-profiling"
 
-default_system_tests_commit: &default_system_tests_commit 3598f2583cd5c25326cd7a5c01097374d94c1993
+default_system_tests_commit: &default_system_tests_commit 7655009d9d26d28ff387f009d20c0d41154bde40
```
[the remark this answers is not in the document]
ygree: Good point! I'll add a TODO here not to forget about it.

telemetry/src/main/java/datadog/telemetry/EventSource.java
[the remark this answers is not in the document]
ygree: The contract is that it next* method shouldn't be called without checking has* first. It's needed because serialized need to check presence of elements without extracting them from the corresponding queue.

telemetry/src/main/java/datadog/telemetry/TelemetryService.java
ygree: I've also introduced a config setting to be able to enable debug mode via configuration

telemetry/src/main/java/datadog/telemetry/EventSource.java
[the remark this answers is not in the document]
PerfectSlayer: Yes, I recognized the Iterator-like pattern.

My two cents would be to push it further by having a consistent behavior, like throwing a `NoSuchElementException`, and enforcing the _check `hasXXX()` first behavior_ in the Javadoc. But that's nitpicking. You can resolve the comment if you are fine with the current behavior 😉

PerfectSlayer (approved):
Except for the system-tests change, LGTM 👍

bantonsson (approved):

PR #5686, opened 2023-08-07
WIP: Scope context R&D week

PerfectSlayer:
# What Does This Do

Introduce ScopeContext and Baggage.
Introduce uniform API for context propagation.

# Motivation

# Additional Notes

Continuation of past R&D weeks:

- #4795 : the original idea implementation
- #5210 : put the concept in application -- some baggage / context upgrade were not ported to this RP

Superseed by: #7343


dd-trace-core/src/main/java/datadog/trace/core/scopemanager/ScopeContext.java
```diff
@@ -0,0 +1,132 @@
...
+    } else if (key == CONTEXT_KEY) {
+      return (T) this.span;
+    } else if (key == BAGGAGE_KEY) {
+      return (T) this.baggage;
+    } else if (this.store != null) {
+      for (int index = 0; index < this.store.length; index += 2) {
+        if (this.store[index] == key) {
+          return (T) this.store[index + 1];
+        }
+      }
```
richardstartin, line 82: What about just storing the values in the array (including the span and the baggage) and making the user register a key before using it for setting/lookups, which contains an offset into the array. `CONTEXT_KEY` and `BAGGAGE_KEY` can be pre-registered. This whole thing reduces to `return key.offset < store.length ? store[key.offset] : null`, and you don't waste space per context storing keys.

internal-api/src/main/java/datadog/trace/bootstrap/instrumentation/api/ContextKey.java
```diff
@@ -0,0 +1,30 @@
+package datadog.trace.bootstrap.instrumentation.api;
+
+/**
+ * The key to store and retrieve values from ScopedContext key-value storage.
+ *
+ * @param <T> The type of the value to store.
+ */
+public class ContextKey<T> {
```
richardstartin, line 8: If you maintain an atomic counter starting from zero, and bump it each time you allocate one, you can make the key store an offset (or global value numbering), which can be used for cheap lookups as mentioned above.

bantonsson: I agree that an _offset_ `ContextKey<T>` would be way more efficient.

PerfectSlayer: Thank you both for the feedback, that could be a good improvement 👍 

Sadly, I feel it is still not ready to make any performance test.
The whole DSM stuff is still in my way and I think I will end up spending the week to clean the tracer from it...
Hopefully, it will be the last R&D week lost as I already lost the last one cleaning the DSM context extraction, and now it’s time to clean the injection part.

I might end up working on context on 2024 at this pace 😒

DDJavierSantos (approved):

PR #5691, opened 2023-08-07
Add configuration option to allow Jax-RS exceptions to not be handled as errors

bm1549:
# What Does This Do
Adds a new configuration option via `DD_TRACE_JAX_RS_EXCEPTION_AS_ERROR_ENABLED` which, when set to `false` forces exceptions from JAX-RS to not be treated as errors. This does not drop the underlying exception if it exists

# Motivation
This is helpful for situations for situations where a 3xx level response returned triggers a custom exception in the underlying code since the span should not be considered an error

# Additional Notes

### DIFF
--- a/dd-java-agent/instrumentation/jax-rs-client-2.0/connection-error-handling-jersey/src/main/java/org/glassfish/jersey/client/WrappingResponseCallback.java
+++ b/dd-java-agent/instrumentation/jax-rs-client-2.0/connection-error-handling-jersey/src/main/java/org/glassfish/jersey/client/WrappingResponseCallback.java
@@ -1,5 +1,6 @@
 package org.glassfish.jersey.client;
 
+import datadog.trace.api.Config;
 import datadog.trace.bootstrap.instrumentation.api.AgentSpan;
 import datadog.trace.instrumentation.jaxrs.ClientTracingFilter;
 import javax.ws.rs.ProcessingException;
@@ -30,8 +31,14 @@ public final class WrappingResponseCallback implements ResponseCallback {
     final Object prop = request.getProperty(ClientTracingFilter.SPAN_PROPERTY_NAME);
     if (prop instanceof AgentSpan) {
       final AgentSpan span = (AgentSpan) prop;
-      span.setError(true);
       span.addThrowable(error);
+
+      @SuppressWarnings("deprecation")
+      final boolean isJaxRsExceptionAsErrorEnabled = Config.get().isJaxRsExceptionAsErrorEnabled();
+      if (!isJaxRsExceptionAsErrorEnabled) {
+        span.setError(false);
+      }
+
       span.finish();
     }
   }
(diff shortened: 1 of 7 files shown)


dd-java-agent/instrumentation/jax-rs-client-2.0/connection-error-handling-jersey/src/main/java/org/glassfish/jersey/client/WrappingResponseCallback.java
```diff
@@ -30,8 +31,14 @@ public static void handleProcessingException(ClientRequest request, ProcessingEx
...
     if (prop instanceof AgentSpan) {
       final AgentSpan span = (AgentSpan) prop;
-      span.setError(true);
       span.addThrowable(error);
+
+      @SuppressWarnings("deprecation")
+      final boolean isJaxRsExceptionAsErrorEnabled = Config.get().isJaxRsExceptionAsErrorEnabled();
+      if (!isJaxRsExceptionAsErrorEnabled) {
+        span.setError(false);
+      }
```
bantonsson, line 40: Can't this if statement be reduced to `span.setError(Confgi.get().isJaxRsExceptionAsErrorEnabled());`?

bm1549: It could be, but then SuppressWarnings would need to be on the function, which I didn't want. Let me know if that would be preferred

bantonsson: That's a good reason to have the boolean.

You can still skip the if and use `span.setError(isJaxRsExceptionAsErrorEnabled);`.

bantonsson (approved):

dd-java-agent/instrumentation/jax-rs-client-2.0/connection-error-handling-jersey/src/test/groovy/WrappingResponseCallbackTest.groovy
```diff
@@ -0,0 +1,46 @@
...
+import org.glassfish.jersey.internal.MapPropertiesDelegate
+
+class WrappingResponseCallbackTest extends AgentTestRunner {
+  def "handleProcessingException properly utilizes the config"() {
+    setup:
+    if (jaxRsExceptionAsErrorEnabled != null) {
+      injectSysConfig(JAX_RS_EXCEPTION_AS_ERROR_ENABLED, "$jaxRsExceptionAsErrorEnabled")
+    }
+    def testSpan = TEST_TRACER.buildSpan("testInstrumentation", "testSpan").start()
+    def props = Map.of(ClientTracingFilter.SPAN_PROPERTY_NAME, testSpan)
```
bantonsson, line 19: It seems this `Map.of` is a Java 9 feature that fails the Java 8 tests. You could to it the `groovy` way instead:
```
def props = [(ClientTracingFilter.SPAN_PROPERTY_NAME): testSpan]
```

bantonsson:
@bm1549 I just realized that there is a similar instrumentation for resteasy that would also need this flag if I'm not misstaken.

PR #5692, opened 2023-08-08
Fix sampling when log probe is evaluation at Exit

jpbempel:
# What Does This Do
Change the instrumentation to call in any cases readyToCapture method that invoke sampling except for SpanDecorationProbe.

# Motivation
With evaluateAt == EXIT for a log probe, sampling was not invoked. 
# Additional Notes


ojung (approved):
Looks good. Tested in staging and it seems to have the desired effect.

PR #5698, opened 2023-08-09
Report dd-trace-java and its dependencies to telemetry

smola:
# What Does This Do

* Remove the telemetry check we previously added to avoid reporting dd-trace-java vendored dependencies to telemetry.
* Amend the exclusion of `META-INF/maven` to not include `pom.properties` (introduced in 5346d91d34c), which we use to report dependencies to telemetry.

# Motivation
While this might be sometimes confusing, we want to report the runtime dependencies as accurately as possible, include those that are introduced by the tracer.

# Additional Notes
* This leads to a ~0.5ms regression in telemetry startup, presumably because now loading classes from the tracer hits the regular codepath that hits `DDCache#computeIfAbsent` for `seenDomains`.

### DIFF
--- a/dd-java-agent/build.gradle
+++ b/dd-java-agent/build.gradle
@@ -30,7 +30,10 @@ targetCompatibility = JavaVersion.VERSION_1_7
 ext.generalShadowJarConfig = {
   mergeServiceFiles()
 
-  exclude '**/META-INF/maven/'
+  // Remove some cruft from the final jar.
+  // These patterns should NOT include **/META-INF/maven/**/pom.properties, which is
+  // used to report our own dependencies.
+  exclude '**/META-INF/maven/**/pom.xml'
   exclude '**/META-INF/proguard/'
   exclude '**/META-INF/*.kotlin_module'
   exclude '**/module-info.class'
--- a/dd-java-agent/build.gradle
+++ b/dd-java-agent/build.gradle
@@ -30,6 +30,8 @@ targetCompatibility = JavaVersion.VERSION_1_7
 ext.generalShadowJarConfig = {
   mergeServiceFiles()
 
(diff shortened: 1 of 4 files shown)


dd-java-agent/build.gradle
```diff
@@ -30,7 +30,10 @@ targetCompatibility = JavaVersion.VERSION_1_7
 ext.generalShadowJarConfig = {
   mergeServiceFiles()
 
-  exclude '**/META-INF/maven/'
+  // Remove some cruft from the final jar.
+  // These patterns should NOT include **/META-INF/maven/**/pom.properties, which is
+  // used to report our own dependencies.
+  exclude '**/META-INF/maven/**/pom.xml'
```
PerfectSlayer, line 36: I am a little rusty but I think I excluded the whole `META-INF/maven/` folder because when merging archive, it produces duplicate entries, meaning multiple `pom.properties` files entry will be added to the same archive.

It created a bogus archive and kept only one random maven configuration.
How did you fix this issue?

smola: I haven't observed this behaviour. I tried running with `duplicatesStrategy = DuplicatesStrategy.FAIL` on the shadowJar conf and it build did not fail. I didn't find anything wrong when manually inspecting the jar.

When this happened, did it completely correupt the archive (as in, the jar did not run at all)?

smola: Added `duplicatesStrategy = DuplicatesStrategy.FAIL` to ensure we do not accidentally break this in the future.

PerfectSlayer: No, it did not correct the archive but left it with non deterministic content. Not sure metadata would be useful in such case.
If you can't reproduce, it's great. I can approve the PR 👍

PerfectSlayer (approved):

PR #5701, opened 2023-08-09
add _dd.base_service to disambiguate service map

amarziali:
# What Does This Do

When the service name is different to the configured one (DD_SERVICE) we add a tag `_dd.base_service` that's used internally to improve the service map

# Motivation

# Additional Notes

### DIFF
--- a/dd-trace-core/src/main/java/datadog/trace/core/DDSpanContext.java
+++ b/dd-trace-core/src/main/java/datadog/trace/core/DDSpanContext.java
@@ -2,7 +2,6 @@ package datadog.trace.core;
 
 import static datadog.trace.api.cache.RadixTreeCache.HTTP_STATUSES;
 
-import datadog.trace.api.Config;
 import datadog.trace.api.DDTags;
 import datadog.trace.api.DDTraceId;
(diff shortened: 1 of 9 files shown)


dd-trace-core/src/main/java/datadog/trace/core/DDSpanContext.java
```diff
@@ -87,6 +88,8 @@ public class DDSpanContext
   private static final TagsPostProcessor postProcessor =
       new PostProcessorChain(
           new PeerServiceCalculator(),
+          new BaseServiceAdder(
+              Config.get().isServiceNameSetByUser() ? Config.get().getServiceName() : null),
```
devinsba, line 92: Do we not want the default service name to work here too? IE: the jar name

devinsba (approved):

dd-trace-core/src/main/java/datadog/trace/core/DDSpanContext.java
amarziali: yes actually I put this to limit core test refactoring. I changed the approach and now it's working also for the default service name. Thanks for having spotted it out

PR #5705, opened 2023-08-09
Clean up DSM API from Core Tracer and introduce a new dedicate Propagation implementation 

PerfectSlayer:
# What Does This Do

This PR tries to separate both DataStreamMonitoring (DSM) concerns and context propagation capabilities from the core tracer.

# Motivation

Other than providing a better design and separation, this change is a preparation work to [a better context tracking API for the core module](https://github.com/DataDog/dd-trace-java/pull/5686). As DSM heavily interacts with context (through its pathway context and tags), putting it aside with clear separation makes easier to introduce a proper context API.
Similarly, creating a proper propagation implementation, that is responsible of all extractors and injectors, makes easier to refactor and generify them later.

Here are the high-level changes the PR introduce:

* Move `Continuation capture()` from `AgentPropagation` API to an helper method in `AgentTracer`.

* `TracerAPI` no more extends `AgentPropagation` API:
  TracerAPI and Propagation implementations were merged while TracerAPI offered a `AgentPropagation propagate()` getter. 
  `CoreTracer` will now return a (new) `CorePropagation` instance when calling `propagate()` 
* `TracerAPI` no more extends `DataStreamCheckpointer`:
  CoreTracer and DataStreamCheckpointer implementations were merged while DataStreamCheckpointer only relies on DSM implementation.
  `DataStreamMonitoring` will now inherits from `DataStreamCheckpointer`, moving the implementation of checkpointer into DSM implementation rather than Tracer implementation.
* Move `void setDataStreamCheckpoint(span, sortedTags, defaultTimestamp)` from to `TracerAPI` to `DataStreamMonitoring` API.

* Introduce `CorePropagation` as the `AgentPropagation` implementation.
  Its long term goal is to be able to manage the generic context propagation using generic injectors / extractors.
  For now, it gathers all the existing propagators even if they don't have the same interface yet.
* Introduce `DataStreamContextInjector` to inject DSM specific context

The whole PR is mainly moving code parts. Only DSM received implementation improvement for its injector (as I did for extractor during the last R&D week).

To help visualize, here are two class diagrams about the changes: the first as before changes, the last as after changes.

##### Before
before

##### After
after

# Additional Notes

Moving out context propagation and DSM implementation from core tracer also points out that those core parts were not tested as `CoreTracer` is excluded from code coverage check.
I added tests for the new context propagation implementation but will leave the DSM tests to its team.

image

### DIFF
--- a/dd-java-agent/agent-bootstrap/src/main/java/datadog/trace/bootstrap/instrumentation/decorator/HttpServerDecorator.java
+++ b/dd-java-agent/agent-bootstrap/src/main/java/datadog/trace/bootstrap/instrumentation/decorator/HttpServerDecorator.java
@@ -143,9 +143,7 @@ public abstract class HttpServerDecorator<REQUEST, CONNECTION, RESPONSE, REQUEST
     }
     AgentPropagation.ContextVisitor<REQUEST_CARRIER> getter = getter();
     if (null != carrier && null != getter) {
-      tracer()
-          .getDataStreamsMonitoring()
-          .setDataStreamCheckpoint(span, SERVER_PATHWAY_EDGE_TAGS, 0);
+      tracer().getDataStreamsMonitoring().setCheckpoint(span, SERVER_PATHWAY_EDGE_TAGS, 0);
     }
     return span;
   }
--- a/dd-java-agent/agent-bootstrap/src/main/java/datadog/trace/bootstrap/instrumentation/decorator/HttpServerDecorator.java
+++ b/dd-java-agent/agent-bootstrap/src/main/java/datadog/trace/bootstrap/instrumentation/decorator/HttpServerDecorator.java
@@ -143,7 +143,9 @@ public abstract class HttpServerDecorator<REQUEST, CONNECTION, RESPONSE, REQUEST
     }
     AgentPropagation.ContextVisitor<REQUEST_CARRIER> getter = getter();
     if (null != carrier && null != getter) {
-      tracer().setDataStreamCheckpoint(span, SERVER_PATHWAY_EDGE_TAGS, 0);
+      tracer()
+          .getDataStreamsMonitoring()
+          .setDataStreamCheckpoint(span, SERVER_PATHWAY_EDGE_TAGS, 0);
     }
     return span;
   }
--- a/internal-api/build.gradle
+++ b/internal-api/build.gradle
@@ -47,7 +47,6 @@ excludedClassesCoverage += [
   "datadog.trace.bootstrap.instrumentation.api.Tags",
   "datadog.trace.bootstrap.instrumentation.api.CommonTagValues",
   "datadog.trace.bootstrap.instrumentation.api.AgentTracer.NoopAgentPropagation",
-  "datadog.trace.bootstrap.instrumentation.api.AgentPropagation",
   "datadog.trace.bootstrap.instrumentation.api.AgentTracer",
   "datadog.trace.bootstrap.instrumentation.api.AgentTracer.NoopContext",
   "datadog.trace.bootstrap.instrumentation.api.InstrumentationTags",
--- a/internal-api/build.gradle
+++ b/internal-api/build.gradle
@@ -47,6 +47,7 @@ excludedClassesCoverage += [
   "datadog.trace.bootstrap.instrumentation.api.Tags",
   "datadog.trace.bootstrap.instrumentation.api.CommonTagValues",
   "datadog.trace.bootstrap.instrumentation.api.AgentTracer.NoopAgentPropagation",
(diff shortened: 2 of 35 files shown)


dd-java-agent/agent-bootstrap/src/main/java/datadog/trace/bootstrap/instrumentation/decorator/HttpServerDecorator.java
```diff
@@ -143,7 +143,9 @@ public AgentSpan startSpan(
     }
     AgentPropagation.ContextVisitor<REQUEST_CARRIER> getter = getter();
     if (null != carrier && null != getter) {
-      tracer().setDataStreamCheckpoint(span, SERVER_PATHWAY_EDGE_TAGS, 0);
+      tracer()
+          .getDataStreamsMonitoring()
+          .setDataStreamCheckpoint(span, SERVER_PATHWAY_EDGE_TAGS, 0);
```
devinsba, line 148: Since these methods are now not in the tracer class can we rename them to not be repetitive? Not sure if these are exposed as an API to customers

PerfectSlayer: Sure thing! It will improve readability 👍

PerfectSlayer:
Hi @kr-igor  👋 
You might be interested by those change. I will add you to the possible reviewers.
Feel free to ask questions if you want more context.

internal-api/src/main/java/datadog/trace/bootstrap/instrumentation/api/AgentPropagation.java
```diff
@@ -4,18 +4,18 @@
...
 
   <C> void inject(AgentSpan span, C carrier, Setter<C> setter, TracePropagationStyle style);
 
   // The input tags should be sorted.
-  <C> void injectBinaryPathwayContext(
-      AgentSpan span, C carrier, BinarySetter<C> setter, LinkedHashMap<String, String> sortedTags);
+  default <C> void injectBinaryPathwayContext(
+      AgentSpan span, C carrier, BinarySetter<C> setter, LinkedHashMap<String, String> sortedTags) {
+    //noinspection unchecked
+    injectPathwayContext(span, carrier, (Setter<C>) setter, sortedTags);
```
ygree, line 17: This default implementation is confusing. How do we guarantee that this type casting is possible? Can we avoid it or at least add a comment to clarify this type casting.

internal-api/build.gradle
```diff
@@ -47,6 +47,7 @@ excludedClassesCoverage += [
   "datadog.trace.bootstrap.instrumentation.api.Tags",
   "datadog.trace.bootstrap.instrumentation.api.CommonTagValues",
   "datadog.trace.bootstrap.instrumentation.api.AgentTracer.NoopAgentPropagation",
+  "datadog.trace.bootstrap.instrumentation.api.AgentPropagation",
```
ygree, line 50: Is it because of a default method? Can we get rid of that default implementation that uses unchecked type casting?

internal-api/src/main/java/datadog/trace/bootstrap/instrumentation/api/AgentTracer.java
```diff
@@ -572,10 +516,10 @@ public AgentHistogram newHistogram(double relativeAccuracy, int maxNumBins) {
     }
   }
 
-  public static final class NoopAgentSpan implements AgentSpan {
+  public static class NoopAgentSpan implements AgentSpan {
```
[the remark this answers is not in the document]
PerfectSlayer: Good catch. There was one in #5686 but it is not included in this patch. I will revert!

internal-api/src/main/java/datadog/trace/bootstrap/instrumentation/api/AgentPropagation.java
PerfectSlayer: There is one change I did not include on this refactoring but it could help to get things clearer make `BinarySetter<C> implements Setter<C>`.

Currently, there is only one implementation of `BinarySetter` which is `TextMapInjectAdapter` from Kafka integration. It implements both `Setter<Headers>` and `BinarySetter<Headers>`.

Doing this, I can remove the whole `injectBinaryPathwayContext()` to only keep the `injectPathwayContext()` method that takes a `Setter` that can implements `BinarySetter`.

I push 8f99633 to address it. What do you think about it?

internal-api/build.gradle
PerfectSlayer: Yes, it is due to the default method. I removed it whatever we decide to do about pathway context injection and revert this line.

bantonsson (approved):
👍🏻

PR #5709, opened 2023-08-11
[WIP] Circle CI dynamic config

smola:
# What Does This Do

# Motivation

# Additional Notes


bantonsson:
Very interesting indeed. Do you know if rerun-from-start will regenerate the dynamic config?

smola:
> @bantonsson Do you know if rerun-from-start will regenerate the dynamic config?

It will not. Picking up changes from labels requires touching the branch.

.circleci/config.continue.yml.j2
```diff
@@ -0,0 +1,1327 @@
...
+      requires:
+        - ok_to_test
+      name: test_semeru11_smoke
+      gradleTarget: "stageMainDist :smokeTest"
+      gradleParameters: "-PskipFlakyTests"
+      stage: smoke
+      cacheType: smoke
+      parallelism: 4
+      maxWorkers: 3
+      testJvm: "semeru11"
```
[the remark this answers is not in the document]
smola: Ooops. Leftovers from the previous configuration. I have removed these blocks.
Semeru 11 and 17 were previously a special case that ran only for smoke tests, but I have added them to the general JDK matrix.

.gitignore
```diff
@@ -68,6 +68,8 @@ replay_pid*
 ############
 _circle_ci_cache_*
 upstream.env
+.circleci/config.continue.yml
+!.circleci/config.continue.yml.j2
```
PerfectSlayer, line 72: Do using an absolute path `/circleci/config.continue.yml` simpler than pattern + exclusion pattern?

smola: Indeed. Changed.

PerfectSlayer (approved):
LGTM. 
I will be in favor of pushing further the logic to Python if we commit to the dynamic configuration idea 👌 

Just one thing: can you try to amend commit to make sure the new `.circleci/config.continue.yml.j2` is marked as renamed from `.circleci/config.yml`? I would rather keep the file commit history.

smola:
> @PerfectSlayer I will be in favor of pushing further the logic to Python if we commit to the dynamic configuration idea 👌

:100: 

> can you try to amend commit to make sure the new `.circleci/config.continue.yml.j2` is marked as renamed from `.circleci/config.yml`? I would rather keep the file commit history.

Done. The PR diff will not see the rename properly, but if you look at commit level, you'll see the proper diff: https://github.com/DataDog/dd-trace-java/pull/5709/commits/e0de49edb984392b7bf96256083052e1eabc3281

`git blame` will also work properly after this.

bantonsson (approved):

nikita-tkachenko-datadog (approved):

jbachorik (approved):

PR #5711, opened 2023-08-11
Add git command line client builder to GitInfoProvider

nikita-tkachenko-datadog:
# What Does This Do
Adds a new `GitInfoBuilder` instance to `GitInfoProvider`.
The new instance fetches git data using shell commands.

Also repository URL in Git data upload now uses `GitInfoProvider` instead of directly invoking the shell command.

# Motivation
Previously Git data upload relied solely on executing a shell command to obtain Git remote URL, while the other logic that used Git remote URL extracted it from a variety of sources: CI provider or user-supplied environment variables, locally available `.git` folder, etc.
This change makes Git data upload consistent with the other places in the code that use Git info.

In addition to that, the other places can now benefit from info fetched with shell commands, in case it cannot be obtained from other sources.

# Additional notes
The new shell-command-based `GitInfoBuilder` has lower priority than the other existing sources.

### DIFF
--- a/dd-java-agent/agent-ci-visibility/build.gradle
+++ b/dd-java-agent/agent-ci-visibility/build.gradle
@@ -7,8 +7,6 @@ apply from: "$rootDir/gradle/java.gradle"
 apply from: "$rootDir/gradle/version.gradle"
 
 minimumBranchCoverage = 0.7
-minimumInstructionCoverage = 0.8
-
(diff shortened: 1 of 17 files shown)


romainkomorndatadog (commented):
Looks sound enough to me but I'm still not enough of a Java dev to approve. 😅

PerfectSlayer (approved):

PR #5723, opened 2023-08-16
Unify spark metrics naming

paul-laffon-dd:
# What Does This Do

Spark metrics had a separate prefix for each operation (application, job, stage) which turned out to be complicated to use from the UI. This PR unifies the prefix to "spark" for all operations

There is already a processing in the backend to rename those prefix, so spans already appear with the name "spark" in datadog UI

# Motivation

# Additional Notes


PerfectSlayer (approved):
LGTM

PR #5733, opened 2023-08-16
Reduce log level for WAF timeouts

cataphract:


ValentinZakharov (approved):

dd-java-agent/appsec/src/main/java/com/datadog/appsec/powerwaf/PowerWAFModule.java
```diff
@@ -410,23 +411,26 @@ public void onDataAvailable(
...
+      StandardizedLogging.executingWAF(log);
+      long start = 0L;
+      if (log.isDebugEnabled()) {
+        start = System.currentTimeMillis();
+      }
 
+      try {
+        resultWithData = doRunPowerwaf(reqCtx, newData, ctxAndAddr, isTransient);
+      } catch (TimeoutPowerwafException tpe) {
+        log.info("Timeout calling the WAF", tpe);
```
smola, line 424:
```suggestion
        log.debug("Timeout calling the WAF", tpe);
```

@cataphract Let's reduce timeouts to debug level, since they might be spammy in some cases and will be better handled by metrics.

smola (changes requested):

smola (approved):

PR #5736, opened 2023-08-17
Telemetry v2 config origin

ygree:
# What Does This Do

Adds config origin.

# Motivation

Telemetry V2 implementation.

# Additional Notes

Adds a specification change to include `jvm.prop` as a valid config origin for Java when configuration passed via JVM property: https://github.com/DataDog/instrumentation-telemetry-api-docs/pull/71


internal-api/src/main/java/datadog/trace/bootstrap/config/provider/PropertiesConfigSource.java
```diff
@@ -25,4 +26,12 @@ public String getConfigFileStatus() {
   protected String get(String key) {
     return props.getProperty(useSystemPropertyFormat ? propertyNameToSystemPropertyName(key) : key);
   }
+
+  @Override
+  public ConfigOrigin origin() {
+    // TODO there is no value for props in the spec
+    // https://github.com/DataDog/instrumentation-telemetry-api-docs/blob/main/GeneratedDocumentation/ApiDocs/v2/SchemaDocumentation/Schemas/conf_key_value.md#conf_key_value\
+    // maybe another value is needed similar to app.config that only applies to .NET
```
bantonsson, line 34: I agree that the spec needs to be updated with a `props` value.

internal-api/src/main/java/datadog/trace/api/Config.java
```diff
@@ -3458,7 +3458,10 @@ private static String getProp(String name) {
   private static String getProp(String name, String def) {
     String value = System.getProperty(name, def);
     if (value != null) {
-      ConfigCollector.get().put(name, value);
+      // TODO there is no value for props in the spec
```
ygree, line 3461: For some reason this PR breaks `./build.sh java --weblog-variant spring-boot-3-native` in system-tests.
I can't see what exactly could cause this. A workaround would be adding `<buildArg>--initialize-at-build-time=datadog.trace.api.ConfigCollector</buildArg>` into https://github.com/DataDog/system-tests/blob/main/utils/build/docker/java/spring-boot-3-native/pom.xml#L70 
Turns out that is caused by this line https://github.com/DataDog/dd-trace-java/pull/5736/files#diff-0d563029c90e2adb7f10f281111b5cd2cb02ca73515f104ec6dbd6c951c4c752R186
I tried to hide this line behind a condition check, a for loop, a method but neither helped to resolve this issue.

internal-api/src/main/java/datadog/trace/bootstrap/config/provider/PropertiesConfigSource.java
ygree: Okay. I've opened [a PR](https://github.com/DataDog/instrumentation-telemetry-api-docs/pull/71) to the spec to propose this change

internal-api/src/main/java/datadog/trace/api/Config.java
ygree: Now some system-test fail because of schema validation because of the `jvm.prop` value.

internal-api/src/main/java/datadog/trace/util/AgentThreadFactory.java
```diff
@@ -104,7 +104,7 @@ public static Thread newAgentThread(
           @Override
           public void uncaughtException(final Thread thread, final Throwable e) {
             LoggerFactory.getLogger(runnable.getClass())
-                .error("Uncaught exception in {}", agentThread.threadName, e);
+                .error("Uncaught exception {} in {}", e, agentThread.threadName, e);
```
PerfectSlayer, line 107: Is this change related to config origin?

PerfectSlayer (approved):
LGTM. Only minor comments

internal-api/build.gradle
```diff
@@ -24,6 +24,7 @@ tasks.compileTestJava.configure {
 minimumBranchCoverage = 0.7
 minimumInstructionCoverage = 0.8
 excludedClassesCoverage += [
+  "datadog.trace.api.ConfigSetting",
```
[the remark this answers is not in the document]
bantonsson: I agree that this should be covered by the tests.

internal-api/src/main/java/datadog/trace/api/ConfigOrigin.java
```diff
@@ -0,0 +1,17 @@
+package datadog.trace.api;
+
+public enum ConfigOrigin {
+  /** configurations that are set through environment variables */
+  ENV("env_var"),
+  /** values that are set using remote config */
+  REMOTE("remote_config"),
+  JVM_PROP("jvm.prop"),
```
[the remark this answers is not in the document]
bantonsson: Another question is why then naming is different here. Why `jvm.prop` and not `jvm_prop`?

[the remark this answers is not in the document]
bantonsson: I see. The `.net` team already added the `app.config` origin, so `jvm.prop` isn't that strange.

bantonsson (approved):

internal-api/src/main/java/datadog/trace/util/AgentThreadFactory.java
ygree: Not directly. While debugging I noticed that it doesn't log an exception, so I added it.

internal-api/src/main/java/datadog/trace/api/ConfigOrigin.java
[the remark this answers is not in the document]
ygree: Yeah, I just followed `.NET` style. I agree that it should either use a snail case or a dot notation not both. I'll change `jvm.prop` to `jvm_prop` then (https://github.com/DataDog/instrumentation-telemetry-api-docs/pull/71)

internal-api/src/main/java/datadog/trace/bootstrap/config/provider/ConfigProvider.java
```diff
@@ -200,6 +216,9 @@ public List<String> getList(String key) {
   public List<String> getList(String key, List<String> defaultValue) {
     String list = getString(key);
     if (null == list) {
+      if (collectConfig) {
+        ConfigCollector.get().put(key, String.join(",", defaultValue), ConfigOrigin.DEFAULT);
```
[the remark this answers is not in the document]
ygree: Good catch!

internal-api/src/main/java/datadog/trace/util/AgentThreadFactory.java
bantonsson: Maybe you missed it somehow. I did a quick check by throwing an exception in the `TraceProcessingWorker` without your changes, and this is what I got in the log.

```
[dd.trace 2023-09-21 09:12:26:816 +0200] [dd-trace-processor] ERROR datadog.trace.agent.common.writer.TraceProcessingWorker$DaemonTraceSerializingHandler - Uncaught exception in dd-trace-processor
java.lang.OutOfMemoryError: SRSLY??
        at datadog.trace.agent.common.writer.TraceProcessingWorker$DaemonTraceSerializingHandler.runDutyCycle(TraceProcessingWorker.java:154)
        at datadog.trace.agent.common.writer.TraceProcessingWorker$DaemonTraceSerializingHandler.run(TraceProcessingWorker.java:145)
        at java.base/java.lang.Thread.run(Thread.java:833)
```

PR #5737, opened 2023-08-18
Add generic async result type support for OpenTelemetry annotations and its Reactive Streams extension

PerfectSlayer:
# What Does This Do

This PR is a follow up of #5593 to provide support for other asynchronous result types (than the ones from JDK8).
This brings the support for Reactive Streams type: `Producer`.

# Motivation

This first type was chosen according the most used libraries.

# Additional Notes


dd-java-agent/agent-bootstrap/src/main/java/datadog/trace/bootstrap/instrumentation/decorator/AsyncResultDecorator.java
```diff
@@ -0,0 +1,117 @@
...
+   * asynchronous, it will be return unmodified and span will be finished.
+   *
+   * @param result The result to check type.
+   * @param span The related span to finish.
+   * @return An asynchronous result that will finish the span if the result is asynchronous, the
+   *     original result otherwise.
+   */
+  public Object wrapAsyncResultOrFinishSpan(final Object result, final AgentSpan span) {
+    if (result != null) {
+      for (AsyncResultSupportExtension extension : EXTENSIONS) {
```
bantonsson, line 45: I'm wondering if this can't be stored in a `ClassValue` and only looked up once for each type?

bantonsson, line 45: I'm wondering if this can't be stored in a `ClassValue` and only looked up once for each type?

bantonsson, line 45: I'm wondering if this can't be stored in a `ClassValue` and only looked up once for each type?

bantonsson (approved):
Nice separation

bantonsson:
GitHub apparently has the hiccups. Commenting once is enough.

dd-java-agent/agent-bootstrap/src/main/java/datadog/trace/bootstrap/instrumentation/decorator/AsyncResultDecorator.java
PerfectSlayer: It can and I will do a patch with this solution but it comes with the following drawback:

- invalid cache: if something goes wrong with the extension activation within an instrumentation or comes after `ClassValue` entry is computed, I will not have a way to find all types to remove their values from `ClassValue`.
- multiple extensions: the current solution is able to run multiple extensions if the first ones do not apply to the instance. So I introduced a `CompoundExtension` to avoid having a `ClassValue<List<Extension>` for readability concern.

bantonsson: Do you have a concrete example when multiple extensions are needed? I fail to see how they could interact in any controlled way if there is no guaranteed ordering to them when registered et.c.

PerfectSlayer: I don’t for now but OpenTelemetry has the case of some parent interfaces, like `Publisher`, implemented by different instrumentations.
I preferred to split it into different instrumentations and keep it simple.
I will remove the support for multiple extension then.

bantonsson (approved):

PR #5747, opened 2023-08-21
Refactor CI Visibility to better encapsulate internal APIs

nikita-tkachenko-datadog:
# What Does This Do
Moves some CI Visibility classes from `internal-api` module to `agent-ci-visibility` module.

# Motivation
Better encapsulation: internal CI Visibility details should be hidden as much as possible in `agent-ci-visibility` module.
As the code evolves, some classes that previously had to be exposed for instrumentations to use can now be moved inside the product module.


PerfectSlayer (approved):
LGTM. Thanks for splitting the refactoring in a dedicated PR 🙏

PR #5749, opened 2023-08-21
Elasticsearch and Opensearch should omit `params` by default

### LINKED ISSUE #5718: Don't write the body of ElasticSearch traces by default
**Describe what happened:**
When you integrate ElasticSearch with the Datadog Agent, it will write the body of all ElasticSearch Traces into the metadata.

**Describe what you expected:**
As the body can contain sensitive information, it shouldn't write the body into the traces as default, like in every other HTTP Request. This behavior should be explicitly set.

**Steps to reproduce the issue:**
1. Integrate ElasticSearch with the Datadog Agent
2. Send some request to ElasticSearch
3. Go to "Traces" -> "your ElasticSearch Service"
4. Click on a request and go to "METADATA" and you will see the body of the request

bm1549:
# What Does This Do
Makes it so that `elasticsearch.params` and `opensearch.params` are omitted by default

The original behavior can be maintained by setting `trace.elasticsearch.body-and-params.enabled` to `true

# Motivation
This is beneficial in the case that they contain sensitive information
Also closes #5718 

# Additional Notes

### DIFF
--- a/dd-trace-api/src/main/java/datadog/trace/api/ConfigDefaults.java
+++ b/dd-trace-api/src/main/java/datadog/trace/api/ConfigDefaults.java
@@ -182,7 +182,7 @@ public final class ConfigDefaults {
 
   static final float DEFAULT_TRACE_FLUSH_INTERVAL = 1;
 
-  static final boolean DEFAULT_ELASTICSEARCH_BODY_AND_PARAMS_ENABLED = true;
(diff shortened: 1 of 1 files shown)


am312 (approved):

bantonsson:
@bm1549 Please use the word `solves` instead of `fixes` or `closes`, and add the relevant issue to the same milestone as the PR. GitHub automatic closing of issues can unfortunately not be configured properly.

PR #5732, opened 2023-08-16
Fixed NPE in User Events Tracking

ValentinZakharov:
# What Does This Do
Fixed SpringSecurity NullPointerException when user creation happens outside request context.

# Motivation
Reported exception message in debug log:
```
[dd.trace 2023-08-16 12:27:42:204 +0000] [main] DEBUG datadog.trace.bootstrap.ExceptionLogger - Failed to handle exception in instrumentation for org.springframework.security.provisioning.InMemoryUserDetailsManager
java.lang.NullPointerException
        at datadog.trace.agent.core.CoreTracer.getTraceSegment(CoreTracer.java:1232)
        at datadog.trace.bootstrap.instrumentation.decorator.AppSecUserEventDecorator.getSegment(AppSecUserEventDecorator.java:68)
        at datadog.trace.bootstrap.instrumentation.decorator.AppSecUserEventDecorator.onSignup(AppSecUserEventDecorator.java:41)
        at datadog.trace.instrumentation.springsecurity5.SpringSecurityUserEventDecorator.onSignup(SpringSecurityUserEventDecorator.java:38)
        at org.springframework.security.provisioning.InMemoryUserDetailsManager.createUser(InMemoryUserDetailsManager.java:96)
        at org.springframework.security.provisioning.InMemoryUserDetailsManager.<init>(InMemoryUserDetailsManager.java:71)
        at org.springframework.boot.autoconfigure.security.servlet.UserDetailsServiceAutoConfiguration.inMemoryUserDetailsManager(UserDetailsServiceAutoConfiguration.java:81)
        at java.base/jdk.internal.reflect.NativeMethodAccessorImpl.invoke0(Native Method)
        at java.base/jdk.internal.reflect.NativeMethodAccessorImpl.invoke(Unknown Source)
        at java.base/jdk.internal.reflect.DelegatingMethodAccessorImpl.invoke(Unknown Source)
        at java.base/java.lang.reflect.Method.invoke(Unknown Source)
        at org.springframework.beans.factory.support.SimpleInstantiationStrategy.instantiate(SimpleInstantiationStrategy.java:154)
```

PR #5757, opened 2023-08-23
Fix NPE in user events tracking (#5732)

smola:
# What Does This Do
* Fixed SpringSecurity NullPointerException when user creation happens outside request.
* That exception used to be caught and logged to debug level.

# Motivation

# Additional Notes
Backport of #5732

### DIFF
--- a/dd-trace-core/src/main/java/datadog/trace/core/CoreTracer.java
+++ b/dd-trace-core/src/main/java/datadog/trace/core/CoreTracer.java
@@ -1229,7 +1229,11 @@ public class CoreTracer implements AgentTracer.TracerAPI {
 
   @Override
   public TraceSegment getTraceSegment() {
-    AgentSpan.Context ctx = activeSpan().context();
+    AgentSpan activeSpan = activeSpan();
+    if (activeSpan == null) {
+      return null;
+    }
+    AgentSpan.Context ctx = activeSpan.context();
     if (ctx instanceof DDSpanContext) {
       return ((DDSpanContext) ctx).getTraceSegment();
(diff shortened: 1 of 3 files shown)


dd-trace-core/src/main/java/datadog/trace/core/CoreTracer.java
```diff
@@ -1229,7 +1229,11 @@ public ProfilingContextIntegration getProfilingContext() {
 
   @Override
   public TraceSegment getTraceSegment() {
-    AgentSpan.Context ctx = activeSpan().context();
+    AgentSpan activeSpan = activeSpan();
```
cataphract, line 1232: This method leaves me a bit uncomfortable, because it behaves differently from  `RequestContext#getTraceSegment`, which never returns null. There is a `TraceSegment.Noop` that can be used here, but at the cost doing unnecessary work if there's no active span. The question is whether there being no active span is a frequent enough occurrence, that it's working risking NPEs

smola: @cataphract Agreed. Since this was the only instance where we had no active spans. However, given that this is a backport PR, I would prefer to move on with this option, which was already merged and tested in master.

I'll open a separate PR to master to follow this up.

smola: On the other hand, the `CoreTracer#getTraceSegment` usage is pretty limited. All _current_ usages consider `null`.

cataphract (approved):

PR #5763, opened 2023-08-23
Fix multi-buffers response case

val06:
# What Does This Do

- Support response that is split into multiple buffers
- limited the Payload buffer to 160 bytes (max buffer size supported by USM on a kernel side)

# Motivation

# Additional Notes

### DIFF
--- a/dd-java-agent/instrumentation/sslengine/src/main/java/datadog/trace/instrumentation/sslengine/SslEngineInstrumentation.java
+++ b/dd-java-agent/instrumentation/sslengine/src/main/java/datadog/trace/instrumentation/sslengine/SslEngineInstrumentation.java
@@ -121,7 +121,7 @@ public final class SslEngineInstrumentation extends Instrumenter.Usm
         return;
       }
       if (result.bytesProduced() > 0 && dst.limit() >= result.bytesProduced()) {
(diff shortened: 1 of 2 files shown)


dd-java-agent/instrumentation/sslengine/src/main/java/datadog/trace/instrumentation/sslengine/SslEngineInstrumentation.java
```diff
@@ -55,27 +56,51 @@ public static void wrap(
...
+            // copy the slice into the destination buffer
+            dstBuffer.put(slice);
+          }
+          // restore the original position in the src buffer
           srcs[i].position(oldPos);
+          // update number of total copied bytes so far
           consumed += oldPos;
         }
-
+        dstBuffer.flip();
```
guyarb, line 101: add the comment here as well

val06: removed this line entirely, we are getting the underlying byte array in the next line and don't perform any additional operations on the ByteBuffer object

guyarb (approved):

PR #5767, opened 2023-08-23
Fix instrumentation when bytecode generation fails

jpbempel:
If bytecode generation fails for any reason we should returns null for cancelling the current class transformation

# What Does This Do

# Motivation

# Additional Notes


shatzi (approved):
Awesome fix.

PR #5769, opened 2023-08-23
Implement sending total coverage percentage for sessions and modules

nikita-tkachenko-datadog:
# What Does This Do
Adds code coverage percentage tag to session and module spans.

# Motivation
Code coverage percentage is used by CI Visibility backend / UI (for displaying coverage trends, in code quality gates, and in other places).

# Additional Notes
Code coverage data is added whenever Jacoco is available (does not matter whether the project already has it, or whether it is configured to be injected by CI Visibility).


dd-trace-api/src/main/java/datadog/trace/api/config/CiVisibilityConfig.java
```diff
@@ -22,6 +22,10 @@ public final class CiVisibilityConfig {
       "civisibility.compiler.plugin.auto.configuration.enabled";
   public static final String CIVISIBILITY_CODE_COVERAGE_ENABLED =
       "civisibility.code.coverage.enabled";
+  public static final String CIVISIBILITY_CODE_COVERAGE_PERCENTAGE_CALCULATION_ENABLED =
```
juan-fernandez, line 25: is this used as a kill switch?

juan-fernandez (approved):

dd-trace-api/src/main/java/datadog/trace/api/config/CiVisibilityConfig.java
nikita-tkachenko-datadog: Exactly, so that the mechanism for calculating coverage percentage could be disabled if there are unexpected errors or the overhead is too big

PR #5771, opened 2023-08-23
Separate config for Elasticsearch body and params

bm1549:
# What Does This Do
Expands upon #5749 to separate the configuration options for `body` and `params` so that they can be individually selected

After this, the usage of `dd.trace.elasticsearch.body-and-params.enabled` will be discouraged in favor of the individual options, `dd.trace.elasticsearch.body.enabled` and `dd.trace.elasticsearch.params.enabled`

# Motivation
Gives users of the library increased flexibility when it comes to configuring their elasticsearch instrumentation

# Additional Notes


PerfectSlayer:
> dd.trace.elasticsearch.body.enabled and dd.trace.elasticsearch.trace.enabled

Do you mean `dd.trace.elasticseach.params.enabled` instead?

But overall, why use two boolean flags to control one behavior rather than an unique settings with meaningful value: `body`, `params` or `none`?

bantonsson:
@PerfectSlayer I think that two separate `*.enabled` settings makes it clearer than the quad-value setting for `body`, `params`, body-params`, `none`. But maybe that's just me.

bantonsson (approved):

PR #5772, opened 2023-08-23
Fix for Vert.x 4.0 instrumentation to close span on timeout

ygree:
# What Does This Do

Fix for an issue with Vert.x HTTP timed out client requests. It was keeping a request span unfinished.

Vert.x uses Netty under the hood. Netty instrumentation already has timeout handling but turns out that Vert.x has its own timeout handling that have not been covered with instrumentation.

There are two main pieces. First, closing a span in Netty instrumentation in case of the `channelInactive` event. Second, decorating a span on Vert.x instrumentation side with the original exception, so it's clear that the request has been timed out.

# Motivation

A client reported issue about created but never finished client http spans in case of timeout.

# Additional Notes

Example of a timed out span:

```json
[
  {
    "service": "service-name",
    "name": "netty.client.request",
    "resource": "GET /?",
    "trace_id": "2178227018975193174",
    "span_id": "7989050593220702817",
    "parent_id": 0,
    "start": "1692834467128191917",
    "duration": 1017764708,
    "type": "http",
    "error": 1,
    "metrics": {
      "_sampling_priority_v1": 1,
      "_dd.measured": 1,
      "_dd.top_level": 1,
      "thread.id": 27,
      "process_id": 55085,
      "_dd.profiling.enabled": 0,
      "_dd.agent_psr": 1,
      "peer.port": 80,
      "_dd.trace_span_attribute_schema": 0
    },
    "meta": {
      "_dd.p.dm": "-1",
      "thread.name": "vert.x-eventloop-thread-0",
      "http.url": "http://httpstat.us/200",
      "language": "jvm",
      "env": "local-test",
      "version": "1",
      "peer.ipv4": "20.40.202.3",
      "component": "netty-client",
      "error.type": "io.vertx.core.http.impl.NoStackTraceTimeoutException",
      "span.kind": "client",
      "error.message": "The timeout period of 1000ms has been exceeded while executing GET /200?sleep=2000 for server httpstat.us:80",
      "error.stack": "io.vertx.core.http.impl.NoStackTraceTimeoutException: The timeout period of 1000ms has been exceeded while executing GET /200?sleep=2000 for server httpstat.us:80\n",
      "runtime-id": "aaf5dacb-af13-449d-a82d-9d7bd45422b8",
      "peer.hostname": "httpstat.us",
      "http.method": "GET"
    }
  }
]
```

### DIFF
--- a/dd-java-agent/instrumentation/netty-4.0/src/main/java/datadog/trace/instrumentation/netty40/client/HttpClientResponseTracingHandler.java
+++ b/dd-java-agent/instrumentation/netty-4.0/src/main/java/datadog/trace/instrumentation/netty40/client/HttpClientResponseTracingHandler.java
@@ -64,4 +64,23 @@ public class HttpClientResponseTracingHandler extends ChannelInboundHandlerAdapt
       super.exceptionCaught(ctx, cause);
     }
   }
+
+  @Override
+  public void channelInactive(ChannelHandlerContext ctx) throws Exception {
+    final Attribute<AgentSpan> parentAttr = ctx.channel().attr(CLIENT_PARENT_ATTRIBUTE_KEY);
+    parentAttr.setIfAbsent(noopSpan());
+    final AgentSpan parent = parentAttr.get();
+    final AgentSpan span = ctx.channel().attr(SPAN_ATTRIBUTE_KEY).getAndSet(parent);
+    if (span != null) {
+      try (final AgentScope scope = activateSpan(span)) {
+        DECORATE.beforeFinish(span);
+        span.finish();
+      }
+    }
+    // We want the callback in the scope of the parent, not the client span
+    try (final AgentScope scope = activateSpan(parent)) {
+      scope.setAsyncPropagation(true);
+      super.channelInactive(ctx);
+    }
+  }
 }
--- a/dd-java-agent/instrumentation/vertx-web-4.0/src/main/java/datadog/trace/instrumentation/vertx_4_0/client/HttpClientRequestBaseInstrumentation.java
+++ b/dd-java-agent/instrumentation/vertx-web-4.0/src/main/java/datadog/trace/instrumentation/vertx_4_0/client/HttpClientRequestBaseInstrumentation.java
@@ -2,7 +2,6 @@ package datadog.trace.instrumentation.vertx_4_0.client;
 
 import static datadog.trace.agent.tooling.bytebuddy.matcher.NameMatchers.named;
 import static datadog.trace.bootstrap.instrumentation.api.AgentTracer.activateSpan;
-import static datadog.trace.bootstrap.instrumentation.decorator.HttpServerDecorator.DD_SPAN_ATTRIBUTE;
 import static datadog.trace.bootstrap.instrumentation.httpurlconnection.HttpUrlConnectionDecorator.DECORATE;
 import static net.bytebuddy.matcher.ElementMatchers.isMethod;
 import static net.bytebuddy.matcher.ElementMatchers.isPackagePrivate;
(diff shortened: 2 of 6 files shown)


dd-java-agent/instrumentation/vertx-web-4.0/src/main/java/datadog/trace/instrumentation/vertx_4_0/client/HttpClientRequestBaseInstrumentation.java
```diff
@@ -0,0 +1,63 @@
...
+  }
+
+  public static class ResetAdvice {
+    @Advice.OnMethodExit(suppress = Throwable.class)
+    public static void onExit(
+        @Advice.Argument(value = 0) Throwable cause,
+        @Advice.FieldValue("stream") final HttpClientStream stream,
+        @Advice.Return(readOnly = false) boolean result) {
+      if (result) {
+        AttributeKey<AgentSpan> spanAttr = AttributeKey.valueOf(DD_SPAN_ATTRIBUTE);
```
bantonsson, line 53: It is probably ok get the attribute key directly here in a vertx setting, but there have been issues with the `AttributeKey` class being loaded in multiple class loaders leading to confusing results. See [AttributeKeys](https://github.com/DataDog/dd-trace-java/blob/master/dd-java-agent/instrumentation/netty-4.1-shared/src/main/java/datadog/trace/instrumentation/netty41/AttributeKeys.java#L12)

dd-java-agent/instrumentation/netty-4.0/src/main/java/datadog/trace/instrumentation/netty40/client/HttpClientResponseTracingHandler.java
```diff
@@ -64,4 +64,23 @@ public void exceptionCaught(ChannelHandlerContext ctx, Throwable cause) throws E
       super.exceptionCaught(ctx, cause);
     }
   }
+
+  @Override
+  public void channelInactive(ChannelHandlerContext ctx) throws Exception {
```
bantonsson, line 69: I guess this is to ensure that the netty versions behave the same, since vertx seem to need netty 4.1.

bantonsson (approved):
👍🏻 Would like some clarification around the long test timeouts.

dd-java-agent/instrumentation/netty-4.0/src/main/java/datadog/trace/instrumentation/netty40/client/HttpClientResponseTracingHandler.java
ygree: Correct, Vert.x uses Netty 4.1. I added this for Netty 4.0 too just because there might be a similar to Vert.x case leading to an unfinished span. Making sure we finish it here will potentially prevent this type of an issue.

dd-java-agent/instrumentation/vertx-web-4.0/src/main/java/datadog/trace/instrumentation/vertx_4_0/client/HttpClientRequestBaseInstrumentation.java
ygree: Hmm, interesting. Would be interesting to know more about possible issues. From what I can see AbstractConstant implementation if there are two AttributeKeys from different class loaders they won't be treated the same key. I didn't notice any issues with Vert.x use case. In the worst case scenario with Vert.x it won't be able to set an error on the span but the span will still be finished by Netty instrumentation.
Actually, I think I can reuse `netty41.AttributeKeys`. Will push this change to this PR.

dd-java-agent/instrumentation/vertx-web-4.0/src/main/java/datadog/trace/instrumentation/vertx_4_0/client/HttpClientRequestBaseInstrumentation.java
```diff
@@ -12,7 +11,7 @@
 import datadog.trace.agent.tooling.Instrumenter;
 import datadog.trace.bootstrap.instrumentation.api.AgentScope;
 import datadog.trace.bootstrap.instrumentation.api.AgentSpan;
-import io.netty.util.AttributeKey;
+import datadog.trace.instrumentation.netty41.AttributeKeys;
```
bantonsson, line 14: Does this need to be declared as a helper class here as well?

ygree: Yep, otherwise it'll fail at muzzle check

PR #5773, opened 2023-08-24
Fix CI triggers with dynamic config

smola:
# What Does This Do
Regular triggers do not play well with dynamic config in Circle CI. We need to use pipeline triggers at:
https://app.circleci.com/settings/project/github/DataDog/dd-trace-java/triggers
# Motivation

# Additional Notes


bantonsson (approved):

.circleci/config.continue.yml.j2
```diff
@@ -32,6 +32,13 @@ profiling_modules: &profiling_modules "dd-java-agent/agent-profiling"
 default_system_tests_commit: &default_system_tests_commit 9659a5e409cf8bacabd739b6af07c5f03c9bcbd6
 
 parameters:
+  nightly:
+    type: boolean
+    default: false
+  weekly:
+    type: boolean
+    default: false
```
PerfectSlayer, line 40: There is an enum type parameters. 
What about using it to encode a the trigger type? (`push`, `nightly`, `weekly`, maybe `release` also?)

smola: Not sure how to make push/release (not triggered by us) set such parameter in `config.yml`. Also note that these are only used in `config.yml`, but Circle CI will pass them to the generated pipeline too, and it will fail if they are not declared.

PR #5774, opened 2023-08-24
Enable ByteCode verification by default

jpbempel:
# What Does This Do
Once we have generated the new classfile we run a static analysis of the bytecode using the ASM library's analyzer. It helps find out before sending back the modified classfile to the JVM if the bytecode is not correct and then cancelling the instrumentation to not impact the user code.

# Motivation
avoid verifier errors preventing class to be loaded

# Additional Notes


dd-java-agent/agent-debugger/src/test/java/com/datadog/debugger/agent/SpanProbeInstrumentationTest.java
```diff
@@ -80,6 +80,17 @@ public void lineRangeSimpleSpan() throws IOException, URISyntaxException {
     assertArrayEquals(new String[] {SPAN_PROBEID_TAG}, span.tags);
   }
 
+  @Test
+  public void lineRangeErrorSimpleSpan() throws IOException, URISyntaxException {
```
shatzi, line 84: This test, while verify the code changes here, is very hard to understand.

this test doesn't not verify we run the verification logic, but instead that we didn't generate a span. so this can hide bug and not protect against regression. For example, let assume we got a different code that fails the instrumentation, this test will pass. but we don't have a code that verify that we run the static code analysis.

Can we create a more specific test verify we call static code analysis function? or maybe there is a better test to verify our safeguards are on by default?

dd-java-agent/agent-debugger/src/test/java/com/datadog/debugger/agent/LogProbesInstrumentationTest.java
```diff
@@ -486,7 +486,6 @@ private DebuggerTransformerTest.TestSnapshotListener installProbes(
     Config config = mock(Config.class);
     when(config.isDebuggerEnabled()).thenReturn(true);
     when(config.isDebuggerClassFileDumpEnabled()).thenReturn(true);
-    when(config.isDebuggerVerifyByteCode()).thenReturn(true);
```
shatzi, line 489: Why we no longer use "true" here? but "true" in other tests? what is the logic here?

dd-java-agent/agent-debugger/src/test/java/com/datadog/debugger/agent/SpanProbeInstrumentationTest.java
```diff
@@ -128,6 +139,7 @@ private MockTracer installSpanProbes(Configuration configuration) {
     Config config = mock(Config.class);
     when(config.isDebuggerEnabled()).thenReturn(true);
     when(config.isDebuggerClassFileDumpEnabled()).thenReturn(true);
+    when(config.isDebuggerVerifyByteCode()).thenReturn(true);
```
shatzi, line 142: if DEFAULT_DEBUGGER_VERIFY_BYTECODE = true; - do we still need this?

shatzi (approved):
Awesome stuff + few comments

dd-java-agent/agent-debugger/src/test/java/com/datadog/debugger/agent/SpanProbeInstrumentationTest.java
jpbempel: yes that's true that this test does not specially test that the static analysis is run. but it's really its purpose it's more that if an error during instrumentation happens, we dont' blow up the program and it still run correctly.
So in a sense it's more an integration test than a unit test.
I don't really care if this is the static analysis that is run or not, just that the program run correctly with or without ant instrumentation failure. here this is one.

dd-java-agent/agent-debugger/src/test/java/com/datadog/debugger/agent/SpanProbeInstrumentationTest.java
jpbempel: yes because the default mock have all its fields initialize with default value, i.e. false.

dd-java-agent/agent-debugger/src/test/java/com/datadog/debugger/agent/LogProbesInstrumentationTest.java
ygree: I'm also confused about this change and would appreciate an explanation.

dd-java-agent/agent-debugger/src/test/java/com/datadog/debugger/agent/SpanProbeInstrumentationTest.java
ygree: Maybe assert log for "Verify result"?

ygree (approved):
Overall looks good. One suggestion for adding a validation log record test assertion if possible.

dd-java-agent/agent-debugger/src/test/java/com/datadog/debugger/agent/LogProbesInstrumentationTest.java
jpbempel: This is not necessary for those tests, and all tests passed without it. It was copy/paste from `CapturedSnapshotTest`

dd-java-agent/agent-debugger/src/test/java/com/datadog/debugger/agent/SpanProbeInstrumentationTest.java
jpbempel: Added an additional test that can only pass with bytecode verification.

PR #5775, opened 2023-08-24
Get databricks cluster name from spark conf, instead of job properties

paul-laffon-dd:
# What Does This Do

Rely on the `spark.databricks.clusterUsageTags.clusterName` present in the SparkConf instead of the one present in the jobProperties

After some trial and errors, it turned out that the value in jobProperties can sometimes be missing, while it is always present in the SparkConf

# Motivation


raphaelgavache (approved):

am312:
In cases where both values are present, do we know that they're guaranteed to be equal?  Should this new clusterName value be a fallback used only when the current value in jobProperties is missing?  Just trying not to disrupt any expected behaviors here.

paul-laffon-dd:
Yes, if both values are equals, they will have the same value. We can use the clusterName present in the spark conf as a fallback if the value is not present in the job properties. Changed the PR to reflect it

am312 (approved):

PR #5778, opened 2023-08-24
Improve IAST tainted map performance

manuel-alvarez-alvarez:
# What Does This Do
Tries to improve the performance of the tainted map when counting the elements and clearing the instance

# Motivation

# Additional Notes

### DIFF
--- a/dd-java-agent/agent-iast/src/main/java/com/datadog/iast/taint/TaintedMap.java
+++ b/dd-java-agent/agent-iast/src/main/java/com/datadog/iast/taint/TaintedMap.java
@@ -3,7 +3,6 @@ package com.datadog.iast.taint;
 import com.datadog.iast.util.NonBlockingSemaphore;
 import java.lang.ref.Reference;
 import java.lang.ref.ReferenceQueue;
-import java.util.Arrays;
 import java.util.Iterator;
 import java.util.NoSuchElementException;
 import java.util.concurrent.atomic.AtomicInteger;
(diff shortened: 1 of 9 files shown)


dd-java-agent/agent-iast/src/main/java/com/datadog/iast/taint/TaintedMap.java
```diff
@@ -233,11 +232,21 @@ private int remove(final TaintedObject entry) {
...
-    Arrays.fill(table, null);
+    int count = 0;
+    for (int i = 0; i < table.length; i++) {
+      TaintedObject entry = table[i];
+      while (entry != null) {
+        count++;
+        entry = entry.next;
+      }
+      table[i] = null;
+    }
```
richardstartin, line 247: This is very surprising unless the table is far too large. If the table has moderate occupancy `Arrays.fill` will compile to fast vectorized code, whereas the new code is pointer chasing with an unpredictable branch.

manuel-alvarez-alvarez: We expect the table to be pretty sparse most of the time and only grow in certain scenarios like http posts with heavy forms or JSON uploads (it has a default capacity of 16384). 

We always need to traverse the table to generate some metrics, so that is the reason to join the cleaning loop with the counting one, but I'll review the JIT optimizations and will benchmark it.

Thanks for the feedback!, you've given me some literature for tonight

smola (commented):
Merging the two steps seems appropriate since counting will be done anyway. Although when telemetry is off, the counting is not necessary?

Also, let's give this a title and description that matches what the PR actually does.

manuel-alvarez-alvarez:
We will postpone this modification for now

PR #5779, opened 2023-08-24
Allow data streams monitoring to be controlled by dynamic config

randomanderson:
# What Does This Do

Allows DSM to be enabled or disabled by dynamic (remote) config.

The main change is that `DefaultDataStreamsMonitoring` is always used. At the normal bucket report interval, it enables or disables itself depending upon the state of `TraceConfig` (which is updated via dynamic config).

Additionally, when extracting or injecting data streams data into a trace context, it will check the version of `TraceConfig` stamped with the trace to decide whether or not dsm is enabled for the trace.

Most of the changes in this PR are changes to tests. I will add a smoke test in a followup PR because it is quite a bit of work: decoding dsm payloads in the test server, enabling bucket duration to be configurable, using the manual dsm api, etc.

### DIFF
--- a/dd-trace-core/src/main/java/datadog/trace/core/datastreams/DefaultDataStreamsMonitoring.java
+++ b/dd-trace-core/src/main/java/datadog/trace/core/datastreams/DefaultDataStreamsMonitoring.java
@@ -134,15 +134,14 @@ public class DefaultDataStreamsMonitoring implements DataStreamsMonitoring, Even
       features.discoverIfOutdated();
     }
 
-    checkDynamicConfig();
     agentSupportsDataStreams = features.supportsDataStreams();
+    checkDynamicConfig();
 
     if (!configSupportsDataStreams) {
       log.debug("Data streams is disabled");
     } else if (!agentSupportsDataStreams) {
       log.debug("Data streams is disabled or not supported by agent");
     }
-    supportsDataStreams = configSupportsDataStreams && agentSupportsDataStreams;
 
     nextFeatureCheck = timeSource.getCurrentTimeNanos() + FEATURE_CHECK_INTERVAL_NANOS;
 


dd-trace-core/src/main/java/datadog/trace/core/CoreTracer.java
```diff
@@ -1134,17 +1133,6 @@ private static String[] generateConstantTags(final Config config) {
...
   }
 
-  @SuppressForbidden
-  private static DataStreamsMonitoring createDataStreamsMonitoring(
-      Config config, SharedCommunicationObjects sharedCommunicationObjects, TimeSource timeSource) {
-    if (config.isDataStreamsEnabled()) {
-      return new DefaultDataStreamsMonitoring(config, sharedCommunicationObjects, timeSource);
-    } else {
-      log.debug("Data streams monitoring not enabled.");
-      return new NoopDataStreamsMonitoring();
```
ygree, line 1144: Can we remove `NoopDataStreamsMonitoring` now or it's still needed?

dd-trace-core/src/main/java/datadog/trace/core/datastreams/DefaultDataStreamsMonitoring.java
```diff
@@ -117,12 +134,15 @@ public void start() {
...
-      supportsDataStreams = false;
+    checkDynamicConfig();
+    agentSupportsDataStreams = features.supportsDataStreams();
+
+    if (!configSupportsDataStreams) {
+      log.debug("Data streams is disabled");
+    } else if (!agentSupportsDataStreams) {
       log.debug("Data streams is disabled or not supported by agent");
     }
+    supportsDataStreams = configSupportsDataStreams && agentSupportsDataStreams;
```
ygree, line 145: Looks like these checks are already done in `checkDynamicConfig()`. Can we just call `checkDynamicConfig()`?

dd-trace-core/build.gradle
```diff
@@ -23,16 +23,14 @@ excludedClassesCoverage += [
...
-  'datadog.trace.core.Base64Encoder',
   // FIXME: more test coverage needed
   'datadog.trace.core.CoreTracer',
   'datadog.trace.core.CoreTracer.1',
   'datadog.trace.core.DDSpan.1',
   'datadog.trace.core.tagprocessor.QueryObfuscator.1',
   'datadog.trace.common.writer.TraceProcessingWorker.NonDaemonTraceSerializingHandler',
   // FIXME(DSM): test coverage needed
-  'datadog.trace.core.datastreams.DefaultDataStreamsMonitoring',
+  'datadog.trace.core.datastreams.NoopDataStreamsMonitoring',
```
ygree, line 33: Is `NoopDataStreamsMonitoring` still needed?

dd-trace-core/src/main/java/datadog/trace/core/CoreTracer.java
randomanderson: I went ahead and removed `NoopDataStreamsMonitoring`. It was only used by a test

dd-trace-core/src/main/java/datadog/trace/core/datastreams/DefaultDataStreamsMonitoring.java
randomanderson: It already calls `checkDynamicConfig()` a few lines up. This assignment was superfluous

ygree (approved):
Looks good to me!

PR #5783, opened 2023-08-25
Check for overhead constraints in weak randomness module

manuel-alvarez-alvarez:
# What Does This Do
Check for overhead limits before reporting weak randomness vulnerability

# Motivation

# Additional Notes


smola (approved):
Looks good, but we have to review quotas, possibly setting a separate one per vulnerability type. Approving to avoid the excessive overhead at the moment.

Issue #5782, opened 2023-08-25
VerifyError at `org/eclipse/jetty/server/HttpChannel.handle()` when instrumenting metabase


vr:
Hello,

We are used to include dd-trace-java in our metabase instance and it was working flawlessly, but with the latest release it did not work and crash at startup of the application, here is the [Stack trace](https://sebsauvage.net/paste/?5676ed3cd5c42075#mfi1/9/NNntRTsZBr4Z4+rWMmBusIgICfyaKh/42cpY=)

It works without the agent so should be linked to it.

Last time we build the image was 3 weeks ago, and it worked

smola:
@vr Thank you for reporting this. What version of dd-trace-java are you using?

smola:
@vr We are currently investigating this:
* I am currently assuming this is happening with dd-trace-java v1.19.0 or higher.
* Knowing the exact metabase version will help.
* You may try setting `DD_APPSEC_ENABLED=false` environment variable to `-Ddd.appsec.enabled=false` system property as a workaround.

smola:
Confirmed we can reproduce this with latest metabase docker image (`metabase/metabase:latest`) and `DD_APPSEC_ENABLED=false` works as a workaround.

vr:
I can confirm this is working with the workaround, thanks !

PS: yes it was with the latest dd-trace-java and the latest metabase 0.47.0 !

smola:
This is now fixed in dd-trace-java v1.20.0.

PR #5787, opened 2023-08-25
Fix simultaneous jetty 10/11 instrumentation when jakarta/javax servlet are both present

cataphract:
See https://github.com/DataDog/dd-trace-java/issues/5782


smola (approved):

manuel-alvarez-alvarez (approved):
LGTM, maybe a smoke test or system test might be interesting to ensure that this does not happen with other instrumentations in the future.

smola:
Verified this fix with metabase v0.47.0.

bantonsson (approved):

PR #5790, opened 2023-08-28
Refactor CI Visibility to use separate domain objects for build systems and for test frameworks

nikita-tkachenko-datadog:
# What Does This Do
Splits `DDTestSessionImpl` and `DDTestModuleImpl` classes into versions that are used by build system instrumentations and those that are used by test framework instrumentations.

# Motivation
Build systems and test frameworks instrumentations interact with test sessions and modules differently.
Having both build-system- and test-framework-specific logic located in the same class makes it difficult to comprehend and maintain.

# Additional Notes
The change creates two pairs of new interfaces:
* `DDBuildSystemSession` / `DDBuildSystemModule`
* `DDTestFrameworkSession` / `DDTestFrameworkModule`
Existing `DDTestSession` / `DDTestModule` interfaces are kept as they are exposed to customers.


romainkomorndatadog (approved):
LGTM from the CI Visibility side.

PR #5795, opened 2023-08-28
Reports instrumentation failure

jpbempel:
# What Does This Do
Send probe status error for all probes applied on the same class indicating instrumentation could not be applied for this class.

# Motivation

# Additional Notes


dd-java-agent/agent-debugger/src/main/java/com/datadog/debugger/agent/DebuggerTransformer.java
```diff
@@ -432,6 +442,15 @@ private void reportLocationNotFound(
       // on a separate class files because probe was set on an inner/top-level class
       return;
     }
+    reportError(definitions, format, className, location);
```
shatzi, line 445: can be simplifed:
```
private void reportLocationNotFound(List<ProbeDefinition> definitions, String className, String methodName) {
    if (methodName != null) {
      reportError(definitions, CANNOT_FIND_METHOD, className, methodName);
    } else {
      // This is a line probe, so we don't report line not found because the line may be found later
      // on a separate class files because probe was set on an inner/top-level class
      return;
    }
}
```

shatzi (approved):
Looks awesome.

PR #5806, opened 2023-08-29
Correctly handle cases when there are multiple JVM forks per single test module

nikita-tkachenko-datadog:
# What Does This Do
Updates build system instrumentations to correctly handle cases when tests for a specific module are executed in multiple JVMs.
Also handles cases when tests for a specific module use different testing frameworks (this is orthogonal to multiple forks, can happen even if there is only one fork per module).

# Motivation
When tests for one module run in multiple forked JVMs, we want the reported module span to contain the aggregated data (that is, all the test frameworks used in all the forks, combined code coverage data from all the forks, the total number of tests skipped by ITR for all the forks, etc).

# Additional Notes
Depending on the build configuration it is possible to have multiple JVM forks started for the one module: in Gradle this is achieved with `maxParallelForks` or `forkEvery` properties of the `Test` task, in Maven - with `forkCount` property of the Surefire plugin. The forks can execute either concurrently or sequentially.


romainkomorndatadog (approved):
Approving with my usual "get someone who knows Java..." caveat.

PR #5809, opened 2023-08-30
Merge span decoration and log instrumentation

jpbempel:
# What Does This Do
Span decoration probes and log probes share the same instrumentor so we need to generate the same instrumentation for both otherwise we fall into the same issues than duplicated probes, and local var access issue with scoping.
CapturedContextInstrumentor is therefore the only instrumentor for both span decoration and log probes

# Motivation

# Additional Notes


dd-java-agent/agent-debugger/src/main/java/com/datadog/debugger/probe/SpanDecorationProbe.java
```diff
@@ -137,7 +138,8 @@ public void instrument(
       MethodNode methodNode,
       List<DiagnosticMessage> diagnostics,
       List<String> probeIds) {
-    new SpanDecorationInstrumentor(this, classLoader, classNode, methodNode, diagnostics, probeIds)
+    new CapturedContextInstrumentor(
+            this, classLoader, classNode, methodNode, diagnostics, probeIds, false, Limits.DEFAULT)
```
shatzi, line 142: it seems we are calling istrument of the first probe in the captureContextProbe list.

what are the side-effects if a span-decoration is first and the second is a log probe with isCaptureSnapshot() set to true?

I think same for different captureLimits

shatzi (approved):
LGTM

dd-java-agent/agent-debugger/src/main/java/com/datadog/debugger/probe/SpanDecorationProbe.java
jpbempel: Fixed that by picking always the first log probe first even though we have specific order dictated by https://github.com/DataDog/dd-trace-java/blob/967c9d72a5c33b293687a10c14d999c63fe2002e/dd-java-agent/agent-debugger/src/main/java/com/datadog/debugger/agent/Configuration.java#L135-L150

PR #5819, opened 2023-08-31
Add support for armeria-grpc

devinsba:
# What Does This Do

Adds support for armeria-grpc


dd-java-agent/instrumentation/armeria-grpc/src/main/java/datadog/trace/instrumentation/armeria/grpc/client/ClientCallImplInstrumentation.java
```diff
@@ -0,0 +1,236 @@
...
+public final class ClientCallImplInstrumentation extends Instrumenter.Tracing
+    implements Instrumenter.ForSingleType {
+
+  public ClientCallImplInstrumentation() {
+    super("armeria-grpc-client", "armeria-grpc", "armeria", "grpc-client", "grpc");
+  }
+
+  @Override
+  public Map<String, String> contextStore() {
+    return Collections.singletonMap("io.grpc.ClientCall", AgentSpan.class.getName());
```
amarziali, line 43: Should a continuation be preferred here ?

amarziali (approved):
Looks good to me. I left a question

dd-java-agent/instrumentation/armeria-grpc/src/main/java/datadog/trace/instrumentation/armeria/grpc/client/ClientCallImplInstrumentation.java
devinsba: In this case because I'll be activating the span on different threads during the client lifecycle, and for the streaming use cases a span is the right object to save

amarziali: thank you

PR #5822, opened 2023-09-01
Add span smoke tests

jpbempel:
# What Does This Do
Also refactor the status of the instrumentation to be returned by instrument method. It avoids to send a probe status INSTALLED first then received later the ERROR one.
Report a probe status error when trying to instrument a single line span probe.

# Motivation

# Additional Notes

### DIFF
--- a/dd-java-agent/agent-debugger/src/main/java/com/datadog/debugger/agent/ConfigurationUpdater.java
+++ b/dd-java-agent/agent-debugger/src/main/java/com/datadog/debugger/agent/ConfigurationUpdater.java
@@ -182,6 +182,9 @@ public class ConfigurationUpdater
 
   private void recordInstrumentationProgress(
       ProbeDefinition definition, InstrumentationResult instrumentationResult) {
+    if (instrumentationResult.isError()) {
+      return;
+    }
     instrumentationResults.put(definition.getId(), instrumentationResult);
     if (instrumentationResult.isInstalled()) {
       sink.addInstalled(definition.getProbeId());
--- a/dd-java-agent/agent-debugger/src/main/java/com/datadog/debugger/instrumentation/SpanInstrumentor.java
+++ b/dd-java-agent/agent-debugger/src/main/java/com/datadog/debugger/instrumentation/SpanInstrumentor.java
@@ -36,25 +36,24 @@ public class SpanInstrumentor extends Instrumentor {
   }
 
   @Override
-  public void instrument() {
+  public InstrumentationResult.Status instrument() {
     if (isLineProbe) {
       fillLineMap();
-      addRangeSpan(lineMap);
-    } else {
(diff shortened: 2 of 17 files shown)


dd-java-agent/agent-debugger/src/main/java/com/datadog/debugger/agent/ConfigurationUpdater.java
```diff
@@ -182,6 +182,9 @@ private void installNewDefinitions() {
 
   private void recordInstrumentationProgress(
       ProbeDefinition definition, InstrumentationResult instrumentationResult) {
+    if (instrumentationResult.isError()) {
```
shatzi, line 185: I find this contact kind of weird, with isError() not report a messages. 
1. does this is a mixed responsibility for classes? 
2. how does the error status message get to the sink?

dd-java-agent/agent-debugger/src/main/java/com/datadog/debugger/instrumentation/SpanInstrumentor.java
```diff
@@ -94,25 +93,29 @@ private InsnList createSpan(LabelNode initSpanLabel) {
     return insnList;
   }
 
-  private void addRangeSpan(LineMap lineMap) {
+  private InstrumentationResult.Status addRangeSpan(LineMap lineMap) {
     Where.SourceLine[] targetLines = definition.getWhere().getSourceLines();
     if (targetLines == null || targetLines.length == 0) {
-      // no line capture to perform
-      return;
+      reportError("Missing line(s) in probe definition.");
```
shatzi, line 99: Maybe consider change the return value of instrument function?
I find it wried that we have both reportError and return InstrumentationResult.Status.

shatzi (approved):
Great we have integration tests for spans.

shatzi:
After talking about it, refactor the internal logic would happen on later PR, this PR goal is to create a set of integration tests to validate the logic of the current code.

dd-java-agent/agent-debugger/src/main/java/com/datadog/debugger/agent/ConfigurationUpdater.java
jpbempel: Yep will refactor it in a next PR to centralize the handling

dd-java-agent/agent-debugger/src/main/java/com/datadog/debugger/instrumentation/SpanInstrumentor.java
jpbempel: sure, will do in the next refactoring PR

jpbempel:
Merge as is, but will refactor the diagnostic messages handling in a next PR

PR #5828, opened 2023-09-04
Register all AppSec instrumentation callbacks

smola:
# What Does This Do
Always register every AppSec instrumentation callback, regardless of configuration rules.

# Motivation
This comes from a time where AppSec could not be configured remotely. Instrumentation callbacks are registered only once, so it currently does not make sense to skip some callbacks based on a configuration that can change later.

# Additional Notes
There may be a performance hit for users using `DD_APPSEC_RULES` with a small static rules files.


ValentinZakharov (approved):

ValentinZakharov:
I think, we need to simplify/rework instrumentation gateway. Initially it was designed to serve many modules, but in practice it used only for ASM, so, I think some simpler implementation, would be better.

smola:
Needs some work to make it up to date again. I might retake this after some refactoring of the gateway callbacks.

PR #5831, opened 2023-09-04
Allow a single subscriber per AppSec event

smola:
# What Does This Do

# Motivation
We only use one listener per event, so we don't need the logic for multiple events.

# Additional Notes

### DIFF
--- a/dd-java-agent/appsec/src/main/java/com/datadog/appsec/AppSecSystem.java
+++ b/dd-java-agent/appsec/src/main/java/com/datadog/appsec/AppSecSystem.java
@@ -156,7 +156,7 @@ public class AppSecSystem {
       }
 
       for (AppSecModule.EventSubscription sub : module.getEventSubscriptions()) {
-        eventSubscriptionSet.addSubscription(sub.eventType, sub);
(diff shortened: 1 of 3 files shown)


dd-java-agent/appsec/src/main/java/com/datadog/appsec/AppSecSystem.java
```diff
@@ -156,7 +156,7 @@ private static void loadModules(EventDispatcher eventDispatcher) {
       }
 
       for (AppSecModule.EventSubscription sub : module.getEventSubscriptions()) {
-        eventSubscriptionSet.addSubscription(sub.eventType, sub);
+        eventSubscriptionSet.setSubscription(sub.eventType, sub);
```
manuel-alvarez-alvarez, line 159: It will be nice if we could log a warn/error message in case there are multiple subscribers, otherwise they will be silently ignored.

manuel-alvarez-alvarez (approved):

smola:
Close in favor of https://github.com/DataDog/dd-trace-java/pull/5840

PR #5857, opened 2023-09-07
Fix blocking for jetty 10.0.16 and 11.0.16

cataphract:
The change in jetty was also applied on jetty 12.0.1, but we don't support AppSec on jetty-12 yet.


smola (approved):

smola:
The remaining failing test is unrelated, see https://github.com/DataDog/dd-trace-java/pull/5864 so this is ready to merge.

PR #5860, opened 2023-09-08
Make peer.service manually overridden by component

amarziali:
# What Does This Do

Add the possibility to statically map values for peer.service by component name.

The mapping can be provided by setting the property `dd.trace.peer.service.component.overrides` with `key:value,key:value`.
For instance, to set all couchbase spans peer service to `couchbase` one can provide `couchbase-client:couchbase`

When overridden, the peer service source will be set to `config_override`

# Motivation

# Additional Notes


PerfectSlayer (approved):
LGTM. Is the feature documented somewhere?

amarziali:
> LGTM. Is the feature documented somewhere (even if it's experimental)?

I'll add the doc to the official doc for inferred services (it's a private beta so behind a link)

PR #5861, opened 2023-09-08
Update Spock version to 2.2

nikita-tkachenko-datadog:
# What Does This Do
Updates Spock testing framework version from `2.1` to `2.2`.

# Motivation
Spock 2.2 is needed to test a CI Visibility feature that cannot be supported in older Spock framework versions ([tagging tests](https://spockframework.org/spock/docs/2.3/all_in_one.html#test-tag-extension)).

# Additional Notes
The version is changed to `2.2` instead of the latest available one, since CI Visibility benefits from running its tests with the oldest framework version possible (and I could not find a way to run CI Vis instrumentation tests with Spock 2.2 while allowing the rest of the project's modules to use a newer version).


nikita-tkachenko-datadog:
> Do you think we should upgrade buildSrc versions too for consistency? Or is it fine to keep a 2.0 version?

Sounds good, I completely forgot we have a separate dependency in `buildSrc`. Upgraded that to 2.2 as well.

smola (approved):
A Spock upgrade that does not break our tests? Nice.

PR #5862, opened 2023-09-08
Check out and build merge commit in GitLab CI

bantonsson:
# What Does This Do

Tries to create a merge commit in GitLab based on the current commit and the closest base branch, and then builds that (this built artifact is used for Benchmark tests).

# Motivation

If pull requests drift from the base branch, then the artifacts used for benchmarking and other testing are not representative of the resulting merge back of the PR. This also aligns with CircleCI where the tests are run on the merge commit.

# Additional Notes


ygree (commented):
LGTM

ygree (approved):
LGTM

PR #5867, opened 2023-09-11
Fix request body tainting in spring boot 2.7.5

DDJavierSantos:
# What Does This Do
Fixes instrumentation for spring boot 2.7.5 

# Motivation
Maintenance bug fix for the new version of spring boot using a new class NonClosingStream that was in the instrumentation exclussion list.

# Additional Notes

### DIFF
--- a/dd-java-agent/agent-tooling/src/main/resources/datadog/trace/agent/tooling/bytebuddy/matcher/ignored_class_name.trie
+++ b/dd-java-agent/agent-tooling/src/main/resources/datadog/trace/agent/tooling/bytebuddy/matcher/ignored_class_name.trie
@@ -288,6 +288,8 @@
 0 org.springframework.core.io.buffer.DataBuffer
 0 org.springframework.core.io.buffer.DefaultDataBuffer
 0 org.springframework.core.io.buffer.NettyDataBuffer
+# Need for IAST so propagation of tainted objects is complete in spring 2.7.5
+0 org.springframework.util.StreamUtils$NonClosingInputStream
 # There are some Mono implementation that get instrumented
 0 org.springframework.http.server.reactive.*
 2 org.springframework.instrument.*


dd-java-agent/agent-tooling/src/main/resources/datadog/trace/agent/tooling/bytebuddy/matcher/ignored_class_name.trie
```diff
@@ -310,7 +310,7 @@
 2 org.springframework.stereotype.*
 2 org.springframework.transaction.*
 2 org.springframework.ui.*
-2 org.springframework.util.*
+0 org.springframework.util.*
```
smola, line 313: It's probably better to be more precise, setting this to `0` only for the specific class we want to instrument. Also adding a comment to note it's needed for IAST. See, for example:

https://github.com/DataDog/dd-trace-java/pull/5867/files#diff-c154dcf8283beb4f1b67579ba190f14675ea92fe9fb3a20fa34753468743069cR324-R325

smola (changes requested):

dd-java-agent/agent-tooling/src/main/resources/datadog/trace/agent/tooling/bytebuddy/matcher/ignored_class_name.trie
DDJavierSantos: I have removed the package and added the specific class:

dd-java-agent/agent-tooling/src/main/resources/datadog/trace/agent/tooling/bytebuddy/matcher/ignored_class_name.trie
```diff
@@ -288,6 +288,7 @@
 0 org.springframework.core.io.buffer.DataBuffer
 0 org.springframework.core.io.buffer.DefaultDataBuffer
 0 org.springframework.core.io.buffer.NettyDataBuffer
+0 org.springframework.util.StreamUtils$NonClosingInputStream
```
smola, line 291: Please, add a comment stating this is for the IAST instrumenter. Otherwise it's going to be hard figuring out why we need this, since there is no explicit mention to this class in the instrumenters.

smola (approved):

smola (approved):
Benchmark looks good now.

mcculls (approved):

PR #5870, opened 2023-09-12
Custom spark spans tags at runtime

paul-laffon-dd:
# What Does This Do

Add custom spark spans tags that can be changed at runtime using the sparkContext localProperty

```scala
// All spark computations launched will have the extra tag
sparkContext.setLocalProperty("spark.datadog.tags.key", "value")
spark.read.parquet(...)

// Remove the custom tag for all next spark computations
sparkContext.setLocalProperty("spark.datadog.tags.key", null)
```

# Motivation

Asked by a design partner on spark monitoring. Spark application can be running for a long time and perform multiple computations that custom tags allow to distinguish. 

[setLocalProperty](https://github.com/apache/spark/blob/d8298bffd91de01299f9456b37e4454e8b4a6ae8/core/src/main/scala/org/apache/spark/SparkContext.scala#L784-L795) is the usual spark way to propagate properties to the next submitted jobs

The [current methods](https://docs.datadoghq.com/tracing/trace_collection/custom_instrumentation/java) for custom tagging in the trace are not supporting this use case:
- using `-Ddd.tags` the tags are common for all the spans in the spark application
- `GlobalTracer.get().activeSpan()` only capture the current span. Spark is using [lazy evaluation](https://spark.apache.org/docs/latest/rdd-programming-guide.html#rdd-operations), spans are created when computations are actually launched, which is not directly in the control of the user

### DIFF
--- a/dd-java-agent/instrumentation/spark/src/main/java/datadog/trace/instrumentation/spark/DatadogSparkListener.java
+++ b/dd-java-agent/instrumentation/spark/src/main/java/datadog/trace/instrumentation/spark/DatadogSparkListener.java
@@ -42,6 +42,7 @@ public class DatadogSparkListener extends SparkListener {
   public static volatile boolean finishTraceOnApplicationEnd = true;
 
   private final int MAX_COLLECTION_SIZE = 1000;
+  private final String RUNTIME_TAGS_PREFIX = "spark.datadog.tags.";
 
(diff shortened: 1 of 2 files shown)


raphaelgavache (approved):
nice, lgtm

am312 (approved):

PR #5872, opened 2023-09-12
Use bytes instead of strings in IAST metrics tags

manuel-alvarez-alvarez:
# What Does This Do
Uses bytes instead of strings for the tags values in IAST metrics, it's not until building the final metric that the byte value is converted to its string representation.

# Motivation
We detected a performance bottleneck due to IAST tag computation due to the heavy usage of metrics.

# Additional Notes


internal-api/src/main/java/datadog/trace/api/iast/telemetry/IastMetric.java
```diff
@@ -50,7 +47,6 @@ public static int count() {
   private final Tag tag;
   private final Verbosity verbosity;
   private int index;
-  private Map<String, Integer> tagIndexes;
```
manuel-alvarez-alvarez, line 53: Main performance improvement is removing this lookup map:

-  we have reversed the lookup `byte -> String `
- the lookup only happens when building the final telemetry request
- the lookup is done via two switch clauses that compile into `TABLESWITCH` bytecodes

smola (approved):

mcculls (approved):

PR #5886, opened 2023-09-13
Fix Telemetry Config Collector to collect String and Map values

ygree:
# What Does This Do

Fix Telemetry Config Collector to collect String and Map values

# Motivation

`trace_span_attribute_schema`  and `trace_peer_service_mapping` are not reported

# Additional Notes


ValentinZakharov (approved):

internal-api/src/main/java/datadog/trace/bootstrap/config/provider/ConfigProvider.java
```diff
@@ -218,6 +227,9 @@ public Map<String, String> getMergedMap(String key) {
     // We reverse iterate to allow overrides
     for (int i = sources.length - 1; 0 <= i; i--) {
       String value = sources[i].get(key);
+      if (collectConfig && value != null) {
+        ConfigCollector.get().put(key, value);
```
bantonsson, line 231: If there are settings from multiple sources, this will overwrite the values in the config collector, but the setting used in the library will merge them.

internal-api/src/main/java/datadog/trace/bootstrap/config/provider/ConfigProvider.java
```diff
@@ -231,6 +243,9 @@ public Map<String, String> getOrderedMap(String key) {
...
```
bantonsson, line 247: Same thing here.

internal-api/src/main/java/datadog/trace/bootstrap/config/provider/ConfigProvider.java
```diff
@@ -246,6 +261,9 @@ public Map<String, String> getMergedMapWithOptionalMappings(
     for (String key : keys) {
       for (int i = sources.length - 1; 0 <= i; i--) {
         String value = sources[i].get(key);
+        if (collectConfig && value != null) {
+          ConfigCollector.get().put(key, value);
```
bantonsson, line 265: And here.

bantonsson (commented):
Should we report the setting used in the merged maps?

internal-api/src/main/java/datadog/trace/bootstrap/config/provider/ConfigProvider.java
ygree: Good point!

PR #5891, opened 2023-09-14
Do not start profiler if running in AWS Lambda.

purple4reina:
# What Does This Do

When it detects the tracer running in AWS Lambda, refuse to start the profiler.

# Motivation

Profiling is not yet supported in Java in AWS Lambda. Node and Python though are currently in public beta. We are specifically not charging customers in this public beta. However, if a customer were to turn on profiling with the java tracer, they would be billed. We want to prevent the situation where the customer is billed for profiles created in serverless environments.

# Additional Notes

### DIFF
--- a/dd-java-agent/agent-bootstrap/src/main/java/datadog/trace/bootstrap/Agent.java
+++ b/dd-java-agent/agent-bootstrap/src/main/java/datadog/trace/bootstrap/Agent.java
@@ -846,6 +846,11 @@ public class Agent {
   private static void startProfilingAgent(final boolean isStartingFirst) {
     StaticEventLogger.begin("ProfilingAgent");
 
+    if (isAwsLambdaRuntime()) {
+      log.info("Profiling not supported in AWS Lambda runtimes");
+      return;
+    }
+
     final ClassLoader contextLoader = Thread.currentThread().getContextClassLoader();
     try {
       Thread.currentThread().setContextClassLoader(AGENT_CLASSLOADER);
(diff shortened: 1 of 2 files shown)


richardstartin:
This looks fine, but could you move the code to `datadog.trace.bootstrap.Agent#start` instead, to avoid creating the profiling agent at all in AWS lambda please? It should be a lift and shift.

richardstartin (approved):

dd-java-agent/agent-bootstrap/src/main/java/datadog/trace/bootstrap/Agent.java
```diff
@@ -911,6 +916,11 @@ public void withTracer(TracerAPI tracer) {
     StaticEventLogger.end("ProfilingAgent");
   }
 
+  private static boolean isAwsLambdaRuntime() {
+    String val = System.getenv("AWS_LAMBDA_FUNCTION_NAME");
+    return val != null && !val.isEmpty();
+  }
```
PerfectSlayer, line 922: Isn’t duplicate of the `ServerlessInfo` class?

purple4reina: It is, but apologies for my naiveté and limited java experience, when I attempt to import `datadog.common.container.ServerlessInfo` I get the error `package datadog.common.container does not exist`. I am unsure how to include the package in the build. As you recommended, I'd much prefer to use the `ServerlessInfo` class if you can show me how to include it.

PerfectSlayer: This class would need to be relocated into a common module first.
I see the PR was merged while under review? (I was on ER last week)
So what should we do about it?

purple4reina: I'll open another PR to add the update as you suggest.

PR #5893, opened 2023-09-15
Refactor diagnostic messages

jpbempel:
# What Does This Do
Using directly DebuggerSink instead going through Sink interface 
Centralize sending probe status

# Motivation

# Additional Notes


dd-java-agent/agent-debugger/src/main/java/com/datadog/debugger/agent/DebuggerTransformer.java
```diff
@@ -95,8 +101,9 @@ public DebuggerTransformer(
     }
   }
 
+  // Used only for tests
```
ojung, line 104: Can we make the constructor package-private?

jpbempel: ah yes, true

PR #5900, opened 2023-09-15
Resource name matching

dougqh:
# What Does This Do
This changes adds resource name matching to the trace sampling rules.

This change also changes matching on service and operation name to use glob matching -- which does introduce a small breaking change if someone used * or ? in their name previously

The breakage is small, since a glob pattern always matches itself.  Any existing rules will continue match all the traces that they matched previously, but they may match additional traces now.

# Motivation
This change will allow users to sample traces using a resource pattern match

# Additional Notes
In addition to the core change to trace sampling rules, this code also cleans up the handling of glob matching

New util classes Matcher and Matchers have now been introduced that handle most of the glob related code

This clean-up allowed me to eliminate some existing duplication in the single span code.

In a subsequent changeset, I may replace the pattern based glob matching with a proper glob matcher.

Parametric tests: https://github.com/DataDog/system-tests/pull/1816


dd-trace-core/src/main/java/datadog/trace/common/sampling/SamplingRule.java
```diff
@@ -106,43 +111,16 @@ public SpanSamplingRule(
...
-
     @Override
     public <T extends CoreSpan<T>> boolean matches(T span) {
-      return (serviceExactName == null || serviceExactName.equals(span.getServiceName()))
-          && (servicePattern == null || servicePattern.matcher(span.getServiceName()).matches())
-          && (operationExactName == null
-              || operationExactName.contentEquals(span.getOperationName()))
-          && (operationPattern == null
-              || operationPattern.matcher(span.getOperationName()).matches());
+      return Matchers.matches(serviceMatcher, span.getServiceName())
```
ygree, line 122: Nice clean up!

ygree (approved):
Looks good to me

mcculls (approved):
LGTM

dd-trace-core/src/main/java/datadog/trace/core/util/Matchers.java
```diff
@@ -0,0 +1,69 @@
+package datadog.trace.core.util;
+
+import java.util.regex.Pattern;
+
+public final class Matchers {
+  private Matchers() {}
+
+  public static final Matcher compileGlob(String glob) {
+    if (glob == null || glob.equals("*")) {
```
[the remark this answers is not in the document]
ygree: I don't think so, because "?" is for matching a single character only and will be converted to the "^.$" regex?

dd-trace-core/src/main/java/datadog/trace/core/util/Matchers.java
```diff
@@ -19,15 +19,15 @@ public static final Matcher compileGlob(String glob) {
     }
   }
 
-  public static final boolean matches(Matcher matcher, String str) {
```
ygree: Unnecessary `final` in the final class

dd-trace-core/src/main/java/datadog/trace/core/util/GlobPattern.java
```diff
@@ -5,20 +5,11 @@
 public final class GlobPattern {
 
   public static Pattern globToRegexPattern(String globPattern) {
-    if (globPattern == null) {
```
ygree, line 8: Removed now unnecessary checks.

dd-trace-core/src/main/java/datadog/trace/core/util/Matchers.java
```diff
@@ -0,0 +1,69 @@
+package datadog.trace.core.util;
+
+import java.util.regex.Pattern;
+
+public final class Matchers {
```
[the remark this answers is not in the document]
ygree: Tests are ready for another review

bm1549 (approved):

PR #5901, opened 2023-09-18
Add Thymeleaf support to IAST XSS vulnerability 

jandro996:
# What Does This Do
Instrument Thymeleaf classes that allow template generation avoiding escape functions

# Additional Notes
Is not possible to get tainted tex, templateName and line in the same class, so two instrumentation sharing context are needed.

org.thymeleaf.standard.processor.StandardUtextTagProcessor#doProcess -> Set IElementTagStructureHandler as potentially dangerous storing  it with template name and line in the instrumentation context

org.thymeleaf.processor.element.IElementTagStructureHandler#setBody -> if is in the instrumentation context check if the input is tainted to report the vulnerability


dd-java-agent/instrumentation/thymeleaf/src/main/java/datadog/trace/instrumentation/thymeleaf/ThymeleafContext.java
```diff
@@ -0,0 +1,21 @@
+package datadog.trace.instrumentation.thymeleaf;
+
+public class ThymeleafContext {
+
+  private String templateName;
```
manuel-alvarez-alvarez, line 5: You most likely want these two to be `final`

dd-smoke-tests/springboot-thymeleaf/src/test/groovy/datadog/smoketest/springboot/IastSpringBootThymeleafSmokeTest.groovy
```diff
@@ -0,0 +1,49 @@
...
+    final request = new Request.Builder().url(url).get().build()
+
+    when:
+    client.newCall(request).execute()
+
+    then:
+    hasVulnerability { vul -> vul.type == 'XSS' && vul.location.path == method && vul.location.line == line }
+
+    where:
+    method     | param | line
```
manuel-alvarez-alvarez, line 46: The name `method` here is a bit misleading, in this case the controller method and the template name both have the same value `utext`, but the location of the vulnerability is actually the template name.

manuel-alvarez-alvarez (approved):

dd-java-agent/agent-iast/src/main/java/com/datadog/iast/sink/XssModuleImpl.java
```diff
@@ -92,4 +92,35 @@ public void onXss(@Nonnull String format, @Nullable Object[] args) {
     checkInjection(
         span, VulnerabilityType.XSS, rangesProviderFor(to, format), rangesProviderFor(to, args));
   }
+
+  @Override
+  public void onXss(@Nonnull CharSequence s, @Nonnull String file, int line) {
+    if (!canBeTainted(s) || !canBeTainted(file)) {
```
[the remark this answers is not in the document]
jandro996: Reuse functionality to check null or emty values, but you are right, conceptually it's not correct. I'll change it

dd-smoke-tests/springboot-thymeleaf/src/test/groovy/datadog/smoketest/springboot/IastSpringBootThymeleafSmokeTest.groovy
jandro996: I will add more parameters to avoid confusion

bantonsson (commented):
Looks ok from `apm-java`.

PR #5911, opened 2023-09-19
Use _component_override for peer service source

amarziali:
# What Does This Do

.... when the peer service value has been manually overridden by component

# Motivation

# Additional Notes


PerfectSlayer:
Related to #5860

PerfectSlayer (approved):

PR #5912, opened 2023-09-19
Fix Span decoration smoke tests flakiness

jpbempel:
# What Does This Do
make sure we are asserting on the right span (name + resource)

# Motivation
Fix smoke test flakiness

# Additional Notes


dd-smoke-tests/debugger-integration-tests/src/test/java/datadog/smoketest/SpanDecorationProbesIntegrationTests.java
```diff
@@ -233,6 +230,11 @@ void testMethodMultiTagValueError() throws Exception {
     processRequests();
   }
 
+  private boolean isTracedFullMethodSpan(DecodedSpan span) {
```
shatzi, line 233: nit, maybe have something like spanHaveNameAndResource(TRACE_NAME,TRACE_RESOURCE) ?

shatzi (approved):

dd-smoke-tests/debugger-integration-tests/src/test/java/datadog/smoketest/SpanDecorationProbesIntegrationTests.java
jpbempel: maybe later if by adding more smoke tests the suggested abstraction make more sense, but for now sounds enough for the usage

PR #5916, opened 2023-09-19
Fix smoke tests for J9 JVMs

jpbempel:
# What Does This Do
skip tests that requires accessing arguments/local vars
fix flakiness of probe statuses and sending probe config

# Motivation

# Additional Notes

### DIFF
--- a/dd-smoke-tests/debugger-integration-tests/src/test/java/datadog/smoketest/SpanProbesIntegrationTest.java
+++ b/dd-smoke-tests/debugger-integration-tests/src/test/java/datadog/smoketest/SpanProbesIntegrationTest.java
@@ -8,6 +8,7 @@ import com.datadog.debugger.probe.SpanProbe;
 import datadog.trace.test.agent.decoder.DecodedSpan;
 import java.nio.file.Path;
 import java.util.List;
+import java.util.concurrent.atomic.AtomicBoolean;
 import org.junit.jupiter.api.DisplayName;
 import org.junit.jupiter.api.Test;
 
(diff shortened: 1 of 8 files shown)


dd-smoke-tests/debugger-integration-tests/src/test/java/datadog/smoketest/SpanProbesIntegrationTest.java
```diff
@@ -55,12 +56,20 @@ void testSingleLineSpan() throws Exception {
...
-    ProbeStatus status = retrieveProbeStatusRequest();
-    assertEquals(ProbeStatus.Status.RECEIVED, status.getDiagnostics().getStatus());
-    status = retrieveProbeStatusRequest();
-    assertEquals(ProbeStatus.Status.ERROR, status.getDiagnostics().getStatus());
-    assertEquals(
-        "Single line span is not supported, you need to provide a range.",
-        status.getDiagnostics().getException().getMessage());
+    AtomicBoolean received = new AtomicBoolean(false);
+    AtomicBoolean error = new AtomicBoolean(false);
+    registerProbeStatusListener(
```
shatzi, line 61: nit. can we have  `registerProbeStatusListener(ProbeStatus.Status.RECEIVED, probeStatus -> {  received.set(true) });`,  simpler logic per a single probe type status.

Not sure where ` return received.get() && error.get();` would be...

shatzi (approved):
LGTM

dd-smoke-tests/debugger-integration-tests/src/test/java/datadog/smoketest/SpanProbesIntegrationTest.java
jpbempel: yeah I have the return of the listener that indicate to stop polling for message 
But I keep that in mind we I need to refactor more depending on the usage

PR #5926, opened 2023-09-21
Update IAST vulnerability hash

jandro996:
# What Does This Do

`VulnarabilityType#calculateHash` behaviour was modified to be type-dependent:

- All types use the type name.
- `INSECURE_COOKIE`, `NO_HTTPONLY_COOKIE` and `NO_SAMESITE_COOKIE` use the evidence value (cookie name).
- `XCONTENTTYPE_HEADER_MISSING`, `HSTS_HEADER_MISSING` use the service name .
- By default vulnerability location is used.

# Motivation

Be able to calculate vulnerability hash without location

# Additional Notes

This may cause some vulnerability duplicates for the changed types when upgrading to v1.22.0.


dd-java-agent/agent-iast/build.gradle
```diff
@@ -16,6 +16,7 @@ dependencies {
   implementation project(':internal-api')
   implementation project(':internal-api:internal-api-9')
   implementation group: 'com.squareup.moshi', name: 'moshi', version: versions.moshi
+  implementation project(':dd-trace-core')
```
smola, line 19: I think this dependency might lead to duplicated classes in the jar? Needs to be checked.

dd-java-agent/agent-iast/src/main/java/com/datadog/iast/sink/HstsMissingHeaderModuleImpl.java
```diff
@@ -44,7 +44,7 @@ public void onRequestEnd(final Object iastRequestContextObject, final IGSpanInfo
         final AgentSpan span = AgentTracer.activeSpan();
         if (overheadController.consumeQuota(Operations.REPORT_VULNERABILITY, span)) {
           reporter.report(
-              span, new Vulnerability(VulnerabilityType.HSTS_HEADER_MISSING, null, null));
+              span, new Vulnerability(VulnerabilityType.HSTS_HEADER_MISSING, getServiceName(span)));
```
manuel-alvarez-alvarez, line 47: The span can be null for vulnerabilities not inside a request, can we delay this further down the line? (since we will have the newly created span later on)

dd-java-agent/agent-iast/src/main/java/com/datadog/iast/sink/SinkModuleBase.java
```diff
@@ -242,6 +243,16 @@ protected StackTraceElement getCurrentStackTrace() {
     return stackWalker.walk(SinkModuleBase::findValidPackageForVulnerability);
   }
 
+  protected String getServiceName(final AgentSpan span) {
+    if (span != null) {
+      AgentSpan.Context context = span.context();
+      if (context instanceof DDSpanContext) {
```
manuel-alvarez-alvarez, line 249: With `MutableSpan` you can have access to the service name without the new dependency

dd-java-agent/agent-iast/src/main/java/com/datadog/iast/sink/HstsMissingHeaderModuleImpl.java
jandro996: current implementation updated the vulnerability with the new span before report it, check dd-java-agent/agent-iast/src/main/java/com/datadog/iast/Reporter.java

this suits you?

manuel-alvarez-alvarez (changes requested):
There are some methods `HttpResponseHeaderModuleImpl` that are not reporting the service name, I thing we should sightly update the PR to the following:

1. Move the service name to the location class
2. Change the method `Location.forSpanAndStack(long spanId, StacktraceElement stack)` to `Location.forSpanAndStack(AgentSpan span, StacktraceElement stack)` and extrac the span id and service name from there.
3. Make sure that all vulnerabilities are created with a non null location.

smola (approved):

dd-java-agent/agent-iast/src/main/java/com/datadog/iast/model/Vulnerability.java
```diff
@@ -37,6 +38,15 @@ public long getHash() {
     return hash;
   }
 
+  public void updateSpan(final AgentSpan newSpan) {
+    if (newSpan != null) {
+      location.updateSpan(newSpan);
+      if (type instanceof VulnerabilityType.HeaderVulnerabilityType) {
+        hash = type.calculateHash(this);
```
manuel-alvarez-alvarez, line 45: It might be more future proof to always recalculate the hash, but it's OK for me.

manuel-alvarez-alvarez (approved):

PR #5928, opened 2023-09-21
Add redefinition listener to disable taintable visitor in case of failure

manuel-alvarez-alvarez:
# What Does This Do
Introduces a new `AgentBuilder.RedefinitionStrategy.Listener` to disable our `datadog.trace.agent.tooling.bytebuddy.iast.TaintableVisitor` in cases of a failure while retransforming already loaded classes.

# Motivation
Since classes are retransformed in batches, a failure transforming a class with the visitor might cause the whole batch to fail. This PR ensures that the taintable visitor will be disabled and a new attempt at retransformation will occur.

# Additional Notes


smola (approved):

ygree (approved):
Looks good to me!

jandro996 (approved):

PR #5936, opened 2023-09-25
propagate span into lettuce CommandHandler event loop

richardstartin:
# What Does This Do

Currently lettuce spans don't propagate to where the network IO is done, which prevents us from attributing CPU samples to the decoding loop, and prevents wallclock sampling of the event loop at all. This pair of instrumentations solves this problem by attaching the active span to an `AsyncCommand` when it is created (the existing `LettuceAsyncCommandsInstrumentation` currently creates the span before this happens). The `CommandHandler` processes commands and is made to reactivate the span by the `CommandHandlerInstrumentation`. 

This has no functional impact on traces, but allows for better attribution of resource consumption to the trace.

# Motivation

# Additional Notes

### DIFF
--- a/dd-java-agent/instrumentation/lettuce-5/src/main/java/datadog/trace/instrumentation/lettuce5/AsyncCommandInstrumentation.java
+++ b/dd-java-agent/instrumentation/lettuce-5/src/main/java/datadog/trace/instrumentation/lettuce5/AsyncCommandInstrumentation.java
@@ -1,0 +1,93 @@
+package datadog.trace.instrumentation.lettuce5;
+
+import static datadog.trace.agent.tooling.bytebuddy.matcher.NameMatchers.named;
+import static datadog.trace.agent.tooling.bytebuddy.matcher.NameMatchers.namedOneOf;
+import static datadog.trace.bootstrap.instrumentation.java.concurrent.AdviceUtils.capture;
+import static datadog.trace.bootstrap.instrumentation.java.concurrent.AdviceUtils.endTaskScope;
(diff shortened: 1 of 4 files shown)


dd-java-agent/instrumentation/lettuce-5/src/main/java/datadog/trace/instrumentation/lettuce5/AsyncCommandInstrumentation.java
```diff
@@ -0,0 +1,93 @@
...
+import java.util.Map;
+import net.bytebuddy.asm.Advice;
+
+/**
+ * Instrumentation attaches the active span to an {@code io.lettuce.core.protocol.AsyncCommand} so
+ * the span can propagate with it into the {@code io.lettuce.core.protocol.CommandHandler} event
+ * loop.
+ */
+@AutoService(Instrumenter.class)
+public class AsyncCommandInstrumentation extends Instrumenter.Profiling
```
bantonsson, line 30: Is there any downside to having this behavior enabled always? I can see that since it doesn't create any new spans, the span graph would look the same either way, but I'm just curious.

bantonsson (approved):

dd-java-agent/instrumentation/lettuce-5/src/main/java/datadog/trace/instrumentation/lettuce5/AsyncCommandInstrumentation.java
richardstartin: there was a request to only enable this when profiling is enabled, but I don't think there is any downside to switching it on without profiling.

PR #5940, opened 2023-09-26
Telemetry V2 Agent/Intake Routing

ygree:
# What Does This Do

Implements Telemetry V2 routing logic between Agent and Intake depending on their availability.
Sends telemetry to the Intake endpoint when Agent is not available or fails and both Telemetry Intake URL and API_KEY are specified.
Switches to the Agent telemetry endpoint once it becomes available or when Intake fails to receive telemetry requests.

# Motivation

Telemetry V2 implementation.

# Additional Notes


telemetry/src/main/java/datadog/telemetry/TelemetryRequest.java
```diff
@@ -11,57 +13,90 @@
...
+  public Request httpRequest(HttpUrl url, String apiKey) {
+    long bodySize = requestBody.endRequest();
+
+    Request.Builder builder =
+        new Request.Builder()
+            .addHeader("Content-Type", String.valueOf(JSON))
+            .addHeader("Content-Length", String.valueOf(bodySize))
+            .addHeader("DD-Telemetry-API-Version", API_VERSION)
+            .addHeader("DD-Telemetry-Request-Type", String.valueOf(this.requestType))
+            .addHeader("DD-Client-Library-Language", "jvm")
```
PerfectSlayer, line 54: What about using `DDTags.LANGUAGE_TAG_VALUE` instead?
Is it the same semantic?

telemetry/src/main/java/datadog/telemetry/TelemetryHttpClient.java
```diff
@@ -0,0 +1,141 @@
...
+    switch (telemetryReceiver) {
+      case AGENT:
+        if (requestFailed) {
+          reportErrorOnce(requestUrl, result);
+          if (apiKey != null && intakeUrl != null) {
+            log.info("Agent Telemetry endpoint failed. Telemetry will be sent to Intake.");
+            telemetryReceiver = TelemetryReceiver.INTAKE;
+            errorReported = false;
+          } else if (!missingApiKeyReported) {
+            log.error("Cannot use Intake to send telemetry because unset API_KEY.");
```
PerfectSlayer, line 87: Should we log as error a telemetry connectivity issue?

telemetry/src/main/java/datadog/telemetry/TelemetryHttpClient.java
```diff
@@ -0,0 +1,141 @@
...
+    boolean requestFailed = result != Result.SUCCESS;
+    switch (telemetryReceiver) {
+      case AGENT:
+        if (requestFailed) {
+          reportErrorOnce(requestUrl, result);
+          if (apiKey != null && intakeUrl != null) {
+            log.info("Agent Telemetry endpoint failed. Telemetry will be sent to Intake.");
+            telemetryReceiver = TelemetryReceiver.INTAKE;
+            errorReported = false;
+          } else if (!missingApiKeyReported) {
```
PerfectSlayer, line 86: Should we report missing API key when `intakeUrl` is `null`?

PerfectSlayer (commented):
Looks good :+1: Few concerns about logging level and annoying customers.

Neatpick: I would move the construction of the `TelemetryHttpClient` to a dedicated method in `TelemetrySystem` and directly pass it to the `TelemetryService` construcor (making its test constructor the only default constructor).
It should decluter all the constructor args passed from one class to another, and you can nicely move all the related Javadoc to the `TelemetryHttpClient` constructor.

telemetry/src/main/java/datadog/telemetry/TelemetryHttpClient.java
ygree: Good point. I'll separate it to two cases (missing intake url and missing api key) to avoid this misleading error.

telemetry/src/main/java/datadog/telemetry/TelemetryHttpClient.java
ygree: I'll change it to a warn.

PerfectSlayer (approved):

PR #5944, opened 2023-09-26
register netty event loops known to do IO for wallclock profiling

richardstartin:
# What Does This Do

This ensures that any syscalls done in these event loops are wallclock profiled. Activity in these event loops may be batched and therefore not associated with a particular span, so our current model of context is not powerful enough to tie this activity to traces.

# Motivation

# Additional Notes


ygree (approved):
LGTM

PR #5946, opened 2023-09-27
Prevent calling size method from unknown classes

jpbempel:
# What Does This Do
we are preventing to call on unsafe implementation. We are limiting to JDK implementation first.

# Motivation
Collection and Map interfaces can be implemented by anyone and therefore may not be safe to call. example: persistent layer that load lazily large amount of rows from DB.

# Additional Notes

### DIFF
--- a/dd-java-agent/agent-debugger/src/test/java/com/datadog/debugger/agent/CapturedSnapshotTest.java
+++ b/dd-java-agent/agent-debugger/src/test/java/com/datadog/debugger/agent/CapturedSnapshotTest.java
@@ -1544,7 +1544,6 @@ public class CapturedSnapshotTest {
     assertEquals(
         "Unsupported Collection class: com.datadog.debugger.CapturedSnapshot24$Holder",
         snapshot.getEvaluationErrors().get(0).getMessage());
-    assertEquals("len(holder)", snapshot.getEvaluationErrors().get(0).getExpr());
   }
 
   @Test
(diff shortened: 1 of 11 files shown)


dd-java-agent/agent-debugger/src/test/java/com/datadog/debugger/agent/CapturedSnapshotTest.java
```diff
@@ -1537,6 +1537,42 @@ public void recursiveCaptureException() throws IOException, URISyntaxException {
...
+            .when(
+                new ProbeCondition(
+                    DSL.when(DSL.ge(DSL.len(DSL.ref("holder")), DSL.value(0))), "len(holder) >= 0"))
+            .build();
+    DebuggerTransformerTest.TestSnapshotListener listener = installProbes(CLASS_NAME, logProbe);
+    Class<?> testClass = compileAndLoadClass(CLASS_NAME);
+    int result = Reflect.on(testClass).call("main", "").get();
+    Assertions.assertEquals(1, result);
+    Snapshot snapshot = assertOneSnapshot(listener);
+    assertEquals(1, snapshot.getEvaluationErrors().size());
```
shatzi, line 1572: maybe also add something like this: `assertEquals("len(holder)", snapshot.getEvaluationErrors().get(0).getExpr());`

shatzi (approved):
nice MVP. I like that we only capture java.*

dd-java-agent/agent-debugger/src/test/java/com/datadog/debugger/agent/CapturedSnapshotTest.java
jpbempel: done

PR #5949, opened 2023-09-27
Remove application key header from settings and ITR requests

nikita-tkachenko-datadog:
# What Does This Do
Removes application key header from requests that are made to CI Visibility endpoints.

# Motivation
The endpoints have been updated recently, and the API key is no longer required.


tonyredondo (approved):
LGTM, do you have support for EVP proxy? if yes, I think you need to remove the `X-Datadog-NeedsAppKey` header

nikita-tkachenko-datadog:
> X-Datadog-NeedsAppKey

Interestingly, I was never using it - somehow things used to work without it

PR #5950, opened 2023-09-27
Fix the X-Content-Type-Options header string

DDJavierSantos:
# What Does This Do
Fix the header string for the X-Content-Type-Options vulnerability detection

# Motivation
The functionality wasn't working

# Additional Notes

### DIFF
--- a/dd-java-agent/agent-iast/src/main/java/com/datadog/iast/util/HttpHeader.java
+++ b/dd-java-agent/agent-iast/src/main/java/com/datadog/iast/util/HttpHeader.java
@@ -58,7 +58,7 @@ public class HttpHeader {
           }
         };
     public static final HttpHeader X_CONTENT_TYPE_OPTIONS =
-        new ContextAwareHeader("X-Content-Type") {
+        new ContextAwareHeader("X-Content-Type-Options") {
           @Override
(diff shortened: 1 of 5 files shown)


dd-java-agent/agent-iast/src/main/java/com/datadog/iast/util/HttpHeader.java
```diff
@@ -58,7 +58,7 @@ public void onHeader(final IastRequestContext ctx, final String value) {
           }
         };
     public static final HttpHeader X_CONTENT_TYPE_OPTIONS =
-        new ContextAwareHeader("X-Content-Type") {
+        new ContextAwareHeader("X-Content-Type-Options") {
```
smola, line 61: The fact that no test caught this means that we need some additional test here.

smola (changes requested):

dd-java-agent/agent-iast/src/main/java/com/datadog/iast/util/HttpHeader.java
DDJavierSantos: I have added a false positive test. We were testing for the vulnerability if the header we look for was missing, but since we were looking for the wrong header, it was always missing.

smola (approved):

PR #5961, opened 2023-09-29
Add Databricks default service name if not set by the user

paul-laffon-dd:
# What Does This Do

Add a default databricks service, if not set by the user based on
- the databricks `RunName` [available for job clusters](https://docs.databricks.com/en/administration-guide/account-settings/usage-detail-tags.html)
- the databricks cluster name for all other cases

# Motivation

If not set by the user, the service name is inferred by the tracer using the class name that was launched. For databricks this name is always `com.databricks.backend.daemon.driver.driverdaemon` which is not very informative

# Additional Notes


raphaelgavache (approved):
very nice, lgtm

mcculls (approved):

PR #5966, opened 2023-09-30
Add DSM implementation for kinesis in SDKv2

devinsba:
# Additional Notes

Kinesis SDK v2 expects the endpoint to support HTTP/2 downgrade so I've updated the jetty dependency for the test server and had to add an extra exclude to all the places where jetty-server was being excluded

JIRA [AIT-8165] 

[AIT-8165]: https://datadoghq.atlassian.net/browse/AIT-8165?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ

### DIFF
--- a/dd-java-agent/instrumentation/aws-java-sdk-2.2/src/main/java/datadog/trace/instrumentation/aws/v2/TracingExecutionInterceptor.java
+++ b/dd-java-agent/instrumentation/aws-java-sdk-2.2/src/main/java/datadog/trace/instrumentation/aws/v2/TracingExecutionInterceptor.java
@@ -104,12 +104,10 @@ public class TracingExecutionInterceptor implements ExecutionInterceptor {
     final AgentSpan span = executionAttributes.getAttribute(SPAN_ATTRIBUTE);
     if (span != null) {
       executionAttributes.putAttribute(SPAN_ATTRIBUTE, null);
-      try (AgentScope ignored = activateSpan(span)) {
-        // Call onResponse on both types of responses:
-        DECORATE.onSdkResponse(span, context.response(), executionAttributes);
-        DECORATE.onResponse(span, context.httpResponse());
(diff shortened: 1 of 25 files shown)


dd-java-agent/instrumentation/aws-java-sdk-2.2/src/main/java/datadog/trace/instrumentation/aws/v2/TracingExecutionInterceptor.java
```diff
@@ -54,9 +55,10 @@ public void afterMarshalling(
       final Context.AfterMarshalling context, final ExecutionAttributes executionAttributes) {
     final AgentSpan span = executionAttributes.getAttribute(SPAN_ATTRIBUTE);
     if (span != null) {
-      DECORATE.onRequest(span, context.httpRequest());
-      DECORATE.onSdkRequest(span, context.request());
-      DECORATE.onAttributes(span, executionAttributes);
+      try (AgentScope ignored = activateSpan(span)) {
```
ygree, line 58: It's not clear to me why it needs an active scope here?

amarziali: I think also there is no particular need. I removed

amarziali: Speaking with Brian this is actually required since the datastream checkpointer needs an active span to work with

ygree (approved):
Looks good to me!

PR #5971, opened 2023-10-02
Fix boolean expression eval as template parameter

jpbempel:
# What Does This Do
Add support of any predicate or boolean expression inside a log template message is evaluated as 'true' or 'false' by converting this expression as value expression with an adapter
Refactor the parsing of the DSL to support all boolean functions

# Motivation
Support having a boolean expression inside a log template parameter liek `{contains(arg, 'foo')}`
# Additional Notes


shatzi (approved):
LGTM

PR #5982, opened 2023-10-03
[AIT-8165] Add DSM API changes to support kinesis use case

devinsba:
# What Does This Do

Modified the DSM API to allow adding consumer checkpoints for kinesis. Where we don't have a full pathway context but only have a timestamp

# Motivation

# Additional Notes

### DIFF
--- a/internal-api/src/main/java/datadog/trace/bootstrap/instrumentation/api/AgentDataStreamsMonitoring.java
+++ b/internal-api/src/main/java/datadog/trace/bootstrap/instrumentation/api/AgentDataStreamsMonitoring.java
@@ -18,4 +18,14 @@ public interface AgentDataStreamsMonitoring extends DataStreamsCheckpointer {
    */
   void setCheckpoint(
       AgentSpan span, LinkedHashMap<String, String> sortedTags, long defaultTimestamp);
+
+  /**
+   * Not added to DataStreamsCheckpointer to not add it to the public API
+   *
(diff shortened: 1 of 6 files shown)


internal-api/src/main/java/datadog/trace/bootstrap/instrumentation/api/AgentDataStreamsMonitoring.java
```diff
@@ -18,4 +18,14 @@ public interface AgentDataStreamsMonitoring extends DataStreamsCheckpointer {
    */
   void setCheckpoint(
       AgentSpan span, LinkedHashMap<String, String> sortedTags, long defaultTimestamp);
+
+  /**
+   * Not added to DataStreamsCheckpointer to not add it to the public API
+   *
+   * @param type The type of the checkpoint, usually the streaming technology being used. Examples:
+   *     kafka, kinesis, sns etc.
+   * @paren source The source of data. For instance: topic, exchange or stream name.
```
bantonsson, line 27: `@paren`?

bantonsson (approved):
Only a minor typo

internal-api/src/main/java/datadog/trace/bootstrap/instrumentation/api/AgentDataStreamsMonitoring.java
devinsba: :+1: I'll fix the one I copied too

PR #5988, opened 2023-10-04
Fix race condition for applying probe rate limits 

jpbempel:
# What Does This Do
Apply rate limit for a probe before applying configuration change
that generate instrumentation and retransformation because execution
of the new instrumented code can happen before rate limit settings
are applied and then create default (snapshot ones) rate limit.

# Motivation
Fix flaky sampling smoke tests

# Additional Notes


ojung (approved):
👍

PR #6001, opened 2023-10-06
[AIT-8165] Add DSM API changes to support kinesis use case

devinsba:
# What Does This Do

Revised #5982 to include only the NoOp context carrier, and move interface methods to make them accessible from instrumentation. The consume method did not work as expected

# Motivation

# Additional Notes


ygree (approved):
LGTM

PR #6006, opened 2023-10-09
Disable sampling for span decoration probe

jpbempel:
# What Does This Do
As span decoration probes are sharing the same instrumented code than Log probes, we also evaluate the sampling during the execution which does not bring any value in the case of Span decoration probes. We set the rate limiter for span decoration probe to a constant one.

# Motivation

# Additional Notes

Jira ticket: [DEBUG-1767]

[DEBUG-1767]: https://datadoghq.atlassian.net/browse/DEBUG-1767?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ

### DIFF
--- a/dd-java-agent/agent-debugger/debugger-bootstrap/src/main/java/datadog/trace/bootstrap/debugger/ProbeRateLimiter.java
+++ b/dd-java-agent/agent-debugger/debugger-bootstrap/src/main/java/datadog/trace/bootstrap/debugger/ProbeRateLimiter.java
@@ -1,6 +1,8 @@
 package datadog.trace.bootstrap.debugger;
 
 import datadog.trace.api.sampling.AdaptiveSampler;
+import datadog.trace.api.sampling.ConstantSampler;
+import datadog.trace.api.sampling.Sampler;
 import java.time.Duration;
 import java.time.temporal.ChronoUnit;
 import java.util.concurrent.ConcurrentHashMap;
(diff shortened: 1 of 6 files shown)


dd-java-agent/agent-debugger/debugger-bootstrap/src/main/java/datadog/trace/bootstrap/debugger/ProbeRateLimiter.java
```diff
@@ -64,7 +65,10 @@ public static void resetAll() {
     resetGlobalRate();
   }
 
-  private static AdaptiveSampler createSampler(double rate) {
+  private static Sampler createSampler(double rate) {
+    if (rate < 0) {
+      return new ConstantSampler(true);
+    }
     if (rate < 1) {
       int intRate = (int) Math.round(rate * 10);
```
shatzi, line 73: Is there a bug here.?if rate is less then 0.05 this then intRate rate would be rounded to zero. what happen if a rate == 0 ?

shatzi: if rate > epsilon, we can do AdaptiveSampler(ONE_SECOND_WINDOW / rate, 1, ...) - aka, change the size of the window instead of the rate.

shatzi (approved):
awesome!

dd-java-agent/agent-debugger/debugger-bootstrap/src/main/java/datadog/trace/bootstrap/debugger/ProbeRateLimiter.java
jpbempel: Bug I don't know but yes don't support less than 0.1.
I am not even sure why I support less than 1 per second.
also, not sure about the effect to reduce the sampling window for a high rate for example, but for rate < 1 that could be more viable option.

PR #6008, opened 2023-10-10
Flush data streams stats when lambda stops

piochelepiotr:
# What Does This Do

Flush Data Streams Monitoring stats when tracer is stopped.
For example, when Lambda function exits.

# Motivation

Data Streams stats were missing when executing Lambda functions

# Additional Notes

Jira ticket: [PROJ-IDENT]

### DIFF
--- a/dd-trace-core/src/main/java/datadog/trace/core/datastreams/DataStreamsMonitoring.java
+++ b/dd-trace-core/src/main/java/datadog/trace/core/datastreams/DataStreamsMonitoring.java
@@ -38,5 +38,5 @@ public interface DataStreamsMonitoring extends AgentDataStreamsMonitoring, AutoC
   @Override
   void close();
 
-  void flushAll();
+  void flush();
 }
--- a/dd-trace-core/src/main/java/datadog/trace/core/datastreams/DataStreamsMonitoring.java
+++ b/dd-trace-core/src/main/java/datadog/trace/core/datastreams/DataStreamsMonitoring.java
@@ -37,4 +37,6 @@ public interface DataStreamsMonitoring extends AgentDataStreamsMonitoring, AutoC
 
   @Override
   void close();
+
+  void flushAll();
 }
--- a/dd-trace-core/src/main/java/datadog/trace/core/datastreams/DefaultDataStreamsMonitoring.java
+++ b/dd-trace-core/src/main/java/datadog/trace/core/datastreams/DefaultDataStreamsMonitoring.java
@@ -52,7 +52,8 @@ public class DefaultDataStreamsMonitoring implements DataStreamsMonitoring, Even
 
   static final long FEATURE_CHECK_INTERVAL_NANOS = TimeUnit.MINUTES.toNanos(5);
 
-  private static final StatsPoint REPORT = new StatsPoint(Collections.emptyList(), 0, 0, 0, 0, 0, 0);
+  private static final StatsPoint REPORT =
+      new StatsPoint(Collections.emptyList(), 0, 0, 0, 0, 0, 0);
   private static final StatsPoint FLUSH = new StatsPoint(Collections.emptyList(), 0, 0, 0, 0, 0, 0);
   private static final StatsPoint POISON_PILL =
       new StatsPoint(Collections.emptyList(), 0, 0, 0, 0, 0, 0);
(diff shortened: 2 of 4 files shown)


dd-trace-core/src/main/java/datadog/trace/core/datastreams/DataStreamsMonitoring.java
```diff
@@ -37,4 +37,6 @@ public interface DataStreamsMonitoring extends AgentDataStreamsMonitoring, AutoC
 
   @Override
   void close();
+
+  void flushAll();
```
PerfectSlayer, line 41: Why `flushAll()` rather than `flush()`. Is there some partial flush in DSM?
Can you add documentation too? (blocking or not, best effort or reliable, …)

PerfectSlayer (commented):
I reached @mcculls out before continuing the review.

dd-trace-core/src/main/java/datadog/trace/core/datastreams/DataStreamsMonitoring.java
piochelepiotr: Flush already exists. I can also rename the current flush instead.

PerfectSlayer: You can overload methods in Java. Meaning you can have both `flush()` and `flush(long)`.
And I would rather having a clear public interface (with `flush()`) rather than using an alternate name due to implementation details (`flush(long)` being private). WDYT?

dd-trace-core/src/main/java/datadog/trace/core/datastreams/DefaultDataStreamsMonitoring.java
```diff
@@ -342,6 +351,37 @@ private void flush(long timestampNanos) {
...
+    boolean signaled = inbox.offer(FLUSH);
+    while (!signaled && supportsDataStreams && !closed) {
+      yieldOrSleep(loop++);
+      signaled = inbox.offer(FLUSH);
+    }
+    int newCount = flushCounter.get();
+    while (count >= newCount && supportsDataStreams && !closed) {
+      yieldOrSleep(loop++);
+      newCount = flushCounter.get();
+    }
```
PerfectSlayer, line 384: You might never been able to add the flush signal (queue full, thread died in the meantime), nor consume it.
You need to consider a timeout for the whole flush mechanism.

dd-trace-core/src/test/groovy/datadog/trace/core/datastreams/DataStreamsWritingTest.groovy
```diff
@@ -101,6 +101,50 @@ class DataStreamsWritingTest extends DDCoreSpecification {
     dataStreams.close()
   }
 
+
+  def "Test flushing stats"() {
```
PerfectSlayer, line 105: What about testing the complete behavior:
- that nothing is sent when not flushed,
- that flush work when no data/multiple time
- how flush work when the queue is full
- the timeout behavior (currently missing)

dd-trace-core/src/main/java/datadog/trace/core/datastreams/DefaultDataStreamsMonitoring.java
```diff
@@ -293,6 +298,10 @@ public void run() {
               flush(Long.MAX_VALUE);
             }
             break;
+          } else if (payload == FLUSH) {
+            if (supportsDataStreams) {
```
PerfectSlayer, line 302: It seems that `flush(long)` is always called behind a `if (supportedDataStreams)` guard.
It might be worthy to move it into the `flush(long)` method itself.

PerfectSlayer (commented):
Few comment about the flush logic

dd-trace-core/src/main/java/datadog/trace/core/datastreams/DataStreamsMonitoring.java
piochelepiotr: Makes sense, updated.

dd-trace-core/src/main/java/datadog/trace/core/datastreams/DefaultDataStreamsMonitoring.java
piochelepiotr: I added a check against `thread.isAlive()`, if the queue is full, I want the code to keep retrying. I think if we can't flush stats, there is an issue somewhere.

dd-trace-core/src/main/java/datadog/trace/core/datastreams/DefaultDataStreamsMonitoring.java
piochelepiotr: Makes sense. Updated.

bm1549:
Hi @piochelepiotr  👋 
We noticed that your pull request has been inactive for a while. In order to keep the project clean and organized, we’re going to close it for now.
Feel free to reopen it at any time when you’re ready to continue working on it.

PR #6009, opened 2023-10-10
Update span links implementation

PerfectSlayer:
# What Does This Do

- Add API to add span links to spans
- Add trace flags field
- Add dedicated attributes structure
- Add new attribute encoding
- Add max tag size encoding
- Add late encoding at serialization

# Motivation

Implement the new span links RFC

# Additional Notes

Jira ticket: [APMJAVA-1073]

[APMJAVA-1073]: https://datadoghq.atlassian.net/browse/APMJAVA-1073?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ

### DIFF
--- a/dd-java-agent/instrumentation/opentelemetry/opentelemetry-0.3/src/test/groovy/TypeConverterTest.groovy
+++ b/dd-java-agent/instrumentation/opentelemetry/opentelemetry-0.3/src/test/groovy/TypeConverterTest.groovy
@@ -23,8 +23,8 @@ class TypeConverterTest extends AgentTestRunner {
 
   def "should avoid extra allocation for a span wrapper"() {
     def context = createTestSpanContext()
-    def span1 = new DDSpan("test", 0, context)
-    def span2 = new DDSpan("test", 0, context)
+    def span1 = new DDSpan("test", 0, context, null)
+    def span2 = new DDSpan("test", 0, context, null)
     expect:
     // return the same wrapper for the same span
     typeConverter.toSpan(span1) is typeConverter.toSpan(span1)
@@ -47,8 +47,8 @@ class TypeConverterTest extends AgentTestRunner {
   def "should avoid extra allocation for a scope wrapper"() {
     def scopeManager = new ContinuableScopeManager(0, false, true)
     def context = createTestSpanContext()
-    def span1 = new DDSpan("test", 0, context)
-    def span2 = new DDSpan("test", 0, context)
+    def span1 = new DDSpan("test", 0, context, null)
+    def span2 = new DDSpan("test", 0, context, null)
     def scope1 = scopeManager.activate(span1, ScopeSource.MANUAL)
     def scope2 = scopeManager.activate(span2, ScopeSource.MANUAL)
     expect:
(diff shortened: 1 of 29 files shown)


dd-trace-core/src/main/java/datadog/trace/core/DDSpan.java
```diff
@@ -793,6 +809,11 @@ public TraceConfig traceConfig() {
     return context.getTrace().getTraceConfig();
   }
 
+  @Override
+  public void addLink(AgentSpanLink link) {
+    this.links.add(link);
```
bantonsson, line 814: Null checks and concurrency?

dd-java-agent/instrumentation/opentelemetry/opentelemetry-0.3/src/test/groovy/TypeConverterTest.groovy
```diff
@@ -23,8 +23,8 @@ class TypeConverterTest extends AgentTestRunner {
 
   def "should avoid extra allocation for a span wrapper"() {
     def context = createTestSpanContext()
-    def span1 = new DDSpan("test", 0, context)
-    def span2 = new DDSpan("test", 0, context)
+    def span1 = new DDSpan("test", 0, context, null)
```
bantonsson, line 26: Can't there be a constructor without the `links` parameter so the number of incidental changes are kept low.

PerfectSlayer: It is supposed to be a private constructor.
I would have done it for public API but avoid getting things more complex than needed.
I leant into more minor changes in this PR rather than adding debt in core...

If you think it is still better to add it, comment that you need it and it will be done.

bantonsson: Not a big issue. Keep the current constructor.

internal-api/src/test/groovy/datadog/trace/bootstrap/instrumentation/api/SpanLinkTest.groovy
```diff
@@ -0,0 +1,133 @@
...
+    map['string-empty'] == ''
+    !map.containsKey('string-null')
+    map['bool'] == 'true'
+    map['bool-false'] == 'false'
+    map['long'] == '12345'
+    map['long-negative'] == '-12345'
+    map['double'] == '67.89'
+    map['double-negative'] == '-67.89'
+
+    map['string-array.0'] == 'abc'
```
bantonsson, line 91: I'm still confused about this thing. Should the link be in JSON, or in this weird in-between state?

PerfectSlayer: This is a requirement of the RFC, whatever the encoding we use.

bantonsson: I think the RFC is broken, and there has been no follow up on my comments in it. I will approve this as it is, but this part needs to be sorted out for the JSON encoding.

PerfectSlayer: According to @dumontg it was decided during the Paris APM meeting to go with this encoding, even for JSON format fallback.
I will merge the PR as is then.

PR #6029, opened 2023-10-11
Rename spark streaming query attributes

paul-laffon-dd:
# What Does This Do

Rename spark streaming query attributes to make it clear that it is coming from a streaming query
- id -> streaming_query.id
- name -> streaming_query.name
- run_id -> streaming_query. run_id
- batch_id -> streaming_query.batch_id

The renaming is already done in the backend

# Motivation

Make it clear that those fields are coming from the streaming_query

# Additional Notes


raphaelgavache (approved):

ygree (approved):
Looks good to me

PR #6030, opened 2023-10-11
remove recommendation to install deprecated "action on save" intelliJ plugin

vandonr:
the plugin is not maintained anymore since Dec 2021, and is not compatible with recent IntelliJ version. 
https://plugins.jetbrains.com/plugin/7642-save-actions

Moreover, its features are mostly covered by the built-in Tools/Actions on Save settings and the git pre-commit hook.


PerfectSlayer (approved):
Thanks for taking care of the documentation 🤗

PR #6034, opened 2023-10-11
Append sql comments instead of prepend to prevent exceptions on CallableStatements

jmeunier28:
# What Does This Do

This PR updates the logic for the APM-DBM link feature introduced in https://github.com/DataDog/dd-trace-java/pull/4847. It changes from prepending the sql comments, containing the APM trace context, to appending them to the raw sql. This requires no changes on the DBM side of things in either the agent or our backend.

# Motivation

```java
  callableStmt.setInt(1,10);
  callableStmt.setInt(2,11);
  callableStmt.execute();
  callableStmt.close(); 
```

# Proof of Fix

```tsql
CREATE PROCEDURE foo
BEGIN
-- do stuff
END;
```


jmeunier28:
Looks like the tests failures aren't related to my PR?
```
* What went wrong:
Execution failed for task ':dd-java-agent:instrumentation:grizzly-http-2.3.20:test'.
> There were failing tests. See the report at: file:///home/circleci/dd-trace-java/workspace/dd-java-agent/instrumentation/grizzly-http-2.3.20/build/reports/tests/test/index.html
```

... although I'm not sure why this is failing now:
```
self = <test_db_integrations_sql.Test_Agent_Mssql_db_integration object at 0x7f05030259d0>

    def test_obfuscate_query(self):
        """ All queries come out obfuscated from agent """
        for db_operation, request in self.requests[self.db_service].items():
            span = self._get_sql_span_for_request(request)
            # We launch all queries with two parameters (from weblog)
            if db_operation == "insert":
                expected_obfuscation_count = 1
            elif db_operation == "procedure":
                # Insert and procedure:These operations also receive two parameters, but are obfuscated as only one.
                # Nodejs: The proccedure has a input parameter, but we are calling through method `execute`` and we can't see the parameters in the traces
                expected_obfuscation_count = 0 if context.library.library == "nodejs" else 2
            else:
                expected_obfuscation_count = 2
    
>           observed_obfuscation_count = span["meta"]["sql.query"].count("?")
E           TypeError: 'NoneType' object is not subscriptable

tests/integrations/test_db_integrations_sql.py:556: TypeError
```
https://app.circleci.com/pipelines/github/DataDog/dd-trace-java/30177/workflows/06d05af0-63ff-41e8-809e-a2c72d369297/jobs/1022674

justiniso (approved):

ygree (changes requested):
Looks good! Except for `integrations-core`. If it wasn't updated on purpose then follow https://github.com/DataDog/dd-trace-java/blob/master/CONTRIBUTING.md#git-submodule-setup to exclude it from the change.

jmeunier28:
> Looks good! Except for integrations-core. If it wasn't updated on purpose then follow https://github.com/DataDog/dd-trace-java/blob/master/CONTRIBUTING.md#git-submodule-setup to exclude it from the change.

@ygree  ah okay, I ran `git config --local submodule.recurse true` and then rebase on master, is this correct?

PerfectSlayer:
Looks like `integration-core` is still in your changeset.
Try `git submodule update` to update it and commit it.
You can then squash your 3 commits together and force push it to your PR branch.

ygree (approved):
Looks good to me!

PR #6044, opened 2023-10-12
Extract Kafka cluster ID as tag for DSM

hokitam:
# What Does This Do

Added new logic to Kafka instrumentation to collect Kafka cluster ID, and propagate that ID in the Data Streams Monitoring (DSM) payloads. This cluster ID will allow Data Streams Monitoring to perform more fine-tuned monitoring for Kafka workloads that happen across different clusters.

# Motivation
The Data Streams Monitoring metrics that track end-to-end latency and Kafka lag today do not keep track of the clusters that the workloads are communicating with. As such, if two workload in separate clusters have similar metadata (for example, similarly named services talking to similarly named Kafka topics that are situated in different regions), DSM is unable to automatically differentiate between them today. Adding the cluster ID as a tag to the DSM metrics will allow DSM to monitor workload across different Kafka clusters separately.

# Additional Notes

Jira ticket: [PROJ-IDENT]


dd-java-agent/instrumentation/kafka-clients-0.11/src/latestDepTest/groovy/KafkaClientTest.groovy
```diff
@@ -774,8 +791,11 @@ class KafkaClientTest extends AgentTestRunner {
     setup:
     def conditions = new PollingConditions(timeout: 10)
     def producerProps = KafkaTestUtils.producerProps(embeddedKafka.getBrokersAsString())
+    producerProps.put(ProducerConfig.METADATA_MAX_AGE_CONFIG, 1000)
     def producerFactory = new DefaultKafkaProducerFactory<String, String>(producerProps)
     def kafkaTemplate = new KafkaTemplate<String, String>(producerFactory)
+    kafkaTemplate.flush()
```
hokitam, line 797: We are setting the `METADATA_MAX_AGE_CONFIG` to 1s, such that `Metadata` updates once every second. We call kafkaTemplate.flush() to ensure that the producer is actually created under the hood (otherwise it wouldn't start updating the `Metadata`. Sleeping for 1.5s allow us to ensure that the metadata we need in the instrumentation would be updated within the 1s timeframe..

kr-igor: This test may be flaky, depending on how the metadata update is done. Maybe we can check for the data we need in a while loop, until we get the data or some reasonable timeout is reached?

dd-java-agent/instrumentation/kafka-clients-0.11/src/main/java/datadog/trace/instrumentation/kafka_clients/KafkaConsumerInstrumentation.java
```diff
@@ -110,8 +139,20 @@ public static void wrap(
         @Advice.Return(readOnly = false) Iterator<ConsumerRecord<?, ?>> iterator,
         @Advice.This ConsumerRecords records) {
       if (iterator != null) {
-        String group = InstrumentationContext.get(ConsumerRecords.class, String.class).get(records);
-        iterator = new TracingIterator(iterator, KAFKA_CONSUME, CONSUMER_DECORATE, group);
+        String group = null;
```
kr-igor, line 142: nit: it seems like this code is repeated several times, maybe we can extract it to a static method? Something like  `props = KafkaConsumerInfo.get(record)`

kr-igor (approved):

dd-java-agent/instrumentation/kafka-clients-0.11/src/latestDepTest/groovy/KafkaClientTest.groovy
hokitam: good idea, updated!

dd-java-agent/instrumentation/kafka-clients-0.11/src/main/java/datadog/trace/instrumentation/kafka_clients/KafkaConsumerInstrumentation.java
hokitam: updated by passing ContextStore as recommended by the APM Java team

dd-java-agent/instrumentation/kafka-clients-0.11/src/main/java/datadog/trace/instrumentation/kafka_clients/KafkaConsumerInfo.java
```diff
@@ -0,0 +1,60 @@
...
+  public String getConsumerGroup() {
+    return consumerGroup;
+  }
+
+  @Nullable
+  public Metadata getClientMetadata() {
+    return clientMetadata;
+  }
+
+  public static class Builder {
```
dougqh, line 26: Builders are rather wasteful in terms of overhead, but I suppose this is fine for now.

dd-java-agent/instrumentation/kafka-clients-0.11/src/main/java/datadog/trace/instrumentation/kafka_clients/KafkaConsumerInfo.java
```diff
@@ -0,0 +1,60 @@
...
+    if (this == o) return true;
+    if (o == null || getClass() != o.getClass()) return false;
+    KafkaConsumerInfo consumerInfo = (KafkaConsumerInfo) o;
+    return Objects.equals(consumerGroup, consumerInfo.consumerGroup)
+        && Objects.equals(clientMetadata, consumerInfo.clientMetadata);
+  }
+
+  @Override
+  public int hashCode() {
+    return Objects.hash(consumerGroup, clientMetadata);
```
dougqh, line 58: Objects.hash performs poorly.  If this is on the hot path, you should should implement hashCode in a more direct fashion.

dd-java-agent/instrumentation/kafka-clients-0.11/src/main/java/datadog/trace/instrumentation/kafka_clients/KafkaConsumerInstrumentation.java
```diff
@@ -30,7 +32,11 @@ public KafkaConsumerInstrumentation() {
 
   @Override
   public Map<String, String> contextStore() {
-    return singletonMap("org.apache.kafka.clients.consumer.ConsumerRecords", "java.lang.String");
+    Map<String, String> contextStores = new HashMap<>();
```
dougqh, line 35: Size the collection

dd-java-agent/instrumentation/kafka-clients-0.11/src/main/java/datadog/trace/instrumentation/kafka_clients/KafkaConsumerInstrumentation.java
```diff
@@ -96,8 +110,14 @@ public static void wrap(
         @Advice.Return(readOnly = false) List<ConsumerRecord<?, ?>> iterable,
         @Advice.This ConsumerRecords records) {
       if (iterable != null) {
-        String group = InstrumentationContext.get(ConsumerRecords.class, String.class).get(records);
-        iterable = new TracingList(iterable, KAFKA_CONSUME, CONSUMER_DECORATE, group);
+        Pair<String, String> data =
```
dougqh, line 113: This looks like extra overhead that is always on -- even when DSM is disabled.
That's a violation of our requirement that DSM imposes no overhead when not enabled.
I'd like to see this benchmarked to determine the cost.

dougqh (approved):
I'm concerned by the overhead being added into the instrumentation.
I'm particularly concerned that it looks like this overhead is being added all the time not just when DSM is enabled.

dd-java-agent/instrumentation/kafka-clients-0.11/src/main/java/datadog/trace/instrumentation/kafka_clients/KafkaConsumerInstrumentation.java
hokitam: I benchmarked this yesterday using JMH. For the produce call I simply called `kafkaProducer.send` in the benchmark method, whereas to benchmark the consume call I pre-sent 100k messages in setup and then called `kafkaConsume.poll` in the benchmark method. Here is the results:

Screen Shot 2023-10-25 at 9 42 20 AM

dd-java-agent/instrumentation/kafka-clients-0.11/src/test/groovy/KafkaClientTestBase.groovy
```diff
@@ -95,7 +95,19 @@ abstract class KafkaClientTestBase extends VersionedNamingTestBase {
...
+    if (isDataStreamsEnabled()) {
+      senderProps.put(ProducerConfig.METADATA_MAX_AGE_CONFIG, 1000)
+    }
+    KafkaProducer<String, String> producer = new KafkaProducer<>(senderProps, new StringSerializer(), new StringSerializer())
+    String clusterId = ""
+    if (isDataStreamsEnabled()) {
+      producer.flush()
+      clusterId = producer.metadata.cluster.clusterResource().clusterId()
+      while (clusterId == null || clusterId.isEmpty()) {
+        Thread.sleep(1500)
```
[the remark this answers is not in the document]
hokitam: We can also just remove the Thread.sleep and it would still work, because we're just waiting for the data to come back from Embedded kafka. Otherwise, we might also be able to add resource listeners by directly accessing it from the `metadata` (https://github.com/apache/kafka/blob/0.11.0/clients/src/main/java/org/apache/kafka/clients/Metadata.java#L99) But then we'll have to run the test inside the listener, which I'm not quite sure if it's feasible.

lmk what your thoughts are!

dd-java-agent/instrumentation/kafka-clients-0.11/src/main/java/datadog/trace/instrumentation/kafka_clients/KafkaConsumerInstrumentationHelper.java
```diff
@@ -0,0 +1,25 @@
+package datadog.trace.instrumentation.kafka_clients;
+
+import datadog.trace.api.Pair;
+import datadog.trace.bootstrap.ContextStore;
+import org.apache.kafka.clients.Metadata;
+import org.apache.kafka.clients.consumer.ConsumerRecords;
+
+public class KafkaConsumerInstrumentationHelper {
+  public static Pair<String, String> extractGroupAndClusterId(
```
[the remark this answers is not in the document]
hokitam: refactored to not use a Pair anymore. Though we technically still have a Pair-like structure in KafkaConsumerInfo

dd-java-agent/instrumentation/kafka-clients-0.11/src/main/java/datadog/trace/instrumentation/kafka_clients/KafkaConsumerInstrumentation.java
hokitam: done

dd-java-agent/instrumentation/kafka-clients-0.11/src/main/java/datadog/trace/instrumentation/kafka_clients/KafkaConsumerInfo.java
hokitam: we only use KafkaConsumerInfo as a value to the context store, not as a key, so it shouldn't be on the hot path, but updated the implementation to match what `Pair` does just in case.

dd-java-agent/instrumentation/kafka-clients-0.11/src/main/java/datadog/trace/instrumentation/kafka_clients/KafkaConsumerInfo.java
hokitam: removed the Builder

PR #6045, opened 2023-10-13
Track payload size for Kafka integration

vandonr:
This is mirroring part of the features that were added to the .net tracer in https://github.com/DataDog/dd-trace-dotnet/pull/4520, on DSM team's request.

tested locally with a sample producer and consumer, I can see the metric being emited in my DD org
image

Jira ticket: [AIT-8557]

[AIT-8557]: https://datadoghq.atlassian.net/browse/AIT-8557?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ


dd-java-agent/instrumentation/kafka-streams-0.11/src/main/java/datadog/trace/instrumentation/kafka_streams/KafkaStreamTaskInstrumentation.java
```diff
@@ -330,6 +333,16 @@ public static void start(
     }
   }
 
+  private static long computePayloadSizeBytes(ConsumerRecord<?, ?> val) {
```
vandonr, line 336: this code is duplicated from the instrumentation for kafka_clients, but I don't really know where to put it to share it between the two integrations

dd-java-agent/instrumentation/rabbitmq-amqp-2.7/src/main/java/datadog/trace/instrumentation/rabbitmq/amqp/RabbitDecorator.java
```diff
@@ -258,6 +260,22 @@ public static AgentScope startReceivingSpan(
...
+    for (Map.Entry<String, Object> kv : headers.entrySet()) {
+      size += kv.getKey().getBytes().length;
+
+      Object v = kv.getValue();
+      if (v instanceof String) {
+        size += ((String) v).getBytes().length;
+      } else if (v instanceof byte[]) {
+        size += ((byte[]) v).length;
+      }
+      // else no idea how to handle the value, don't compute size.
```
vandonr, line 278: didn't know how to handle `Object` so I took inspiration from how it's done in .net
https://github.com/DataDog/dd-trace-dotnet/blob/a0f98c75c25e195c39bd53157ad2a138128ac159/tracer/src/Datadog.Trace/ClrProfiler/AutoInstrumentation/RabbitMQ/RabbitMQIntegration.cs#L92-L116

dd-trace-core/src/main/java/datadog/trace/core/datastreams/DefaultDataStreamsMonitoring.java
```diff
@@ -204,7 +205,10 @@ public void trackBacklog(LinkedHashMap<String, String> sortedTags, long value) {
 
   @Override
   public void setCheckpoint(
-      AgentSpan span, LinkedHashMap<String, String> sortedTags, long defaultTimestamp) {
+      AgentSpan span,
+      LinkedHashMap<String, String> sortedTags,
+      long defaultTimestamp,
+      long payloadSizeBytes) {
```
vandonr, line 211: maybe it'd make sense to make this new parameter optional given how many places I had to put zero.

dd-java-agent/instrumentation/kafka-streams-0.11/src/main/java/datadog/trace/instrumentation/kafka_streams/KafkaStreamTaskInstrumentation.java
```diff
@@ -309,7 +312,7 @@ public static void start(
         sortedTags.put(TYPE_TAG, "kafka");
         AgentTracer.get()
             .getDataStreamsMonitoring()
-            .setCheckpoint(span, sortedTags, record.timestamp());
+            .setCheckpoint(span, sortedTags, record.timestamp(), 0);
```
vandonr, line 315: This one, the version `27` of the method, has a type that does't expose much (`ProcessorRecordContext`), so I didn't see how to compute the payload size for this.

internal-api/src/main/java/datadog/trace/bootstrap/instrumentation/api/StatsPoint.java
```diff
@@ -9,20 +9,23 @@ public class StatsPoint implements InboxItem {
...
-      long edgeLatencyNano) {
+      long edgeLatencyNano,
+      long payloadSizeBytes) {
     this.edgeTags = edgeTags;
     this.hash = hash;
     this.parentHash = parentHash;
     this.timestampNanos = timestampNanos;
     this.pathwayLatencyNano = pathwayLatencyNano;
     this.edgeLatencyNano = edgeLatencyNano;
+    this.payloadSizeBytes = payloadSizeBytes;
```
vandonr, line 28: I still have to propagate this somewhere where it'll be sent to DD

dd-trace-core/src/main/java/datadog/trace/core/datastreams/DefaultPathwayContext.java
```diff
@@ -94,15 +94,18 @@ public long getHash() {
...
+    setCheckpoint(sortedTags, pointConsumer, 0, payloadSizeBytes);
   }
 
   @Override
   public void setCheckpoint(
       LinkedHashMap<String, String> sortedTags,
       Consumer<StatsPoint> pointConsumer,
-      long defaultTimestamp) {
+      long defaultTimestamp,
+      long payloadSizeBytes) {
```
amarziali, line 108: I would probably have left the original method signature since it's public. Instead, you can create another signature with also this parameter so it won't create signature mismatches.

dd-java-agent/instrumentation/kafka-clients-0.11/src/main/java/datadog/trace/instrumentation/kafka_clients/TracingIterator.java
```diff
@@ -111,6 +113,16 @@ protected void startNewRecordSpan(ConsumerRecord<?, ?> val) {
     }
   }
 
+  private static long computePayloadSizeBytes(ConsumerRecord<?, ?> val) {
+    long headersSize = 0;
+    Headers headers = val.headers();
+    if (headers != null)
+      for (Header h : headers) {
+        headersSize += h.value().length + h.key().getBytes().length;
```
amarziali, line 121: looking on kafka-clients code (in the part where the headers are written):
```
            byte[] utf8Bytes = Utils.utf8(headerKey);
```
So the key (that's a string in java) it's assumed to be encoded in UTF8. In this charset, generally speaking, the length of a string is not equal to the number of bytes needed to encode it

vandonr: Yeah ok, I just need to specify the UTF8 encoding as a parameter of `getBytes` then

dd-java-agent/instrumentation/grpc-1.5/src/main/java/datadog/trace/instrumentation/grpc/server/TracingServerInterceptor.java
```diff
@@ -67,7 +67,9 @@ public <ReqT, RespT> ServerCall.Listener<ReqT> interceptCall(
     CallbackProvider cbp = tracer.getCallbackProvider(RequestContextSlot.APPSEC);
     final AgentSpan span = startSpan(GRPC_SERVER, spanContext).setMeasured(true);
 
-    AgentTracer.get().getDataStreamsMonitoring().setCheckpoint(span, SERVER_PATHWAY_EDGE_TAGS, 0);
+    AgentTracer.get()
+        .getDataStreamsMonitoring()
+        .setCheckpoint(span, SERVER_PATHWAY_EDGE_TAGS, 0, 0);
```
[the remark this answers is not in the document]
amarziali: If you don't know the payload size (or if you don't need to calculate for some integrations) what should be reported as payload statistics? 0 or just nothing? Since in this case the semantic to use on the API is different.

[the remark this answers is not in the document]
vandonr: I think we can register it as zero internally, and then exclude zeros in the reporting code since I don't think a payload of size zero is a realistic value (given the fact that we consider headers)

dd-java-agent/instrumentation/armeria-grpc/src/main/java/datadog/trace/instrumentation/armeria/grpc/server/TracingServerInterceptor.java
```diff
@@ -68,7 +68,9 @@ public <ReqT, RespT> ServerCall.Listener<ReqT> interceptCall(
     final AgentSpan span =
         startSpan(DECORATE.instrumentationNames()[0], GRPC_SERVER, spanContext).setMeasured(true);
 
-    AgentTracer.get().getDataStreamsMonitoring().setCheckpoint(span, SERVER_PATHWAY_EDGE_TAGS, 0);
+    AgentTracer.get()
+        .getDataStreamsMonitoring()
+        .setCheckpoint(span, SERVER_PATHWAY_EDGE_TAGS, 0, 0);
```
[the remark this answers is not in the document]
tabgok: The focus is on Kafka for now, so I believe gRPC can be put aside for now.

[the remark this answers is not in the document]
vandonr: yes I realized that later :p

dd-java-agent/instrumentation/kafka-streams-0.11/src/main/java/datadog/trace/instrumentation/kafka_streams/KafkaStreamTaskInstrumentation.java
kr-igor: I believe you can get precise size by using `serializedKeySize+serializedValueSize` from `RecordMetadata` (ProcessorRecordContext is inherited from it)
https://github.com/a0x8o/kafka/blob/54eff6af115ee647f60129f2ce6a044cb17215d0/clients/src/main/java/org/apache/kafka/clients/producer/RecordMetadata.java#L112

vandonr: Ha, so this is troubling, the inheritance of `RecordMetadata` was added in https://github.com/apache/kafka/commit/69790a1463bafc1f63e3c288a6636b3f4586c1b4#diff-85c25edc2dbc410907fe3d4706dc5799005e0e70950af191ce8a4d70dbf3f650L32
which was merged 3 years ago. 
However, the jar I have in my local project, which was resolved by gradle I think, is kafka-streams 0.11 from Jun 22, 2017. I suppose this is because that's the minimal version we're supporting ?

What should we do from there ? Can we bump the version ? How can we do something where clients get the size only if they have a more recent kafka ?

vandonr: If I judge just from the dates, that inheritance should be there for versions >= 2.6.1

PerfectSlayer:
Context? Goal? Description?

vandonr:
> Context? Goal? Description?

I added a little paragraph in the description, but I don't have a whole lot of info to add.

internal-api/src/main/java/datadog/trace/bootstrap/instrumentation/api/AgentDataStreamsMonitoring.java
```diff
@@ -19,6 +19,24 @@ public interface AgentDataStreamsMonitoring extends DataStreamsCheckpointer {
...
+   *     Value will be ignored for checkpoints happening not at the start of the pipeline.
+   * @param payloadSizeBytes size of the message (body + headers) in bytes. Zero should be passed if
+   *     the size cannot be evaluated.
+   */
+  void setCheckpoint(
+      AgentSpan span,
+      LinkedHashMap<String, String> sortedTags,
+      long defaultTimestamp,
+      long payloadSizeBytes);
+
```
PerfectSlayer, line 39: Can you avoid overloading this method? (not creating a duplicate with different parameters)
It is usually a good practice but we already have too many methods in the _tracer_ classes.

There are less than 10 different calls for those methods so it will be fine to add 0 as last default parameter in all calls. Thanks! 

Note: It's fine in `PathwayContext` as it will only impact DSM object only 👌

dd-java-agent/instrumentation/kafka-clients-0.11/src/main/java/datadog/trace/instrumentation/kafka_clients/Utils.java
```diff
@@ -0,0 +1,19 @@
+package datadog.trace.instrumentation.kafka_clients;
+
+import java.nio.charset.StandardCharsets;
+import org.apache.kafka.clients.consumer.ConsumerRecord;
+import org.apache.kafka.common.header.Header;
+import org.apache.kafka.common.header.Headers;
+
+public class Utils {
```
PerfectSlayer, line 8: What about making the class final with a private constructor?

PerfectSlayer (changes requested):
One change request for internal tracer API.
I will let the two other default reviewiers review the PR.

internal-api/src/main/java/datadog/trace/bootstrap/instrumentation/api/AgentDataStreamsMonitoring.java
vandonr: hmm ok, the initial reasoning was that since this code is in `/api/` we wanted to avoid breaking changes to the interfaces.

PerfectSlayer: You are inside `internal-api`. So it is safe to change it if you take care of other teams relying on it (only DSM here). Thanks for caring 🙏

PerfectSlayer (commented):
Thanks for the API edition

kr-igor (approved):
LGTM

dd-java-agent/instrumentation/rabbitmq-amqp-2.7/src/main/java/datadog/trace/instrumentation/rabbitmq/amqp/RabbitDecorator.java
```diff
@@ -247,7 +247,9 @@ public static AgentScope startReceivingSpan(
       sortedTags.put(DIRECTION_TAG, DIRECTION_IN);
       sortedTags.put(TOPIC_TAG, queue);
       sortedTags.put(TYPE_TAG, "rabbitmq");
-      AgentTracer.get().getDataStreamsMonitoring().setCheckpoint(span, sortedTags, produceMillis);
+      AgentTracer.get()
+          .getDataStreamsMonitoring()
+          .setCheckpoint(span, sortedTags, produceMillis, 0);
```
amarziali, line 252: Nit: since you have the signature without the payload size you can just use it (and leave that class unchanged). It will generate a smaller diff (the same applied for all the other decorators)

amarziali (approved):
Looks ok. I left a comment. Please let apm-java also review that PR

dd-java-agent/instrumentation/rabbitmq-amqp-2.7/src/main/java/datadog/trace/instrumentation/rabbitmq/amqp/RabbitDecorator.java
vandonr: no I don't have it anymore, @PerfectSlayer asked in https://github.com/DataDog/dd-trace-java/pull/6045#discussion_r1369804322 that I remove it !

dd-trace-core/src/main/java/datadog/trace/core/datastreams/StatsGroup.java
```diff
@@ -12,18 +12,23 @@ public class StatsGroup {
...
+    payloadSize = Histograms.newHistogram();
   }
 
-  public void add(long pathwayLatencyNano, long edgeLatencyNano) {
+  public void add(long pathwayLatencyNano, long edgeLatencyNano, long payloadSizeBytes) {
     pathwayLatency.accept(((double) pathwayLatencyNano) / NANOSECONDS_TO_SECOND);
     edgeLatency.accept(((double) edgeLatencyNano) / NANOSECONDS_TO_SECOND);
+    // payload size is set to zero when we cannot compute it
+    // in that case, it's probably better to have an empty histogram than filling it with zeros
+    if (payloadSizeBytes != 0) payloadSize.accept((double) payloadSizeBytes);
```
PerfectSlayer, line 31: I don't even know spotless would agree to use such kind of one-liner... 🤔

PerfectSlayer (approved):
LGTM

PR #6048, opened 2023-10-13
Introduce PII redaction based on keywords

jpbempel:
# What Does This Do
Redacts values based on a list of predefined words name are normalized in lower case and by removing special characters: `@`, `$`, `-` and `_`
Redaction happening at:
 - expression language for reference, index and getmember operations
 - serialization for snapshot or template values including maps
 - capture of the values from instrumentation For conditions (log or span decoration probes) evaluation of a redacted values triggers an evaluation exception that will be reported as evaluation error into a snapshot.

# Motivation
Prevent leakage of sensitive information like passwords, api keys, tokens, ...

# Additional Notes

Jira ticket: [DEBUG-1838]

[DEBUG-1838]: https://datadoghq.atlassian.net/browse/DEBUG-1838?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ

### DIFF
--- a/dd-java-agent/agent-debugger/debugger-el/src/main/java/com/datadog/debugger/el/ProbeCondition.java
+++ b/dd-java-agent/agent-debugger/debugger-el/src/main/java/com/datadog/debugger/el/ProbeCondition.java
@@ -11,6 +11,7 @@ import datadog.trace.bootstrap.debugger.el.DebuggerScript;
 import datadog.trace.bootstrap.debugger.el.ValueReferenceResolver;
 import edu.umd.cs.findbugs.annotations.NonNull;
 import java.io.IOException;
+import java.util.Map;
 
 /** Implements expression language for probe condition */
 public final class ProbeCondition implements DebuggerScript<Boolean> {
@@ -103,14 +104,55 @@ public final class ProbeCondition implements DebuggerScript<Boolean> {
     if (when == null) {
       return true;
     }
-    if (when.evaluate(valueRefResolver)) {
-      then.evaluate(valueRefResolver);
+    ProbeConditionResolver conditionResolver = new ProbeConditionResolver(valueRefResolver);
+    if (when.evaluate(conditionResolver)) {
+      then.evaluate(conditionResolver);
+      conditionResolver.checkRedacted();
       return true;
     }
+    conditionResolver.checkRedacted();
     return false;
   }
 
   public void accept(Visitor visitor) {
     when.accept(visitor);
+  }
+
+  private static class ProbeConditionResolver implements ValueReferenceResolver {
+    private final ValueReferenceResolver delegate;
+    private String expr;
+    private boolean redacted;
+
+    public ProbeConditionResolver(ValueReferenceResolver delegate) {
+      this.delegate = delegate;
+    }
+
+    @Override
+    public Object lookup(String name) {
+      return delegate.lookup(name);
+    }
+
+    @Override
+    public Object getMember(Object target, String name) {
+      return delegate.getMember(target, name);
+    }
+
+    @Override
+    public void redacted(String expr) {
+      this.redacted = true;
+      this.expr = expr;
+    }
+
+    public void checkRedacted() {
+      if (redacted) {
+        throw new EvaluationException(
+            "Could not evaluate the expression because '" + expr + "' was redacted", expr);
+      }
+    }
+
+    @Override
+    public ValueReferenceResolver withExtensions(Map<String, Object> extensions) {
+      return delegate.withExtensions(extensions);
+    }
   }
 }
--- a/dd-java-agent/agent-debugger/debugger-el/src/main/java/com/datadog/debugger/el/ProbeCondition.java
+++ b/dd-java-agent/agent-debugger/debugger-el/src/main/java/com/datadog/debugger/el/ProbeCondition.java
@@ -11,7 +11,6 @@ import datadog.trace.bootstrap.debugger.el.DebuggerScript;
 import datadog.trace.bootstrap.debugger.el.ValueReferenceResolver;
 import edu.umd.cs.findbugs.annotations.NonNull;
 import java.io.IOException;
-import java.util.Map;
 
 /** Implements expression language for probe condition */
 public final class ProbeCondition implements DebuggerScript<Boolean> {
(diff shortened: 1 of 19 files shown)


shatzi (commented):
Overall I like the implementation. I think we can simplify the expression evaluation logic by simple throw an exception and avoid all the redacted callback.

dd-java-agent/agent-debugger/debugger-el/src/main/java/com/datadog/debugger/el/ProbeCondition.java
```diff
@@ -103,14 +104,55 @@ public Boolean execute(ValueReferenceResolver valueRefResolver) {
...
+
+    @Override
+    public void redacted(String expr) {
+      this.redacted = true;
+      this.expr = expr;
+    }
+
+    public void checkRedacted() {
+      if (redacted) {
+        throw new EvaluationException(
```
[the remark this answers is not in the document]
jpbempel: Yes as stated in the RFC

dd-java-agent/agent-debugger/debugger-el/src/main/java/com/datadog/debugger/el/expressions/GetMemberExpression.java
```diff
@@ -23,7 +25,13 @@ public Value<?> evaluate(ValueReferenceResolver valueRefResolver) {
       return targetValue;
     }
     try {
-      return Value.of(valueRefResolver.getMember(targetValue.getValue(), memberName));
+      Object member = valueRefResolver.getMember(targetValue.getValue(), memberName);
+      if (member == Redaction.REDACTED_VALUE) {
```
shatzi, line 29: so, if there is a normal string with value "REDACTED"? I think we might need to have something that is not comparable to normal strings.

dd-java-agent/agent-debugger/debugger-el/src/main/java/com/datadog/debugger/el/expressions/ValueRefExpression.java
```diff
@@ -22,7 +22,9 @@ public Value<?> evaluate(ValueReferenceResolver valueRefResolver) {
     try {
       Object symbol = valueRefResolver.lookup(symbolName);
       if (symbol == Redaction.REDACTED_VALUE) {
-        valueRefResolver.redacted(PrettyPrintVisitor.print(this));
+        String expr = PrettyPrintVisitor.print(this);
+        throw new EvaluationException(
```
shatzi, line 26: the expr value would be overwritten by the catch expression as symbolName

dd-java-agent/agent-debugger/debugger-el/src/main/java/com/datadog/debugger/el/expressions/IndexExpression.java
```diff
@@ -34,8 +34,9 @@ public Value<?> evaluate(ValueReferenceResolver valueRefResolver) {
       if (targetValue instanceof MapValue) {
         Object objKey = keyValue.getValue();
         if (objKey instanceof String && Redaction.isRedactedKeyword((String) objKey)) {
-          valueRefResolver.redacted(PrettyPrintVisitor.print(this));
-          result = Value.of(Redaction.REDACTED_VALUE);
+          String expr = PrettyPrintVisitor.print(this);
+          throw new EvaluationException(
```
shatzi, line 38: the catch exception would call PrettyPrintVisitor the second time.

dd-java-agent/agent-debugger/debugger-el/src/main/java/com/datadog/debugger/el/expressions/IndexExpression.java
```diff
@@ -31,7 +32,14 @@ public Value<?> evaluate(ValueReferenceResolver valueRefResolver) {
...
     try {
       if (targetValue instanceof MapValue) {
-        result = ((MapValue) targetValue).get(keyValue.getValue());
+        Object objKey = keyValue.getValue();
+        if (objKey instanceof String && Redaction.isRedactedKeyword((String) objKey)) {
+          String expr = PrettyPrintVisitor.print(this);
+          throw new EvaluationException(
+              "Could not evaluate the expression because '" + expr + "' was redacted", expr);
+        } else {
+          result = ((MapValue) targetValue).get(objKey);
```
shatzi, line 41: is the get(objKey) is an alien call?

dd-java-agent/agent-debugger/debugger-el/src/main/java/com/datadog/debugger/el/expressions/GetMemberExpression.java
shatzi: Maybe consider that CaptureContext.getMember() would throw the exception?
That can avoid refactoring this class

shatzi (commented):
I think we have a bug with REDACTED_VALUE comparison.
Few comments about refactoring.

dd-java-agent/agent-debugger/debugger-el/src/main/java/com/datadog/debugger/el/expressions/GetMemberExpression.java
jpbempel: >s o, if there is a normal string with value "REDACTED"? I think we might need to have something that is not comparable to normal strings.
that's why I am using == and not `equals`. we are doing reference equality here, so the instance but be exactly the same. though, I need to create a specific string to avoid internalization by the JVM.

jpbempel: > Maybe consider that CaptureContext.getMember() would throw the exception? That can avoid refactoring this class

Yes in this case I think it would work

dd-java-agent/agent-debugger/debugger-el/src/main/java/com/datadog/debugger/el/expressions/ValueRefExpression.java
jpbempel: true I need to fix the catch handler

dd-java-agent/agent-debugger/debugger-el/src/main/java/com/datadog/debugger/el/expressions/IndexExpression.java
jpbempel: indirectly in `MapValue::get` for a `Map`
same for ListValue
But allowing `Map::get` or `List::get` only for JDK classes maybe too restrictive.

dd-java-agent/agent-debugger/debugger-el/src/main/java/com/datadog/debugger/el/expressions/GetMemberExpression.java
jpbempel: well after thought it feels clunky here, because it will be the only place where I throw a runtime exception for that from the CapturedContext class, while for the other case we are passing through the `REDACTED_VALUE`. so it's less uniform.

shatzi (approved):

PR #6033, opened 2023-10-11
Refactor of the PropagationModule and removal of WebModule

manuel-alvarez-alvarez:
# What Does This Do
Refactoring of the tainting API by moving all tainting operations to the `PropagationModule` with methods according to:

1. `taint` methods are used to taint values : `taint(context?, target, source_attributes+)`
2. `taintIfTainted` methods are used to taint values only if the input is tainted : `taintIfTainted(context?, target, input, source_attributes+)`
3. `taintIfAnyTainted` methods are used to taint values only if any of the inputs is tainted : `taintIfAnyTainted(context?, target, input[], source_attributes+)`

# Motivation
Both modules `WebModule` and `PropagationModule` have grown organically bringing a lot of duplication, this PR tries to merge all tainting operations providing a cohesive API.

# Additional Notes
This PR is quite big but it does not introduce any new functionality, just a reorganization of the tainting API.

Jira ticket: [APPSEC-10454]

[APPSEC-10454]: https://datadoghq.atlassian.net/browse/APPSEC-10454?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ

PR #6031, opened 2023-10-11
IAST Servlet.getRequestURL instrumentation

DDJavierSantos:
# What Does This Do
Adds instrumentation to Servlet.getServletURL call

# Motivation
To be able to taint the output of the function so it can be used by IAST to detect vulnerabilities.

# Additional Notes

Jira ticket: [APPSEC-11729](https://datadoghq.atlassian.net/browse/APPSEC-11729)

[APPSEC-11729]: https://datadoghq.atlassian.net/browse/APPSEC-11729?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ

PR #6050, opened 2023-10-16
Fix IAST ranges for StringBuffer.toString

smola:
# What Does This Do
A StringBuffer or StringBuilder object might be tainted directly. In that case, we'll use an open Range (signaling that the full object is tainted, without a specific length). On StringBuffer#toString, this will lead to a String with an open Range, which the rest of our code base does not expect.

# Motivation
This is required to be able to taint StringBuffers in IAST sources (but probably cannot be triggered with any released version).

# Additional Notes

Jira ticket: [APPSEC-11732](https://datadoghq.atlassian.net/browse/APPSEC-11732)

[APPSEC-11732]: https://datadoghq.atlassian.net/browse/APPSEC-11732?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ

### DIFF
--- a/dd-java-agent/agent-iast/src/main/java/com/datadog/iast/model/Range.java
+++ b/dd-java-agent/agent-iast/src/main/java/com/datadog/iast/model/Range.java
@@ -81,4 +81,9 @@ public final class Range implements Ranged {
   public boolean isMarked(final int mark) {
     return (marks & mark) != NOT_MARKED;
   }
+
+  /** Check if this is an "unbound" range, covering the whole object. */
+  public boolean isUnbound() {
+    return start == 0 && length == Integer.MAX_VALUE;
+  }
 }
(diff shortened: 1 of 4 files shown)


manuel-alvarez-alvarez (approved):
Tainting a `StringBuilder` `StringBuffer` without bounds should not be the case anymore with the new refactor of the sources (they should be treated as `CharSequence`):
https://github.com/DataDog/dd-trace-java/pull/6033

smola:
@manuel-alvarez-alvarez Indeed. In fact, I think I'd prefer to not merge this PR, and hold https://github.com/DataDog/dd-trace-java/pull/6031, and instead go for tainting StringBuilder/StringBuffer as CharSequence, with proper bound start/length ranges. Then `StringBuffer#toString()` would not need to special-case this.

jandro996 (approved):

smola:
Not a proper fix. This will be superseded by handling of `CharSequence` objects in sources.

PR #6061, opened 2023-10-18
[SVLS-579] Add span error metadata to DD Lambda Extension request headers

DylanLovesCoffee:
# What Does This Do
Adds span error tags (`error.msg`, `error.type` and `error.stack`) as request headers made to the DD Lambda Extension's `/lambda/end-invocation` endpoint. 

This is a part of the changes made to the DD Extension: https://github.com/DataDog/datadog-agent/pull/20235 

# Motivation
Explicitly setting error tags in Lambda handler code does not currently get carried over to the Lambda extension.

# Additional Notes

Jira ticket: [SVLS-579]

### DIFF
--- a/dd-trace-core/src/main/java/datadog/trace/lambda/LambdaHandler.java
+++ b/dd-trace-core/src/main/java/datadog/trace/lambda/LambdaHandler.java
@@ -11,6 +11,7 @@ import datadog.trace.bootstrap.instrumentation.api.AgentSpan;
 import datadog.trace.core.propagation.ExtractedContext;
 import datadog.trace.core.propagation.PropagationTags;
 import java.io.ByteArrayInputStream;
+import java.util.Base64;
 import okhttp3.MediaType;
 import okhttp3.OkHttpClient;
 import okhttp3.Request;
@@ -36,6 +37,9 @@ public class LambdaHandler {
   private static final String DATADOG_SPAN_ID = "x-datadog-span-id";
   private static final String DATADOG_SAMPLING_PRIORITY = "x-datadog-sampling-priority";
   private static final String DATADOG_INVOCATION_ERROR = "x-datadog-invocation-error";
+  private static final String DATADOG_INVOCATION_ERROR_MSG = "x-datadog-invocation-error-msg";
+  private static final String DATADOG_INVOCATION_ERROR_TYPE = "x-datadog-invocation-error-type";
+  private static final String DATADOG_INVOCATION_ERROR_STACK = "x-datadog-invocation-error-stack";
   private static final String DATADOG_TAGS_KEY = "x-datadog-tags";
 
   private static final String START_INVOCATION = "/lambda/start-invocation";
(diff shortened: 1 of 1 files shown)


dd-trace-core/src/main/java/datadog/trace/lambda/LambdaHandler.java
```diff
@@ -115,12 +119,28 @@ public static boolean notifyEndInvocation(AgentSpan span, Object result, boolean
     Request.Builder builder =
         new Request.Builder()
             .url(EXTENSION_BASE_URL + END_INVOCATION)
-            .addHeader(DATADOG_META_LANG, "java")
```
DylanLovesCoffee, line 118: removed a duplication here

purple4reina (changes requested):
Tests?

purple4reina (approved):

dd-trace-core/src/main/java/datadog/trace/lambda/LambdaHandler.java
```diff
@@ -115,12 +119,28 @@ public static boolean notifyEndInvocation(AgentSpan span, Object result, boolean
...
+    }
+
+    Object errorType = span.getTag("error.type");
+    if (errorType != null) {
+      builder.addHeader(DATADOG_INVOCATION_ERROR_TYPE, errorType.toString());
+    }
+
+    Object errorStack = span.getTag("error.stack");
+    if (errorStack != null) {
+      String encodedErrStack = Base64.getEncoder().encodeToString(errorStack.toString().getBytes());
```
PerfectSlayer, line 140: Is it safe to rely on default charset? It changes according Java versions.

dd-trace-core/src/test/groovy/datadog/trace/lambda/LambdaHandlerTest.groovy
```diff
@@ -127,6 +127,49 @@ class LambdaHandlerTest extends DDCoreSpecification {
...
+    server.lastRequest.headers.get("x-datadog-trace-id") == tIdHeaderValue
+    server.lastRequest.headers.get("x-datadog-span-id") == sIdHeaderValue
+    server.lastRequest.headers.get("x-datadog-sampling-priority") == sPIdHeaderValue
+    result == expected
+
+    cleanup:
+    server.close()
+
+    where:
+    expected | eHeaderValue | eMsgHeaderValue        | eTypeHeaderValue      | eStackHeaderValue          | tIdHeaderValue | sIdHeaderValue | sPIdHeaderValue | lambdaResult | boolValue
```
PerfectSlayer, line 168: Why using data tables if values does not changes and are not used in your test definition?

dd-trace-core/src/main/java/datadog/trace/lambda/LambdaHandler.java
DylanLovesCoffee: Ah, thanks. Good catch. I think we'll need to be explicit here.

dd-trace-core/src/test/groovy/datadog/trace/lambda/LambdaHandlerTest.groovy
DylanLovesCoffee: 👍 good point

PerfectSlayer (approved):
Thanks for the changes! LGTM

PR #6056, opened 2023-10-18
Taint gson sources (deserialization)

jandro996:
# What Does This Do

Add instrumentations related with gson v.16 and greater versions --> `com.google.gson.stream.JsonReader`

We also need propagation in java.io.StringReader#<init>(java.lang.String) as it's used internally by gson

# Motivation

Add IAST propagation to gson library

# Additional Notes

v1.1 and v1.4 will be supported in different PRs due to the need of a mechanism to remove tainted objects

Jira ticket: [APPSEC-11185]

[APPSEC-11185]: https://datadoghq.atlassian.net/browse/APPSEC-11185?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ

PR #6082, opened 2023-10-23
Taint gson sources (deserialization) v1.1

jandro996:
# What Does This Do

Add  instrumentation to `com.google.gson.JsonParser`

# Motivation

Add gson source propagation for v1.1

# Additional Notes

related with https://github.com/DataDog/dd-trace-java/pull/6056

This PR is not completed, for ReInit method instrumentation  we need to check if the input is tainted and if check if the JsonParser instance is tainted and remove it from the tainted objects.
Right now there is no mechanism to allow the removal of tainted objects (This will impact performance)

Jira ticket: [PROJ-IDENT]


jandro996:
This PR will be resumed one we are able to remove elements from tainted objects

PR #6088, opened 2023-10-24
[WIP] Add support for dependency detection in Knopflerfish

smola:
# What Does This Do

# Motivation

# Additional Notes

* Jira ticket: [APPSEC-11213]
* Related PR: https://github.com/DataDog/dd-trace-java/pull/6085

PR #6090, opened 2023-10-24
Refactor dependency resolving in telemetry

smola:
# What Does This Do
* Avoid back and forth conversions between URL, String, URI, just
  use URL directly and do early checks to enforce it's either file:
  or jar:file: by the time it is enqueued.
* Remove code redundancies, and move dependency resolving logic to
  DependencyResolver.

# Motivation
Originally meant to ease Knopflerfish support (https://github.com/DataDog/dd-trace-java/pull/6088), but moved to a more general clean up.

# Additional Notes

Jira ticket: [PROJ-IDENT]


smola:
Given the benchmark results, I'm closing this. I'll just bring back some smaller changes if/when really required.

PR #6093, opened 2023-10-24
Jmeunier/dbmon 2759 take two

jmeunier28:
# What Does This Do

This PR updates the logic for the APM-DBM link feature introduced in https://github.com/DataDog/dd-trace-java/pull/4847. It changes from prepending the sql comments, containing the APM trace context, to appending them to the raw sql. This requires no changes on the DBM side of things in either the agent or our backend.

# Motivation

```java
  callableStmt.setInt(1,10);
  callableStmt.setInt(2,11);
  callableStmt.execute();
  callableStmt.close(); 
```

# Proof of Fix

```tsql
CREATE PROCEDURE foo
BEGIN
-- do stuff
END;
```


PerfectSlayer:
Duplicate of #6034

PR #6094, opened 2023-10-24
Improve IAST coverage by not relying on an active context for tainting purposes

manuel-alvarez-alvarez:
# What Does This Do
Adds a new feature flag to IAST in order to scope the tainted map globally instead of a map per request. 

The main changes are:

- New feature flag `dd.iast.context.mode` with values `REQUEST` (by default) or `GLOBAL`
- When in GLOBAL mode, sampling of requests is automatically removed (concurrent requests stays the same)
- When in GLOBAL mode, there is no pool of tainted objects, only a singleton instance exists
- When in GLOBAL mode, the size of the tainted objects is calculated as `4 * default_size`
- When in GLOBAL mode, we lose the metric for the number of tainteds created in a request 

# Motivation
When the tainted map is scoped to a request, only taint operations with an active span are executed, which brings two problems:

- If there's is no active context we will lose the taint
- Accessing the thread local per taint operation can be slow in high propagation scenarios

A global instance will help solving both issues but it will introduce other constraints of its own, e.g. concurrency increases and the possibility of losing taints due to collisions in the map is way bigger.

# Additional Notes

Jira ticket: [APPSEC-11483]

[APPSEC-11483]: https://datadoghq.atlassian.net/browse/APPSEC-11483?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ

### DIFF
--- a/benchmark/load/insecure-bank/benchmark.json
+++ b/benchmark/load/insecure-bank/benchmark.json
@@ -24,6 +24,12 @@
         "JAVA_OPTS": "-javaagent:${TRACER} -Ddd.iast.enabled=true"
       }
     },
+    "iast_GLOBAL": {
+      "env": {
+        "VARIANT": "iast_GLOBAL",
+        "JAVA_OPTS": "-javaagent:${TRACER} -Ddd.iast.enabled=true -Ddd.iast.context.mode=GLOBAL"
+      }
+    },
     "iast_FULL": {
       "env": {
         "VARIANT": "iast_FULL",
--- a/dd-java-agent/agent-iast/src/main/java/com/datadog/iast/IastGlobalContext.java
+++ b/dd-java-agent/agent-iast/src/main/java/com/datadog/iast/IastGlobalContext.java
@@ -1,0 +1,53 @@
+package com.datadog.iast;
+
+import com.datadog.iast.taint.TaintedMap;
+import com.datadog.iast.taint.TaintedObjects;
+import datadog.trace.api.iast.IastContext;
+import java.util.concurrent.TimeUnit;
+import javax.annotation.Nonnull;
+import javax.annotation.Nullable;
+
+public class IastGlobalContext implements IastContext {
+
+  private final TaintedObjects taintedObjects;
+
+  public IastGlobalContext(final TaintedObjects taintedObjects) {
+    this.taintedObjects = taintedObjects;
+  }
+
+  @SuppressWarnings("unchecked")
+  @Nonnull
+  @Override
+  public TaintedObjects getTaintedObjects() {
+    return taintedObjects;
+  }
+
+  public static class Provider extends IastContext.Provider {
+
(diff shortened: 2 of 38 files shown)


dd-java-agent/agent-iast/src/main/java/com/datadog/iast/IastGlobalContext.java
```diff
@@ -0,0 +1,49 @@
...
+  @Nonnull
+  @Override
+  public TaintedObjects getTaintedObjects() {
+    return taintedObjects;
+  }
+
+  public static class Provider extends IastContext.Provider {
+
+    static final int MAP_SIZE =
+        IastRequestContext.Provider.MAP_SIZE * ConfigDefaults.DEFAULT_IAST_MAX_CONCURRENT_REQUESTS;
```
smola, line 28: Unless it is changed, currently TaintedMap requires that the size is a power of two, which can be broken by this.

smola (commented):
Looks good, except the map size constraint.

benchmark/load/insecure-bank/benchmark.json
```diff
@@ -24,6 +24,12 @@
         "JAVA_OPTS": "-javaagent:${TRACER} -Ddd.iast.enabled=true"
       }
     },
+    "iast_GLOBAL": {
+      "env": {
+        "VARIANT": "iast_GLOBAL",
+        "JAVA_OPTS": "-javaagent:${TRACER} -Ddd.iast.enabled=true -Ddd.iast.context.mode=GLOBAL"
```
smola, line 30: I don't see the benchmark picked up this?

dd-java-agent/agent-iast/src/main/java/com/datadog/iast/IastGlobalContext.java
manuel-alvarez-alvarez: Yes, by default `ConfigDefaults.DEFAULT_IAST_MAX_CONCURRENT_REQUESTS` is 4, but it's true that this solution is a bit fragile, I'll change it.

benchmark/load/insecure-bank/benchmark.json
manuel-alvarez-alvarez: It looks like the summary with our table+charts is fully gone for the **load** reports, I'll have a look into this.

dd-java-agent/agent-iast/src/main/java/com/datadog/iast/IastGlobalContext.java
manuel-alvarez-alvarez: Changed!

jandro996 (approved):

smola (approved):

PR #6108, opened 2023-10-27
Fix suite name calculation for jqwik framework

nikita-tkachenko-datadog:
# What Does This Do
Fixes instrumentation for [jqwik](https://jqwik.net/) framework.

# Motivation
The framework is used by one of the DD repos that are used for dogfooding.

# Additional Notes
The specific problem being fixed is incorrect suite name calculation in testcase start/finish events.
Previous logic calculated suite name based on the class where the given test method was defined.
It was incorrect in some cases, since the test case can be defined in a parent class, while the suite that runs this test case can be defined in the child class.

Jira ticket: [CIVIS-7825]

[CIVIS-7825]: https://datadoghq.atlassian.net/browse/CIVIS-7825?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ

### DIFF
--- a/dd-java-agent/instrumentation/junit-5.3/src/main/java/datadog/trace/instrumentation/junit5/TracingListener.java
+++ b/dd-java-agent/instrumentation/junit-5.3/src/main/java/datadog/trace/instrumentation/junit5/TracingListener.java
@@ -107,7 +107,19 @@ public class TracingListener implements EngineExecutionListener {
   }
 
   private void testMethodExecutionStarted(TestDescriptor testDescriptor, MethodSource testSource) {
-    String testSuitName = testSource.getClassName();
+    TestDescriptor suiteDescriptor = getSuiteDescriptor(testDescriptor);
+
+    Class<?> testClass;
+    String testSuiteName;
+    if (suiteDescriptor != null) {
+      testClass = JUnitPlatformUtils.getJavaClass(suiteDescriptor);
+      testSuiteName =
+          testClass != null ? testClass.getName() : suiteDescriptor.getLegacyReportingName();
+    } else {
+      testClass = JUnitPlatformUtils.getTestClass(testSource);
+      testSuiteName = testSource.getClassName();
+    }
+
     String displayName = testDescriptor.getDisplayName();
     String testName = testSource.getMethodName();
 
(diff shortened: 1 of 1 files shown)


dd-java-agent/instrumentation/junit-5.3/src/main/java/datadog/trace/instrumentation/junit5/TracingListener.java
```diff
@@ -107,20 +107,31 @@ private void testCaseExecutionStarted(final TestDescriptor testDescriptor) {
...
+    String testSuiteName;
+    if (suiteDescriptor != null) {
+      testClass = JUnitPlatformUtils.getJavaClass(suiteDescriptor);
+      testSuiteName =
+          testClass != null ? testClass.getName() : suiteDescriptor.getLegacyReportingName();
+    } else {
+      testClass = JUnitPlatformUtils.getTestClass(testSource);
+      testSuiteName = testSource.getClassName();
+    }
+
```
romainkomorndatadog, line 122: Seems like you could factor this out into a utility method?

romainkomorndatadog (commented):
This feels like something that wants a unit test to make sure someone doesn't come along and break the functionality later on.

dd-java-agent/instrumentation/junit-5.3/src/main/java/datadog/trace/instrumentation/junit5/TracingListener.java
nikita-tkachenko-datadog: The problem is Java doesn't really have tuples and given that there are two values that are computed (testClass/testSuiteName) I'd have to return an array or some other object that encapsulates both. I did not want to introduce the overhead of doing so since this method is invoked for every test case

nikita-tkachenko-datadog:
> This feels like something that wants a unit test to make sure someone doesn't come along and break the functionality later on.

I will add a proper suite of tests for `jqwik` once there is "official" support for it in the tracer. For now this is just a quick fix to avoid breaking the dogfooding we have in logs-backend

romainkomorndatadog (approved):

PR #6112, opened 2023-10-30
Support SSRF in Apache HttpClient V4

jandro996:
# What Does This Do

Add new IAST instrumentation for `org.apache.http.client.HttpClient`

# Motivation

Add SSRF support for Apache HttpClient V4

# Additional Notes

ApacheHttpClientInstrumentation

As `org.apache.http.client.HttpClient` is currently instrumented for tracing implementing `Instrumenter.CanShortcutTypeMatching` we will keep the same approach

Jira ticket: [APPSEC-11925]

[APPSEC-11925]: https://datadoghq.atlassian.net/browse/APPSEC-11925?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ


dd-java-agent/instrumentation/apache-httpclient-4/src/main/java/datadog/trace/instrumentation/apachehttpclient/IastHelperMethods.java
```diff
@@ -0,0 +1,38 @@
...
+  }
+
+  public static void doMethodEnter(HttpHost httpHost, HttpRequest request) {
+    final int callDepth = CallDepthThreadLocalMap.incrementCallDepth(IastHelperMethods.class);
+    if (callDepth > 0) {
+      return;
+    }
+    final SsrfModule module = InstrumentationBridge.SSRF;
+    if (module != null) {
+      module.onURLConnection(
```
manuel-alvarez-alvarez, line 30: You could create a new method with both params, you are not force to reuse the single param one :wink:

dd-java-agent/instrumentation/apache-httpclient-4/src/main/java/datadog/trace/instrumentation/apachehttpclient/IastApacheHttpClientInstrumentation.java
```diff
@@ -0,0 +1,172 @@
...
+  }
+
+  @Override
+  public boolean onlyMatchKnownTypes() {
+    return isShortcutMatchingEnabled(false);
+  }
+
+  @Override
+  public String[] knownMatchingTypes() {
+    return new String[] {
```
manuel-alvarez-alvarez, line 35: I wonder if we could reuse the list of classes to instrument, this way  it will be clear that both instrumenters are targeting the same.

dd-smoke-tests/apache-httpclient-4/build.gradle
```diff
@@ -0,0 +1,30 @@
+plugins {
```
manuel-alvarez-alvarez, line 1: Have you tried to do this inside the actual module of the instrumentation?, for instance you could create a `datadog.trace.agent.test.base.HttpServer` (see `test.boot.SpringBootServer`), and do exactly the same without having to create a new smoke test module (smoke tests are a bit harder to debug and slower to test).

dd-java-agent/agent-iast/src/main/java/com/datadog/iast/sink/SsrfModuleImpl.java
```diff
@@ -26,4 +34,49 @@ public void onURLConnection(@Nullable final Object url) {
     }
     checkInjection(span, ctx, VulnerabilityType.SSRF, url);
   }
+
+  /*
+   * if the host or the uri are tainted, we report the url as tainted as well
+   * A new range is created covering all the value string in order to simplify the algorithm
```
bantonsson, line 40: Is there any particular reason that this is not a proper JavaDoc comment?

dd-java-agent/instrumentation/apache-httpcore-4/src/main/java/datadog/trace/instrumentation/apachehttpcore/IastHttpHostInstrumentation.java
```diff
@@ -0,0 +1,49 @@
...
+
+  @Override
+  public String instrumentedType() {
+    return "org.apache.http.HttpHost";
+  }
+
+  @Override
+  public void adviceTransformations(AdviceTransformation transformation) {
+    transformation.applyAdvice(
+        isConstructor()
```
DDJavierSantos, line 30: isConstructor().and(takesArguments(String.class, int.class, String.class))
makes for more compact code

dd-java-agent/instrumentation/iast-instrumenter/src/main/resources/datadog/trace/instrumentation/iastinstrumenter/iast_exclusion.trie
```diff
@@ -189,6 +189,8 @@
 1 org.aopalliance.*
 1 org.antlr.*
 1 org.apache.*
+#apache httpClient needs URI propagation
+0 org.apache.http.client.methods.*
```
DDJavierSantos, line 193: We are not instrumenting all of the classes on that package. My understanding is that the policy is to explicit the classes that are instrumented:
0 org.apache.http.client.methods.HttpUriRequest

jandro996: The change in the exclusion list is not necessary for HttpUriRequest, it aims to instrument the call to URI#create in all the http verb classes like:

HttpTrace
HttpPut
HttpPost
HttpOptions
HttpHead
HttpGet
HttpDelete

This list change between version, so I decided to exclude all the package

Let me know if this is a valid approach or we need to invest in look for each class in all versions

smola (approved):

dd-java-agent/agent-iast/src/main/java/com/datadog/iast/sink/SsrfModuleImpl.java
```diff
@@ -26,4 +34,53 @@ public void onURLConnection(@Nullable final Object url) {
...
+    final AgentSpan span = AgentTracer.activeSpan();
+    final IastRequestContext ctx = IastRequestContext.get(span);
+    if (ctx == null) {
+      return;
+    }
+    TaintedObject taintedObject = getTaintedObject(ctx, host, uri);
+    if (taintedObject == null) {
+      return;
+    }
+    Range[] ranges =
```
manuel-alvarez-alvarez, line 60: Ideally we should use something like

```java
checkInjection(
        span,
        VulnerabilityType.SSRF,
        rangesProviderFor(to, host),
        rangesProviderFor(to, url));
```

Let's refactor the reporting facility in another PR so you don't have to do it by hand.

dd-java-agent/instrumentation/apache-httpclient-4/build.gradle
```diff
@@ -23,11 +23,32 @@ muzzle {
 apply from: "$rootDir/gradle/java.gradle"
 
 addTestSuiteForDir('latestDepTest', 'test')
+addTestSuite('iastIntegrationTest')
```
manuel-alvarez-alvarez, line 26: Nice!

dd-java-agent/instrumentation/apache-httpclient-4/src/iastIntegrationTest/java/SsrfController.java
```diff
@@ -0,0 +1,98 @@
...
+      final String url,
+      final String clientClassName,
+      final String method,
+      final String requestType,
+      final String scheme) {
+    try {
+      HttpClient httpClient = getHttpClient(clientClassName);
+      execute(httpClient, url, ExecuteMethod.valueOf(method), requestType, scheme);
+    } catch (Exception e) {
+    }
```
manuel-alvarez-alvarez, line 23: We should probably log the exception and return something like `"NO_OK"` to ease troubleshooting in case of error.

manuel-alvarez-alvarez (approved):

dd-java-agent/agent-iast/src/main/java/com/datadog/iast/sink/SsrfModuleImpl.java
jandro996: I've created a JIRA ticket with this task

dd-java-agent/instrumentation/apache-httpclient-4/build.gradle
jandro996: thanks for your help! :)

PR #6122, opened 2023-10-31
Match tainted objects with sources when checking unbounded vulnerabilities

manuel-alvarez-alvarez:
# What Does This Do
When reporting a vulnerability where the source tainted value has a single unbounded range (happens when we lose propagation over objects like `URL`, `URI`, ...), try to match the original source value with the final representation of the tainted object to do a better reporting,.

# Motivation
From a customer standpoint sometimes is difficult to correlate the vulnerability with its original source in the cases that we lose exact propagation (for instance when creating URLs).

# Additional Notes

Jira ticket: [APPSEC-11922]

[APPSEC-11922]: https://datadoghq.atlassian.net/browse/APPSEC-11922?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ

### DIFF
--- a/dd-java-agent/agent-iast/src/main/java/com/datadog/iast/taint/Ranges.java
+++ b/dd-java-agent/agent-iast/src/main/java/com/datadog/iast/taint/Ranges.java
@@ -59,6 +59,15 @@ public final class Ranges {
     return new Range[] {new Range(0, Integer.MAX_VALUE, source, mark)};
   }
 
+  @Nullable
+  public static Range findUnbound(@Nonnull final Range[] ranges) {
+    if (ranges.length != 1) {
+      return null;
+    }
+    final Range range = ranges[0];
+    return range.getStart() == 0 && range.getLength() == Integer.MAX_VALUE ? range : null;
+  }
+
   public static void copyShift(
       final @Nonnull Range[] src, final @Nonnull Range[] dst, final int dstPos, final int shift) {
     copyShift(src, dst, dstPos, shift, src.length);
--- a/dd-java-agent/agent-iast/src/main/java/com/datadog/iast/sink/SinkModuleBase.java
+++ b/dd-java-agent/agent-iast/src/main/java/com/datadog/iast/sink/SinkModuleBase.java
@@ -9,6 +9,7 @@ import com.datadog.iast.Reporter;
 import com.datadog.iast.model.Evidence;
 import com.datadog.iast.model.Location;
 import com.datadog.iast.model.Range;
+import com.datadog.iast.model.Source;
 import com.datadog.iast.model.Vulnerability;
 import com.datadog.iast.model.VulnerabilityType;
 import com.datadog.iast.model.VulnerabilityType.InjectionType;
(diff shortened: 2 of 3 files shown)


jandro996 (approved):

dd-java-agent/agent-iast/src/main/java/com/datadog/iast/taint/Ranges.java
```diff
@@ -59,6 +59,14 @@ public static Range[] forObject(final @Nonnull Source source, final int mark) {
     return new Range[] {new Range(0, Integer.MAX_VALUE, source, mark)};
   }
 
+  public static boolean isUnbound(@Nonnull final Range[] ranges) {
+    if (ranges.length != 1) {
+      return false;
+    }
+    final Range range = ranges[0];
+    return range.getStart() == 0 && range.getLength() == Integer.MAX_VALUE;
```
jandro996, line 67: Probably it will be a super corner case, but it will be possible that the range start is != 0 and the length will be a negative number

I know that this is not possible from a new source point of view but, maybe dealing with the redaction or propagation...

manuel-alvarez-alvarez: This PR only tries to deal with ranges created by `forObject`:

```java
  public static Range[] forObject(final @Nonnull Source source, final int mark) {
    return new Range[] {new Range(0, Integer.MAX_VALUE, source, mark)};
  }

  public static boolean isUnbound(@Nonnull final Range[] ranges) {
    if (ranges.length != 1) {
      return false;
    }
    final Range range = ranges[0];
    return range.getStart() == 0 && range.getLength() == Integer.MAX_VALUE;
  }
```
For other corner cases I'm not sure if we should try to do the matching

dd-java-agent/agent-iast/src/main/java/com/datadog/iast/taint/Ranges.java
```diff
@@ -59,6 +59,15 @@ public static Range[] forObject(final @Nonnull Source source, final int mark) {
     return new Range[] {new Range(0, Integer.MAX_VALUE, source, mark)};
   }
 
+  @Nullable
+  public static Range findUnbound(@Nonnull final Range[] ranges) {
+    if (ranges.length != 1) {
+      return null;
+    }
+    final Range range = ranges[0];
```
ValentinZakharov, line 67: Shouldn't we iterate over all ranges instead of checking only the first one?

manuel-alvarez-alvarez: Sorry for the delay in the response, this PR only covers the case of unbounded tainted values that are created via:

```java
  public static Range[] forObject(final @Nonnull Source source, final int mark) {
    return new Range[] {new Range(0, Integer.MAX_VALUE, source, mark)};
  }
```

ValentinZakharov (approved):

PR #6123, opened 2023-10-31
IAST support for Spring with gson

jandro996:
# What Does This Do

Add propagation:

- void java.io.InputStreamReader.<init>(java.io.InputStream, java.nio.charset.Charset)
- java.lang.String org.springframework.util.StreamUtils.copyToString(java.io.InputStream, java.nio.charset.Charset)

Move spring gson test cases to AbstractGsonIastSpringBootSmokeTest

# Motivation

Add Spring gson support

# Additional Notes

Allow instrumentation for org.springframework.http.converter.json.AbstractJsonHttpMessageConverter as org.springframework.http.converter.json.GsonHttpMessageConverter in some versions delegates to create a new InputStreamReader

Jira ticket: [APPSEC-11945]

[APPSEC-11945]: https://datadoghq.atlassian.net/browse/APPSEC-11945?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ


dd-java-agent/agent-tooling/src/main/resources/datadog/trace/agent/tooling/bytebuddy/matcher/ignored_class_name.trie
```diff
@@ -330,6 +333,8 @@
 0 org.springframework.web.servlet.*
 # Need for IAST so propagation of tainted objects is complete in spring 2.7.5
 0 org.springframework.util.StreamUtils$NonClosingInputStream
+# Need for IAST so propagation
```
manuel-alvarez-alvarez, line 336: Looks like part of the reason is missing :smile:

dd-smoke-tests/spring-boot-2.6-webmvc/src/test/groovy/GsonIastSpringBootSmokeTest.groovy
```diff
@@ -0,0 +1,9 @@
+import datadog.smoketest.AbstractGsonIastSpringBootSmokeTest
+
+class GsonIastSpringBootSmokeTest extends  AbstractGsonIastSpringBootSmokeTest{
+
+  @Override
+  String[] getCustomSpringProperties(){
+    return ['-Dspring.mvc.converters.preferred-json-mapper=gson']
```
manuel-alvarez-alvarez, line 7: Is there any reason to not define this property in the base class?

jandro996: We have different properties to change the json mapper between spring versions, in our case:

`spring.mvc.converters.preferred-json-mapper` vs `spring.http.converters.preferred-json-mapper`

dd-java-agent/instrumentation/java-io/src/main/java/datadog/trace/instrumentation/java/lang/InputStreamReaderCallSite.java
```diff
@@ -0,0 +1,30 @@
+package datadog.trace.instrumentation.java.lang;
```
manuel-alvarez-alvarez, line 1: Mental note, we should rename this package to `datadog.trace.instrumentation.java.io`

jandro996: I've created a ticket https://datadoghq.atlassian.net/browse/APPSEC-12176  to do it in another PR

dd-java-agent/instrumentation/java-io/src/main/java/datadog/trace/instrumentation/java/lang/InputStreamReaderCallSite.java
```diff
@@ -0,0 +1,30 @@
...
+
+  @CallSite.After(
+      "void java.io.InputStreamReader.<init>(java.io.InputStream, java.nio.charset.Charset)")
+  public static InputStreamReader afterInit(
+      @CallSite.AllArguments @Nonnull final Object[] params,
+      @CallSite.Return @Nonnull final InputStreamReader result) {
+    final PropagationModule module = InstrumentationBridge.PROPAGATION;
+    if (module != null) {
+      try {
+        module.taintIfTainted(result, params[0]);
```
DDJavierSantos, line 23: What would happen if you get called with a null on the first parameter?

jandro996: I've reviewed it and it can't be null

```
protected Reader(Object lock) {
        if (lock == null) {
            throw new NullPointerException();
        }
        this.lock = lock;
    }
```

any way taintIfTainted supports null elements

I will change the afterInit method to receive explicitly the  java.io.InputStream

jandro996: we need to keep the current implementation as the use of  @CallSite.AllArguments is mandatory in constructor callsite instrumentation to access arguments in after advices

```
[⨉] @CallSite datadog.trace.instrumentation.java.lang.InputStreamReaderCallSite
  [ADVICE_AFTER_CONSTRUCTOR_ALL_ARGUMENTS] After advice in constructors should use @AllArguments
```

PR #6126, opened 2023-10-31
Try to locate tracer JAR using class loader resource lookup when bootstrapping the agent

nikita-tkachenko-datadog:
# What Does This Do
Updates agent bootstrap logic to try to locate tracer JAR by finding the `AgentBootstrap.class` resource using the class' classloader.

# Motivation
Currently the lookup is done using `ProtectionDomain`/`CodeSource` and by parsing the `-javaagent` argument if the first approach fails.
However the second approach may fail too if multiple agents are specified (which, for example, is possible in CI Visibility when a test JVM is forked both with the tracer and with the Jacoco agent).

Jira ticket: [CIVIS-7865]

### DIFF
--- a/dd-java-agent/src/main/java/datadog/trace/bootstrap/AgentBootstrap.java
+++ b/dd-java-agent/src/main/java/datadog/trace/bootstrap/AgentBootstrap.java
@@ -10,6 +10,7 @@ import java.lang.instrument.Instrumentation;
 import java.lang.management.ManagementFactory;
 import java.lang.reflect.Field;
 import java.lang.reflect.Method;
+import java.net.URI;
 import java.net.URISyntaxException;
 import java.net.URL;
 import java.nio.charset.StandardCharsets;
@@ -203,13 +204,10 @@ public final class AgentBootstrap {
 
   private static synchronized URL installAgentJar(final Instrumentation inst)
       throws IOException, URISyntaxException {
-    URL ddJavaAgentJarURL = null;
-
     // First try Code Source
     final CodeSource codeSource = thisClass.getProtectionDomain().getCodeSource();
-
     if (codeSource != null) {
-      ddJavaAgentJarURL = codeSource.getLocation();
+      URL ddJavaAgentJarURL = codeSource.getLocation();
       if (ddJavaAgentJarURL != null) {
         final File ddJavaAgentJarPath = new File(ddJavaAgentJarURL.toURI());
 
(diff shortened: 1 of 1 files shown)


dd-java-agent/src/main/java/datadog/trace/bootstrap/AgentBootstrap.java
```diff
@@ -222,7 +221,44 @@ private static synchronized URL installAgentJar(final Instrumentation inst)
...
+    } else {
+      thisClassUrl = classLoader.getResource(thisClassResourceName);
+    }
+
+    if (thisClassUrl == null) {
+      throw new IllegalStateException(
+          "Could not locate agent bootstrap class resource, not installing tracing agent");
+    }
 
+    javaagentFile = new File(new URI(thisClassUrl.getFile().split("!")[0]));
```
bric3, line 248: nitpick: maybe extract this strategy to another method.

dd-java-agent/src/main/java/datadog/trace/bootstrap/AgentBootstrap.java
```diff
@@ -222,7 +221,44 @@ private static synchronized URL installAgentJar(final Instrumentation inst)
...
+    if (!javaagentFile.isDirectory()) {
+      URL ddJavaAgentJarURL = javaagentFile.toURI().toURL();
+      checkJarManifestMainClassIsThis(ddJavaAgentJarURL);
+      inst.appendToBootstrapClassLoaderSearch(new JarFile(javaagentFile));
+      return ddJavaAgentJarURL;
+    }
```
bric3, line 254: This method is similar to [L225:L230](https://github.com/DataDog/dd-trace-java/pull/6126/files#diff-b1a8f8f62f4935e26683f3300e46b355fc41aad5ed91821c27b83958630c45afR225-R230)

I wonder if the code could be changed to avoid the duplication.

bric3 (commented):
I have the same issue with IntelliJ classloader, which strips the protection domain. So when two agents are loaded at the same time (e.g the Kotlin Coroutine debugger agent) dd-trace-java bails out.

Coincidentally I was trying to fix `AgentBootstrap` for that reason, using a property as a fall back instead of short circuiting when there are multiple agents. However the proposed approach seems to work better without "user" intervention.

For reference I have reproduced the failure with IJ there [bric3/dd-trace-java-with-ij-classloading-and-another-agent](https://github.com/bric3/dd-trace-java-with-ij-classloading-and-another-agent), and trying your change avoid the issue.

nikita-tkachenko-datadog:
> I have the same issue with IntelliJ classloader, which strips the protection domain. So when two agents are loaded at the same time (e.g the Kotlin Coroutine debugger agent) dd-trace-java bails out.
Yes, this is exactly the issue that I am fixing here :) I tested this with the DD Intellij Plugin repo, and the fix works without the need for additional changes in the project

bric3:
Would you be interested in me adding a smoke test for this ?

nikita-tkachenko-datadog:
> Would you be interested in me adding a smoke test for this ?

Of course, that'd be great!

mcculls (approved):
LGTM - same comments as @bric3, but no blockers

dd-java-agent/src/main/java/datadog/trace/bootstrap/AgentBootstrap.java
nikita-tkachenko-datadog: Makes sense, done

dd-java-agent/src/main/java/datadog/trace/bootstrap/AgentBootstrap.java
nikita-tkachenko-datadog: Extracted to a separate method

PR #6135, opened 2023-11-02
Fix sampling for line probe without condition

jpbempel:
# What Does This Do
In the case of a line probe MethodLocation is DEFAULT and we should not consider the evaluteAt attribute as it is meant for method probe only

# Motivation
no sampling for line probe without condition
# Additional Notes

Jira ticket: [DEBUG-1908]

[DEBUG-1908]: https://datadoghq.atlassian.net/browse/DEBUG-1908?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ

### DIFF
--- a/dd-java-agent/agent-debugger/debugger-bootstrap/src/main/java/datadog/trace/bootstrap/debugger/MethodLocation.java
+++ b/dd-java-agent/agent-debugger/debugger-bootstrap/src/main/java/datadog/trace/bootstrap/debugger/MethodLocation.java
@@ -7,7 +7,7 @@ public enum MethodLocation {
 
   public static boolean isSame(MethodLocation methodLocation, MethodLocation evaluateAt) {
     if (methodLocation == MethodLocation.DEFAULT) {
-      // line probe always assume we are the right location
+      // line probe, no evaluation of probe's evaluateAt
       return true;
     }
     if (methodLocation == MethodLocation.ENTRY) {
--- a/dd-java-agent/agent-debugger/debugger-bootstrap/src/main/java/datadog/trace/bootstrap/debugger/MethodLocation.java
+++ b/dd-java-agent/agent-debugger/debugger-bootstrap/src/main/java/datadog/trace/bootstrap/debugger/MethodLocation.java
@@ -6,6 +6,10 @@ public enum MethodLocation {
   EXIT;
 
   public static boolean isSame(MethodLocation methodLocation, MethodLocation evaluateAt) {
+    if (methodLocation == MethodLocation.DEFAULT) {
+      // line probe always assume we are the right location
+      return true;
+    }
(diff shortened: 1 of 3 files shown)


dd-java-agent/agent-debugger/debugger-bootstrap/src/main/java/datadog/trace/bootstrap/debugger/MethodLocation.java
```diff
@@ -6,6 +6,10 @@ public enum MethodLocation {
   EXIT;
 
   public static boolean isSame(MethodLocation methodLocation, MethodLocation evaluateAt) {
+    if (methodLocation == MethodLocation.DEFAULT) {
```
shatzi, line 9: if we do this only for line probe. maybe the line probe logic should not call this instead of adding this case in the method?

Aka, the evaluate function should already have the probe definition, right? so that function can simply skip this if the probe location does not need to check method location

shatzi (approved):

dd-java-agent/agent-debugger/debugger-bootstrap/src/main/java/datadog/trace/bootstrap/debugger/MethodLocation.java
jpbempel: it centralizes the logic for both calls:
```
    LogStatus logStatus = (LogStatus) status;
    if (!hasCondition()) {
      // sample when no condition associated
      sample(logStatus, methodLocation);
    }
    logStatus.setCondition(evaluateCondition(context, logStatus));
    CapturedContext.CapturedThrowable throwable = context.getThrowable();
    if (logStatus.hasConditionErrors() && throwable != null) {
      logStatus.addError(
          new EvaluationError(
              "uncaught exception", throwable.getType() + ": " + throwable.getMessage()));
    }
    if (hasCondition() && logStatus.getCondition()) {
      // sample if probe has condition and condition is true
      sample(logStatus, methodLocation);
    }
```

jpbempel: actually it's better to centralize also the other isSame call

shatzi: I would explain this in the comment. that MethodLocation.DEFAULT is used to mark that this is a line probe and therefore we don't check the probeLocation as it meaningless.

shatzi (approved):

PR #6138, opened 2023-11-02
Add multiple context extraction support

PerfectSlayer:
# What Does This Do

* Extract multiple propagated context according the new W3C trace context initiative,
* Add `DD_TRACE_PROPAGATION_EXTRACT_FIRST` configuration to limit to the first valid one,
* Propagate W3C tracestate if matches the extracted context, even if not a Datadog one,
* Add span links of terminated contexts (terminated but not matching).

# Motivation

This will improve W3C tracecontext standard compatibility.
W3C trace state will be extracted from W3C extracted context even using Datadog headers. 

# Additional Notes

An addition PR is coming to documentation repository for the new configuration key and behavior change.

Jira tickets:
* [APMJAVA-1123]
* [APMJAVA-1088]

Few comments that might help reviewing:
* Context interpreters are now aware of the propagation style --> Required to handle specific W3C behavior
* Same for extracted context --> Add a new parameter to constructor. I did not want to create another overloaded constructor as the constructor was not call that often (it mostly impacts tests and I am fine with updating them rather than creating multiple constructors that would make the code harder to read)
* The new extract first trace context logic will only impact the `CompoundExtractor` --> The RFC has a diagram to help understand the expected behavior. I tried to comment it the most possible for futur maintainability. 
* W3C tracestate is added to PTags --> It should now be supported whatever the main propagation style, not only W3C. So `original` from W3CPTags was removed in favor of the new generic field.
* W3CPtagsCodec now supports PTags (not only W3CPTags) for tracestate injection. --> I also add the 32 member limit check, with and without dd member presence. 
* W3CPTagsCodec now use the generic error field for inconsistent trace id --> I removed logic override to rely on the parent field
* B3 extractor test are fixed --> They failed to extract partial tag content when trace id is invalid.  
* Moved propagation tags factory to the parent `ContextInterpreter` and make sure PTags are always defined (empty at least) --> PTags factory was the same for all context interpreter using it. Moving it to parent allow to make empty initialization if not done by the child interpreter. It does not introduce performance overhead as PTags will always be creating when creating a span from an extracted context (even an empty instance). Creating it at `ExtractedContext` level ensure we can propagate `tracestate` in `CompoundExtractor`. 

[APMJAVA-1123]: https://datadoghq.atlassian.net/browse/APMJAVA-1123?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ

[APMJAVA-1088]: https://datadoghq.atlassian.net/browse/APMJAVA-1088?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ

### DIFF
--- a/dd-java-agent/instrumentation/opentelemetry/opentelemetry-0.3/src/test/groovy/OpenTelemetryTest.groovy
+++ b/dd-java-agent/instrumentation/opentelemetry/opentelemetry-0.3/src/test/groovy/OpenTelemetryTest.groovy
@@ -4,6 +4,8 @@ import datadog.trace.api.DDTags
 import datadog.trace.api.DDTraceId
 import datadog.trace.api.interceptor.MutableSpan
 import datadog.trace.core.propagation.PropagationTags
+
+import static datadog.trace.api.TracePropagationStyle.NONE
 import static datadog.trace.api.sampling.PrioritySampling.*
 import static datadog.trace.api.sampling.SamplingMechanism.*
 import datadog.trace.context.TraceScope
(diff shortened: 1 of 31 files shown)


PerfectSlayer:
I marked the PR ready for review as it should be feature compete and pass [the related system tests](https://github.com/DataDog/system-tests/pull/1795) (currently in development).
It still need to performance measurement though.

mcculls (approved):
LGTM

PR #6142, opened 2023-11-03
Fix tracer enabled detection default value

jpbempel:
# What Does This Do
Checking for trace enabled needs to be true by default if no token

# Motivation

# Additional Notes


OmerRaviv (approved):
LGTM

PR #6153, opened 2023-11-03
DBM-APM: Selectively pre/append sql comments depending on DBMS type

jmeunier28:
# What Does This Do

This PR is a follow on to https://github.com/DataDog/dd-trace-java/pull/6034, which should fix the failed mysql integration tests that were failing as a result of it. For example, this failing test case for mysql https://github.com/DataDog/system-tests/actions/runs/6701163011/job/18213623612. The failure is similar to the one we saw from SQL Server, which #6034 intended to fix. E.g.: Running `{call test_procedure(?,?)}` with DBM full propagation mode enabled, results in the query failing to execute & no trace collected. 

In order to fix both issues, this PR introduces a new class, which either prepends or appends the comments based on a set of concrete class types. SQL Server types will append the comments, where all others will prepend like they were doing prior to #6034. 

# Motivation

# Additional Notes

Jira ticket: [PROJ-IDENT]


bm1549:
@jmeunier28 can this be closed in favor of https://github.com/DataDog/dd-trace-java/pull/6160 ?

jmeunier28:
@bm1549 yep! sorry about that!

PR #6154, opened 2023-11-06
Fix accessing array field by reflection

jpbempel:
# What Does This Do
When some fields are not accessible directly because of the visibility we emit a relfection call to get the value but arrays was not supported

# Motivation

# Additional Notes

### DIFF
--- a/dd-java-agent/agent-debugger/src/test/java/com/datadog/debugger/agent/CapturedSnapshotTest.java
+++ b/dd-java-agent/agent-debugger/src/test/java/com/datadog/debugger/agent/CapturedSnapshotTest.java
@@ -1234,10 +1234,11 @@ public class CapturedSnapshotTest {
     Snapshot snapshot = assertOneSnapshot(listener);
     Map<String, CapturedContext.CapturedValue> staticFields =
         snapshot.getCaptures().getReturn().getStaticFields();
-    assertEquals(5, staticFields.size());
+    assertEquals(6, staticFields.size());
     assertEquals("barfoo", getValue(staticFields.get("strValue")));
     assertEquals("48", getValue(staticFields.get("intValue")));
     assertEquals("6.28", getValue(staticFields.get("doubleValue")));
+    assertEquals("[1, 2, 3, 4]", getValue(staticFields.get("longValues")));
   }
 
   @Test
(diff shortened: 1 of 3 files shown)


dd-java-agent/agent-debugger/src/test/java/com/datadog/debugger/agent/CapturedSnapshotTest.java
```diff
@@ -2140,13 +2141,48 @@ private static String getValue(CapturedContext.CapturedValue capturedValue) {
         Assertions.fail("NotCapturedReason: " + valued.getNotCapturedReason());
       }
       Object obj = valued.getValue();
+      if (obj != null && obj.getClass().isArray()) {
+        if (obj.getClass().getComponentType().isPrimitive()) {
+          return primitiveArrayToString(obj);
+        }
+        return Arrays.toString((Object[]) obj);
```
shatzi, line 2148: what is the difference between calling `Arrays.toString((Object[]) obj)` and `primitiveArrayToString`?
And what if we have an array of String[]? would that be rendered in a meaningful way to the user?
Lastly, does Array.toString is calling the objects toString() function?

dd-java-agent/agent-debugger/src/test/java/com/datadog/debugger/agent/CapturedSnapshotTest.java
```diff
@@ -1234,10 +1234,11 @@ public void staticInheritedFields() throws IOException, URISyntaxException {
     Snapshot snapshot = assertOneSnapshot(listener);
     Map<String, CapturedContext.CapturedValue> staticFields =
         snapshot.getCaptures().getReturn().getStaticFields();
-    assertEquals(5, staticFields.size());
+    assertEquals(6, staticFields.size());
     assertEquals("barfoo", getValue(staticFields.get("strValue")));
     assertEquals("48", getValue(staticFields.get("intValue")));
     assertEquals("6.28", getValue(staticFields.get("doubleValue")));
+    assertEquals("[1, 2, 3, 4]", getValue(staticFields.get("longValues")));
```
shatzi, line 1241: why the array is not serialized as other arrays? with size and elements?

shatzi (approved):
Maybe adding some specific unit tests to check that we support both primitive arrays and object arrays?

dd-java-agent/agent-debugger/src/test/java/com/datadog/debugger/agent/CapturedSnapshotTest.java
shatzi: not relevant

dd-java-agent/agent-debugger/src/test/java/com/datadog/debugger/agent/CapturedSnapshotTest.java
shatzi: not relevant

PR #6156, opened 2023-11-06
Add DSM for google pubsub

amarziali:
# What Does This Do

Adds DSM support to google pubsub instrumentation

# Motivation

# Additional Notes

Jira ticket: [APMJAVA-932]

[APMJAVA-932]: https://datadoghq.atlassian.net/browse/APMJAVA-932?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ

### DIFF
--- a/dd-java-agent/instrumentation/google-pubsub/src/main/java/datadog/trace/instrumentation/googlepubsub/PubSubDecorator.java
+++ b/dd-java-agent/instrumentation/google-pubsub/src/main/java/datadog/trace/instrumentation/googlepubsub/PubSubDecorator.java
@@ -2,17 +2,24 @@ package datadog.trace.instrumentation.googlepubsub;
 
 import static datadog.trace.bootstrap.instrumentation.api.AgentTracer.propagate;
 import static datadog.trace.bootstrap.instrumentation.api.AgentTracer.startSpan;
+import static datadog.trace.core.datastreams.TagsProcessor.DIRECTION_IN;
+import static datadog.trace.core.datastreams.TagsProcessor.DIRECTION_TAG;
+import static datadog.trace.core.datastreams.TagsProcessor.SUBSCRIPTION_TAG;
+import static datadog.trace.core.datastreams.TagsProcessor.TYPE_TAG;
 
+import com.google.protobuf.Timestamp;
 import com.google.pubsub.v1.PubsubMessage;
 import datadog.trace.api.Functions;
 import datadog.trace.api.cache.DDCache;
 import datadog.trace.api.cache.DDCaches;
 import datadog.trace.api.naming.SpanNaming;
 import datadog.trace.bootstrap.instrumentation.api.AgentSpan;
+import datadog.trace.bootstrap.instrumentation.api.AgentTracer;
 import datadog.trace.bootstrap.instrumentation.api.InternalSpanTypes;
 import datadog.trace.bootstrap.instrumentation.api.Tags;
 import datadog.trace.bootstrap.instrumentation.api.UTF8BytesString;
 import datadog.trace.bootstrap.instrumentation.decorator.MessagingClientDecorator;
+import java.util.LinkedHashMap;
 import java.util.function.Function;
 import java.util.regex.Matcher;
 import java.util.regex.Pattern;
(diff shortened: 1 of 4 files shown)


dd-java-agent/instrumentation/google-pubsub/src/main/java/datadog/trace/instrumentation/googlepubsub/PubSubDecorator.java
```diff
@@ -107,14 +129,40 @@ public AgentSpan onConsume(final PubsubMessage message, final String subscriptio
...
+    final PathwayContext pathwayContext = dataStreamsMonitoring.newPathwayContext();
+
+    final String parsedSubscription = extractSubscription(subscription);
+    final LinkedHashMap<String, String> sortedTags = new LinkedHashMap<>(3);
+    sortedTags.put(DIRECTION_TAG, DIRECTION_IN);
+    sortedTags.put(SUBSCRIPTION_TAG, parsedSubscription);
+    sortedTags.put(TYPE_TAG, "google-pubsub");
+    final Timestamp publishTime = message.getPublishTime();
+    // FIXME: use full nanosecond resolution when this method will accept nanos
+    pathwayContext.setCheckpoint(
```
hokitam, line 142: why do we use the `newPathwayContext -> setCheckpoint -> mergePathwayContext` calls for consume, but `propagate().injectPathwayContext()` for publish?

dd-java-agent/instrumentation/google-pubsub/src/main/java/datadog/trace/instrumentation/googlepubsub/PubSubDecorator.java
```diff
@@ -107,14 +129,40 @@ public AgentSpan onConsume(final PubsubMessage message, final String subscriptio
...
         propagate().extract(message, TextMapExtractAdapter.GETTER);
     final AgentSpan span = startSpan(PUBSUB_CONSUME, spanContext);
-
+    final AgentDataStreamsMonitoring dataStreamsMonitoring = get().getDataStreamsMonitoring();
+    final PathwayContext pathwayContext = dataStreamsMonitoring.newPathwayContext();
+
+    final String parsedSubscription = extractSubscription(subscription);
+    final LinkedHashMap<String, String> sortedTags = new LinkedHashMap<>(3);
+    sortedTags.put(DIRECTION_TAG, DIRECTION_IN);
+    sortedTags.put(SUBSCRIPTION_TAG, parsedSubscription);
```
hokitam, line 138: just a heads up: we will need to add the `subscription` tag to our allowlist in the backend, without it, the tag won't show up in the metrics we create

dd-java-agent/instrumentation/google-pubsub/src/main/java/datadog/trace/instrumentation/googlepubsub/PubSubDecorator.java
hokitam: here is an example of how we did the consume call for `Kafka`: https://github.com/DataDog/dd-trace-java/blob/7baec8590f5f629b003857d10343b33162f84cbd/dd-java-agent/instrumentation/kafka-clients-0.11/src/main/java/datadog/trace/instrumentation/kafka_clients/TracingIterator.java#L73 we used `propagate().extract(...)`

Should we switch to using this API in the consume call here too? I was looking at the metrics generated by your test runs, and I noticed that `parent_hash` for the consume calls are still set to 0, meaning that the context isn't being propagated correctly in the current code:
Screen Shot 2023-11-06 at 1 30 43 PM

dd-java-agent/instrumentation/google-pubsub/src/main/java/datadog/trace/instrumentation/googlepubsub/PubSubDecorator.java
amarziali: Yes ok. As per our chat I can leave it as `subscription` since closer to the gcp pubsub semantic. However, if needed to change just please let me know I can do quickly (e.g. use consumer group semantic,..)

kr-igor (approved):

dd-java-agent/instrumentation/google-pubsub/src/main/java/datadog/trace/instrumentation/googlepubsub/PubSubDecorator.java
```diff
@@ -107,14 +121,36 @@ public AgentSpan onConsume(final PubsubMessage message, final String subscriptio
...
     final AgentSpan span = startSpan(PUBSUB_CONSUME, spanContext);
-
+    final CharSequence parsedSubscription = extractSubscription(subscription);
+    final LinkedHashMap<String, String> sortedTags = new LinkedHashMap<>(3);
+    sortedTags.put(DIRECTION_TAG, DIRECTION_IN);
+    sortedTags.put(SUBSCRIPTION_TAG, parsedSubscription.toString());
+    sortedTags.put(TYPE_TAG, "google-pubsub");
+    final Timestamp publishTime = message.getPublishTime();
+    // FIXME: use full nanosecond resolution when this method will accept nanos
+    get()
```
mcculls, line 131: minor: I'd leave this as `AgentTracer.get()` because otherwise it's not clear what you're getting

mcculls (approved):

dd-java-agent/instrumentation/google-pubsub/src/main/java/datadog/trace/instrumentation/googlepubsub/PubSubDecorator.java
amarziali: sure suggestion applied

PR #6164, opened 2023-11-07
Improve snapshot pruning algorithm

jpbempel:
# What Does This Do
Instead of decreasing the depth of the collect by re-parsing the snapshot with the Json parser, we are scanning the the json string and building a tree structure representing the json structure with indexed string segments. WE have special treatment for notCapturedReason nodes that helps us in the pruning algorithm:
Leaves of this structure are retrieved and then sorted through a priority queue where the sort key is the following:
 - notCapturedReason depth node
 - level
 - notCapturedReason
 - size of the node 

Then following this, we are pruning first leaves with notCapturedReason depth then removing deepest level leaves, notCapturedReason leaves and finally the leaves with largest size. if the leaves of the original snapshot are not enough for reducing the size we are considering parents.
Nodes pruned are replaced by `{"pruned":true}`.

# Motivation
Improving the way snapshots are pruned when size is larger than 1MB

# Additional Notes

Jira ticket: [DEBUG-1832]

[DEBUG-1832]: https://datadoghq.atlassian.net/browse/DEBUG-1832?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ


dd-java-agent/agent-debugger/src/main/java/com/datadog/debugger/util/SnapshotPruner.java
```diff
@@ -0,0 +1,242 @@
...
+  private Node root;
+
+  public static String prune(String snapshot, int maxTargetedSize, int minLevel) {
+    int delta = snapshot.length() - maxTargetedSize;
+    if (delta <= 0) {
+      return snapshot;
+    }
+    SnapshotPruner snapshotPruner = new SnapshotPruner(snapshot);
+    Collection<Node> leaves = snapshotPruner.getLeaves();
+    leaves = leaves.stream().filter(node -> node.level >= minLevel).collect(Collectors.toList());
```
shatzi, line 36: nit - we can move this to getLeaves(minLevel:int) - so we don't need to collect the leaves and then filter them.

shatzi (approved):
nice!!!

dd-java-agent/agent-debugger/src/main/java/com/datadog/debugger/util/SnapshotPruner.java
jpbempel: done

shatzi (approved):

PR #6170, opened 2023-11-08
Make sure AbstractServletInputStreamWrapper does not propagate exceptions

manuel-alvarez-alvarez:
# What Does This Do
Ensure the class `AbstractServletInputStreamWrapper` does not leak exceptions when called from customer code.

# Motivation
We have received a couple of tickets due to the following exception in their services after activating appsec:
```
java.lang.IllegalStateException: Current state = CODING_END, new state = CODING
at java.base/java.nio.charset.CharsetDecoder.throwIllegalStateException(Unknown Source)
at java.base/java.nio.charset.CharsetDecoder.decode(Unknown Source)
at datadog.trace.api.http.StoredByteBody.commit(StoredByteBody.java:139)
at datadog.trace.api.http.StoredByteBody.appendData(StoredByteBody.java:49)
at datadog.trace.instrumentation.servlet.AbstractServletInputStreamWrapper.read(AbstractServletInputStreamWrapper.java:31)
```

# Additional Notes

Jira ticket: [APPSEC-12101]

[APPSEC-12101]: https://datadoghq.atlassian.net/browse/APPSEC-12101?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ

### DIFF
--- a/internal-api/src/main/java/datadog/trace/api/http/StoredByteBody.java
+++ b/internal-api/src/main/java/datadog/trace/api/http/StoredByteBody.java
@@ -14,9 +14,13 @@ import java.nio.charset.CodingErrorAction;
 import java.nio.charset.StandardCharsets;
 import java.util.function.BiFunction;
 import javax.annotation.Nullable;
+import org.slf4j.Logger;
+import org.slf4j.LoggerFactory;
 
 /** @see StoredCharBody */
 public class StoredByteBody implements StoredBodySupplier {
+
+  private static final Logger LOGGER = LoggerFactory.getLogger(StoredByteBody.class);
 
   static final Charset UTF_8 = StandardCharsets.UTF_8;
   static final Charset ISO_8859_1 = StandardCharsets.ISO_8859_1;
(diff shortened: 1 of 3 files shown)


smola (approved):

jandro996 (approved):

smola (approved):
Looks good to me if tests pass.

dd-java-agent/instrumentation/servlet-common/src/main/java/datadog/trace/instrumentation/servlet/AbstractServletInputStreamWrapper.java
```diff
@@ -16,44 +23,67 @@ public AbstractServletInputStreamWrapper(ServletInputStream is, StoredByteBody s
...
-      storedByteBody.maybeNotifyAndBlock();
+    try {
+      if (numRead > 0) {
+        storedByteBody.appendData(b, 0, numRead);
+      } else if (numRead == -1) {
+        storedByteBody.maybeNotifyAndBlock();
+      }
+    } catch (final BlockingException block) {
+      throw block;
+    } catch (final Throwable e) {
```
amarziali, line 34: I may misunderstand the fix on this PR but I think that suppressing the potential IOException at this level will break the servlet handling. Indeed I think that, according to the stacktrace in the description, the issue is on our class `StoredByteBody`. I would probably had a look fixing the root cause or silent the exception at this point:
```
charsetDecoder.decode(this.undecodedData, this.decodedData, endOfInput);
```

amarziali (approved):
it looks ok. As a nit I would have probably silented upstream on the StoredByteBody class in case is or will be used elsewhere it will save some bugfixes in the future

cataphract (approved):

manuel-alvarez-alvarez:
> it looks ok. As a nit I would have probably silented upstream on the StoredByteBody class in case is or will be used elsewhere it will save some bugfixes in the future

Yep, you are right, I didn't do it as silencing possible exceptions at `StoredByteBody` is not as clean, but I've moved the try/catch logic and silenced `StoredCharBody` too (just in case).

jandro996 (approved):

PR #6178, opened 2023-11-09
Add support for synthetic vars for metric probes

jpbempel:
# What Does This Do

Add support of `@return` and `@duration` for method metric probes if the return value of the method is compatible with long or double (like integer, byte, short or float) we can use `@return` as input for the metric.
`@duration` is now supported as input for metric also the value is a double representing the ms elapsed for the execution of the method

# Motivation
There was no support for `@return` and `@duration` for metric probes

# Additional Notes

Jira ticket: [DEBUG-1854]

[DEBUG-1854]: https://datadoghq.atlassian.net/browse/DEBUG-1854?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ

### DIFF
--- a/dd-java-agent/agent-debugger/src/main/java/com/datadog/debugger/instrumentation/MetricInstrumentor.java
+++ b/dd-java-agent/agent-debugger/src/main/java/com/datadog/debugger/instrumentation/MetricInstrumentor.java
@@ -11,6 +11,8 @@ import static com.datadog.debugger.instrumentation.ASMHelper.isStaticField;
 import static com.datadog.debugger.instrumentation.ASMHelper.ldc;
 import static com.datadog.debugger.instrumentation.Types.*;
 import static datadog.trace.util.Strings.getClassName;
+import static org.objectweb.asm.Type.DOUBLE_TYPE;
+import static org.objectweb.asm.Type.LONG_TYPE;
 
 import com.datadog.debugger.el.InvalidValueException;
 import com.datadog.debugger.el.Visitor;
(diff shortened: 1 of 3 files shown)


dd-java-agent/agent-debugger/src/main/java/com/datadog/debugger/instrumentation/MetricInstrumentor.java
```diff
@@ -727,5 +752,58 @@ private ASMHelper.Type tryRetrieveField(
...
+
+    private ASMHelper.Type tryRetrieveSynthetic(String name, InsnList insnList) {
+      if (name.equals(ValueReferences.RETURN_REF)) {
+        if (instrumentor.metricProbe.getEvaluateAt() != MethodLocation.EXIT) {
+          return null;
+        }
+        if (returnContext == null) {
+          return null;
+        }
+        VarInsnNode varInsnNode =
```
shatzi, line 764: crazy cool.

dd-java-agent/agent-debugger/src/main/java/com/datadog/debugger/instrumentation/MetricInstrumentor.java
```diff
@@ -727,5 +752,58 @@ private ASMHelper.Type tryRetrieveField(
...
+        insnList.add(varInsnNode);
+        // stack [return_value]
+        return new ASMHelper.Type(returnContext.type);
+      }
+      if (name.equals(ValueReferences.DURATION_REF)) {
+        if (instrumentor.metricProbe.getEvaluateAt() != MethodLocation.EXIT) {
+          return null;
+        }
+        // call System.nanoTime at the beginning of the method
+        int var = instrumentor.newVar(LONG_TYPE);
```
shatzi, line 775: even more crazy cool.

shatzi (approved):

shatzi:
what if someone use @return on line method? does it fails to compille? maybe add a unit tests for that?

jpbempel:
> what if someone use @return on line probe? does it fails to compile? maybe add a unit tests for that?

yes it fails to compile saying that the symbol cannot be resolved because we don't have the ReturnContext like a void method

PR #6179, opened 2023-11-09
Avoid calculate Kafka message size when DSM is not enabled

amarziali:
# What Does This Do

Skip costly message size calculation for kafka when DSM is off

Note: I could directly shortcut the whole checkpointing but I preferred to keep code changes minimal

# Motivation

# Additional Notes


PerfectSlayer (approved):

vandonr (approved):
thx

mcculls (approved):

PR #6186, opened 2023-11-09
Context propagation improvements

PerfectSlayer:
# What Does This Do

This PR improves the context propagation behavior by default. It will: 
* Change the default propagator style from `datadog` to `datadog,tracecontext`,
* Enable the generation of 128-bit trace identifier (instead of 64-bit only) by default,
* Extract multiple propagated context according the new W3C trace context initiative,
* Add `DD_TRACE_PROPAGATION_EXTRACT_FIRST` configuration to limit to the first valid one,
* Propagate W3C tracestate if matches the extracted context, even if not a Datadog one,
* Add span links of terminated contexts (terminated but not matching).

# Motivation

This will improve W3C tracecontext standard compatibility by default and interoperability with other tracing systems.
W3C trace state will be extracted from W3C extracted context even using Datadog headers. 

# Additional Notes

Jira ticket: [APMJAVA-1049]

There are multiple fixes coming for the system-tests but they are not ready yet.

[APMJAVA-1049]: https://datadoghq.atlassian.net/browse/APMJAVA-1049?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ


mcculls (approved):
déjà vu... LGTM

PR #6201, opened 2023-11-13
Disable most smoke tests for J9 JVMs

jpbempel:
# What Does This Do
Disable tests for now on those JVMs as this is not priority for fixing them use annotation `@DisabledIf` and `datadog.trace.api.Platform#isJ9`

# Motivation
Lot of flakiness running smoke tests on IBM JVMs. 

# Additional Notes


ojung (approved):
Nice! TIL 👍

PR #6209, opened 2023-11-13
Fix SQL Server connection instrumentation driver class allow list

jmeunier28:
# What Does This Do

SQL Server classes were added in both whitelists, due to a mistake made during a refactor of PR https://github.com/DataDog/dd-trace-java/pull/6160

The PR had to be rebased several times, and so I think these classes snuck back into the `DBMCompatibleConnectionInstrumentation` class erroneously 

# Motivation

Unfortunately, this means that #6160 didn't actually fix the issue described in that PR & we still see these errors 

```
exception running jdbc Incorrect syntax near '{'.
```

for SQL Server. This fix has been confirmed internally. 

# Additional Notes

Jira ticket: [PROJ-IDENT]


dd-java-agent/instrumentation/jdbc/src/main/java/datadog/trace/instrumentation/jdbc/DefaultConnectionInstrumentation.java
```diff
@@ -38,6 +38,8 @@ public class DefaultConnectionInstrumentation extends AbstractConnectionInstrume
     "oadd.org.apache.calcite.avatica.AvaticaConnection",
     // SAP HANA in-memory DB
     "com.sap.db.jdbc.ConnectionSapDB",
+    // IBM Informix
+    "com.informix.jdbc.IfmxConnection",
```
jmeunier28, line 42: This DBMS is not supported by DBM at all, so it should be in the default connection whitelist

mcculls (approved):
LGTM, thanks

mcculls (approved):

PR #6225, opened 2023-11-15
Revert "google-pubsub temporary disabled by default (#6222)"

amarziali:
# What Does This Do

**PLEASE DO NOT MERGE BEFORE Dec 2**

This reverts commit 4131e509a94db430b47104769800ec14de5f0a0d.

# Motivation

# Additional Notes

Jira ticket: [PROJ-IDENT]


dd-java-agent/instrumentation/google-pubsub/src/main/java/datadog/trace/instrumentation/googlepubsub/PublisherInstrumentation.java
```diff
@@ -29,12 +29,7 @@ public final class PublisherInstrumentation extends Instrumenter.Tracing
     implements Instrumenter.ForSingleType, ExcludeFilterProvider {
 
   public PublisherInstrumentation() {
-    super("google-pubsub", "google-pubsub-publisher");
```
amarziali, line 32: I will add a commit to preserve this double instrumentation name that's handy

mcculls (approved):

PR #6228, opened 2023-11-15
add instrumentation to allow sending kafka payload size on produce

vandonr:
I worked on this topic on #6045 but upon closer review, realized that I forgot to tool the produce side, which should send the payload size as well, according to [the spec](https://docs.google.com/document/d/1AD3AB4nd-aTpEZ_AjWPJZlKHMCD7SYGIvHX5Zr5lnx8/edit?pli=1#heading=h.rax57dhxcx9x)

To minimize the performance impact of computing the payload size, I'm reusing the result of such a computation that already happens in the kafka client as part of the `send` code, [here](https://github.com/apache/kafka/blob/e7f4f5dfe7ef2c95f3e9a1dbbeded8502d3e1e05/clients/src/main/java/org/apache/kafka/clients/producer/KafkaProducer.java#L1024), by instrumenting this method.

To do that, the solution I found was to hijack the code where the metric is sent, to save it until we know the payload size, and send it only then.

I see 3 ways to save the stats before we send them:
 * a ContextStore of <Span, StatsPoint>, but I think we don't want to use spans as keys of a context store
 * serializing the StatsPoint in a string, and storing it as a `baggageItem` in the span, but that's unnecessary string manipulations
 * adding a field in the pathway context, which is what I did.


internal-api/src/main/java/datadog/trace/bootstrap/instrumentation/api/AgentPropagation.java
```diff
@@ -14,6 +14,9 @@ public interface AgentPropagation {
   <C> void injectPathwayContext(
       AgentSpan span, C carrier, Setter<C> setter, LinkedHashMap<String, String> sortedTags);
 
+  /** This method does not set a checkpoint */
+  <C> void injectPathwayContext(AgentSpan span, C carrier, Setter<C> setter);
```
vandonr, line 18: should I use a more explicit name like `injectPathwayContextNoCheckpoint` ? Or is this good enough ?

dd-java-agent/instrumentation/kafka-clients-0.11/src/main/java/datadog/trace/instrumentation/kafka_clients/KafkaProducerInstrumentation.java
```diff
@@ -130,4 +139,32 @@ public static void stopSpan(
...
+
+    /**
+     * Instrumentation for the method KafkaProducer.ensureValidRecordSize that is called as part of
+     * sending a kafka payload. This gives us access to an estimate of the payload size "for free",
+     * that we send as a metric.
+     */
+    @Advice.OnMethodEnter(suppress = Throwable.class)
+    public static void onEnter(@Advice.Argument(value = 0) int estimatedPayloadSize) {
+      final AgentSpan encompassingSpan = activeSpan();
+      System.out.println("current span = " + encompassingSpan);
```
vandonr, line 153: this is currently always null :(

amarziali: probably because the span is not activated on the other hook point.  Some of the changes you introduced seem to break all the kafka tests (see circleci). In particular this seems to bubble up the advice code before the span is actually activated:
```
16:55:05.999 [Test worker] DEBUG datadog.trace.core.propagation.HttpCodec - Inject context DDSpan [ t_id=2, s_id=3, p_id=1 ] trace=kafka/kafka.produce/Produce Topic shared.topic *measured* tags={component=java-kafka, messaging.kafka.bootstrap.servers=127.0.0.1:41827, span.kind=producer, thread.id=1, thread.name=Test worker}
16:55:06.001 [Test worker] DEBUG datadog.trace.bootstrap.ExceptionLogger - Failed to handle exception in instrumentation for org.apache.kafka.clients.producer.KafkaProducer
java.lang.IllegalStateException: Context must be started to encode
	at datadog.trace.core.datastreams.DefaultPathwayContext.encode(DefaultPathwayContext.java:196)
	at datadog.trace.core.datastreams.DataStreamContextInjector.injectBinaryPathwayContext(DataStreamContextInjector.java:64)
	at datadog.trace.core.datastreams.DataStreamContextInjector.doInject(DataStreamContextInjector.java:52)
	at datadog.trace.core.datastreams.DataStreamContextInjector.injectPathwayContext(DataStreamContextInjector.java:39)
```

vandonr: yes I just saw that with Richard. Turns out setting the checkpoint is doing more than I thought :/

amarziali (commented):
If payload size stats are really required on producer side, separating the stat calculation from the injection is a required step at this point.
It looks ok but please ask an explicit review to apm-java and dsm folks (because of the API change).

dd-trace-core/src/main/java/datadog/trace/core/datastreams/DataStreamContextInjector.java
```diff
@@ -34,14 +34,37 @@ public <C> void injectPathwayContext(
       LinkedHashMap<String, String> sortedTags,
       long defaultTimestamp,
       long payloadSizeBytes) {
-    PathwayContext pathwayContext = span.context().getPathwayContext();
+    injectPathwayContext(
+        span, carrier, setter, sortedTags, defaultTimestamp, payloadSizeBytes, false);
```
bantonsson, line 38: This looks like a change in behavoir. Is that intentional?

bantonsson (commented):
Looks good. Only one minor question.

dd-trace-core/src/main/java/datadog/trace/core/datastreams/DataStreamContextInjector.java
vandonr: hmm, nope... good catch !
This also means that tests are lacking here 😬

bantonsson (approved):

PR #6239, opened 2023-11-17
Store OpenTelemetry span kind as tag instead of span type and drop reserved attributes from span tags

PerfectSlayer:
# What Does This Do

This PR is a port of #6232 bug fix.
It also prevents reserved attributes being sent as span tags.

> [!WARNING]  
> This will change the span type for span issued from OpenTelemetry instrumentations.

# Motivation

This behavior was recently describes in the OTel API support RFC. 

# Additional Notes

Jira ticket: [APMJAVA-1062]


ygree (approved):
LGTM

PR #6240, opened 2023-11-17
add option to collect heap histogram after GC or periodically

richardstartin:
# What Does This Do

Currently enabling the heap histogram triggers a GC once every 60s, which has the advantage of frequently collecting data about heap occupancy. This change adds a mode to only collect the histogram when there is a natural GC and makes it the default.

# Motivation

# Additional Notes

Jira ticket: [PROJ-IDENT]

### DIFF
--- a/dd-java-agent/agent-profiling/profiling-controller-openjdk/src/main/java/com/datadog/profiling/controller/openjdk/OpenJdkController.java
+++ b/dd-java-agent/agent-profiling/profiling-controller-openjdk/src/main/java/com/datadog/profiling/controller/openjdk/OpenJdkController.java
@@ -20,6 +20,8 @@ import static com.datadog.profiling.controller.ProfilingSupport.isObjectCountPar
 import static datadog.trace.api.Platform.isJavaVersionAtLeast;
 import static datadog.trace.api.config.ProfilingConfig.PROFILING_HEAP_HISTOGRAM_ENABLED;
 import static datadog.trace.api.config.ProfilingConfig.PROFILING_HEAP_HISTOGRAM_ENABLED_DEFAULT;
+import static datadog.trace.api.config.ProfilingConfig.PROFILING_HEAP_HISTOGRAM_MODE;
+import static datadog.trace.api.config.ProfilingConfig.PROFILING_HEAP_HISTOGRAM_MODE_DEFAULT;
 import static datadog.trace.api.config.ProfilingConfig.PROFILING_ULTRA_MINIMAL;
 
 import com.datadog.profiling.controller.ConfigurationException;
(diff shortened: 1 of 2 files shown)


dd-trace-api/src/main/java/datadog/trace/api/config/ProfilingConfig.java
```diff
@@ -190,5 +190,8 @@ public final class ProfilingConfig {
   public static final String PROFILING_HEAP_HISTOGRAM_ENABLED = "profiling.heap.histogram.enabled";
   public static final boolean PROFILING_HEAP_HISTOGRAM_ENABLED_DEFAULT = false;
 
+  public static final String PROFILING_HEAP_HISTOGRAM_MODE = "profiling.heap.histogram.mode";
```
jbachorik, line 193: Can you add a comment here what are the expected values?

dd-java-agent/agent-profiling/profiling-controller-openjdk/src/main/java/com/datadog/profiling/controller/openjdk/OpenJdkController.java
```diff
@@ -110,7 +112,15 @@ public OpenJdkController(final ConfigProvider configProvider)
             "enabling Datadog heap histogram on JVM without an efficient implementation of the jdk.ObjectCount event. "
                 + "This may increase p99 latency. Consider upgrading to JDK 17.0.9+ or 21+ to reduce latency impact.");
       }
-      enableEvent(recordingSettings, "jdk.ObjectCount", "user enabled histogram heap collection");
+      String mode =
+          configProvider.getString(
+              PROFILING_HEAP_HISTOGRAM_MODE, PROFILING_HEAP_HISTOGRAM_MODE_DEFAULT);
+      if ("periodic".equalsIgnoreCase(mode)) {
+        enableEvent(recordingSettings, "jdk.ObjectCount", "user enabled histogram heap collection");
```
jbachorik, line 119: Do you think allowing to specify the period could be useful, or is it just 'meh'?

dd-trace-api/src/main/java/datadog/trace/api/config/ProfilingConfig.java
richardstartin: ok

dd-java-agent/agent-profiling/profiling-controller-openjdk/src/main/java/com/datadog/profiling/controller/openjdk/OpenJdkController.java
richardstartin: meh

jbachorik (approved):

PR #6241, opened 2023-11-17
Remove ReferenceQueue from TaintedMap implementation and purge inline

manuel-alvarez-alvarez:
# What Does This Do
Replaces the original tainted map implementation based on a queue by a map that purges stale entries while reading or writing data.

# Motivation
We want to move towards a global instance of the tainted map (this way we wont require access to the current tracing context) in order to interact with the map. Previous implementation might introduce delays (when the map is going to be purged) that makes it not very suitable for a global instance.

# Additional Notes

Jira ticket: [PROJ-IDENT]

Some benchmarks

**Before**

| Benchmark | Mode | Cnt | Score | Error | Units |
| --------------- | -------- | ---- | -------- | ------- | ------- |
| TaintedMapEmptyBenchmark.getFromEmptyMap  | avgt  |  9  | 2.628 | ± 0.096 | ns/op |
| TaintedMapGetsBenchmark.gets | avgt | 9 | 4.485 | ± 0.949 | ns/op |
| TaintedMapGetsBenchmark.getsBaseline | avgt | 9 | 2.586 |± 0.091 |  ns/op |
| TaintedMapPutsBenchmark.puts | avgt | 9  | 56.547 | ± 2.568 | ns/op |
| TaintedMapPutsBenchmark.putsBaseline | avgt | 9 | 54.387 | ± 2.855 | ns/op |

**After**

| Benchmark | Mode | Cnt | Score | Error | Units |
| --------------- | -------- | ---- | -------- | ------- | ------- |
| TaintedMapEmptyBenchmark.baseline |     avgt   |  9   |  2.987 | ±  0.180 |  ns/op | 
| TaintedMapEmptyBenchmark.purgeInline |  avgt  |   9  |   3.320 | ±  0.157  | ns/op | 
| TaintedMapEmptyBenchmark.purgeQueue |   avgt |    9 |    2.872 |  ±  0.123 |  ns/op | 
| TaintedMapGetsBenchmark.baseline  |     avgt |    9  |   2.649 |  ±  0.264 |  ns/op | 
| TaintedMapGetsBenchmark.purgeInline  |  avgt  |   9 |    6.902 |  ±  0.370 |  ns/op | 
| TaintedMapGetsBenchmark.purgeQueue  |   avgt   |  9  |   5.036 | ±  0.555 |  ns/op | 
| TaintedMapPutsBenchmark.baseline   |    avgt |   15  |  90.647 | ±  3.215 |  ns/op | 
| TaintedMapPutsBenchmark.purgeInline |   avgt  |  15  | 116.038 | ±  3.248 |  ns/op | 
| TaintedMapPutsBenchmark.purgeQueue   |  avgt  |  15  | 184.431 | ± 10.006 |  ns/op |


dd-java-agent/agent-iast/src/main/java/com/datadog/iast/taint/TaintedMap.java
```diff
@@ -29,294 +34,582 @@
...
- * the head of the bucket, and any tail (colliding entries) will be discarded. Once a map switches
- * to flat mode, it never goes back from it. Note that entries for garbage collected entries are
- * removed before this threshold is checked.
+ * <p><i>Capacity</i> is fixed, so there is no rehashing.
  *
  * <p>This implementation works reasonably well under high concurrency, but it will lose some writes
  * in that case.
  */
-public final class TaintedMap implements Iterable<TaintedObject> {
+public interface TaintedMap extends Iterable<TaintedObject> {
```
smola, line 42: If the inline purge seems to be better, it'd probably be better to make it the only implementation and avoid all the extra code to be able to switch them. This would also allow us to remove more code (stuff related to flat mode, reference queue, etc).

manuel-alvarez-alvarez: Done, only one implementation remains, let's see how the perf benchmarks react :wink:

dd-java-agent/agent-iast/src/main/java/com/datadog/iast/taint/TaintedMap.java
```diff
@@ -29,294 +30,339 @@
...
-          return true;
-        }
-        for (; currentIndex < stop; currentIndex++) {
-          if (table[currentIndex] != null) {
-            return true;
-          }
-        }
-        return false;
+    @Override
+    public void put(@Nonnull final TaintedObject entry) {
```
DDJavierSantos, line 242: Maybe we could add extra debug information in this case. It would be interesting to log the actual identity hash of the object being tainted, so we could compare two tainted strings with the same content. I can remember at least a couple of cases where this would have allowed an easier troubleshooting.

DDJavierSantos (approved):

dd-java-agent/agent-iast/src/main/java/com/datadog/iast/taint/TaintedMap.java
manuel-alvarez-alvarez: Let's discuss offline what other metrics we can add, and do an specific PR to add support for them.

PR #6249, opened 2023-11-20
Spring scheduling: have actuator display the right task class name

### LINKED ISSUE #5651: "Wrong" class names are shown in Actuator overview of scheduled tasks in Spring Boot application
Instead of real class and method names the class `datadog.trace.instrumentation.springscheduling.SpringSchedulingRunnableWrapper` is shown when you open the scheduled tasks overview from Actuator (`/actuator/scheduledtasks`):

image

There was a [similar issue in `opentelemetry-java-instrumentation`](https://github.com/open-telemetry/opentelemetry-java-instrumentation/issues/6103), I would suggest to peek in their fix implementation.

amarziali:
# What Does This Do

Solves #5651 

# Motivation

# Additional Notes

Also add tests for spring 6

### DIFF
--- a/dd-java-agent/instrumentation/spring-scheduling-3.1/build.gradle
+++ b/dd-java-agent/instrumentation/spring-scheduling-3.1/build.gradle
@@ -1,4 +1,7 @@
-
+ext {
+  spring6TestMinJavaVersionForTests = JavaVersion.VERSION_17
+  spring6LatestDepTestMinJavaVersionForTests = JavaVersion.VERSION_17
+}
 muzzle {
   pass {
     group = 'org.springframework'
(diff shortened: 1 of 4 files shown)


dd-java-agent/instrumentation/spring-scheduling-3.1/build.gradle
```diff
@@ -1,35 +1,103 @@
...
+}
+
 
 dependencies {
   // choose a recent version so that we can test both lambdas (JDK8)
   // @Async requires proxying and older versions can't read classfile versions > 51
   // we muzzle older versions of spring anyway
-  compileOnly group: 'org.springframework', name: 'spring-context', version: '5.0.0.RELEASE'
-  testImplementation group: 'org.springframework', name: 'spring-context', version: '5.0.0.RELEASE'
+  main_java17CompileOnly group: 'org.springframework', name: 'spring-context', version: '5.0.0.RELEASE'
```
ygree, line 71: Could you clarify why do we need to compile instrumentation with Java 17?
Could we still compile it with Java 8 and use Java 17 for tests only?

ygree (approved):
Looks good to me. I have one question about Java 17 though.

dd-java-agent/instrumentation/spring-scheduling-3.1/build.gradle
amarziali: Indeed you are right. I will adjust  thanks

PR #6251, opened 2023-11-20
Implement coverage for Cucumber .feature files

nikita-tkachenko-datadog:
# What Does This Do
When running instrumented Cucumber scenarios, their corresponding .feature files are added to the list of covered files.

# Motivation
Needed for ITR to work in Cucumber projects: if a .feature file is modified, ITR should not be skipping tests that are defined in that file. 
Therefore the files should be included in the tests' coverage sets.

# Additional Notes

Jira ticket: [CIVIS-7993]

[CIVIS-7993]: https://datadoghq.atlassian.net/browse/CIVIS-7993?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ


dd-java-agent/agent-ci-visibility/src/main/java/datadog/trace/civisibility/source/index/SourceType.java
```diff
@@ -1,23 +1,30 @@
...
+  JAVA(".java", false),
+  GROOVY(".groovy", false),
+  KOTLIN(".kt", false),
+  SCALA(".scala", false),
+  GHERKIN(".feature", true);
 
   private static final SourceType[] UNIVERSE = SourceType.values();
 
   private final String extension;
+  private final boolean resource;
```
romainkomorndatadog, line 13: very nitpicky but I find the mismatch between `resource` and `isResource` disturbing.

romainkomorndatadog (approved):

dd-java-agent/agent-ci-visibility/src/main/java/datadog/trace/civisibility/source/index/SourceType.java
nikita-tkachenko-datadog: This is one of the standard naming conventions in Java: if there's a boolean field named `something` then its getter method is usually named `isSomething()`.

PR #6252, opened 2023-11-20
Add Remote Config based enablement for Symbol DB

jpbempel:
# What Does This Do
If SymbolDB is enabled we subscribe to Remote Config new product LIVE_DEBUGGING_SYMBOL_DB to receive the RC record indicating if we extract and upload symbol for the current instance. As most classes may have been loaded by the time we receive the upload symbol flag we need to scan all loaded classes that matches our criteria for symbol extraction and scan the actual jar for all those allowed classes and parse it and extract symbols

# Motivation

# Additional Notes

Jira ticket: [DEBUG-1916]

[DEBUG-1916]: https://datadoghq.atlassian.net/browse/DEBUG-1916?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ


dd-java-agent/agent-debugger/src/main/java/com/datadog/debugger/symbol/SymbolExtractionTransformer.java
```diff
@@ -77,64 +39,29 @@ public byte[] transform(
...
-    String jarName = "DEFAULT";
-    if (protectionDomain != null) {
-      CodeSource codeSource = protectionDomain.getCodeSource();
-      if (codeSource != null) {
-        URL location = codeSource.getLocation();
-        if (location != null) {
-          jarName = location.getFile();
+    try {
+      if (allowListHelper.isAllowAll()) {
+        if (className.startsWith("java/")
```
shatzi, line 44: are we sure we don't want to black list those domain even if the user provides an include list?

dd-java-agent/agent-debugger/src/test/java/com/datadog/debugger/symbol/SymDBEnablementTest.java
```diff
@@ -0,0 +1,95 @@
...
+  }
+
+  @Test
+  public void enableDisableSymDBThroughRC() throws IOException {
+    SymDBEnablement symDBEnablement =
+        new SymDBEnablement(instr, config, new SymbolAggregator(symbolSink, 1));
+    symDBEnablement.accept(ParsedConfigKey.parse(CONFIG_KEY), UPlOAD_SYMBOL_TRUE, null);
+    verify(instr).addTransformer(any(SymbolExtractionTransformer.class), eq(true));
+
+    symDBEnablement.accept(ParsedConfigKey.parse(CONFIG_KEY), UPlOAD_SYMBOL_FALSE, null);
```
shatzi, line 52: add test for config removed?

shatzi (approved):
overall I think this is great. few comments regarding handling errors and return fast when accept config is called.

dd-java-agent/agent-debugger/src/main/java/com/datadog/debugger/symbol/SymDBEnablement.java
```diff
@@ -0,0 +1,180 @@
...
+  }
+
+  private void extractSymbolForLoadedClasses(AllowListHelper allowListHelper) {
+    Class<?>[] classesToExtract = null;
+    try {
+      classesToExtract =
+          Arrays.stream(instrumentation.getAllLoadedClasses())
+              .filter(clazz -> allowListHelper.isAllowed(clazz.getTypeName()))
+              .filter(instrumentation::isModifiableClass)
+              .toArray(Class<?>[]::new);
```
ojung, line 116: Ah, so we're scanning the class loader for all loaded classes and get the jar path from them instead of walking the file system. It's more safe I'm sure but what are the edge cases here? If symbol extraction is triggered very early in the application (e.g. before the first HTTP request was made to an API), it might be that we're missing the class / jar, right?

dd-java-agent/agent-debugger/src/main/java/com/datadog/debugger/symbol/SymDBEnablement.java
```diff
@@ -0,0 +1,180 @@
...
+      }
+      try {
+        try (JarFile jarFile = new JarFile(jarPath.toFile())) {
+          jarFile.stream()
+              .filter(jarEntry -> jarEntry.getName().endsWith(".class"))
+              .filter(
+                  jarEntry ->
+                      allowListHelper.isAllowed(
+                          Strings.getClassName(trimPrefixes(jarEntry.getName()))))
+              .forEach(jarEntry -> parseJarEntry(jarEntry, jarFile, jarPath, baos, buffer));
```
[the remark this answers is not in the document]
jpbempel: `parseJarEntry` handle exceptions

dd-java-agent/agent-debugger/src/main/java/com/datadog/debugger/symbol/SymDBEnablement.java
```diff
@@ -0,0 +1,180 @@
...
+    return SYM_DB_JSON_ADAPTER.fromJson(
+        Okio.buffer(Okio.source(new ByteArrayInputStream(content))));
+  }
+
+  public void stopSymbolExtraction() {
+    LOGGER.debug("Stopping symbol extraction.");
+    instrumentation.removeTransformer(symbolExtractionTransformer);
+  }
+
+  public void startSymbolExtraction() {
```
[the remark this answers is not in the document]
jpbempel: done

dd-java-agent/agent-debugger/src/main/java/com/datadog/debugger/symbol/SymDBEnablement.java
```diff
@@ -0,0 +1,180 @@
...
+      } else {
+        stopSymbolExtraction();
+      }
+    } else {
+      throw new IOException("unsupported configuration id " + configKey.getConfigId());
+    }
+  }
+
+  @Override
+  public void remove(
```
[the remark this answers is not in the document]
jpbempel: done

dd-java-agent/agent-debugger/src/main/java/com/datadog/debugger/agent/DebuggerAgent.java
```diff
@@ -128,9 +137,13 @@ private static void loadFromFile(
   }
 
   private static void subscribeConfigurationPoller(
-      Config config, ConfigurationUpdater configurationUpdater) {
+      Config config, ConfigurationUpdater configurationUpdater, SymDBEnablement symDBEnablement) {
     configurationPoller.addListener(
```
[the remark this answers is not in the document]
jpbempel: done

dd-java-agent/agent-debugger/src/main/java/com/datadog/debugger/symbol/SymDBEnablement.java
```diff
@@ -0,0 +1,180 @@
...
+              .filter(clazz -> allowListHelper.isAllowed(clazz.getTypeName()))
+              .filter(instrumentation::isModifiableClass)
+              .toArray(Class<?>[]::new);
+    } catch (Throwable ex) {
+      LOGGER.debug("Failed to get all loaded classes", ex);
+      return;
+    }
+    Set<String> alreadyScannedJars = new HashSet<>();
+    byte[] buffer = new byte[4096];
+    ByteArrayOutputStream baos = new ByteArrayOutputStream(8192);
```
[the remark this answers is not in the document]
jpbempel: static buffer of 4K for each read
stream can be resized
moved as constant

dd-java-agent/agent-debugger/src/main/java/com/datadog/debugger/symbol/SymbolExtractionTransformer.java
jpbempel: for me includes overrides everything it allows to have symbol for somehting that regularly we try to avoid, but in some case you may want to have them. not a security problem like redaction

dd-java-agent/agent-debugger/src/test/java/com/datadog/debugger/symbol/SymDBEnablementTest.java
jpbempel: done

dd-java-agent/agent-debugger/src/main/java/com/datadog/debugger/symbol/SymDBEnablement.java
jpbempel: we register the transformer for that...

ojung: Ah! missed that.

dd-java-agent/agent-debugger/src/main/java/com/datadog/debugger/symbol/SymDBEnablement.java
```diff
@@ -0,0 +1,180 @@
...
+      throws IOException {
+    if (configKey.getConfigId().equals("symDb")) {
+      SymDbRemoteConfigRecord symDb = deserializeSymDb(content);
+      if (symDb.isUploadSymbols()) {
+        startSymbolExtraction();
+      } else {
+        stopSymbolExtraction();
+      }
+    } else {
+      throw new IOException("unsupported configuration id " + configKey.getConfigId());
```
ojung, line 69: We probably shouldn't throw here since it might be that we're adding more `symDB_*` configs in the future. I would log and ignore instead.

dd-java-agent/agent-debugger/src/main/java/com/datadog/debugger/symbol/SymDBEnablement.java
```diff
@@ -0,0 +1,180 @@
...
+          LocalDateTime.ofInstant(
+              Instant.ofEpochMilli(lastUploadTimestamp), ZoneId.systemDefault()));
+      return;
+    }
+    String includes = config.getDebuggerSymbolIncludes();
+    AllowListHelper allowListHelper = new AllowListHelper(buildFilterList(includes));
+    symbolExtractionTransformer =
+        new SymbolExtractionTransformer(allowListHelper, symbolAggregator);
+    instrumentation.addTransformer(symbolExtractionTransformer, true);
+    extractSymbolForLoadedClasses(allowListHelper);
```
ojung, line 105: We need to make sure that we upload no class twice. Is this taken care of in some way?

dd-java-agent/agent-debugger/src/main/java/com/datadog/debugger/symbol/SymDBEnablement.java
jpbempel: disagree because we are indirectly informing the RC backend that we accept this configuration, I prefer to reject it until we have an instance that handles it correctly. That way the backend can still send us again the config later

dd-java-agent/agent-debugger/src/main/java/com/datadog/debugger/symbol/SymDBEnablement.java
jpbempel: not completely: there is still a race between adding the transformer and snapshoting the classes (though very short)
I need to add a special mechanism to handle this case to synchronize the transformer and the reload of existing classes. Will do in a next PR.

dd-java-agent/agent-debugger/src/main/java/com/datadog/debugger/symbol/SymDBEnablement.java
ojung: This is a problem though because when we add new config ids, older versions of the library will be unusable. It's actually the same issue with DI configs IIRC.

jpbempel: ok, good point.
I make the modification for SymDB at least, will do DI configs in a separate PR

dd-java-agent/agent-debugger/src/main/java/com/datadog/debugger/symbol/SymDBEnablement.java
ojung: With your latest change making reloading existing classes async this race is made worse, right?

jpbempel: no nothing change because this is still the same thread that register the transformer and snapshot the loaded classes

ojung: Right, I see. Cool, let's go with this then. The effect of this is when a class is uploaded twice, it will show up in search and completions twice. It's more difficult to filter this in the backend so it would be great to find a way to guarantee that each class is only uploaded once or at least make it very unlikely that it's uploaded twice.

dd-java-agent/agent-debugger/src/test/java/com/datadog/debugger/symbol/SymDBEnablementTest.java
shatzi: I don't think this is a good test as you upload both false config and null config once after the other. we should test that only one of them is also causing removing of transformer. WDYT?

shatzi (approved):

PR #6255, opened 2023-11-21
Upgrade ddprof to 0.88.0

richardstartin:
# What Does This Do

Avoids unwinding Java stacks from within `call_stub`

# Motivation

# Additional Notes

Jira ticket: [PROJ-IDENT]


bantonsson (approved):
The naming here is so confusing, since this `ddprof` is not [ddprof](https://github.com/DataDog/ddprof) but actually [java-profiler](https://github.com/datadog/java-profiler) (which I keep forgetting).
Would be great with a simple [link for the diff](https://github.com/DataDog/java-profiler/compare/v_0.87.0...v_0.88.0)

PR #6256, opened 2023-11-21
Fix race for SymDB enablement

jpbempel:
# What Does This Do
When we register the symbol extraction transformer and when we snapshot all loaded classes there is a race where a class can be loaded by a classloader and process by the transformer but can also be part of the loaded classes and then be processed twice. we add a temporary concurrent set to track class names that are parsed by the SymbolAggregator

# Motivation

# Additional Notes

Jira ticket: [DEBUG-1916]

[DEBUG-1916]: https://datadoghq.atlassian.net/browse/DEBUG-1916?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ

### DIFF
--- a/dd-java-agent/agent-debugger/src/main/java/com/datadog/debugger/symbol/SymDBEnablement.java
+++ b/dd-java-agent/agent-debugger/src/main/java/com/datadog/debugger/symbol/SymDBEnablement.java
@@ -27,6 +27,7 @@ import java.util.Arrays;
 import java.util.Collections;
 import java.util.HashSet;
 import java.util.Set;
+import java.util.concurrent.atomic.AtomicBoolean;
 import java.util.jar.JarEntry;
 import java.util.jar.JarFile;
 import java.util.regex.Pattern;
@@ -48,6 +49,7 @@ public class SymDBEnablement implements ProductListener {
   private final SymbolAggregator symbolAggregator;
   private SymbolExtractionTransformer symbolExtractionTransformer;
   private volatile long lastUploadTimestamp;
+  private AtomicBoolean starting = new AtomicBoolean();
 
   public SymDBEnablement(
       Instrumentation instrumentation, Config config, SymbolAggregator symbolAggregator) {
@@ -102,25 +104,32 @@ public class SymDBEnablement implements ProductListener {
   }
 
   public void startSymbolExtraction() {
-    LOGGER.debug("Starting symbol extraction...");
-    if (lastUploadTimestamp > 0) {
-      LOGGER.debug(
-          "Last upload was on {}",
-          LocalDateTime.ofInstant(
-              Instant.ofEpochMilli(lastUploadTimestamp), ZoneId.systemDefault()));
+    if (!starting.compareAndSet(false, true)) {
       return;
     }
-    String includes = config.getDebuggerSymbolIncludes();
-    AllowListHelper allowListHelper = new AllowListHelper(buildFilterList(includes));
-    symbolAggregator.startLoadedClasses();
     try {
-      symbolExtractionTransformer =
-          new SymbolExtractionTransformer(allowListHelper, symbolAggregator);
-      instrumentation.addTransformer(symbolExtractionTransformer, true);
-      extractSymbolForLoadedClasses(allowListHelper);
-      lastUploadTimestamp = System.currentTimeMillis();
+      LOGGER.debug("Starting symbol extraction...");
+      if (lastUploadTimestamp > 0) {
+        LOGGER.debug(
+            "Last upload was on {}",
+            LocalDateTime.ofInstant(
+                Instant.ofEpochMilli(lastUploadTimestamp), ZoneId.systemDefault()));
+        return;
+      }
+      String includes = config.getDebuggerSymbolIncludes();
+      AllowListHelper allowListHelper = new AllowListHelper(buildFilterList(includes));
+      symbolAggregator.loadedClassesProcessStarted();
+      try {
+        symbolExtractionTransformer =
+            new SymbolExtractionTransformer(allowListHelper, symbolAggregator);
+        instrumentation.addTransformer(symbolExtractionTransformer);
+        extractSymbolForLoadedClasses(allowListHelper);
+        lastUploadTimestamp = System.currentTimeMillis();
+      } finally {
+        symbolAggregator.loadedClassesProcessEnded();
+      }
     } finally {
-      symbolAggregator.finishLoadedClasses();
+      starting.set(false);
     }
   }
 
--- a/dd-java-agent/agent-debugger/src/main/java/com/datadog/debugger/symbol/SymDBEnablement.java
+++ b/dd-java-agent/agent-debugger/src/main/java/com/datadog/debugger/symbol/SymDBEnablement.java
@@ -112,11 +112,16 @@ public class SymDBEnablement implements ProductListener {
     }
     String includes = config.getDebuggerSymbolIncludes();
     AllowListHelper allowListHelper = new AllowListHelper(buildFilterList(includes));
-    symbolExtractionTransformer =
-        new SymbolExtractionTransformer(allowListHelper, symbolAggregator);
-    instrumentation.addTransformer(symbolExtractionTransformer, true);
-    extractSymbolForLoadedClasses(allowListHelper);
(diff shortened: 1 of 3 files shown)


dd-java-agent/agent-debugger/src/test/java/com/datadog/debugger/symbol/SymDBEnablementTest.java
```diff
@@ -103,6 +109,45 @@ public void parseLoadedClass() throws ClassNotFoundException, IOException, URISy
...
+        .thenAnswer(
+            invocation -> {
+              URL jarFileUrl = getClass().getResource("/debugger-symbol.jar");
+              URL jarUrl = new URL("jar:file:" + jarFileUrl.getFile() + "!/");
+              URLClassLoader urlClassLoader = new URLClassLoader(new URL[] {jarUrl}, null);
+              Class<?> testClass = urlClassLoader.loadClass(Strings.getClassName(CLASS_NAME_PATH));
+              return new Class[] {testClass};
+            });
+    symDBEnablement.startSymbolExtraction();
+    verify(mockSymbolSink).addScope(any());
```
shatzi, line 148: does that ensure we call this only once?

dd-java-agent/agent-debugger/src/main/java/com/datadog/debugger/symbol/SymDBEnablement.java
```diff
@@ -112,11 +112,16 @@ public void startSymbolExtraction() {
...
-    symbolExtractionTransformer =
-        new SymbolExtractionTransformer(allowListHelper, symbolAggregator);
-    instrumentation.addTransformer(symbolExtractionTransformer, true);
-    extractSymbolForLoadedClasses(allowListHelper);
-    lastUploadTimestamp = System.currentTimeMillis();
+    symbolAggregator.startLoadedClasses();
+    try {
+      symbolExtractionTransformer =
+          new SymbolExtractionTransformer(allowListHelper, symbolAggregator);
+      instrumentation.addTransformer(symbolExtractionTransformer, true);
```
shatzi, line 119: we pass to true to canRestansform. does that mean we send symbols again to that class after we add/remove probes? what if pass 'false'? does that fix this issue?

dd-java-agent/agent-debugger/src/main/java/com/datadog/debugger/symbol/SymDBEnablement.java
```diff
@@ -112,11 +112,16 @@ public void startSymbolExtraction() {
     }
```
shatzi, line 112: consider that instead of return we simply just add the transformer to continue upload new loaded classes

dd-java-agent/agent-debugger/src/main/java/com/datadog/debugger/symbol/SymDBEnablement.java
```diff
@@ -112,11 +112,16 @@ public void startSymbolExtraction() {
...
-    instrumentation.addTransformer(symbolExtractionTransformer, true);
-    extractSymbolForLoadedClasses(allowListHelper);
-    lastUploadTimestamp = System.currentTimeMillis();
+    symbolAggregator.startLoadedClasses();
+    try {
+      symbolExtractionTransformer =
+          new SymbolExtractionTransformer(allowListHelper, symbolAggregator);
+      instrumentation.addTransformer(symbolExtractionTransformer, true);
+      extractSymbolForLoadedClasses(allowListHelper);
+      lastUploadTimestamp = System.currentTimeMillis();
```
shatzi, line 121: There is still a race condition here if there a bug in RC/enablement logic and we get a config, remove it before we finish analyze all jars, then we will call start again just before we update lastUploadTimestamp on a new agent task. the add/remove config would register only one transformer but we will call extractSymbolForLoadedClasses twice.

I wonder how we can fix that avoid starting a new thread if we already have a running method

dd-java-agent/agent-debugger/src/main/java/com/datadog/debugger/symbol/SymbolAggregator.java
```diff
@@ -95,4 +110,14 @@ private void addJarScope(Scope jarScope, boolean forceFlush) {
       }
     }
   }
+
+  void startLoadedClasses() {
```
ojung, line 114: I'm a bit confused by the name. What is this starting?

ojung: I understand it from the code but not from the method name.

dd-java-agent/agent-debugger/src/main/java/com/datadog/debugger/symbol/SymbolAggregator.java
```diff
@@ -10,19 +12,23 @@
...
   private static final Logger LOGGER = LoggerFactory.getLogger(SymbolAggregator.class);
+  private static final String CLASS_SUFFIX = ".class";
 
   private final SymbolSink sink;
   private final int symbolFlushThreshold;
   private final Map<String, Scope> jarScopesByName = new HashMap<>();
   private final AgentTaskScheduler.Scheduled<SymbolAggregator> scheduled;
   private final Object jarScopeLock = new Object();
   private int totalClasses;
+  private volatile Set<String> loadedClasses;
```
shatzi, line 31: we can also just keep this list in memory and avoid all the race condition logic. will consume more memory for sure but would be simpler code. I'm sure you thoughts on the tradeoff so this might be null point.

shatzi (approved):
few comments to think about

dd-java-agent/agent-debugger/src/test/java/com/datadog/debugger/symbol/SymDBEnablementTest.java
jpbempel: yes, implicitly `times(1)`

ojung, line 148: Had to look up that `verify` is basically an alias to `verify(.., times(1))`. Maybe worth it to add it explicitly?

ojung (approved):
Thanks!

dd-java-agent/agent-debugger/src/main/java/com/datadog/debugger/symbol/SymDBEnablement.java
jpbempel: yes can add an AtomicBoolean to handle the start phase and prevent any call to start again while executing the phase

dd-java-agent/agent-debugger/src/main/java/com/datadog/debugger/symbol/SymbolAggregator.java
jpbempel: keeping all class strings in memory for just a race is just wasteful

dd-java-agent/agent-debugger/src/test/java/com/datadog/debugger/symbol/SymDBEnablementTest.java
jpbempel: added explicitly

dd-java-agent/agent-debugger/src/main/java/com/datadog/debugger/symbol/SymDBEnablement.java
jpbempel: yes fixed

dd-java-agent/agent-debugger/src/main/java/com/datadog/debugger/symbol/SymDBEnablement.java
jpbempel: it means it was already done before, so no need to add something again.

dd-java-agent/agent-debugger/src/main/java/com/datadog/debugger/symbol/SymbolAggregator.java
jpbempel: renamed

PR #6259, opened 2023-11-22
Add DSM support to SQS v1

vandonr:
# What Does This Do

Add the pathway context injection necessary for DSM support to SQS when used with the v1 java sdk.

DSM is only going to be available for producers in v >= 1.11.106 because that's the first version to introduce a handler context with `AmazonWebServiceRequest`, and this is very helpful to pass the span from `beforeMarshalling` to `beforeRequest`. Not being able to use that increases the complexity a lot, and this version is old enough that it shouldn't be a big ask from customers.
Consequently, I added a new instrumentation for DSM support specifically, that's going to overlap with the existing APM instrumentation for all V1.X versions of the client.

# Motivation

DSM support had already been added to the SQS v2 client, adding support to v1 will allow more clients to benefit from it.

# Additional Notes

Jira ticket: [AIT-8676]

[AIT-8676]: https://datadoghq.atlassian.net/browse/AIT-8676?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ


dd-java-agent/instrumentation/aws-java-sdk-1.11.0/src/main/java/datadog/trace/instrumentation/aws/v0/TracingRequestHandler.java
```diff
@@ -45,6 +46,17 @@ public TracingRequestHandler(ContextStore<Object, String> responseQueueStore) {
 
   @Override
   public AmazonWebServiceRequest beforeMarshalling(final AmazonWebServiceRequest request) {
+    if (isSqsSendRequest(request)) {
+      // for SQS, create the span now so that we can use it for DSM purposes in
+      // datadog.trace.instrumentation.aws.v1.sqs.SqsInterceptor. This forces us to hardcode the
+      // service name because we cannot access it programmatically here.
+      AgentSpan span = startSpan(AwsNameCache.spanName(request, "AmazonSQS"));
+      DECORATE.afterStart(span);
```
vandonr, line 54: I could avoid one context store usage by doing this in SqsInterceptor's `beforeMarshalling` but doing it here allows me to call the `afterStart`, well, after I start it.
We could also leave the afterStart in `beforeRequest`, below. I'm not sure what are all the implications.

dd-java-agent/instrumentation/aws-java-sqs-1.0/src/main/java/datadog/trace/instrumentation/aws/v1/sqs/SqsClientInstrumentation.java
vandonr, line 1: I wrote this by mixing [the V2 instrumentation](https://github.com/DataDog/dd-trace-java/blob/2b28420f512e247394da30d40f7f93bdd588a718/dd-java-agent/instrumentation/aws-java-sqs-2.0/src/main/java/datadog/trace/instrumentation/aws/v2/sqs/SqsClientInstrumentation.java) and what was done in [the aws sdk v1 instrumentation](https://github.com/DataDog/dd-trace-java/blob/2b28420f512e247394da30d40f7f93bdd588a718/dd-java-agent/instrumentation/aws-java-sdk-1.11.0/src/main/java/datadog/trace/instrumentation/aws/v0/HandlerChainFactoryInstrumentation.java)

dd-java-agent/instrumentation/aws-java-sqs-1.0/src/main/java/datadog/trace/instrumentation/aws/v1/sqs/SqsInterceptor.java
vandonr, line 1: written with [the V2 implem](https://github.com/DataDog/dd-trace-java/blob/2b28420f512e247394da30d40f7f93bdd588a718/dd-java-agent/instrumentation/aws-java-sqs-2.0/src/main/java/datadog/trace/instrumentation/aws/v2/sqs/SqsInterceptor.java) as reference

dd-java-agent/instrumentation/aws-java-sdk-1.11.0/src/main/java/datadog/trace/instrumentation/aws/v0/TracingRequestHandler.java
```diff
@@ -45,6 +46,17 @@ public TracingRequestHandler(ContextStore<Object, String> responseQueueStore) {
...
+      FieldBackedContextStores.getContextStore(
```
amarziali, line 55: The right way to obtain a context store is to use the `InstrumentationContext.get(..)` api and not using this directly

vandonr: ah, ok, but `InstrumentationContext` only works within an `Advice`, and it's not the case here :'(

dd-java-agent/instrumentation/aws-java-sqs-1.0/src/test/groovy/SqsClientTest.groovy
```diff
@@ -138,19 +148,39 @@ abstract class SqsClientTest extends VersionedNamingTestBase {
...
+    }
+
     assert messages[0].attributes['AWSTraceHeader'] =~
     /Root=1-[0-9a-f]{8}-00000000${sendSpan.traceId.toHexStringPadded(16)};Parent=${DDSpanId.toHexStringPadded(sendSpan.spanId)};Sampled=1/
 
     cleanup:
     client.shutdown()
   }
 
+  @IgnoreIf({ instance.isDataStreamsEnabled() })
```
amarziali, line 183: why?

dd-java-agent/instrumentation/aws-java-sqs-1.0/src/main/java/datadog/trace/instrumentation/aws/v1/sqs/SqsClientInstrumentation.java
```diff
@@ -0,0 +1,52 @@
...
+
+  public static class HandlerChainAdvice {
+    @Advice.OnMethodExit(onThrowable = Throwable.class, suppress = Throwable.class)
+    public static void addHandler(@Advice.Return final List<RequestHandler2> handlers) {
+      for (RequestHandler2 interceptor : handlers) {
+        if (interceptor instanceof SqsInterceptor) {
+          return; // list already has our interceptor, return to builder
+        }
+      }
+      handlers.add(new SqsInterceptor());
```
amarziali, line 49: here you can obtain a context store and pass it to the interceptor constructor

dd-java-agent/instrumentation/aws-java-sqs-1.0/src/test/groovy/SqsClientTest.groovy
vandonr: Ah yes, it's a mistake on my part when synchronizing this with the V2 tests, will remove that line.

dd-java-agent/instrumentation/aws-java-sdk-1.11.0/src/main/java/datadog/trace/instrumentation/aws/v0/TracingRequestHandler.java
amarziali: You can obtain on an advice and pass where you need it

dd-java-agent/instrumentation/aws-java-sqs-1.0/src/main/java/datadog/trace/instrumentation/aws/v1/sqs/SqsClientInstrumentation.java
vandonr: hmm interesting, I'll try that.

dd-java-agent/instrumentation/aws-java-sqs-1.0/src/test/groovy/SqsClientTest.groovy
vandonr: Ah, it was not a mistake from me ! It's disabled in V2 as well !
I suppose the logic is that whether DSM is enabled or not has no impact on the expecations of the test, so we ignore it to not run the same test twice.
One could also argue that we could run it to make sure we didn't change the behavior with DSM. But in any case, I think it should be coherent between v1 and v2 if tests are ignored or not.

dd-java-agent/instrumentation/aws-java-sqs-1.0/src/main/java/datadog/trace/instrumentation/aws/v1/sqs/SqsInterceptor.java
amarziali: The sdk v1 also has a `beforeExecution` method but only starting from a certain version. It could be an option to support DSM only starting from this version (leaving out legacy sdk versions). That can be a quick win to reuse exactly the same logic

dd-java-agent/instrumentation/aws-java-sdk-1.11.0/src/main/java/datadog/trace/instrumentation/aws/v0/TracingRequestHandler.java
```diff
@@ -56,8 +68,18 @@ public void beforeRequest(final Request<?> request) {
       // so we can tell when receive call is complete without affecting the rest of the trace
       span = noopSpan();
     } else {
-      span = startSpan(AwsNameCache.spanName(request));
-      DECORATE.afterStart(span);
+      if (isSqsSendRequest(request.getOriginalRequest())) {
+        // for SQS, spans are created earlier, in beforeMarshalling just above.
+        span =
```
amarziali, line 73: You should just use the request context (see below) to store and retrieve this span without adding complexity (and unneeded bytecode manipulations with context stores):

```
      request.addHandlerContext(SPAN_CONTEXT_KEY, span);
```

dd-java-agent/instrumentation/aws-java-sdk-1.11.0/src/main/java/datadog/trace/instrumentation/aws/v0/TracingRequestHandler.java
amarziali: what if `beforeMarshalling` fails? That span will be leaked out and never closed

amarziali (changes requested):
I left some comments that should be addressed on this PR. I think that the implementation can be also simplified a bit

dd-java-agent/instrumentation/aws-java-sdk-1.11.0/src/main/java/datadog/trace/instrumentation/aws/v0/TracingRequestHandler.java
vandonr: Yes I would, but the request context is only available on an `Request` object (like the one we have in `beforeRequest`), not on the `AmazonWebServiceRequest` object that we have in `beforeMarshalling`.

amarziali: yes true on the 1.11.0 it's not there. However in the 1.11.116 it has been added and the IRequestHandler supports also `beforeExecute`. Probably supporting dsm only starting from 1.11.116 (released on April 2017) it's a good tradeoff to simplify

vandonr: ok, fair point

dd-java-agent/instrumentation/aws-java-sdk-1.11.0/src/main/java/datadog/trace/instrumentation/aws/v0/TracingRequestHandler.java
```diff
@@ -43,20 +43,22 @@ public TracingRequestHandler(ContextStore<Object, String> responseQueueStore) {
...
+    AgentSpan span = null;
     if (!AWS_LEGACY_TRACING && isPollingRequest(request.getOriginalRequest())) {
       // SQS messages spans are created by aws-java-sqs-1.0 - replace client scope with no-op,
       // so we can tell when receive call is complete without affecting the rest of the trace
       span = noopSpan();
     } else {
-      span = startSpan(AwsNameCache.spanName(request));
+      if (isSqsSendRequest(request.getOriginalRequest())) {
+        // for SQS, spans are created earlier, in beforeMarshalling()
+        span = request.getHandlerContext(SPAN_CONTEXT_KEY);
```
vandonr, line 56: idk if it's better to check on the request type first, or if I should just try getting the span every time, since we have the if null below to handle the case where it's missing.
At least with the explicit type check, we know what's the expected behavior.

dd-java-agent/instrumentation/aws-java-sqs-1.0/src/main/java/datadog/trace/instrumentation/aws/v1/sqs/TracingIterator.java
vandonr, line 1: we only support DSM in v1.11.106 for the producer, but we don't need the new features for the consumer, so it can be in v1.0 (even though it's pretty unlikely that customers would use mismatching versions between producer and consumer ?)

dd-java-agent/instrumentation/aws-java-sqs-1.11.106/src/main/java/datadog/trace/instrumentation/aws/v1/sqs/SqsInterceptor.java
```diff
@@ -0,0 +1,79 @@
...
+
+  @Override
+  public AmazonWebServiceRequest beforeMarshalling(AmazonWebServiceRequest request) {
+    if (request instanceof SendMessageRequest) {
+      SendMessageRequest smRequest = (SendMessageRequest) request;
+
+      String queueUrl = smRequest.getQueueUrl();
+      if (queueUrl == null) return request;
+
+      AgentSpan span = startSpan("aws.sqs.send");
```
vandonr, line 40: hardcoded the span name because I don't want to pull dependencies just to access the cache that is normally used to get the name https://github.com/DataDog/dd-trace-java/blob/62d6d15c3fdb685d049a47679171dc84ca01551e/dd-java-agent/instrumentation/aws-java-sdk-1.11.0/src/main/java/datadog/trace/instrumentation/aws/v0/TracingRequestHandler.java#L60
and since we know what kind of request we're dealing with here, it'd always be the same value anyway.

amarziali: why a span is started here? For the sdk v2 is not happening andt the reason is that the tracing for the send should already done by the aws sdk instrumentation.

vandonr: in the sdk v2 instrumentation, the spans are always started early, in `beforeExecution`, which is a method that doesn't exist in v1, but takes place even before `beforeMarshalling`. This allows access to the span in all methods basically.
https://github.com/DataDog/dd-trace-java/blob/275ed3a6479cddcdb5b7ab42f8c0b43f62d8da72/dd-java-agent/instrumentation/aws-java-sdk-2.2/src/main/java/datadog/trace/instrumentation/aws/v2/TracingExecutionInterceptor.java#L49
(it is activated only later)

In the sdk v1 instrumentation, the span is both created and started in the same method, in `beforeRequest`, which means it's not available in `beforeMarshalling`, where I need it. I could create the span in the `beforeMarshalling` method of the sdk instrumentation, but we already discussed [here](https://github.com/DataDog/dd-trace-java/pull/6259#discussion_r1403239316) that it'd be better to create it right where I need it, in the SQS instrumentation, which saves one use of the context, and also avoids potential leaking span issues.

amarziali:
If it can help an easy implementation can be found in #6281 . It has to be completed for extracting the context on receive side but it will be easy to do

vandonr:
> If it can help an easy implementation can be found in #6281 . It has to be completed for extracting the context on receive side but it will be easy to do

I don't understand, it's more or less what I did initially, but then we discussed the fact that it'd be better to use the handler context that is available from 1.11.106 to store the span in [this discussion](https://github.com/DataDog/dd-trace-java/pull/6259#discussion_r1403357343).

I do note the improvements you added: 
 - changing the span operation name to avoid having to add dependencies where we start it (I can use that here)
 - removing it from the context store when you use it to free memory faster

Are you suggesting that after trying it yourself, you think using a context store is a better idea than relying on the post-1.11.106 handler context ?

dd-java-agent/instrumentation/aws-java-sdk-1.11.0/src/main/java/datadog/trace/instrumentation/aws/v0/HandlerChainFactoryInstrumentation.java
```diff
@@ -42,8 +43,10 @@ public String[] helperClassNames() {
 
   @Override
   public Map<String, String> contextStore() {
-    return singletonMap(
-        "com.amazonaws.services.sqs.model.ReceiveMessageResult", "java.lang.String");
+    Map<String, String> map = new java.util.HashMap<>();
+    map.put("com.amazonaws.services.sqs.model.ReceiveMessageResult", "java.lang.String");
+    map.put(AmazonWebServiceRequest.class.getName(), AgentSpan.class.getName());
```
amarziali, line 48: we use not to refer classes that are not on the bootstrap classloader. This can be changed to the qualified name of the class `com.amazonaws.AmazonWebServiceRequest`

dd-java-agent/instrumentation/aws-java-sdk-1.11.0/src/main/java/datadog/trace/instrumentation/aws/v0/HandlerChainFactoryInstrumentation.java
```diff
@@ -64,7 +67,8 @@ public static void addHandler(@Advice.Return final List<RequestHandler2> handler
       handlers.add(
           new TracingRequestHandler(
               InstrumentationContext.get(
-                  "com.amazonaws.services.sqs.model.ReceiveMessageResult", "java.lang.String")));
+                  "com.amazonaws.services.sqs.model.ReceiveMessageResult", "java.lang.String"),
+              InstrumentationContext.get(AmazonWebServiceRequest.class, AgentSpan.class)));
```
amarziali, line 71: Same as above we can use String class name here

dd-java-agent/instrumentation/aws-java-sqs-1.0/src/main/java/datadog/trace/instrumentation/aws/v1/sqs/SqsClientInstrumentation.java
```diff
@@ -0,0 +1,65 @@
...
+  public String[] helperClassNames() {
+    return new String[] {
+      packageName + ".SqsInterceptor", packageName + ".MessageAttributeInjector"
+    };
+  }
+
+  @Override
+  public Map<String, String> contextStore() {
+    return Collections.singletonMap(
+        AmazonWebServiceRequest.class.getName(), AgentSpan.class.getName());
```
amarziali, line 49: Also here the ` com.amazonaws.AmazonWebServiceRequest` key should be used. The agentSpan.class.getName can be kept as is (it will be on the boostrap classpath)

dd-java-agent/instrumentation/aws-java-sqs-1.0/src/main/java/datadog/trace/instrumentation/aws/v1/sqs/SqsClientInstrumentation.java
```diff
@@ -0,0 +1,65 @@
...
+
+  public static class HandlerChainAdvice {
+    @Advice.OnMethodExit(onThrowable = Throwable.class, suppress = Throwable.class)
+    public static void addHandler(@Advice.Return final List<RequestHandler2> handlers) {
+      for (RequestHandler2 interceptor : handlers) {
+        if (interceptor instanceof SqsInterceptor) {
+          return; // list already has our interceptor, return to builder
+        }
+      }
+      handlers.add(
```
amarziali, line 60: nit: if datastreams is not enabled you can avoid adding this extra handler since it won't be needed

vandonr: yes right. I'll look into how I can get the DSM on/off information at this point.

vandonr:
> You might also ask a double review form the apm java core team (or dsm folks since it's dsm related)

@PerfectSlayer or @mcculls could you give this a look ?

I don't think a DSM review is necessary, because I really copied what was done in the V2 implem, so there was no functional question, only technical ones.

dd-java-agent/instrumentation/aws-java-sqs-1.0/src/main/java/datadog/trace/instrumentation/aws/v1/sqs/MessageAttributeInjector.java
```diff
@@ -0,0 +1,24 @@
...
+public class MessageAttributeInjector
+    implements AgentPropagation.Setter<Map<String, MessageAttributeValue>> {
+
+  public static final MessageAttributeInjector SETTER = new MessageAttributeInjector();
+
+  @Override
+  public void set(
+      final Map<String, MessageAttributeValue> carrier, final String key, final String value) {
+    if (carrier.size() < 10 && !carrier.containsKey(DATADOG_KEY)) {
+      String jsonPathway = String.format("{\"%s\": \"%s\"}", key, value);
```
PerfectSlayer, line 18: Don’t you fear some `key` or `value` value might break the JSON? Like `some"thing`?

dd-java-agent/instrumentation/aws-java-sqs-1.0/src/main/java/datadog/trace/instrumentation/aws/v1/sqs/SqsClientInstrumentation.java
```diff
@@ -0,0 +1,69 @@
...
+import net.bytebuddy.asm.Advice;
+
+/** AWS SDK v1 instrumentation */
+@AutoService(Instrumenter.class)
+public final class SqsClientInstrumentation extends Instrumenter.Tracing
+    implements Instrumenter.ForSingleType {
+  private static final String INSTRUMENTATION_NAME = "aws-sdk";
+
+  public SqsClientInstrumentation() {
+    super(INSTRUMENTATION_NAME);
```
PerfectSlayer, line 23: Should this instrumentation have some aliases more? Like "sqs" or similar? (I am not familiar with the whole AWS ecosystem)

dd-java-agent/instrumentation/aws-java-sqs-1.0/src/main/java/datadog/trace/instrumentation/aws/v1/sqs/MessageAttributeInjector.java
vandonr: that's a good point, but also this line of code was taken from https://github.com/DataDog/dd-trace-java/blob/5afa55c4c750cc298bd6d7d15db14ed79adef4e6/dd-java-agent/instrumentation/aws-java-sqs-2.0/src/main/java/datadog/trace/instrumentation/aws/v2/sqs/MessageAttributeInjector.java#L18
so... idk, maybe I can fix both at the same time.

dd-java-agent/instrumentation/aws-java-sqs-1.0/src/main/java/datadog/trace/instrumentation/aws/v1/sqs/SqsClientInstrumentation.java
vandonr: I was trying to stay consistent with the existing v2 instrumentation that uses the `aws-sdk` name too
https://github.com/DataDog/dd-trace-java/blob/5afa55c4c750cc298bd6d7d15db14ed79adef4e6/dd-java-agent/instrumentation/aws-java-sqs-2.0/src/main/java/datadog/trace/instrumentation/aws/v2/sqs/SqsClientInstrumentation.java#L17-L20

dd-java-agent/instrumentation/aws-java-sqs-1.0/src/main/java/datadog/trace/instrumentation/aws/v1/sqs/MessageAttributeInjector.java
vandonr: Looking at the dotnet implementation, it's implemented that way as well.
https://github.com/DataDog/dd-trace-dotnet/blob/e8735bc26f63534356e7f40b650c5e91a43b111a/tracer/src/Datadog.Trace/ClrProfiler/AutoInstrumentation/AWS/SQS/ContextPropagation.cs#L84
I'd say we keep it like this, and it may be worth it to fix it if we get an escalation for it.
Also, AWS is usually pretty restrictive on what it accepts for names, so we have that on our side.

mcculls (approved):

PR #6263, opened 2023-11-22
Handle WellKnown classes as primitives in EL

jpbempel:
# What Does This Do
WellKnown classes (like `UUID`) need to be treated as string primitives to be use in Expression Language and conditions
Add a list of string primitives

# Motivation
coherency with serialization behavior

# Additional Notes

Jira ticket: [DEBUG-1689]

[DEBUG-1689]: https://datadoghq.atlassian.net/browse/DEBUG-1689?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ

### DIFF
--- a/dd-java-agent/agent-debugger/debugger-el/src/main/java/com/datadog/debugger/el/expressions/ValueRefExpression.java
+++ b/dd-java-agent/agent-debugger/debugger-el/src/main/java/com/datadog/debugger/el/expressions/ValueRefExpression.java
@@ -7,6 +7,7 @@ import com.datadog.debugger.el.Value;
 import com.datadog.debugger.el.Visitor;
 import datadog.trace.bootstrap.debugger.el.ValueReferenceResolver;
 import datadog.trace.bootstrap.debugger.util.Redaction;
+import datadog.trace.bootstrap.debugger.util.WellKnownClasses;
 import java.util.Objects;
 
 /** An expression taking a reference path and resolving to {@linkplain Value} */
@@ -25,9 +26,14 @@ public final class ValueRefExpression implements ValueExpression<Value<?>> {
     } catch (RuntimeException ex) {
       throw new EvaluationException(ex.getMessage(), PrettyPrintVisitor.print(this));
     }
-    if (symbol == Redaction.REDACTED_VALUE
-        || (symbol != null && Redaction.isRedactedType(symbol.getClass().getTypeName()))) {
-      ExpressionHelper.throwRedactedException(this);
+    if (symbol != null) {
+      String typeName = symbol.getClass().getTypeName();
+      if (WellKnownClasses.isStringPrimitive(typeName)) {
+        symbol = symbol.toString();
+      }
+      if (symbol == Redaction.REDACTED_VALUE || (Redaction.isRedactedType(typeName))) {
+        ExpressionHelper.throwRedactedException(this);
+      }
     }
     return Value.of(symbol);
   }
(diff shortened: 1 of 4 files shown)


dd-java-agent/agent-debugger/debugger-el/src/main/java/com/datadog/debugger/el/expressions/ValueRefExpression.java
```diff
@@ -25,9 +26,14 @@ public Value<?> evaluate(ValueReferenceResolver valueRefResolver) {
     } catch (RuntimeException ex) {
       throw new EvaluationException(ex.getMessage(), PrettyPrintVisitor.print(this));
     }
-    if (symbol == Redaction.REDACTED_VALUE
-        || (symbol != null && Redaction.isRedactedType(symbol.getClass().getTypeName()))) {
-      ExpressionHelper.throwRedactedException(this);
+    if (symbol != null) {
+      String typeName = symbol.getClass().getTypeName();
+      if (WellKnownClasses.isStringPrimitive(typeName)) {
```
shatzi, line 31: Should we do this only after we check this is not redacted?

dd-java-agent/agent-debugger/debugger-el/src/test/resources/test_conditional_10.json
```diff
@@ -0,0 +1,21 @@
...
+          },
+          "a3cbe9e7-edd3-4bef-8e5b-59bfcb04cf91"]
+      }, {
+        "eq": [
+          {"ref": "duration"},
+          "PT42S"]
+      }, {
+      "eq": [
+        {"ref": "clazz"},
+        "class java.lang.String"]
```
shatzi, line 17: I find this weird. I expected class to be just fully qualified name of the class. Same thing as we have instance of operator.

Maybe we should also provide a function to string those classes and instead calling toString we should just use the class name for this one

dd-java-agent/agent-debugger/debugger-el/src/test/resources/test_conditional_10.json
```diff
@@ -0,0 +1,21 @@
...
+    "and": [{
+        "eq": [
+          {
+            "ref": "uuid"
+          },
+          "a3cbe9e7-edd3-4bef-8e5b-59bfcb04cf91"]
+      }, {
+        "eq": [
+          {"ref": "duration"},
+          "PT42S"]
```
shatzi, line 13: This is cool

shatzi (approved):

dd-java-agent/agent-debugger/debugger-el/src/main/java/com/datadog/debugger/el/expressions/ValueRefExpression.java
jpbempel: why?

dd-java-agent/agent-debugger/debugger-el/src/test/resources/test_conditional_10.json
jpbempel: yep, I found it also weird.
I will add the special toString later

dd-java-agent/agent-debugger/debugger-el/src/main/java/com/datadog/debugger/el/expressions/ValueRefExpression.java
shatzi: avoid calling to string if the type/value is redact...

dd-java-agent/agent-debugger/debugger-el/src/test/resources/test_conditional_10.json
jpbempel: not a big deal, those toString are very cheap (and that's why we authorize them) and also very low chance that they would be redacted

dd-java-agent/agent-debugger/debugger-el/src/main/java/com/datadog/debugger/el/expressions/ValueRefExpression.java
jpbempel: not a big deal, those toString are very cheap (and that's why we authorize them) and also very low chance that they would be redacted

PR #6264, opened 2023-11-22
Add support of `@exception` for uncaught exception

jpbempel:
# What Does This Do
Uncaught exception captured by method probe (log, snapshot or span decoration) can be accessed through @exception synthetic variable with expression language in conditions, log template or span tags

# Motivation

# Additional Notes

Jira ticket: [DEBUG-1940]

[DEBUG-1940]: https://datadoghq.atlassian.net/browse/DEBUG-1940?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ

### DIFF
--- a/dd-java-agent/agent-debugger/debugger-bootstrap/src/main/java/datadog/trace/bootstrap/debugger/CapturedContext.java
+++ b/dd-java-agent/agent-debugger/debugger-bootstrap/src/main/java/datadog/trace/bootstrap/debugger/CapturedContext.java
@@ -211,7 +211,8 @@ public class CapturedContext implements ValueReferenceResolver {
   }
 
   public void addThrowable(Throwable t) {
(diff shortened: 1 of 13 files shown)


shatzi (commented):
Let's add a test that check if @exception == null. I wonder if the expression works if now exception is as thrown. Because we don't set the extension will the ref resolve return null or throw an error

jpbempel:
> Let's add a test that check if @exception == null. I wonder if the expression works if now exception is as thrown. Because we don't set the extension will the ref resolve return null or throw an error

I have added support for the missing `isUndefined` function that solve the issue. so you can do `isUndefined(@Exception)`

shatzi (approved):

PR #6265, opened 2023-11-22
Add GraalVM native image build instructions

ygree:
# What Does This Do

Adds GraalVM native image build instructions

### DIFF
--- a/CONTRIBUTING.md
+++ b/CONTRIBUTING.md
@@ -220,11 +220,11 @@ Suggested plugins and settings:
 
 To run tests on a different JVM than the one used for doing the build, you need two things:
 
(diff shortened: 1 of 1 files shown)


PerfectSlayer (commented):
Isn't a duplicate of the `Running tests on another JVM` paragraph?
Should we just explicit the required GraalVM JDK from the `prerequisites` instead?

ygree:
> Isn't a duplicate of the `Running tests on another JVM` paragraph? Should we just explicit the required GraalVM JDK from the `prerequisites` instead?

Okay. I've added an example to the Running tests on another JVM paragraph to make it searchable.
Doesn't seem to me as a prerequisites.

PR #6270, opened 2023-11-23
Fix inner class signature matching

jpbempel:
# What Does This Do
the dollar character in inner class is interpreted as regex and replacement special character. Need to be escaped.

# Motivation
Supporting inner classes in probe method signature

# Additional Notes

Jira ticket: [DEBUG-1620]

[DEBUG-1620]: https://datadoghq.atlassian.net/browse/DEBUG-1620?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ

### DIFF
--- a/dd-java-agent/agent-debugger/src/main/java/com/datadog/debugger/probe/Where.java
+++ b/dd-java-agent/agent-debugger/src/main/java/com/datadog/debugger/probe/Where.java
@@ -164,10 +164,17 @@ public class Where {
     }
     String result = targetSignature;
     for (String className : classes) {
-      Pattern classNamePattern = Pattern.compile("L" + className + ";");
-      result = classNamePattern.matcher(result).replaceAll("L" + simplify(className) + ";");
+      Pattern classNamePattern = Pattern.compile("L" + escapeClassName(className) + ";");
+      result =
+          classNamePattern
+              .matcher(result)
+              .replaceAll("L" + escapeClassName(simplify(className)) + ";");
     }
     return result;
+  }
+
+  private static String escapeClassName(String className) {
+    return className.replace("$", "\\$");
   }
 
   private static String simplify(String fqnClass) {
--- a/dd-java-agent/agent-debugger/src/test/java/com/datadog/debugger/probe/WhereTest.java
+++ b/dd-java-agent/agent-debugger/src/test/java/com/datadog/debugger/probe/WhereTest.java
@@ -1,5 +1,7 @@
 package com.datadog.debugger.probe;
 
+import static org.junit.jupiter.api.Assertions.assertTrue;
+
 import com.datadog.debugger.util.MoshiHelper;
 import com.squareup.moshi.JsonAdapter;
 import java.io.IOException;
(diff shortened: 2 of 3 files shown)


dd-java-agent/agent-debugger/src/main/java/com/datadog/debugger/probe/Where.java
```diff
@@ -164,12 +164,19 @@ private static String removeFQN(String targetSignature) {
     }
     String result = targetSignature;
     for (String className : classes) {
-      Pattern classNamePattern = Pattern.compile("L" + className + ";");
-      result = classNamePattern.matcher(result).replaceAll("L" + simplify(className) + ";");
+      Pattern classNamePattern = Pattern.compile("L" + escapeClassName(className) + ";");
```
shatzi, line 167: why are we using regex here? it look for me like a simple string search and replace?

dd-java-agent/agent-debugger/src/test/java/com/datadog/debugger/probe/WhereTest.java
```diff
@@ -78,18 +80,30 @@ public void linesRoundTrip() throws IOException {
   @Test
   public void methodMatching() {
     Where where = new Where("String", "substring", "(int,int)", new String[0], null);
```
shatzi, line 82: the signature doesn't have return value? we only match based on args?

dd-java-agent/agent-debugger/src/test/java/com/datadog/debugger/probe/WhereTest.java
```diff
@@ -78,18 +80,30 @@ public void linesRoundTrip() throws IOException {
...
-    Assertions.assertTrue(where.isMethodMatching("concat", "String (String)"));
+    assertTrue(where.isMethodMatching("concat", "String (String)"));
     where = new Where("String", "concat", " \t", new String[0], null);
-    Assertions.assertTrue(where.isMethodMatching("concat", "String (String)"));
+    assertTrue(where.isMethodMatching("concat", "String (String)"));
+    where =
+        new Where(
+            "Inner",
+            "innerMethod",
+            "(com.datadog.debugger.probe.Outer$Inner)",
```
shatzi, line 100: how does symdb export those inner classes? we export the class symbol as com.datadog.debugger.probe.Outer$Inner?

shatzi (approved):

dd-java-agent/agent-debugger/src/main/java/com/datadog/debugger/probe/Where.java
jpbempel: the method `String.replaceAll` does exactly (and use regex) the same thing but for reason this call is forbidden by the a CI tool configured on this repo

dd-java-agent/agent-debugger/src/test/java/com/datadog/debugger/probe/WhereTest.java
jpbempel: yes, because it is gave by ASM Type class for us

dd-java-agent/agent-debugger/src/test/java/com/datadog/debugger/probe/WhereTest.java
jpbempel: return value is optional because it does not help to distinguish 2 overloaded ones for example.
In Java you couldn't differentiate 2 overloaded method by their return type. Only by args

```
  int process() {}
  void process() {}
  
  process(); // which one is called?
```

PR #6286, opened 2023-11-28
Add basic telemetry support for native image

smola:
# What Does This Do
* Adds basic telemetry support for native image.
* Only supports app started/closed, heartbeats, extended heartbeats, and config.

# Motivation

# Additional Notes

Other PRs will follow for:
* Dependencies: https://github.com/DataDog/dd-trace-java/pull/6302
* Integrations
* Metrics
* Logs

### DIFF
--- a/.circleci/config.continue.yml.j2
+++ b/.circleci/config.continue.yml.j2
@@ -36,7 +36,7 @@ instrumentation_modules: &instrumentation_modules "dd-java-agent/instrumentation
 debugger_modules: &debugger_modules "dd-java-agent/agent-debugger|dd-java-agent/agent-bootstrap|dd-java-agent/agent-builder|internal-api|communication|dd-trace-core"
 profiling_modules: &profiling_modules "dd-java-agent/agent-profiling"
 
-default_system_tests_commit: &default_system_tests_commit 706b4eb3ff053068e961a684d21dfe2ff4487d37
+default_system_tests_commit: &default_system_tests_commit f5409dee787e1758c77d067188b20a2705a679a2
 
 parameters:
   nightly:
--- a/dd-java-agent/agent-bootstrap/src/main/java/datadog/trace/bootstrap/Agent.java
+++ b/dd-java-agent/agent-bootstrap/src/main/java/datadog/trace/bootstrap/Agent.java
@@ -140,7 +140,6 @@ public class Agent {
   private static boolean cwsEnabled = false;
   private static boolean ciVisibilityEnabled = false;
   private static boolean usmEnabled = false;
-  private static boolean telemetryEnabled = true;
   private static boolean debuggerEnabled = false;
 
   public static void start(final Instrumentation inst, final URL agentJarURL, String agentArgs) {
@@ -153,7 +152,6 @@ public class Agent {
       // these default services are not used during native-image builds
       jmxFetchEnabled = false;
       remoteConfigEnabled = false;
-      telemetryEnabled = false;
       // apply trace instrumentation, but skip starting other services
       startDatadogAgent(inst);
       StaticEventLogger.end("Agent.start");
@@ -211,7 +209,6 @@ public class Agent {
     appSecFullyDisabled = isAppSecFullyDisabled();
     remoteConfigEnabled = isFeatureEnabled(AgentFeature.REMOTE_CONFIG);
     cwsEnabled = isFeatureEnabled(AgentFeature.CWS);
-    telemetryEnabled = isFeatureEnabled(AgentFeature.TELEMETRY);
     debuggerEnabled = isFeatureEnabled(AgentFeature.DEBUGGER);
 
     if (profilingEnabled) {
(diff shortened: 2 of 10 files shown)


internal-api/src/main/java/datadog/trace/api/Platform.java
```diff
@@ -7,6 +7,11 @@
 import java.util.Locale;
 
 public final class Platform {
+  // A helper class to capture whether the executable is a native image or not.
```
smola, line 10: From https://github.com/DataDog/dd-trace-java/pull/6145
Temporarily in this PR until this utility hits master.

.circleci/config.continue.yml.j2
```diff
@@ -36,7 +36,7 @@ instrumentation_modules: &instrumentation_modules "dd-java-agent/instrumentation
 debugger_modules: &debugger_modules "dd-java-agent/agent-debugger|dd-java-agent/agent-bootstrap|dd-java-agent/agent-builder|internal-api|communication|dd-trace-core"
 profiling_modules: &profiling_modules "dd-java-agent/agent-profiling"
 
-default_system_tests_commit: &default_system_tests_commit ca7566bf7e66c0afdf29b0d3c07b2972e5869997
+default_system_tests_commit: &default_system_tests_commit f5409dee787e1758c77d067188b20a2705a679a2
```
smola, line 39: To be removed before merge.

dd-java-agent/agent-tooling/src/main/java/datadog/trace/agent/tooling/nativeimage/TracerActivation.java
```diff
@@ -12,13 +14,27 @@ public final class TracerActivation {
   private static final Logger log = LoggerFactory.getLogger(TracerActivation.class);
 
   public static void activate() {
+    SharedCommunicationObjects sco = null;
```
dougqh, line 17: Why the initialization to null?
Could it be left unassigned?  Or does the unassigned / reachability analysis have trouble with this code?

nayeem-kamal (approved):
LGTM. Reminder to change the system tests commit before merging.

internal-api/src/main/java/datadog/trace/api/Config.java
```diff
@@ -1492,16 +1507,19 @@ private Config(final ConfigProvider configProvider, final InstrumenterConfig ins
     telemetryMetricsInterval = telemetryInterval;
 
     telemetryMetricsEnabled =
```
dougqh, line 1509: It feels like there should be a better abstraction here.
I'd really like the code for these subsystems to be more declarative in nature.
Otherwise, we'll deal with this for every new subsystem we add.

But I suppose for now, this is okay.

smola: In these particular cases, it should be fine, since the native image checks will be removed once metrics, logs, and dependencies are supported in native image (subsequent PRs).

dd-java-agent/agent-bootstrap/src/main/java/datadog/trace/bootstrap/Agent.java
```diff
@@ -341,7 +338,7 @@ public static void shutdown(final boolean sync) {
     if (profilingEnabled) {
       shutdownProfilingAgent(sync);
     }
-    if (telemetryEnabled) {
+    if (Config.get().isTelemetryEnabled()) {
```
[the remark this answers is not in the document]
smola: `isFeatureEnabled` replicates the config provider logic, and in this case, it would lead to duplicate more logic about the checks for telemetry enabled at runtime vs build time. It seems that avoiding `Config` calls in `Agent` is only required for subsytems starting earlier, and not by any of the last falls to `maybeStart*`.

So I wanted to rely on `Config.get().isTelemetryEnabled()` to ensure we are using the exact same logic as everywhere else. In fact, we could probably do this for anything that starts late?

dd-java-agent/agent-tooling/src/main/java/datadog/trace/agent/tooling/nativeimage/TracerActivation.java
```diff
@@ -12,13 +14,27 @@ public final class TracerActivation {
...
     try {
       // the JFR based profiling does not allow context propagation - use the noop integration
       ProfilingContextIntegration contextIntegration = ProfilingContextIntegration.NoOp.INSTANCE;
-      TracerInstaller.installGlobalTracer(new SharedCommunicationObjects(), contextIntegration);
+      TracerInstaller.installGlobalTracer(sco, contextIntegration);
       ProfilerInstaller.installProfiler();
     } catch (Throwable e) {
       log.warn("Problem activating datadog tracer", e);
     }
+    if (Config.get().isTelemetryEnabled()) {
```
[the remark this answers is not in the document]
smola: No visibility from this module.

PR #6287, opened 2023-11-28
Adds `env`, `service`, and `version` tags to `Log4j2` instrumentation

zarirhamza:
# What Does This Do
Adds `dd.env`, `dd.service`, and `dd.version` tags to `Log4j2` instrumentation instead of just `dd.trace_id`/`dd.span_id`

# Motivation
All other logging instrumentations (`Log4j`, `Log4j2.7`, `Logback`, `Tinylog`, `jboss-logmodules`) have 5 trace values injected but `Log4j2` only has two. Fixed for parity.


zarirhamza:
Closed due to limitations of current implementation only applying to one thread instead of all. Recommended to upgrade version to use more complete instrumentation

PR #6291, opened 2023-11-29
Armeria Jetty: ensure spans are finished

### LINKED ISSUE #6248: Unfinished spans in armeria
Datadog agent [finishes jetty traces](https://github.com/DataDog/dd-trace-java/blob/v1.24.2/dd-java-agent/instrumentation/jetty-9/src/main/java/datadog/trace/instrumentation/jetty9/JettyServerInstrumentation.java) on `HttpChannel.recycle()`, however Armeria [doesn't invoke that method](https://github.com/line/armeria/blob/armeria-1.26.2/jetty/jetty11/src/main/java/com/linecorp/armeria/server/jetty/JettyService.java#L277), leaving the tracing spans in pending state.

In my (limited) understanding recycling a channel is not functionally necessary, what can be done to make the dd agent work with armeria? 

As a workaround for now, I am finishing the traces on an application level in an outermost response filter, which works, but not ideal.

amarziali:
# What Does This Do

Armeria is using the underlying jetty engine without calling `recycle` that's what we instrument as a span termination pointcut.  Not finishing that root span will prevent the full trace from being sent.

This PR fixes #6248 for jetty 11 and jetty 9

# Motivation

# Additional Notes

I had to shortcut some query encoding tests that are not passing probably because armeria adapt the request to jetty in a different way. However this should be perfectly fine (and anyway out of scope on this PR) since we only want to test that the spans opened in most common use cases are finished.

Jira ticket: [PROJ-IDENT]


nayeem-kamal (approved):
LGTM

PR #6292, opened 2023-11-29
Telemetry API cleanup

smola:
# What Does This Do
* Move `LogCollector` and `IntegrationsCollector` to `datadog.trace.api.telemetry` package, now consistent with metrics collectors.
* Break up test suites per class under test (no actual test case changes).
* Minor clean up: marking final fields, reducing visibility, etc.

# Motivation

# Additional Notes

* `ConfigCollector` is left as is at the moment, since moving it caused some issue with access from `log-injection` smoke test.


bantonsson (approved):

PerfectSlayer (approved):
Thanks for cleaning up! 🧹

PR #6299, opened 2023-11-30
Header injection vulnerability detection

DDJavierSantos:
# What Does This Do
Implements detection of header injection vulnerabilities

# Motivation
To increase the scope of the security vulnerability detection functionality

# Additional Notes
Please note that there is instrumentation for webflux so we can separate the cases where the headers come from the request from the cases where the headers come from the response. Previously all headers were tainted even if they were response headers, this would cause false header injection positives.

Jira ticket: [PROJ-IDENT]

### DIFF
--- a/dd-java-agent/agent-iast/src/main/java/com/datadog/iast/sink/HeaderInjectionModuleImpl.java
+++ b/dd-java-agent/agent-iast/src/main/java/com/datadog/iast/sink/HeaderInjectionModuleImpl.java
@@ -1,0 +1,102 @@
+package com.datadog.iast.sink;
+
+import com.datadog.iast.Dependencies;
+import com.datadog.iast.IastRequestContext;
+import com.datadog.iast.model.Evidence;
+import com.datadog.iast.model.Range;
+import com.datadog.iast.model.VulnerabilityType;
+import com.datadog.iast.overhead.Operations;
+import com.datadog.iast.taint.Ranges;
+import com.datadog.iast.taint.TaintedObject;
+import datadog.trace.api.iast.SourceTypes;
+import datadog.trace.api.iast.sink.HeaderInjectionModule;
+import datadog.trace.bootstrap.instrumentation.api.AgentSpan;
+import datadog.trace.bootstrap.instrumentation.api.AgentTracer;
+import java.util.Arrays;
+import java.util.HashSet;
+import java.util.Set;
+import javax.annotation.Nonnull;
+import javax.annotation.Nullable;
+
+public class HeaderInjectionModuleImpl extends SinkModuleBase implements HeaderInjectionModule {
+
+  private static final Set<String> headerInjectionExclusions =
+      new HashSet<String>(
+          Arrays.asList(
+              "Sec-WebSocket-Location".toUpperCase(),
+              "Sec-WebSocket-Accept".toUpperCase(),
+              "Upgrade".toUpperCase(),
+              "Connection".toUpperCase(),
+              "location".toUpperCase()));
+
+  public HeaderInjectionModuleImpl(final Dependencies dependencies) {
+    super(dependencies);
+  }
+
+  @Override
+  public void onHeader(@Nonnull final String name, @Nullable final String value) {
+    if (null == value) {
+      return;
+    }
+    boolean headerInjectionExclusion = headerInjectionExclusions.contains(name.toUpperCase());
+
+    if (!headerInjectionExclusion) {
+      final AgentSpan span = AgentTracer.activeSpan();
+      final IastRequestContext ctx = IastRequestContext.get(span);
+      if (ctx == null) {
+        return;
+      }
+
+      final TaintedObject taintedObject = ctx.getTaintedObjects().get(value);
+      if (null == taintedObject) {
+        return;
+      }
+
+      final Range[] ranges =
+          Ranges.getNotMarkedRanges(
+              taintedObject.getRanges(), VulnerabilityType.HEADER_INJECTION.mark());
+      if (ranges == null || ranges.length == 0) {
+        return;
+      }
+
+      if ("Access-Control-Allow-Origin".equalsIgnoreCase(name)) {
+        boolean allRangesFromOrigin = true;
+        for (Range range : ranges) {
+          if (null != range.getSource().getName()
+              && range.getSource().getOrigin() == SourceTypes.REQUEST_HEADER_VALUE
+              && !range.getSource().getName().equalsIgnoreCase("origin")) {
+            allRangesFromOrigin = false;
+          }
+        }
+        if (allRangesFromOrigin) {
+          return;
+        }
+      }
+
+      if ("Set-Cookie".equalsIgnoreCase(name)) {
+        boolean allRangesFromCookie = true;
+        for (Range range : ranges) {
+          if (null != range.getSource().getName()
+              && range.getSource().getOrigin() == SourceTypes.REQUEST_HEADER_VALUE
+              && !range.getSource().getName().equalsIgnoreCase("Set-Cookie")) {
+            allRangesFromCookie = false;
+          }
+        }
+        if (allRangesFromCookie) {
+          return;
+        }
+      }
+
+      if (!overheadController.consumeQuota(Operations.REPORT_VULNERABILITY, span)) {
+        return;
+      }
+      final String evidenceString = name + ": " + value;
+      final Range[] shiftedRanges = new Range[ranges.length];
+      for (int i = 0; i < ranges.length; i++) {
+        shiftedRanges[i] = ranges[i].shift(name.length() + 2);
+      }
+      final Evidence result = new Evidence(evidenceString, shiftedRanges);
+      report(span, VulnerabilityType.HEADER_INJECTION, result);
+    }
+  }
+}
--- a/dd-java-agent/agent-iast/src/main/java/com/datadog/iast/sink/HeaderInjectionModuleImpl.java
+++ b/dd-java-agent/agent-iast/src/main/java/com/datadog/iast/sink/HeaderInjectionModuleImpl.java
@@ -40,9 +40,6 @@ public class HeaderInjectionModuleImpl extends SinkModuleBase implements HeaderI
       return;
     }
     HttpHeader header = HttpHeader.from(name);
-    if (null == header) {
-      return;
-    }
     boolean headerInjectionExclusion = headerInjectionExclusions.contains(header);
 
     if (!headerInjectionExclusion) {
--- a/dd-java-agent/agent-iast/src/main/java/com/datadog/iast/sink/HeaderInjectionModuleImpl.java
+++ b/dd-java-agent/agent-iast/src/main/java/com/datadog/iast/sink/HeaderInjectionModuleImpl.java
@@ -8,6 +8,7 @@ import com.datadog.iast.model.VulnerabilityType;
 import com.datadog.iast.overhead.Operations;
 import com.datadog.iast.taint.Ranges;
 import com.datadog.iast.taint.TaintedObject;
+import com.datadog.iast.util.HttpHeader;
 import datadog.trace.api.iast.SourceTypes;
 import datadog.trace.api.iast.sink.HeaderInjectionModule;
 import datadog.trace.bootstrap.instrumentation.api.AgentSpan;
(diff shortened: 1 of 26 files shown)


dd-java-agent/agent-iast/src/main/java/com/datadog/iast/sensitive/HeaderValueRegexpTokenizer.java
```diff
@@ -0,0 +1,26 @@
+package com.datadog.iast.sensitive;
+
+import com.datadog.iast.model.Evidence;
+import com.datadog.iast.util.Ranged;
+import java.util.regex.Pattern;
+
+public class HeaderValueRegexpTokenizer extends AbstractRegexTokenizer {
+
+  private static final Pattern HEADER_PATTERN =
+      Pattern.compile(
```
manuel-alvarez-alvarez, line 10: I think you have it already defined at `Config#getIastRedactionValuePattern()`

dd-java-agent/agent-iast/src/main/java/com/datadog/iast/sink/HttpResponseHeaderModuleImpl.java
```diff
@@ -28,13 +33,23 @@
 public class HttpResponseHeaderModuleImpl extends SinkModuleBase
     implements HttpResponseHeaderModule {
 
+  private static final List<String> headerInjectionExclusions =
```
manuel-alvarez-alvarez, line 36:
```suggestion
  private static final Set<String> headerInjectionExclusions =
```

dd-java-agent/agent-iast/src/main/java/com/datadog/iast/sink/HttpResponseHeaderModuleImpl.java
```diff
@@ -50,6 +65,64 @@ public void onHeader(@Nonnull final String name, final String value) {
         InstrumentationBridge.UNVALIDATED_REDIRECT.onHeader(name, value);
       }
     }
+
+    if (!headerInjectionExclusion && null != value) {
```
manuel-alvarez-alvarez, line 69: This should go on a separate `HeaderInjectionModule`, same as `UNVALIDATED_REDIRECT`

manuel-alvarez-alvarez:
I'm a bit lost with all the new instrumentations in the package `datadog.trace.instrumentation.springwebflux.server.iast`. 

This PR is focused on header injection that should only take into account sinks for response headers, but those new instrumentations take care of request headers sources. Besides, I thought we were fully covering request headers in spring webflux, if we are missing something perhaps it makes more sense to create a separate PR for it.

dd-java-agent/agent-iast/src/main/java/com/datadog/iast/sink/HttpResponseHeaderModuleImpl.java
```diff
@@ -28,13 +33,23 @@
 public class HttpResponseHeaderModuleImpl extends SinkModuleBase
     implements HttpResponseHeaderModule {
 
+  private static final Set<String> headerInjectionExclusions =
+      Arrays.asList(
```
smola, line 37: Wrong type here `List` vs `Set`.

dd-java-agent/agent-iast/src/main/java/com/datadog/iast/sink/HttpResponseHeaderModuleImpl.java
```diff
@@ -50,6 +65,64 @@ public void onHeader(@Nonnull final String name, final String value) {
...
+              && !range.getSource().getName().equalsIgnoreCase("Set-Cookie")) {
+            allRangesFromCookie = false;
+          }
+        }
+        if (allRangesFromCookie) {
+          return;
+        }
+      }
+
+      Range[] ranges =
```
anderruiz, line 109: I think this should be the first check and we should use this ranges instead of calling taintedObject.getRanges() as we are ignore the marks in the previous checks

dd-java-agent/agent-iast/src/main/java/com/datadog/iast/sensitive/HeaderValueRegexpTokenizer.java
DDJavierSantos: fixed

dd-java-agent/agent-iast/src/main/java/com/datadog/iast/sink/HttpResponseHeaderModuleImpl.java
DDJavierSantos: I have created the module and moved this code there

dd-java-agent/agent-iast/src/main/java/com/datadog/iast/sink/HttpResponseHeaderModuleImpl.java
DDJavierSantos: fixed

dd-java-agent/agent-iast/src/main/java/com/datadog/iast/sink/HttpResponseHeaderModuleImpl.java
DDJavierSantos: suggestion applied

dd-java-agent/agent-iast/src/test/groovy/com/datadog/iast/sink/HttpResponseHeaderModuleTest.groovy
```diff
@@ -126,4 +135,129 @@ class HttpResponseHeaderModuleTest extends IastModuleImplTestBase {
     then:
     0 * _
   }
+
+  void 'check header value injection'(final String headerName, final int mark, final String headerValue, final String expected) {
```
jandro996, line 139: It could be great to add test cases with the HEADER_INJECTION_MARK and not tainted value

```
'var'  | VulnerabilityMarks. NOT_MARKED    | 'headerName'             | null
'/==>var<=='  | VulnerabilityMarks.HEADER_INJECTION_MARK  | 'headerName'             | null
```

We also need to modify the then clause to something like this:

```
then:
   if(expected != null) {
    1 * reporter.report(_, _ as Vulnerability) >> { savedVul = it[1] }
    assertEvidence(savedVul, expected, VulnerabilityType.HEADER_INJECTION)
    } else {
0 * reporter.report(_, _)
}
```

dd-java-agent/instrumentation/spring-webflux-5/src/main/java/datadog/trace/instrumentation/springwebflux/server/iast/TaintHttpHeadersToSingleValueMapAdvice.java
```diff
@@ -26,8 +26,8 @@ public static void after(@Advice.Return Map<String, String> values) {
     for (Map.Entry<String, String> e : values.entrySet()) {
       final String name = e.getKey();
       final String value = e.getValue();
-      module.taint(ctx, name, SourceTypes.REQUEST_HEADER_NAME, name);
-      module.taint(ctx, value, SourceTypes.REQUEST_HEADER_VALUE, name);
+      module.taintIfTainted(ctx, name, self, SourceTypes.REQUEST_HEADER_VALUE, name);
```
jandro996, line 29: Is this change correct? I think it should be

` module.taintIfTainted(ctx, name, self, SourceTypes.REQUEST_HEADER_NAME, name);`

dd-java-agent/agent-iast/src/main/java/com/datadog/iast/sink/HeaderInjectionModuleImpl.java
```diff
@@ -0,0 +1,99 @@
...
+import java.util.Set;
+import javax.annotation.Nonnull;
+import javax.annotation.Nullable;
+
+public class HeaderInjectionModuleImpl extends SinkModuleBase implements HeaderInjectionModule {
+
+  private static final Set<String> headerInjectionExclusions =
+      new HashSet<String>(
+          Arrays.asList(
+              "Sec-WebSocket-Location".toUpperCase(),
```
jandro996, line 26: why not write it directly with uppercase?

dd-java-agent/instrumentation/spring-webflux-5/src/main/java/datadog/trace/instrumentation/springwebflux/server/iast/TaintHttpHeadersToSingleValueMapAdvice.java
DDJavierSantos: This is tainting both the name and the value. The name needs propagation too.

dd-java-agent/agent-iast/src/main/java/com/datadog/iast/sink/HeaderInjectionModuleImpl.java
DDJavierSantos: I believe this is more legible, and this static initialization only happens once.

dd-java-agent/instrumentation/spring-webflux-5/src/main/java/datadog/trace/instrumentation/springwebflux/server/iast/TaintHttpHeadersToSingleValueMapAdvice.java
manuel-alvarez-alvarez: But why did you change the source type? this method is tainting the header name so should be `SourceTypes.REQUEST_HEADER_NAME`

dd-java-agent/agent-iast/src/test/groovy/com/datadog/iast/sink/HttpResponseHeaderModuleTest.groovy
DDJavierSantos: Test added

dd-java-agent/instrumentation/spring-webflux-5/src/main/java/datadog/trace/instrumentation/springwebflux/server/iast/TaintHttpHeadersToSingleValueMapAdvice.java
DDJavierSantos: You are right, that is fixed now

dd-java-agent/agent-iast/src/main/java/com/datadog/iast/sink/HeaderInjectionModuleImpl.java
```diff
@@ -0,0 +1,99 @@
...
+              "Upgrade".toUpperCase(),
+              "Connection".toUpperCase(),
+              "location".toUpperCase()));
+
+  public HeaderInjectionModuleImpl(final Dependencies dependencies) {
+    super(dependencies);
+  }
+
+  @Override
+  public void onHeader(@Nonnull final String name, @Nullable final String value) {
```
jandro996, line 37: IMHO

```suggestion
  public void onHeader(@Nonnull final String name, @Nullable final String value) {
  if(value == null){
    return;
  }
  if(!headerInjectionExclusions.contains(name.toUpperCase())){
  ...
  }
  }
```

jandro996 (approved):

dd-java-agent/agent-iast/src/main/java/com/datadog/iast/sink/HeaderInjectionModuleImpl.java
DDJavierSantos: applied

dd-java-agent/agent-iast/src/main/java/com/datadog/iast/sink/HeaderInjectionModuleImpl.java
```diff
@@ -0,0 +1,102 @@
...
+import datadog.trace.api.iast.sink.HeaderInjectionModule;
+import datadog.trace.bootstrap.instrumentation.api.AgentSpan;
+import datadog.trace.bootstrap.instrumentation.api.AgentTracer;
+import java.util.Arrays;
+import java.util.HashSet;
+import java.util.Set;
+import javax.annotation.Nonnull;
+import javax.annotation.Nullable;
+
+public class HeaderInjectionModuleImpl extends SinkModuleBase implements HeaderInjectionModule {
```
manuel-alvarez-alvarez, line 21: We do have a class to deal with headers `com.datadog.iast.util.HttpHeader` it might make sense to add these new ones there and reuse its logic for header detection.

internal-api/src/main/java/datadog/trace/api/iast/VulnerabilityTypes.java
```diff
@@ -81,7 +83,8 @@ private VulnerabilityTypes() {}
     XCONTENTTYPE_HEADER_MISSING,
     NO_SAMESITE_COOKIE,
     XSS,
-    STACKTRACE_LEAK
+    STACKTRACE_LEAK,
```
manuel-alvarez-alvarez, line 86: You also need to add the vulnerability to the list of response headers:

```java
  public static final byte[] RESPONSE_HEADER_TYPES = {
    UNVALIDATED_REDIRECT,
    INSECURE_COOKIE,
    NO_SAMESITE_COOKIE,
    XCONTENTTYPE_HEADER_MISSING,
    HSTS_HEADER_MISSING
     <-- Here
  };
```

dd-java-agent/agent-iast/src/main/java/com/datadog/iast/sink/HeaderInjectionModuleImpl.java
DDJavierSantos: suggestion applied

internal-api/src/main/java/datadog/trace/api/iast/VulnerabilityTypes.java
DDJavierSantos: fixed

manuel-alvarez-alvarez (approved):

PR #6300, opened 2023-11-30
Detect servlet configuration vulnerabilities with IAST

jandro996:
# What Does This Do

Add 6 new vulnerabilities:

VERB_TAMPERING
ADMIN_CONSOLE_ACTIVE
DEFAULT_HTML_ESCAPE_INVALID
SESSION_TIMEOUT
DIRECTORY_LISTING_LEAK
INSECURE_JSP_LAYOUT

These vulnerabilities rely on the web.xml analysis or checking the jsp files location

We instrument the javax.servlet.ServletContext#getRealPath to be able to check the web.xml and the jsp files using the instrumentation context to call the module only once per context

# Motivation

Innovation week project

# Additional Notes

As we run it once per application there is no overheadController check

This PR aims to cover tomcat server

Jira ticket: [APPSEC-17025]

[APPSEC-17025]: https://datadoghq.atlassian.net/browse/APPSEC-17025?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ

### DIFF
--- a/dd-java-agent/agent-iast/src/main/java/com/datadog/iast/sink/ApplicationModuleImpl.java
+++ b/dd-java-agent/agent-iast/src/main/java/com/datadog/iast/sink/ApplicationModuleImpl.java
@@ -18,6 +18,7 @@ import java.nio.file.Path;
 import java.nio.file.Paths;
 import java.util.ArrayList;
 import java.util.List;
+import java.util.stream.Collectors;
 import javax.annotation.Nullable;
 
 public class ApplicationModuleImpl extends SinkModuleBase implements ApplicationModule {
@@ -139,16 +140,16 @@ public class ApplicationModuleImpl extends SinkModuleBase implements Application
     if (jspPaths.isEmpty()) {
       return;
     }
-    StringBuilder sb = new StringBuilder();
-    for (String jsp : jspPaths) {
-      sb.append(File.separatorChar).append(jsp).append('\n');
-    }
+    String result =
+        jspPaths.stream()
+            .map(s -> File.separatorChar + s)
+            .collect(Collectors.joining(System.lineSeparator()));
     reporter.report(
         span,
         new Vulnerability(
             VulnerabilityType.INSECURE_JSP_LAYOUT,
             Location.forSpanAndFileAndLine(span, "web.xml", -1),
-            new Evidence(sb.toString())));
+            new Evidence(result)));
   }
 
   @Nullable
--- a/dd-java-agent/agent-iast/src/main/java/com/datadog/iast/sink/ApplicationModuleImpl.java
+++ b/dd-java-agent/agent-iast/src/main/java/com/datadog/iast/sink/ApplicationModuleImpl.java
@@ -1,0 +1,247 @@
+package com.datadog.iast.sink;
+
+import com.datadog.iast.Dependencies;
+import com.datadog.iast.model.Evidence;
+import com.datadog.iast.model.Location;
+import com.datadog.iast.model.Vulnerability;
+import com.datadog.iast.model.VulnerabilityType;
+import datadog.trace.api.iast.sink.ApplicationModule;
+import datadog.trace.bootstrap.instrumentation.api.AgentSpan;
+import datadog.trace.bootstrap.instrumentation.api.AgentTracer;
+import java.io.BufferedReader;
+import java.io.File;
+import java.io.IOException;
+import java.io.StringReader;
+import java.nio.charset.StandardCharsets;
+import java.nio.file.Files;
+import java.nio.file.Path;
+import java.nio.file.Paths;
+import java.util.ArrayList;
+import java.util.List;
+import javax.annotation.Nullable;
+
+public class ApplicationModuleImpl extends SinkModuleBase implements ApplicationModule {
+
+  public ApplicationModuleImpl(final Dependencies dependencies) {
+    super(dependencies);
+  }
+
+  @Override
+  public void onRealPath(final @Nullable String realPath) {
+
+    if (realPath == null) {
+      return;
+    }
+
+    final AgentSpan span = AgentTracer.activeSpan();
+
+    checkInsecureJSPLayout(realPath, span);
+
+    String webXmlContent = webXmlContent(realPath);
+    if (webXmlContent == null) {
+      return;
+    }
+
+    checkVerbTampering(webXmlContent, span);
+    checkSessionTimeOut(webXmlContent, span);
+    checkDirectoryListingLeak(webXmlContent, span);
+    checkAdminConsoleActive(webXmlContent, span);
+  }
+
+  private void checkAdminConsoleActive(String webXmlContent, AgentSpan span) {
+    if (webXmlContent != null && webXmlContent.contains("Tomcat Manager Application")) {
+      reporter.report(
+          span,
+          new Vulnerability(
+              VulnerabilityType.ADMIN_CONSOLE_ACTIVE,
+              Location.forSpanAndFileAndLine(span, "web.xml", -1),
+              new Evidence("Tomcat Manager Application")));
+    }
+  }
+
+  private void checkDirectoryListingLeak(final String webXmlContent, final AgentSpan span) {
+    if (webXmlContent != null && webXmlContent.contains("<param-name>listings</param-name>")) {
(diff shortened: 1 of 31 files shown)


dd-java-agent/agent-iast/src/main/java/com/datadog/iast/sink/ApplicationModuleImpl.java
```diff
@@ -0,0 +1,278 @@
...
+            new Vulnerability(
+                VulnerabilityType.DEFAULT_HTML_ESCAPE_INVALID,
+                Location.forSpanAndFileAndLine(span, "web.xml", -1),
+                new Evidence("defaultHtmlEscape tag should be set")));
+      }
+    }
+  }
+
+  private void checkAdminConsoleActive(String webXmlContent, AgentSpan span) {
+    if (webXmlContent != null && webXmlContent.contains("Tomcat Manager Application")) {
```
jandro996, line 76: I will improve this

I think that this always is placed into <display-name> tag (<display-name>Tomcat Manager Application</display-name>) so we can relay on the Clue class to report also the line

dd-smoke-tests/springboot/src/main/resources/WEB-INF/web.xml
jandro996, line 1: remove this

dd-java-agent/agent-iast/src/main/java/com/datadog/iast/sink/ApplicationModuleImpl.java
```diff
@@ -0,0 +1,277 @@
...
+
+    if (realPath == null) {
+      return;
+    }
+
+    final AgentSpan span = AgentTracer.activeSpan();
+
+    checkInsecureJSPLayout(realPath, span);
+
+    String webXmlContent = webXmlContent(realPath);
```
smola, line 43: Does this work with `.war` files or only when the deployed app is not archived/compressed?

dd-java-agent/instrumentation/servlet/src/main/java/datadog/trace/instrumentation/servlet/dispatcher/IastServletContextInstrumentation.java
```diff
@@ -0,0 +1,69 @@
...
+
+  @Override
+  public Map<String, String> contextStore() {
+    return Collections.singletonMap(TYPE, String.class.getName());
+  }
+
+  @Override
+  public void adviceTransformations(AdviceTransformation transformation) {
+    transformation.applyAdvice(
+        isMethod().and(isPublic()).and(named("getRealPath")).and(takesArguments(String.class)),
```
manuel-alvarez-alvarez, line 53: Any reason to not instrument the constructor instead?, this will introduce a small overhead in all `getRealPath` operations.

dd-java-agent/agent-iast/src/main/java/com/datadog/iast/sink/ApplicationModuleImpl.java
```diff
@@ -0,0 +1,277 @@
...
+    if (webXmlContent == null) {
+      return;
+    }
+
+    checkVerbTampering(webXmlContent, span);
```
manuel-alvarez-alvarez, line 48: I think we should do it better, I've tried to count the number of unbounded `contains` and `indexOf` calls and the worst case scenario for big web descriptors can be huge. Can we change it so we only traverse the web descriptor only once?

anderruiz: I would say that it will complicate the logic, an easier solution ( I don't think we have a problem ) would be to create a regex with all the tags we are finding and finding the matches and then executing the logic of the analyzer for which we find the tag. But I would not rewrite each logic as it will be more complex to understand

dd-java-agent/agent-iast/src/main/java/com/datadog/iast/sink/ApplicationModuleImpl.java
anderruiz: It works for wars, I don't know if there could be some weird configuration in which it doesn't but it does not require the war to be uncompressed. (or at least it was in that way in Hdiv)

dd-java-agent/instrumentation/servlet/src/main/java/datadog/trace/instrumentation/servlet/dispatcher/IastServletContextInstrumentation.java
anderruiz: Given how the servlet works I could imagine not having everything set in the constructor for some servers. I doubt that this method is going to be called repeatedly

dd-java-agent/agent-iast/src/main/java/com/datadog/iast/sink/ApplicationModuleImpl.java
jandro996: I've made some changes with your suggestions, please let me know if we need to invest more in performance improvements

dd-java-agent/instrumentation/servlet/src/main/java/datadog/trace/instrumentation/servlet/dispatcher/IastServletContextInstrumentation.java
jandro996: As @anderruiz said there is a risk of losing vulnerabilities if the path is not correctly initialized

dd-java-agent/agent-iast/src/main/java/com/datadog/iast/sink/ApplicationModuleImpl.java
jandro996: we decide to remove all the Clue stuff to improve performance, the drawback of this approach is the loss of the web.xml line. Maybe we can do this calc if we found the vulnerability

dd-java-agent/agent-iast/src/main/java/com/datadog/iast/sink/ApplicationModuleImpl.java
jandro996: We will look for default apps in different servers and versions in future PRs

anderruiz (approved):

dd-java-agent/agent-iast/src/main/java/com/datadog/iast/sink/ApplicationModuleImpl.java
```diff
@@ -0,0 +1,289 @@
...
+
+  private static final String TOMCAT_MANAGER_APPLICATION = "Tomcat Manager Application";
+
+  private static final String LISTINGS_PATTERN = "<param-name>listings</param-name>";
+
+  private static final String SESSION_TIMEOUT_START_TAG = "<session-timeout>";
+
+  private static final String SESSION_TIMEOUT_END_TAG = "</session-timeout>";
+
+  private static final String SECURITY_CONSTRAINT_START_TAG = "<security-constraint>";
```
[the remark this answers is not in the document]
jandro996: I think that those cases can't happen:

1.  web.xml is case sensitive
2. you are not allowed to open a tag with a space between the '<' and the first letter
3. The available tags are defined in the schema so you can't have a valid uppercase tag that is defined lowercase in the schema, for instance `<SECURITY-CONSTRAINT>`

What is wrong right now is that we are not covering the `<security-constraint >` for instance that is valid or the `<security-constraint id="test">`

Thanks for your comment!

dd-java-agent/agent-iast/src/main/java/com/datadog/iast/sink/ApplicationModuleImpl.java
```diff
@@ -0,0 +1,289 @@
...
+
+    if (realPath == null) {
+      return;
+    }
+
+    final AgentSpan span = AgentTracer.activeSpan();
+
+    checkInsecureJSPLayout(realPath, span);
+
+    String webXmlContent = webXmlContent(realPath);
```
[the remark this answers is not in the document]
jandro996: I think that is not necessary, check my previous comment

manuel-alvarez-alvarez (approved):

PR #6302, opened 2023-11-30
Dependency collection with telemetry for native image

smola:
# What Does This Do
Adds support for dependency collection with telemetry on native image.

Dependency collection is done during the native image build (analysis stage), and embedded in the final binary within resource files. These are later read at runtime to send them through telemetry.

# Motivation

# Additional Notes

Depends on:
* https://github.com/DataDog/dd-trace-java/pull/6286 for the initial telemetry support

To do:
* [ ] Check if the scenario of multiple analysis runs with multiple dependencies files is possible and works.

### DIFF
--- a/telemetry/src/main/java/datadog/telemetry/TelemetrySystem.java
+++ b/telemetry/src/main/java/datadog/telemetry/TelemetrySystem.java
@@ -3,6 +3,7 @@ package datadog.telemetry;
 import datadog.communication.ddagent.DDAgentFeaturesDiscovery;
 import datadog.communication.ddagent.SharedCommunicationObjects;
 import datadog.telemetry.TelemetryRunnable.TelemetryPeriodicAction;
+import datadog.telemetry.dependency.Dependency;
 import datadog.telemetry.dependency.DependencyPeriodicAction;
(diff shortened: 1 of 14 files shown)


telemetry/src/main/java/datadog/telemetry/TelemetrySystem.java
```diff
@@ -110,4 +125,32 @@ public static void stop() {
...
+  /** Adds dependencies resolved statically. For native image. */
+  private static void addStaticDependencies(final TelemetryService telemetryService) {
+    try {
+      final Enumeration<URL> resources =
+          ClassLoader.getSystemClassLoader().getResources("dd-java-agent.dependencies");
+      while (resources.hasMoreElements()) {
+        final URL url = resources.nextElement();
+        log.debug("Dependency list found found: {}", url);
+        final URLConnection conn = url.openConnection();
+        try (final InputStream is = conn.getInputStream()) {
```
dougqh, line 138: Out of paranoia, I tend to want to make sure the wrappers are closed, too, but I _think_ this is safe.

dougqh (approved):

PR #6303, opened 2023-11-30
Fix config DD_IAST_TRUNCATION_MAX_VALUE_LENGTH value typo

jandro996:
# What Does This Do

Fix config DD_IAST_TRUNCATION_MAX_VALUE_LENGTH value typo

# Motivation

Bug


bm1549 (approved):
LGTM. If we suspect people are using this today, consider marking it as breaking as well

smola:
> @bm1549 If we suspect people are using this today, consider marking it as breaking as well

Close to zero chance since it's a very recent addition and not publicly documented or communicated to anyone through support.

PR #6305, opened 2023-12-01
Add spark SQL plan the spark.stage spans

paul-laffon-dd:
# What Does This Do

Adding two new hidden fields
- `_dd.spark.sql_plan`, containing a JSON representation of the SQL plan for this stage
- `_dd.spark.parent_ids`, containing the ids of the parent stage for the stage hierarchy

Spark is generating the SQL plan at the SQL query level (which can be composed of multiple stages). Using the accumulators of each SQL node to map them to their given stage

# Motivation

Collect information about SQL execution


dd-java-agent/instrumentation/spark/src/main/java/datadog/trace/instrumentation/spark/SparkSQLUtils.java
```diff
@@ -0,0 +1,199 @@
...
+      Collection<SQLMetricInfo> metrics =
+          AbstractDatadogSparkListener.listener.getPlanInfoMetrics(plan);
+
+      // Writing final values of metrics
+      if (metrics.size() > 0) {
+        generator.writeFieldName("metrics");
+        generator.writeStartArray();
+        for (SQLMetricInfo metric : metrics) {
+          long accumulatorId = metric.accumulatorId();
+          AccumulatorWithStage acc = accumulators.get(accumulatorId);
```
raphaelgavache, line 164: shouldn't the accumulator be removed the `accumulators` map here?

dd-java-agent/instrumentation/spark/src/testFixtures/groovy/datadog/trace/instrumentation/spark/AbstractSparkTest.groovy
```diff
@@ -587,4 +589,55 @@ abstract class AbstractSparkTest extends AgentTestRunner {
...
+          spanType "spark"
+          childOf(span(1))
+        }
+        span {
+          operationName "spark.stage"
+          spanType "spark"
+          childOf(span(2))
+          // Exact SQL Plan changes depending on the spark version
+          assert span.tags["_dd.spark.sql_plan"] =~ /.*HashAggregate.*Filter.*SerializeFromObject.*MapElements.*DeserializeToObject.*LocalTableScan.*/
+          assert isJsonValid(span.tags["_dd.spark.sql_plan"].toString())
```
raphaelgavache, line 638: adding a test with metrics to test accumulator logic, or at least the toJson function with different real world spark queries would be useful to enforce the behaviour

dd-java-agent/instrumentation/spark/src/main/java/datadog/trace/instrumentation/spark/SparkSQLUtils.java
```diff
@@ -0,0 +1,199 @@
...
+
+            if (name != null && value != null) {
+              generator.writeStartObject();
+              generator.writeNumberField(name, value);
+              generator.writeStringField("type", metric.metricType());
+              generator.writeEndObject();
+            }
+          }
+        }
+        generator.writeEndArray();
```
raphaelgavache, line 183: nit: it's hard to visually assert that the EndArray is at the right level of indentation, it would be easier of that MetricInfo is added in it's own function

dd-java-agent/instrumentation/spark/src/main/java/datadog/trace/instrumentation/spark/SparkSQLUtils.java
paul-laffon-dd: Accumulators are also used for [the stopping condition](https://github.com/DataDog/dd-trace-java/pull/6305/files#diff-e076f41c2f836b621ad9b08805c6fdb3ed93ecda9aacd1d6f4bd378d750cf1d7R47-R48) when traversing the SQL plan, to know if a node is part of another stage. Meaning we would want to keep the accumulators as long as the SQL query is active

dd-java-agent/instrumentation/spark/src/main/java/datadog/trace/instrumentation/spark/SparkSQLUtils.java
paul-laffon-dd: Agreed, changing it

dd-java-agent/instrumentation/spark/src/testFixtures/groovy/datadog/trace/instrumentation/spark/AbstractSparkTest.groovy
paul-laffon-dd: Looking into adding more testing to the PR

raphaelgavache (approved):
lgtm!

dd-java-agent/instrumentation/spark/src/main/java/datadog/trace/instrumentation/spark/RemoveEldestHashMap.java
```diff
@@ -0,0 +1,17 @@
+package datadog.trace.instrumentation.spark;
+
+import java.util.LinkedHashMap;
+import java.util.Map;
+
+public class RemoveEldestHashMap<K, V> extends LinkedHashMap<K, V> {
```
dougqh, line 6: Generally speaking, it is considered poor form to extend a Collection class.
Also, LinkedHashMap is quite heavy, so I'd typically prefer to see something else used instead.

dd-java-agent/instrumentation/spark/src/main/java/datadog/trace/instrumentation/spark/AbstractDatadogSparkListener.java
```diff
@@ -602,16 +632,44 @@ public void onOtherEvent(SparkListenerEvent event) {
...
+      Class<?> adaptiveExecutionUpdateClass =
+          Class.forName(
+              "org.apache.spark.sql.execution.ui.SparkListenerSQLAdaptiveExecutionUpdate");
+
+      if (adaptiveExecutionUpdateClass.isInstance(event)) {
+        Method executionIdMethod = adaptiveExecutionUpdateClass.getDeclaredMethod("executionId");
+        Method sparkPlanInfoMethod =
+            adaptiveExecutionUpdateClass.getDeclaredMethod("sparkPlanInfo");
+
+        long queryId = (long) executionIdMethod.invoke(event);
```
dougqh, line 651: Depending on the visibility, it would be preferable to use MethodHandles rather than Method.  MethodHandles are lighter and produce fewer warnings (assuming the methods are visible normally from here).

dd-java-agent/instrumentation/spark/src/main/java/datadog/trace/instrumentation/spark/AbstractDatadogSparkListener.java
```diff
@@ -602,16 +632,44 @@ public void onOtherEvent(SparkListenerEvent event) {
...
+
+  @SuppressForbidden // Using reflection to avoid splitting the instrumentation once more
+  private synchronized void updateAdaptiveSQLPlan(SparkListenerEvent event) {
+    try {
+      Class<?> adaptiveExecutionUpdateClass =
+          Class.forName(
+              "org.apache.spark.sql.execution.ui.SparkListenerSQLAdaptiveExecutionUpdate");
+
+      if (adaptiveExecutionUpdateClass.isInstance(event)) {
+        Method executionIdMethod = adaptiveExecutionUpdateClass.getDeclaredMethod("executionId");
```
dougqh, line 647: Typically, it is best to stash the Method / MethodHandle in a static.

dd-java-agent/instrumentation/spark/src/main/java/datadog/trace/instrumentation/spark/AbstractDatadogSparkListener.java
```diff
@@ -75,8 +82,15 @@ public abstract class AbstractDatadogSparkListener extends SparkListener {
...
       new HashMap<>();
   private final HashMap<Long, SparkListenerSQLExecutionStart> sqlQueries = new HashMap<>();
+  protected final HashMap<Long, SparkPlanInfo> sqlPlans = new HashMap<>();
   private final HashMap<String, SparkListenerExecutorAdded> liveExecutors = new HashMap<>();
 
+  // There is no easy way to know if an accumulator is not useful anymore (meaning it is not part of
+  // an active SQL query)
+  // so capping the size of the collection storing them
+  private final Map<Long, SparkSQLUtils.AccumulatorWithStage> accumulators =
+      new RemoveEldestHashMap<>(10000);
```
dougqh, line 92: 10,000 seems quite large.  How did you arrive at that number?
Also, it would be nice for that number to configuration or at least have a constant, so it is more self-documenting.

dd-java-agent/instrumentation/spark/src/main/java/datadog/trace/instrumentation/spark/SparkSQLUtils.java
```diff
@@ -0,0 +1,209 @@
...
+      span.setTag("_dd.spark.sql_plan", json);
+    }
+  }
+
+  public static SparkPlanInfoForStage computeStageInfoForStage(
+      SparkPlanInfo plan,
+      Map<Long, AccumulatorWithStage> accumulators,
+      int stageId,
+      boolean foundStage) {
+    Set<Integer> stageIds = stageIdsForPlan(plan, accumulators);
```
dougqh, line 43: The plan is probably fine, but we need to be careful about attaching SQL to spans because they may contain sensitive information.

dd-java-agent/instrumentation/spark/src/main/java/datadog/trace/instrumentation/spark/SparkSQLUtils.java
```diff
@@ -0,0 +1,209 @@
...
+
+    public SparkPlanInfoForStage(SparkPlanInfo plan, List<SparkPlanInfoForStage> children) {
+      this.plan = plan;
+      this.children = children;
+    }
+
+    public String toJson(Map<Long, AccumulatorWithStage> accumulators) {
+      // Using the jackson JSON lib used by spark
+      // https://mvnrepository.com/artifact/org.apache.spark/spark-core_2.12/3.5.0
+      ByteArrayOutputStream baos = new ByteArrayOutputStream();
```
dougqh, line 144: This style of code can create resource leaks.
close operations need to be in finally block -- or you need to use try-with-resources.

I would suggest try-with-resources.

dd-java-agent/instrumentation/spark/src/main/java/datadog/trace/instrumentation/spark/SparkSQLUtils.java
```diff
@@ -0,0 +1,209 @@
...
+          AccumulatorWithStage acc = accumulators.get(accumulatorId);
+          if (acc != null) {
+            acc.toJson(generator, metric);
+          }
+        }
+        generator.writeEndArray();
+      }
+
+      // Writing child nodes
+      if (children.size() > 0) {
```
dougqh, line 197: This is better written as children.isEmpty()

dougqh (changes requested):
The handling of close is a potential resource leak.
While I think it benign in this case because it is working on a ByteArrayOutputStream, I would still prefer to see it fixed to avoid introducing problematic examples into the codebase.

There are number of things that could be improved from a performance perspective, but mostly, not mandatory.

dd-java-agent/instrumentation/spark/src/main/java/datadog/trace/instrumentation/spark/AbstractDatadogSparkListener.java
paul-laffon-dd: Moved the value to a constant. 

The value is a trade-off between losing old accumulators, which will impact SQL plan computation and not using too much memory. 

From my knowledge, a large SQL query can have more than 1000 accumulators, so using one order of magnitude above. Considering ~100 bytes per Accumulator, this should have a ~1MB memory impact

dd-java-agent/instrumentation/spark/src/main/java/datadog/trace/instrumentation/spark/SparkSQLUtils.java
paul-laffon-dd: Changed it to `try-with-resources`

dd-java-agent/instrumentation/spark/src/main/java/datadog/trace/instrumentation/spark/SparkSQLUtils.java
paul-laffon-dd: Change it to `!children.isEmpty()`

dd-java-agent/instrumentation/spark/src/main/java/datadog/trace/instrumentation/spark/SparkSQLUtils.java
```diff
@@ -0,0 +1,208 @@
...
+
+    public String toJson(Map<Long, AccumulatorWithStage> accumulators) {
+      // Using the jackson JSON lib used by spark
+      // https://mvnrepository.com/artifact/org.apache.spark/spark-core_2.12/3.5.0
+      ObjectMapper mapper =
+          new ObjectMapper().configure(DeserializationFeature.FAIL_ON_UNKNOWN_PROPERTIES, false);
+      try (ByteArrayOutputStream baos = new ByteArrayOutputStream()) {
+        JsonGenerator generator = mapper.getFactory().createGenerator(baos);
+        this.toJson(generator, accumulators);
+        generator.close();
```
dougqh, line 149: baos.close will now be done by the try-with-resources.

But you may or may not need that depending on what generator.close does.
If generator.close closes the InputStream, then you can just make sure generator close happens.
Unfortunately, the generator.close part is still problematic.  It also needs to be done try-with-resources or try / finally.

dd-java-agent/instrumentation/spark/src/main/java/datadog/trace/instrumentation/spark/SparkSQLUtils.java
```diff
@@ -0,0 +1,208 @@
...
+
+      List<SQLMetricInfo> metrics = AbstractDatadogSparkListener.listener.getPlanInfoMetrics(plan);
+
+      // Scala list are immutable, copying to sort it
+      metrics = new ArrayList<>(metrics);
+      // Sorting to have a consistent order of metrics
+      metrics.sort(Comparator.comparing(SQLMetricInfo::name));
+
+      // Writing final values of metrics
+      if (metrics.size() > 0) {
```
dougqh, line 182: !isEmpty() would be better

dd-java-agent/instrumentation/spark/src/main/java/datadog/trace/instrumentation/spark/SparkSQLUtils.java
```diff
@@ -0,0 +1,208 @@
...
+          generator.writeStringField(metadata._1, metadata._2);
+        }
+
+        generator.writeEndObject();
+      }
+
+      List<SQLMetricInfo> metrics = AbstractDatadogSparkListener.listener.getPlanInfoMetrics(plan);
+
+      // Scala list are immutable, copying to sort it
+      metrics = new ArrayList<>(metrics);
```
dougqh, line 177: I find this bit slightly hard to read, since it sets metrics then immediately overwrites it.
I think it might be cleaner to just immediately do the conversion to an ArrayList.

dd-java-agent/instrumentation/spark/src/main/java/datadog/trace/instrumentation/spark/SparkSQLUtils.java
```diff
@@ -0,0 +1,208 @@
...
+      // Sorting to have a consistent order of metrics
+      metrics.sort(Comparator.comparing(SQLMetricInfo::name));
```
dougqh, line 179: Why?  I think is work better left to the consumer.

dd-java-agent/instrumentation/spark/src/main/java/datadog/trace/instrumentation/spark/SparkSQLUtils.java
paul-laffon-dd: Make sense, I changed the try with resource to be on the generator instead, so that it is always closed and will handle the `ByteArrayOutputStream` close

dd-java-agent/instrumentation/spark/src/main/java/datadog/trace/instrumentation/spark/SparkSQLUtils.java
paul-laffon-dd: Changed it to `!isEmpty()`

dd-java-agent/instrumentation/spark/src/main/java/datadog/trace/instrumentation/spark/SparkSQLUtils.java
paul-laffon-dd: I agree, changed it to immediately do the conversion

dd-java-agent/instrumentation/spark/src/main/java/datadog/trace/instrumentation/spark/SparkSQLUtils.java
paul-laffon-dd: This is to have a deterministic order in the metrics, making the unit test much simpler to implement and reducing the chances of flakiness

dd-java-agent/instrumentation/spark/src/main/java/datadog/trace/instrumentation/spark/AbstractDatadogSparkListener.java
```diff
@@ -602,16 +634,64 @@ public void onOtherEvent(SparkListenerEvent event) {
     } else if (event instanceof SparkListenerSQLExecutionEnd) {
       onSQLExecutionEnd((SparkListenerSQLExecutionEnd) event);
     }
+
+    updateAdaptiveSQLPlan(event);
+  }
+
+  private static Class<?> adaptiveExecutionUpdateClass;
```
dougqh, line 641: These can be final

paul-laffon-dd: I was setting those on the first time they were accessed, but I agree that it should be more efficient / easier to just declare those final

dd-java-agent/instrumentation/spark/src/main/java/datadog/trace/instrumentation/spark/SparkSQLUtils.java
dougqh: try-with-resources will perform the close implicitly, you don't need to call close explicitly.

dd-java-agent/instrumentation/spark/src/main/java/datadog/trace/instrumentation/spark/SparkSQLUtils.java
dougqh: Then you'd be better doing the sorting in the test.  Adding overhead to the production code to ease the test is not a great trade off.

dd-java-agent/instrumentation/spark/src/main/java/datadog/trace/instrumentation/spark/SparkSQLUtils.java
paul-laffon-dd: The generator is not flushing the data to the output stream until it is closed. Move the return statement at the end of the function so that the explicit close can be removed

dd-java-agent/instrumentation/spark/src/main/java/datadog/trace/instrumentation/spark/SparkSQLUtils.java
paul-laffon-dd: Removed the sorting by reworking the unit test in the latest commits

dougqh (approved):

PR #6314, opened 2023-12-05
Test that HTTP headers with commas should be truncated at the first comma

bm1549:
# What Does This Do
Adds a new test that confirms that we truncate headers at the first comma
This also applies our default auto-formatting to the updated file

# Motivation
Codifies expected behavior via additional testing

# Additional Notes

Jira ticket: APMS-10909


nayeem-kamal (commented):
Test changes look good to me. There seem to be some weird formatting changes (Looks like spotless CI failed). It would be nice to revert those changes if possible since they seem unnecessary. Let me know what you think.  Other than that LGTM.

nayeem-kamal (approved):
LGTM. Just the formatting changes I mentioned above missing.(probably need to run spotless)

bantonsson (changes requested):
I actually think that we should go in the other direction and be lenient with this, since the header names are end user supplied, and most likely under the users control. I'm working on a PR that ensures this. #6376

PR #6326, opened 2023-12-06
R&D week: blackhole span

amarziali:
# What Does This Do

Introduces the blackhole span who's muting the tracing during its lifetime (as soon as a blackhole scope is opened)
As analogy with the `@Trace` annotation, a `@DoNotTrace` annotation is also provided in order to decorate a method with a blackhole activated span.

It also provides the possibility to obtain a blackhole scope using the DD public API.

Finally showcases how it can be used in an instrumentation. In this case the example is in decoupling the apache http client instrumentation from the aws sdk one

# Motivation

# Additional Notes

Jira ticket: [PROJ-IDENT]


dd-java-agent/instrumentation/trace-annotation/src/main/java/datadog/trace/instrumentation/trace_annotation/DoNotTraceAnnotationInstrumentation.java
```diff
@@ -0,0 +1,44 @@
...
+
+  @Override
+  public ElementMatcher<TypeDescription> hierarchyMatcher() {
+    return declaresMethod(isAnnotatedWith(named(hierarchyMarkerType())));
+  }
+
+  @Override
+  public String[] helperClassNames() {
+    return new String[] {
+      packageName + ".TraceDecorator",
```
amarziali, line 35: should delete this one since it's not needed

internal-api/src/main/java/datadog/trace/bootstrap/instrumentation/api/AgentTracer.java
```diff
@@ -519,7 +536,32 @@ public AgentHistogram newHistogram(double relativeAccuracy, int maxNumBins) {
     }
   }
 
-  public static final class NoopAgentSpan implements AgentSpan {
+  public static final class BlackholeAgentSpan extends NoopAgentSpan {
+    private final DDTraceId ddTraceId;
+
+    public BlackholeAgentSpan(final DDTraceId ddTraceId) {
+      this.ddTraceId = ddTraceId;
```
ygree, line 543: Can you please clarify what is the purpose of the trace ID here?

amarziali: I think I need a real trace ID here (and not zero as no-op span). This will allow to respect the contract of `isSameTrace` (used in some instrumentations) and also continue providing the traceId for logging correlation

dd-java-agent/instrumentation/apache-httpclient-4/src/main/java/datadog/trace/instrumentation/apachehttpclient/HelperMethods.java
```diff
@@ -19,46 +18,30 @@
 import org.apache.http.client.methods.HttpUriRequest;
 
 public class HelperMethods {
-  private static final boolean AWS_LEGACY_TRACING =
-      Config.get().isLegacyTracingEnabled(false, "aws-sdk");
-
   public static AgentScope doMethodEnter(final HttpUriRequest request) {
-    boolean awsClientCall = request.containsHeader("amz-sdk-invocation-id");
-    if (!AWS_LEGACY_TRACING && awsClientCall) {
```
ygree, line 27: Nice to see that the logic associated with AWS is now being moved to the associated instrumentation.

ygree (approved):
Looks good to me

dd-java-agent/instrumentation/apache-httpclient-4/src/main/java/datadog/trace/instrumentation/apachehttpclient/HelperMethods.java
amarziali: thanks. Yes the same can be done for all that kind of instrumentations that are coupled

PR #6340, opened 2023-12-10
API Security configuration via remote-config

ValentinZakharov:
# What Does This Do
Improvement allows API Security to set request sample rate for API Security using remote-config. Assumed API Security can be deactivated if the value set to `0`. Default value for sample rate is `0.1` (means - will be applied to 10% of sampled requests)

The `ASM_FEATURE` product extended with extra object/value
```
{
  "api_security": {
    "request_sample_rate": 0.1
  }  
}
```
In the case that ASM is disabled, the API Security feature will be automatically disabled as well. However, the tracer will keep the last value of `request_sample_rate` to be used as soon as ASM is enabled again.

# Motivation
Customers want to have option to tune or deactivate API Security remotely.

# Additional Notes

Two possible states for the tracer with API Security activated with `DD_EXPERIMENTAL_API_SECURITY_ENABLED=true`

ASM is activated by one click remote config. The tracer must register the `ASM_FEATURES` product and receive all information, both for asm activation and api sample rate.

ASM is activated by an env var `DD_APPSEC_ENABLED=true`. The tracer must register the `ASM_FEATURES` product (this is new from previous tracers behaviour) and only receive information about api sample rate.

Jira ticket: [APPSEC-16674]

[APPSEC-16674]: https://datadoghq.atlassian.net/browse/APPSEC-16674?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ


smola (changes requested):

dd-java-agent/appsec/src/main/java/com/datadog/appsec/config/AppSecConfigServiceImpl.java
```diff
@@ -74,10 +75,10 @@ public AppSecConfigServiceImpl(
 
   private void subscribeConfigurationPoller() {
     // see also close() method
-    if (tracerConfig.getAppSecActivation() == ProductActivation.ENABLED_INACTIVE) {
+    if (tracerConfig.getAppSecActivation() != ProductActivation.FULLY_DISABLED) {
```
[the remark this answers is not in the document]
ValentinZakharov: Rolled back changes according to a new fixed behaviour

dd-java-agent/appsec/src/main/java/com/datadog/appsec/api/security/ApiSecurityRequestSampler.java
```diff
@@ -1,17 +1,55 @@
...
   }
 
+  public ApiSecurityRequestSampler(final Config config, ConfigurationPoller configurationPoller) {
+    this(config);
+    if (configurationPoller != null) {
+      configurationPoller.addListener(
+          Product.ASM_FEATURES,
+          AppSecFeaturesDeserializer.INSTANCE,
+          (configKey, newConfig, pollingRateHinter) -> {
+            if (newConfig != null && newConfig.apiSecurity != null) {
```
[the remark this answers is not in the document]
ValentinZakharov: Added checking for the empty config `AppSecConfigServiceImpl`

.gitlab/benchmarks.yml
```diff
@@ -30,3 +30,40 @@ benchmarks:
 
     KUBERNETES_SERVICE_ACCOUNT_OVERWRITE: dd-trace-java
     FF_USE_LEGACY_KUBERNETES_EXECUTION_STRATEGY: "true"
+
+.dsm-kafka-benchmarks:
```
smola, line 34: Does this belong to this PR?

smola (commented):
I think there was a dirty merge or febase here, so many unrelated changes...

remote-config/src/main/java/datadog/remoteconfig/state/ProductState.java
```diff
@@ -29,18 +30,22 @@ public class ProductState {
...
-  private final ProductListener listener;
+  private final List<ProductListener> listeners;
 
   List<ConfigurationPoller.ReportableException> errors = null;
 
-  public ProductState(Product product, ProductListener listener) {
+  public ProductState(Product product) {
     this.product = product;
-    this.listener = listener;
+    this.listeners = new LinkedList<>();
```
nikita-tkachenko-datadog, line 39: Any specific reason `LinkedList` is used here and not the array list? The latter is more efficient in terms of memory and when it comes to iteration (which is what we seem to be doing here)

remote-config/src/main/java/datadog/remoteconfig/ConfigurationPoller.java
```diff
@@ -129,7 +129,9 @@ public ConfigurationPoller(
   }
 
   public synchronized <T> void addListener(Product product, ProductListener listener) {
-    this.productStates.put(product, new ProductState(product, listener));
+    ProductState productState =
+        this.productStates.computeIfAbsent(product, p -> new ProductState(product));
```
nikita-tkachenko-datadog, line 133: As a side-note, `productStates` could be an `EnumMap`

remote-config/src/main/java/datadog/remoteconfig/state/ProductState.java
```diff
@@ -94,7 +99,9 @@ private void callListenerApplyTarget(
       byte[] content) {
 
     try {
-      listener.accept(configKey, content, hinter);
+      for (ProductListener listener : listeners) {
+        listener.accept(configKey, content, hinter);
```
nikita-tkachenko-datadog, line 103: Not sure whether relevant or not, but if a single listener fails here, those that come after it in the list will not receive the notification. Same goes for remove/commit operations.

dd-java-agent/appsec/src/main/java/com/datadog/appsec/api/security/ApiSecurityRequestSampler.java
```diff
@@ -1,17 +1,55 @@
...
 import java.util.concurrent.atomic.AtomicLong;
+import org.slf4j.Logger;
+import org.slf4j.LoggerFactory;
 
 public class ApiSecurityRequestSampler {
 
-  private final int sampling;
+  private static final Logger log = LoggerFactory.getLogger(ApiSecurityRequestSampler.class);
+
+  private int sampling;
```
nikita-tkachenko-datadog, line 17: There seems to be no synchronization where this field is being read. If I understood correctly, the listener that modifies it will be called in a task scheduler thread, while the `sampleRequest` method is registered as a callback in the `SubscriptionService` which can be called from many different threads. In absence of synchronization changes to the field might not be visible.

.gitlab/benchmarks.yml
ValentinZakharov: Thanks, Rebased

remote-config/src/main/java/datadog/remoteconfig/state/ProductState.java
ValentinZakharov: Right. If I recall correctly ArrayList's default size is 10, and we have only few listeners per product - there will never happens reallocation, so we can easily use ArrayList.

remote-config/src/main/java/datadog/remoteconfig/ConfigurationPoller.java
ValentinZakharov: Replaced with `EnumMap`

remote-config/src/main/java/datadog/remoteconfig/state/ProductState.java
ValentinZakharov: The problem here is - in current implementation config state is a single value, so if problem happens, we can report only one exception. So as a workaround I decided to interrupt iteration and report problem in case if issue arises.
I suppose, we might need to improve remote-config state system in future implementations.

nikita-tkachenko-datadog (approved):

remote-config/src/main/java/datadog/remoteconfig/state/ProductState.java
nikita-tkachenko-datadog: Makes sense. An alternative could be to have a single exception and link exceptions from individual listeners as suppressed, although I'm not familiar enough with the remote config to know whether it'd be acceptable

dd-java-agent/appsec/src/main/java/com/datadog/appsec/config/AppSecConfigServiceImpl.java
```diff
@@ -160,15 +161,16 @@ private void subscribeActivation() {
...
-              newConfig != null && newConfig.asm != null && newConfig.asm.enabled;
-          if (AppSecSystem.isActive() != newState) {
-            log.info("AppSec {} (runtime)", newState ? "enabled" : "disabled");
-            AppSecSystem.setActive(newState);
-            if (AppSecSystem.isActive()) {
-              // On remote activation, we need to re-distribute the last known configuration.
-              // This may trigger initializations, including PowerWAF if it was lazy loaded.
-              this.currentAppSecConfig.dirtyStatus.markAllDirty();
+          if (newConfig != null && newConfig.asm != null) {
+            final boolean newState = newConfig.asm.enabled;
```
ygree, line 165: Looks like ASM can be enabled/disabled via remote config? If so, we need to update the telemetry to report product changes. Now it's only reported once at the start of the sending application. See TelemetryRequest.writeProducts()

ygree (approved):
One question about Telemetry. Otherwise looks good to me!

dd-java-agent/appsec/src/main/java/com/datadog/appsec/config/AppSecConfigServiceImpl.java
ygree: I've created a Jira ticket [APMJAVA-1174] to generate product change events when Appsec is turned on/off at runtime.

dd-java-agent/appsec/src/main/java/com/datadog/appsec/api/security/ApiSecurityRequestSampler.java
```diff
@@ -1,17 +1,56 @@
...
+  private static final Logger log = LoggerFactory.getLogger(ApiSecurityRequestSampler.class);
+
+  private volatile int sampling;
   private final AtomicLong cumulativeCounter = new AtomicLong();
 
   public ApiSecurityRequestSampler(final Config config) {
     sampling = computeSamplingParameter(config.getApiSecurityRequestSampleRate());
   }
 
+  public ApiSecurityRequestSampler(final Config config, ConfigurationPoller configurationPoller) {
```
jandro996, line 24: Not relevant! IMHO you can reduce the cognitive overhead if you avoid nested if statements if it's possible. For instance:

```
if (configurationPoller != null) {
return;
}
```
//continue with the logic

jandro996 (approved):

jandro996 (approved):

smola (approved):
@ValentinZakharov  Looks good to me, assuming the possible interaction with ASM enable/disable via remote config has been tested.

PR #6353, opened 2023-12-12
muzzle: skip bad jersey versions

amarziali:
# What Does This Do

Failing because pom seems to be corrupt

# Motivation

# Additional Notes

Jira ticket: [PROJ-IDENT]


amarziali:
closing since it seems unrelated

PR #6365, opened 2023-12-14
Add emitting status for probes

jpbempel:
# What Does This Do
Add a new status for probes to indicate the fact that they are sending data. a new EMITTING status is added and for each probe implementation we send this status.
- Log probes: when snapshot is sent
- metric probes: when we sent the metric to statsd
- span probes: when the span finishes
- span decoration probes: when all tags were added.

The status is set and sent only the first time.
Refactoring was done for providing full probe id (uuid + version) inside the probe evaluation as encoded string `uuid:version`. 

Refactored also smoke tests to test the status is reached when data are sent.

# Motivation
Indicate when a probe is emitting data

# Additional Notes

Jira ticket: [DEBUG-1893]

[DEBUG-1893]: https://datadoghq.atlassian.net/browse/DEBUG-1893?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ

### DIFF
--- a/dd-java-agent/agent-debugger/src/main/java/com/datadog/debugger/agent/ConfigurationUpdater.java
+++ b/dd-java-agent/agent-debugger/src/main/java/com/datadog/debugger/agent/ConfigurationUpdater.java
@@ -7,6 +7,7 @@ import com.datadog.debugger.probe.ProbeDefinition;
 import com.datadog.debugger.probe.SpanDecorationProbe;
 import com.datadog.debugger.probe.SpanProbe;
 import com.datadog.debugger.sink.DebuggerSink;
+import com.datadog.debugger.sink.ProbeStatusSink;
 import com.datadog.debugger.util.ExceptionHelper;
 import datadog.trace.api.Config;
 import datadog.trace.bootstrap.debugger.DebuggerContext;
@@ -69,7 +70,8 @@ public class ConfigurationUpdater
         instrumentation,
         transformerSupplier,
         config,
-        new DebuggerSink(config, config.getFinalDebuggerSnapshotUrl(), false),
+        new DebuggerSink(
+            config, new ProbeStatusSink(config, config.getFinalDebuggerSnapshotUrl(), false)),
         finder);
   }
 
@@ -240,7 +242,7 @@ public class ConfigurationUpdater
   @Override
   public ProbeImplementation resolve(String id, Class<?> callingClass) {
     ProbeDefinition definition = appliedDefinitions.get(id);
-    if (definition == null) {
+    if (definition == null && callingClass != null) {
       LOGGER.info(
           "Cannot resolve probe id={}, re-transforming calling class: {}",
           id,
--- a/dd-java-agent/agent-debugger/src/main/java/com/datadog/debugger/agent/ConfigurationUpdater.java
+++ b/dd-java-agent/agent-debugger/src/main/java/com/datadog/debugger/agent/ConfigurationUpdater.java
@@ -195,7 +195,7 @@ public class ConfigurationUpdater
     if (instrumentationResult.isError()) {
       return;
     }
-    instrumentationResults.put(definition.getId(), instrumentationResult);
+    instrumentationResults.put(definition.getProbeId().getEncodedId(), instrumentationResult);
     if (instrumentationResult.isInstalled()) {
       sink.addInstalled(definition.getProbeId());
     } else if (instrumentationResult.isBlocked()) {
@@ -230,22 +230,22 @@ public class ConfigurationUpdater
 
   private void storeDebuggerDefinitions(ConfigurationComparer changes) {
     for (ProbeDefinition definition : changes.getRemovedDefinitions()) {
-      appliedDefinitions.remove(definition.getId());
+      appliedDefinitions.remove(definition.getProbeId().getEncodedId());
     }
     for (ProbeDefinition definition : changes.getAddedDefinitions()) {
-      appliedDefinitions.put(definition.getId(), definition);
+      appliedDefinitions.put(definition.getProbeId().getEncodedId(), definition);
     }
     LOGGER.debug("Stored appliedDefinitions: {}", appliedDefinitions.values());
   }
 
   // /!\ This is called potentially by multiple threads from the instrumented code /!\
   @Override
-  public ProbeImplementation resolve(String id, Class<?> callingClass) {
-    ProbeDefinition definition = appliedDefinitions.get(id);
+  public ProbeImplementation resolve(String encodedProbeId, Class<?> callingClass) {
+    ProbeDefinition definition = appliedDefinitions.get(encodedProbeId);
     if (definition == null && callingClass != null) {
       LOGGER.info(
           "Cannot resolve probe id={}, re-transforming calling class: {}",
-          id,
+          encodedProbeId,
           callingClass.getName());
       retransformClasses(Collections.singletonList(callingClass));
       return null;
--- a/dd-java-agent/agent-debugger/src/main/java/com/datadog/debugger/agent/StatsdMetricForwarder.java
+++ b/dd-java-agent/agent-debugger/src/main/java/com/datadog/debugger/agent/StatsdMetricForwarder.java
@@ -1,10 +1,12 @@
 package com.datadog.debugger.agent;
 
+import com.datadog.debugger.sink.ProbeStatusSink;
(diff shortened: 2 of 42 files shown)


dd-java-agent/agent-debugger/src/main/java/com/datadog/debugger/agent/StatsdMetricForwarder.java
```diff
@@ -25,45 +30,63 @@ public StatsdMetricForwarder(Config config) {
...
     statsd.distribution(name, value, tags);
+    sendEmittingStatus(probeId);
   }
 
   @Override
   public void handle(Exception exception) {
     LOGGER.warn("Error when sending metrics: ", exception);
   }
+
+  private void sendEmittingStatus(String probeId) {
```
shatzi, line 84: why we are only passing probeId string here? and not use ProbeId type with id+version?

an instrumentation can pass an old probeId while the resolver return a new probeId version and send emitng status to the wrong probe version.

dd-java-agent/agent-debugger/src/main/java/com/datadog/debugger/instrumentation/Instrumentor.java
```diff
@@ -68,6 +70,8 @@ public Instrumentor(
       argOffset += t.getSize();
     }
     localVarsBySlot = extractLocalVariables(argTypes);
+    this.probeIdFieldName =
```
shatzi, line 73: where this is used?

dd-java-agent/agent-debugger/src/test/java/com/datadog/debugger/agent/SpanDecorationProbeInstrumentationTest.java
```diff
@@ -84,6 +86,7 @@ public void methodActiveSpanSimpleTag() throws IOException, URISyntaxException {
     MutableSpan span = traceInterceptor.getFirstSpan();
     assertEquals("1", span.getTags().get("tag1"));
     assertEquals(PROBE_ID.getId(), span.getTags().get("_dd.di.tag1.probe_id"));
+    verify(probeStatusSink).addEmitting(ArgumentMatchers.eq(PROBE_ID));
```
shatzi, line 89: should we verify test with all received/active/emitting?

dd-smoke-tests/debugger-integration-tests/src/test/java/datadog/smoketest/LogProbesIntegrationTest.java
```diff
@@ -62,27 +66,27 @@ void testFullMethod() throws Exception {
...
+              snapshot.getCaptures().getReturn(),
+              "java.lang.String",
+              "42, foobar, 3.42, {key1=val1, key2=val2, key3=val3}, var1,var2,var3");
+          assertNotNull(snapshot.getCaptures().getReturn().getLocals());
+          assertEquals(1, snapshot.getCaptures().getReturn().getLocals().size()); // @return
+          assertNull(snapshot.getCaptures().getReturn().getThrowable());
+          assertNull(snapshot.getCaptures().getReturn().getFields());
+          snapshotReceived.set(true);
+        });
+    AtomicBoolean statusResult = registerCheckReceivedInstalledEmitting();
```
shatzi, line 88: very cool tests

shatzi (commented):
I think that the method probe status is still broken.

another idea you might want to consider is to use ProbeId.toString() and Probe.fromString(string) that will encode decode probe-id + version so instrumentation can use simple strings as consts but then we can decode them to have a ProbeId class with version. then overwrite emitting status with wrong version will be resolved

dd-java-agent/agent-debugger/src/main/java/com/datadog/debugger/instrumentation/Instrumentor.java
jpbempel: unused. removed.

dd-java-agent/agent-debugger/src/main/java/com/datadog/debugger/agent/StatsdMetricForwarder.java
jpbempel: resolved by sending encoded id: `uuid:version`

dd-java-agent/agent-debugger/src/test/java/com/datadog/debugger/agent/SpanDecorationProbeInstrumentationTest.java
jpbempel: received/installed status are verified in smoke tests, but I will add unit test in ConfigurationUpdaterTest in a next PR

dd-java-agent/agent-debugger/src/main/java/com/datadog/debugger/agent/ConfigurationUpdater.java
```diff
@@ -228,22 +230,22 @@ private void retransformClasses(List<Class<?>> classesToBeTransformed) {
...
     LOGGER.debug("Stored appliedDefinitions: {}", appliedDefinitions.values());
   }
 
   // /!\ This is called potentially by multiple threads from the instrumented code /!\
   @Override
-  public ProbeImplementation resolve(String id, Class<?> callingClass) {
-    ProbeDefinition definition = appliedDefinitions.get(id);
-    if (definition == null) {
+  public ProbeImplementation resolve(String encodedProbeId, Class<?> callingClass) {
+    ProbeDefinition definition = appliedDefinitions.get(encodedProbeId);
```
shatzi, line 244: that is interesting... if a probe+version is not found (probe just got update) we are calling retransform on the class?

Aka, if a probe just bumped it version and a old function was still running we will re transform now? maybe we should just ignore and return null if we found updated version?

dd-java-agent/agent-debugger/src/main/java/com/datadog/debugger/agent/StatsdMetricForwarder.java
```diff
@@ -25,45 +27,57 @@ public StatsdMetricForwarder(Config config) {
...
     statsd.distribution(name, value, tags);
+    sendEmittingStatus(encodedProbeId);
   }
 
   @Override
   public void handle(Exception exception) {
     LOGGER.warn("Error when sending metrics: ", exception);
   }
+
+  private void sendEmittingStatus(String probeId) {
```
shatzi, line 80: encodedProbeId

dd-java-agent/agent-debugger/src/main/java/com/datadog/debugger/sink/ProbeStatusSink.java
```diff
@@ -132,7 +146,7 @@ private int enqueueAllProbesStatusIfNeeded(Clock clock) {
   }
 
   public void removeDiagnostics(ProbeId probeId) {
-    probeStatuses.remove(probeId.getId());
+    probeStatuses.remove(probeId.getEncodedId());
```
shatzi, line 149: before this change we have probeStatus per probe ID only. now probeStatus per probe ID+verison. 
Do we delete the old probe status if a probe version got updated?

shatzi: I assume configurationUpdated just keep calling this removeDiagnostics... so it should all just keep to work?

shatzi (approved):
LGTM, just one comment about retransforming classes if probe not found

dd-java-agent/agent-debugger/src/main/java/com/datadog/debugger/sink/ProbeStatusSink.java
jpbempel: yes

dd-java-agent/agent-debugger/src/main/java/com/datadog/debugger/agent/ConfigurationUpdater.java
jpbempel: The problem that we had was the following: Instrumented code was not removed for unknown reason and the code was still trying to resolve an old probe id and failed. So we are trying to retransform the class to remove this old instrumented code.
The new encoded format of the probe id does not change the anything here. An existing probe that is no longer active but still running will trigger the same behavior. Though, it could be problematic in a tight loop. will try to address the issue in a next PR

PR #6366, opened 2023-12-14
Avoid flakiness on webflux 5 tests

amarziali:
# What Does This Do

be less strict on span completion order when using reactive netty behind. Avoid flakiness of some tests

# Motivation

# Additional Notes

Jira ticket: [PROJ-IDENT]


nayeem-kamal (approved):
LGTM

PR #6370, opened 2023-12-14
Track Kafka Avro schemas using Data Streams

piochelepiotr:
# What Does This Do

This PR adds the Avro schema definition as a tag on the span if the schema can be found in the schema field of the sent object.
This is only done if data streams is enabled.

# Motivation

# Additional Notes

Jira ticket: [PROJ-IDENT]

### DIFF
--- a/dd-java-agent/instrumentation/kafka-clients-0.11/src/main/java/datadog/trace/instrumentation/kafka_clients/AvroSchemaExtractor.java
+++ b/dd-java-agent/instrumentation/kafka-clients-0.11/src/main/java/datadog/trace/instrumentation/kafka_clients/AvroSchemaExtractor.java
@@ -9,7 +9,7 @@ import org.apache.kafka.clients.producer.ProducerRecord;
 
 public class AvroSchemaExtractor {
   public static void tryExtractProducer(ProducerRecord record, AgentSpan span) {
-    Integer prio = span.forceSamplingDecision();
+    Integer prio = span.getSamplingPriority();
     if (prio == null || prio <= 0) {
       // don't extract schema if span is not sampled
       return;
--- a/dd-java-agent/instrumentation/kafka-clients-0.11/src/main/java/datadog/trace/instrumentation/kafka_clients/AvroSchemaExtractor.java
+++ b/dd-java-agent/instrumentation/kafka-clients-0.11/src/main/java/datadog/trace/instrumentation/kafka_clients/AvroSchemaExtractor.java
@@ -1,12 +1,14 @@
 package datadog.trace.instrumentation.kafka_clients;
 
+import datadog.trace.api.DDTags;
 import datadog.trace.bootstrap.instrumentation.api.AgentSpan;
 import datadog.trace.bootstrap.instrumentation.api.AgentTracer;
+import datadog.trace.util.FNV64Hash;
 import java.lang.reflect.Field;
 import org.apache.kafka.clients.producer.ProducerRecord;
 
 public class AvroSchemaExtractor {
-  public static void tryExtract(ProducerRecord record, AgentSpan span) {
+  public static void tryExtractProducer(ProducerRecord record, AgentSpan span) {
     Integer prio = span.forceSamplingDecision();
     if (prio == null || prio <= 0) {
       // don't extract schema if span is not sampled
@@ -18,10 +20,13 @@ public class AvroSchemaExtractor {
     }
     String schema = extract(record.value());
     if (schema != null) {
-      span.setTag("messaging.kafka.value_schema.definition", schema);
-      span.setTag("messaging.kafka.value_schema.weight", weight);
-      span.setTag("messaging.kafka.value_schema.type", "avro");
-      span.setTag("messaging.kafka.value_schema.id", schema.hashCode());
+      span.setTag(DDTags.SCHEMA_DEFINITION, schema);
+      span.setTag(DDTags.SCHEMA_WEIGHT, weight);
+      span.setTag(DDTags.SCHEMA_TYPE, "avro");
+      span.setTag(DDTags.SCHEMA_OPERATION, "serialization");
+      span.setTag(
+          DDTags.SCHEMA_ID,
+          Long.toUnsignedString(FNV64Hash.generateHash(schema, FNV64Hash.Version.v1A)));
     }
   }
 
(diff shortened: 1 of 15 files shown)


dd-java-agent/instrumentation/kafka-clients-0.11/src/main/java/datadog/trace/instrumentation/kafka_clients/AvroSchemaExtractor.java
```diff
@@ -0,0 +1,41 @@
+package datadog.trace.instrumentation.kafka_clients;
+
+import datadog.trace.bootstrap.instrumentation.api.AgentSpan;
+import datadog.trace.bootstrap.instrumentation.api.AgentTracer;
+import java.lang.reflect.Field;
+import org.apache.kafka.clients.producer.ProducerRecord;
+
+public class AvroSchemaExtractor {
+  public static void tryExtract(ProducerRecord record, AgentSpan span) {
+    Integer prio = span.forceSamplingDecision();
```
PerfectSlayer, line 10: Is there any drawback to force decision early?

dd-java-agent/instrumentation/kafka-clients-0.11/src/main/java/datadog/trace/instrumentation/kafka_clients/AvroSchemaExtractor.java
```diff
@@ -0,0 +1,41 @@
...
+    String schema = extract(record.value());
+    if (schema != null) {
+      span.setTag("messaging.kafka.value_schema.definition", schema);
+      span.setTag("messaging.kafka.value_schema.weight", weight);
+      span.setTag("messaging.kafka.value_schema.type", "avro");
+      span.setTag("messaging.kafka.value_schema.id", schema.hashCode());
+    }
+  }
+
+  public static String extract(Object value) {
```
PerfectSlayer, line 28: Can the extraction be cached per type? (ie Class instance)

dd-java-agent/instrumentation/kafka-clients-0.11/src/main/java/datadog/trace/instrumentation/kafka_clients/AvroSchemaExtractor.java
```diff
@@ -0,0 +1,41 @@
...
+    int weight = AgentTracer.get().getDataStreamsMonitoring().shouldSampleSchema(record.topic());
+    if (weight == 0) {
+      return;
+    }
+    String schema = extract(record.value());
+    if (schema != null) {
+      span.setTag("messaging.kafka.value_schema.definition", schema);
+      span.setTag("messaging.kafka.value_schema.weight", weight);
+      span.setTag("messaging.kafka.value_schema.type", "avro");
+      span.setTag("messaging.kafka.value_schema.id", schema.hashCode());
```
PerfectSlayer, line 24: What is supposed to be the `messaging.kafka.value_schema.id`? Because `hashCode()` isn't unique.

piochelepiotr: The goal is to tag spans by schemaID, to generate a timeseries of throughput by schema ID, to see when a change occurs

piochelepiotr: What would you recommend using instead?

dd-java-agent/instrumentation/kafka-clients-0.11/src/main/java/datadog/trace/instrumentation/kafka_clients/AvroSchemaExtractor.java
piochelepiotr: So the type might be the same, but the schema might be different right?
https://github.com/DataDog/dd-trace-java/pull/6370/files#diff-21fcc68f4b2d54a50ae1c667af1df7c3b873c5d02cf4a8056d21e4d7c88b1697R65
For example here, the type is `GenericRecord` but many objects with different schemas could use this type.

dd-java-agent/instrumentation/kafka-clients-0.11/src/main/java/datadog/trace/instrumentation/kafka_clients/AvroSchemaExtractor.java
piochelepiotr: I switched to fnv.1a

dd-java-agent/instrumentation/kafka-clients-0.11/src/main/java/datadog/trace/instrumentation/kafka_clients/AvroSchemaExtractor.java
PerfectSlayer: I am not that familiar with Avro but it seems there is also `SpecificRecord` where schema will be fixed according the subtype.
Have to try the Avro code generator and check which parent type is used for the generated types? 
If it is not using `GenericType`, it might be useful to set up a cache. If not, meaning that all generated user messages fall  under generic, it is not.

dd-java-agent/instrumentation/kafka-clients-0.11/src/main/java/datadog/trace/instrumentation/kafka_clients/AvroSchemaExtractor.java
PerfectSlayer: Hey @bantonsson Can you give us some insight here?

bantonsson: So there is a difference between these two things:
* This needs to be on all spans that will be sampled
* I only want to attach it to a span that I know will be sampled

If it's the first, then you need to force the sampling decision, and if it is the second, then you should read the sampling priority with `getSamplingPriority`.

PerfectSlayer: Thanks for the insight Bjorn 🤟

PerfectSlayer (approved):

hokitam (approved):

PR #6376, opened 2023-12-15
Allow comma in request header tag values

### LINKED ISSUE #6148: JSON headers being truncated in span
Hey folks, we're trying to add request headers to our traces. To do that, we're adding environment variables like this
```
      - name: DD_TRACE_REQUEST_HEADER_TAGS
        value: <key>:<value>,<key>:<value>
```
However, one of our headers is JSON that looks like this
```
{"user_agent_ip":<value>,"user_agent":<value>}
```
The span tags don't seem to like JSON headers, and instead of displaying them as normal it seems to be truncating the value at the comma, so that it looks like this in the trace
```
{"user_agent_ip":"<value>"
```
Is there a way to get the trace to not truncate these JSON headers? Thanks in advance!

bantonsson:
# What Does This Do

Allows the trace header tags values to contain commas (previously the value was split at the first comma, and only the first part was used).

It is possible to switch to the old behavior by setting the Java property `-Ddd.trace.request_header.tags.comma.allowed=false` or the environment variable `DD_TRACE_REQUEST_HEADER_TAGS_COMMA_ALLOWED=false`

# Motivation

Splitting at the first value breaks some encodings that use commas in the values. This was also the default behavior for some propagators up until `v1.4.0`.

# Additional Notes

Although it's not possible in HTTP to distinguish between multiple instances of the same header being combined into a comma separate list of the values, and a single header with a value with comma in it, we could be lenient here since the header names to read are user supplied.

Jira ticket: [APMJAVA-1140]
GitHub Ticket: #6148 

[APMJAVA-1140]: https://datadoghq.atlassian.net/browse/APMJAVA-1140?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ

### DIFF
--- a/dd-java-agent/testing/src/main/groovy/datadog/trace/agent/test/base/HttpClientTest.groovy
+++ b/dd-java-agent/testing/src/main/groovy/datadog/trace/agent/test/base/HttpClientTest.groovy
@@ -790,9 +790,11 @@ abstract class HttpClientTest extends VersionedNamingTestBase {
     }
 
     where:
-    method | header                           | value | tags
-    'GET'  | 'X-Datadog-Test-Both-Header'     | 'foo' | [ 'both_header_tag': 'foo' ]
-    'GET'  | 'X-Datadog-Test-Request-Header'  | 'bar' | [ 'request_header_tag': 'bar' ]
+    method | header                           | value     | tags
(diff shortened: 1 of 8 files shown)


ygree (approved):
Looks good to me! Thanks for a quick fix!

PR #6389, opened 2023-12-18
Add time and bucket size boundaries to the tainted map

manuel-alvarez-alvarez:
# What Does This Do
This PR adds two different boundaries to the tainted map implementation:

- **Max age**: limits the max time an entry can stay in the map. A bug in the code can cause us to create strong references to the original values, in that case, their respective tainted values will stay forever in the map causing memory issues. This limit will ensure that this case does not happen.
- **Max bucket size**: when adding tainted values to the same bucket, the performance can quickly degrade due to the linked list nature of the map. This PR limits the max size of each bucket to prevent the performance degradation.

# Motivation
Maps of tainted objects can grow a lot (specially if we move to a global tainted dictionary strategy). This PR ensures that we put in place the required boundaries to enforce a correct memory footprint and a good enough performance.

# Additional Notes

Jira ticket: [APPSEC-11483]

[APPSEC-11483]: https://datadoghq.atlassian.net/browse/APPSEC-11483?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ

### DIFF
--- a/dd-java-agent/agent-iast/src/main/java/com/datadog/iast/taint/TaintedMap.java
+++ b/dd-java-agent/agent-iast/src/main/java/com/datadog/iast/taint/TaintedMap.java
@@ -1,10 +1,16 @@
 package com.datadog.iast.taint;
 
 import com.datadog.iast.IastSystem;
+import datadog.trace.api.Config;
+import datadog.trace.api.iast.telemetry.IastMetric;
+import datadog.trace.api.iast.telemetry.IastMetricCollector;
+import datadog.trace.api.iast.telemetry.Verbosity;
+import datadog.trace.util.AgentTaskScheduler;
 import java.util.Arrays;
 import java.util.Collections;
 import java.util.Iterator;
 import java.util.NoSuchElementException;
+import java.util.concurrent.TimeUnit;
 import java.util.concurrent.atomic.AtomicLong;
 import javax.annotation.Nonnull;
 import javax.annotation.Nullable;
@@ -28,6 +34,7 @@ import org.slf4j.LoggerFactory;
  *   <li>Entries SHOULD be removed when key objects are garbage-collected.
  *   <li>All operations MUST NOT throw, even with concurrent access and modification.
  *   <li>Put operations MAY be lost under concurrent modification.
+ *   <li>Entries MUST NOT stay in the map forever
  * </ol>
  *
  * <p><i>Capacity</i> is fixed, so there is no rehashing.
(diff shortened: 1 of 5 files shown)


dd-java-agent/agent-iast/src/main/java/com/datadog/iast/taint/TaintedObject.java
```diff
@@ -16,6 +16,12 @@ public class TaintedObject extends WeakReference<Object> {
   @Nullable TaintedObject next;
   private Range[] ranges;
 
+  /** generation of the tainted for max age purging purposes */
+  boolean generation;
+
+  /** true if the tainted caused the bucket to be emptied and flatten on insertion */
+  boolean flatten;
```
smola, line 23: Emitting a metric when this happens makes sense, but do we really need to store the boolean here?

smola (approved):

dd-java-agent/agent-iast/src/main/java/com/datadog/iast/taint/TaintedObject.java
manuel-alvarez-alvarez: Changed it!, I did it to keep it consistent with a tainted objects being in charge of collecting metrics, but being this an exceptional situation the performance implications are not that important.

dd-java-agent/agent-iast/src/main/java/com/datadog/iast/taint/TaintedMap.java
```diff
@@ -43,8 +50,20 @@ public interface TaintedMap extends Iterable<TaintedObject> {
   /** Bitmask to convert hashes to positive integers. */
   int POSITIVE_MASK = Integer.MAX_VALUE;
 
+  /** Max allowed size for the linked list inside a bucket */
+  int DEFAULT_MAX_BUCKET_SIZE = 10;
+
+  /** Max age of entries contained in the map (worst case will be {@code 2 * maxAge}) */
+  int DEFAULT_MAX_AGE = 30;
+
+  TimeUnit DEFAULT_MAX_AGE_UNIT = TimeUnit.MINUTES;
```
DDJavierSantos, line 59: This looks too long. Do we have any benchmarks with the average age?

DDJavierSantos (approved):

dd-java-agent/agent-iast/src/main/java/com/datadog/iast/taint/TaintedMap.java
manuel-alvarez-alvarez: No, but I guess an average age would not be more than a few seconds. This is just a precaution, entries should be automatically removed when the are released by the GC, if we have a bug and keep strong references it should remove those entries from the map allowing the GC to free those references.

PR #6393, opened 2023-12-19
fix logging dependencies for spring-messaging tests

amarziali:
# What Does This Do

Help understanding failures of `SpringListenerSQSTest`. It fixes slf4j dependences that today are too recent and fails binding logback factory. 

# Motivation

# Additional Notes

Jira ticket: [PROJ-IDENT]

### DIFF
--- a/dd-java-agent/instrumentation/spring-messaging-4/build.gradle
+++ b/dd-java-agent/instrumentation/spring-messaging-4/build.gradle
@@ -29,8 +29,12 @@ dependencies {
   testImplementation project(':dd-java-agent:instrumentation:aws-java-sqs-2.0')
 
   testImplementation group: 'jakarta.annotation', name: 'jakarta.annotation-api', version: '2.0.0'
-  testImplementation group: 'io.awspring.cloud', name: 'spring-cloud-aws-sqs', version: '3.0.1'
+  testImplementation group: 'io.awspring.cloud', name: 'spring-cloud-aws-sqs', version: '3.0.1', {
+    exclude group: 'org.slf4j', module: 'slf4j-api'
+  }
   testImplementation group: 'org.elasticmq', name: 'elasticmq-rest-sqs_2.13', version: '1.2.3'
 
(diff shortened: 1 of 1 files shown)


dd-java-agent/instrumentation/spring-messaging-4/src/test/groovy/SpringListenerSQSTest.groovy
```diff
@@ -22,6 +22,11 @@ class SpringListenerSQSTest extends AgentTestRunner {
     injectSysConfig(GeneralConfig.SERVICE_NAME, "my-service")
   }
 
+  @Override
+  boolean useStrictTraceWrites() {
+    false
```
bantonsson, line 27: This is basically saying that the instrumentation is broken, and we might not close some spans/scopes.

amarziali: if we do not close the span is not reported. I put this because some spans might be closed after the root span is closed (due to async misbehaviour). Without this if a span is enqueued after the root span, it's discarded failing the test. Usually this setting is false on the runtime case:

```
    traceStrictWritesEnabled = configProvider.getBoolean(TRACE_STRICT_WRITES_ENABLED, false);
```

bantonsson: Indeed it is false in the runtime case, but the reason that the strict mode was implemented was that we kept leaking `Continuations` and `Scopes` and the instrumentations where an even bigger mess. So the real question is why there are `Continuations` that have not been canceled or child `Spans` that have not been finished when the _root_ span is finished.

amarziali: got it thanks! I only kept the commit for the gradle deps so we can have now more logs to investigate

bantonsson (approved):

PR #6399, opened 2023-12-20
Use OpenTelemetry SDK context key instead of Datadog proprietary one for local root span

PerfectSlayer:
# What Does This Do

This PR changes the Context key to retrieve the local root span from OpenTelemetry `Context`.
It will now use the same key as the OpenTelemetry SDK making it compatible with SDK `LocalRootSpan` behavior. 

# Motivation

This will increase compatibility with OTel SDK and OTel instrumenter API. 

# Additional Notes

This behavior change is following the gap analysis of Micronaut Tracing features.

The related documentation update PR: https://github.com/DataDog/documentation/pull/21122

Jira ticket: [APMJAVA-1151]

[APMJAVA-1151]: https://datadoghq.atlassian.net/browse/APMJAVA-1151?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ

### DIFF
--- a/dd-java-agent/instrumentation/opentelemetry/opentelemetry-1.4/src/main/java/datadog/trace/instrumentation/opentelemetry14/context/OtelContext.java
+++ b/dd-java-agent/instrumentation/opentelemetry/opentelemetry-1.4/src/main/java/datadog/trace/instrumentation/opentelemetry14/context/OtelContext.java
@@ -15,7 +15,7 @@ public class OtelContext implements Context {
   public static final OtelContext ROOT = new OtelContext(OtelSpan.invalid(), OtelSpan.invalid());
 
   private static final String OTEL_CONTEXT_SPAN_KEY = "opentelemetry-trace-span-key";
-  private static final String DATADOG_CONTEXT_ROOT_SPAN_KEY = "datadog-root-span-key";
+  private static final String OTEL_CONTEXT_ROOT_SPAN_KEY = "opentelemetry-traces-local-root-span";
 
   private final Span currentSpan;
   private final Span rootSpan;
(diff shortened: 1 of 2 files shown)


dd-java-agent/instrumentation/opentelemetry/opentelemetry-1.4/src/main/java/datadog/trace/instrumentation/opentelemetry14/context/OtelContext.java
```diff
@@ -15,7 +15,7 @@ public class OtelContext implements Context {
   public static final OtelContext ROOT = new OtelContext(OtelSpan.invalid(), OtelSpan.invalid());
 
   private static final String OTEL_CONTEXT_SPAN_KEY = "opentelemetry-trace-span-key";
-  private static final String DATADOG_CONTEXT_ROOT_SPAN_KEY = "datadog-root-span-key";
```
bantonsson, line 18: Should we keep this one around a while for backwards compatibility?

PerfectSlayer: As the feature is still in Beta, it is not the first time we break things (span naming, span type, etc...) as we move forward and don't add backward compatibility feature flag.
Moreover, Context support is already documented as a WIP 🤷

bantonsson: ok

bantonsson (approved):

PR #6406, opened 2023-12-20
Add specific support for Optional

jpbempel:
# What Does This Do
We access directly to the value of an Optional through the `java.util.Optional` API rather than using the reflection as since JDK 16 it will be prohibited.

# Motivation
Avoid using reflection on some JDK classes that is prohibited since JDK 16

# Additional Notes

Jira ticket: [DEBUG-1980]

[DEBUG-1980]: https://datadoghq.atlassian.net/browse/DEBUG-1980?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ

### DIFF
--- a/dd-java-agent/agent-debugger/debugger-bootstrap/src/main/java/datadog/trace/bootstrap/debugger/CapturedContext.java
+++ b/dd-java-agent/agent-debugger/debugger-bootstrap/src/main/java/datadog/trace/bootstrap/debugger/CapturedContext.java
@@ -8,6 +8,7 @@ import datadog.trace.bootstrap.debugger.el.ValueReferences;
 import datadog.trace.bootstrap.debugger.el.Values;
 import datadog.trace.bootstrap.debugger.util.Redaction;
 import datadog.trace.bootstrap.debugger.util.TimeoutChecker;
+import datadog.trace.bootstrap.debugger.util.WellKnownClasses;
 import java.util.ArrayList;
 import java.util.Collections;
 import java.util.HashMap;
@@ -15,6 +16,7 @@ import java.util.LinkedHashMap;
 import java.util.List;
 import java.util.Map;
 import java.util.Objects;
+import java.util.function.Function;
 
 /** Stores different kind of data (arguments, locals, fields, exception) for a specific location */
 public class CapturedContext implements ValueReferenceResolver {
@@ -127,7 +129,18 @@ public class CapturedContext implements ValueReferenceResolver {
         }
       }
     } else {
-      target = ReflectiveFieldValueResolver.resolve(target, target.getClass(), memberName);
+      Function<Object, WellKnownClasses.SpecialField> specialFieldAccess =
+          WellKnownClasses.getSpecialFieldAccess(target.getClass().getTypeName());
+      if (specialFieldAccess != null) {
+        WellKnownClasses.SpecialField specialField = specialFieldAccess.apply(target);
+        if (specialField != null && specialField.getName().equals(memberName)) {
+          return specialField.getValue();
+        } else {
+          target = Values.UNDEFINED_OBJECT;
+        }
+      } else {
+        target = ReflectiveFieldValueResolver.resolve(target, target.getClass(), memberName);
+      }
     }
     checkUndefined(target, memberName, "Cannot dereference to field: ");
     return target;
--- a/dd-java-agent/agent-debugger/debugger-bootstrap/src/main/java/datadog/trace/bootstrap/debugger/CapturedContext.java
+++ b/dd-java-agent/agent-debugger/debugger-bootstrap/src/main/java/datadog/trace/bootstrap/debugger/CapturedContext.java
@@ -129,10 +129,10 @@ public class CapturedContext implements ValueReferenceResolver {
         }
       }
     } else {
-      Function<Object, WellKnownClasses.SpecialField> specialFieldAccess =
+      Function<Object, CapturedValue> specialFieldAccess =
           WellKnownClasses.getSpecialFieldAccess(target.getClass().getTypeName());
       if (specialFieldAccess != null) {
-        WellKnownClasses.SpecialField specialField = specialFieldAccess.apply(target);
+        CapturedValue specialField = specialFieldAccess.apply(target);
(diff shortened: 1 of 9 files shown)


dd-java-agent/agent-debugger/debugger-bootstrap/src/main/java/datadog/trace/bootstrap/debugger/util/WellKnownClasses.java
```diff
@@ -55,6 +58,12 @@ public class WellKnownClasses {
               "java.time.LocalDateTime",
               "java.util.UUID"));
 
+  private static Map<String, Function<Object, SpecialField>> specialFields = new HashMap<>();
```
shatzi, line 61: I wonder why the map is for a type -> single field.
I think it would be better to have a map for [type,field-name] -> Func<SpecicalField>.
Two things:
1. we can now just inject a special access to a single field in an type.
2. support special types with multiple fields

One more thing we can do is convert the return value to be array of fields instead of a single field. but I like the first direction better.

dd-java-agent/agent-debugger/debugger-bootstrap/src/main/java/datadog/trace/bootstrap/debugger/CapturedContext.java
```diff
@@ -127,7 +129,18 @@ public Object getMember(Object target, String memberName) {
         }
       }
     } else {
-      target = ReflectiveFieldValueResolver.resolve(target, target.getClass(), memberName);
+      Function<Object, WellKnownClasses.SpecialField> specialFieldAccess =
+          WellKnownClasses.getSpecialFieldAccess(target.getClass().getTypeName());
+      if (specialFieldAccess != null) {
+        WellKnownClasses.SpecialField specialField = specialFieldAccess.apply(target);
+        if (specialField != null && specialField.getName().equals(memberName)) {
```
shatzi, line 136: I find this wired that we first try to access the "special field" even before we check it name. see other suggestion on the getSpecialFieldAccess should also get field name and not just type.

dd-java-agent/agent-debugger/debugger-bootstrap/src/main/java/datadog/trace/bootstrap/debugger/util/WellKnownClasses.java
```diff
@@ -98,7 +107,47 @@ public static boolean isSizeSafe(Map<?, ?> map) {
...
+   */
   public static boolean isStringPrimitive(String type) {
     return stringPrimitives.contains(type);
   }
+
+  /**
+   * @return a function to access special field of a type, or null if type is not supported. This is
+   *     used to avoid using reflection to access fields on well known types
+   */
+  public static Function<Object, SpecialField> getSpecialFieldAccess(String type) {
```
shatzi, line 122: one thing that I'm not sure how this works is how about inheritance. assume someone derived from Optional, will this stop working and we would use reflection?

dd-java-agent/agent-debugger/debugger-bootstrap/src/main/java/datadog/trace/bootstrap/debugger/util/WellKnownClasses.java
```diff
@@ -98,7 +107,47 @@ public static boolean isSizeSafe(Map<?, ?> map) {
...
+    return specialFields.get(type);
+  }
+
+  private static SpecialField optionalSpecialField(Object o) {
+    return new SpecialField("value", Object.class.getTypeName(), ((Optional<?>) o).orElse(null));
+  }
+
+  public static class SpecialField {
```
shatzi, line 130: why not use CaptureValue? seems very similar...

dd-java-agent/agent-debugger/src/test/java/com/datadog/debugger/agent/SnapshotSerializationTest.java
```diff
@@ -294,6 +296,8 @@ static class WellKnownClasses {
     UUID uuid = UUID.nameUUIDFromBytes("foobar".getBytes());
     AtomicLong atomicLong = new AtomicLong(123);
     URI uri = URI.create("https://www.datadoghq.com");
+    Optional<Date> maybeUiid = Optional.of(new Date());
```
shatzi, line 299: maybeDate?

shatzi (commented):
I like the direction of having a special handles for fetching values from types, I think I want to ensure this done in a way we can expand it.

dd-java-agent/agent-debugger/debugger-bootstrap/src/main/java/datadog/trace/bootstrap/debugger/util/WellKnownClasses.java
jpbempel: Optional is final

dd-java-agent/agent-debugger/debugger-bootstrap/src/main/java/datadog/trace/bootstrap/debugger/util/WellKnownClasses.java
jpbempel: true. done.

dd-java-agent/agent-debugger/debugger-bootstrap/src/main/java/datadog/trace/bootstrap/debugger/util/WellKnownClasses.java
jpbempel: there is so far no need to handle multiple fields for this mechanism.

dd-java-agent/agent-debugger/src/test/java/com/datadog/debugger/agent/SnapshotSerializationTest.java
jpbempel: done

shatzi (approved):

PR #6414, opened 2023-12-22
Capture the SELinux status as a profiler setting event

jbachorik:
# What Does This Do
It will attempt to get the SELinux status via `getenforce` call and record it as `datadog.ProfilerSetting` event with name of `SELinux Status`.

# Motivation
Provide help in solving support tickets - having SELinux in effect with rules affecting eg. the temp location can negatively affect profiler.

# Additional Notes

Jira ticket: [PROF-8837]

[PROF-8837]: https://datadoghq.atlassian.net/browse/PROF-8837?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ


dd-java-agent/agent-profiling/profiling-controller/src/main/java/com/datadog/profiling/controller/ProfilerSettingsSupport.java
```diff
@@ -103,6 +108,28 @@ protected ProfilerSettingsSupport(ConfigProvider configProvider) {
...
+  private String getSELinuxStatus() {
+    String value = "Not present";
+    if (Platform.isLinux()) {
+      try {
+        Process process = Runtime.getRuntime().exec("getenforce");
+        try (BufferedReader reader =
+            new BufferedReader(new InputStreamReader(process.getInputStream()))) {
+          String line = reader.readLine();
+
+          if (line != null) {
```
bantonsson, line 124: Is it ok to return `null`?

richardstartin (approved):

dd-java-agent/agent-profiling/profiling-controller/src/main/java/com/datadog/profiling/controller/ProfilerSettingsSupport.java
jbachorik: I assume you are asking about the `line` value?
If yes, then `null` returned by `readLine` is used as an indicator of EOF (https://docs.oracle.com/javase/8/docs/api/java/io/BufferedReader.html#readLine--)

Otherwise, the `getSELinuxStatus` will never return `null` - the return value is initialized to `Not present`.

bantonsson: What I meant to ask was: If `readLine()` returns `null`, then `line` will be `null`, and this method will return `null`. Is that ok?

jbachorik: Ok. I am a bit confused here because it should not return null.

Let's step over the execution.

1. The `value` is initialized to `Not present`
2. `getenforce` is executed and its output is processed via a `BufferedReader`
3. A line is read from the `BufferedReader` via `readLine()` and stored in `line` variable
4. If the `line` variable is not null it will be copied over to `value`
5. If exception happens during the execution or reading the output the `value` will not be changed
6. The `value`, containing either the output of `getenforce` or `Not present` is returned

Does it sound ok? I apologise, I am not trying to be difficult but I am failing to see where `null` can be returned when `line` is null.

jbachorik: Hm, it won't return null but your instinct is right - the code should be resilient to `getenforce` not returning within a reasonable amount of time. Let me try improving that - it would be really nasty to hang the agent because of misbihaving external tool ...

bantonsson (approved):

PR #6417, opened 2023-12-26
Use connection pool for http communications with extension.

purple4reina:
# What Does This Do

Instead of creating a new tcp connection with each call to the extension (ie start and end invocation), use a pool of available connections.

# Motivation

This change reduces `aws.lambda.enhanced.runtime_duration` by approximately 2%.

Screenshot 2023-12-26 at 2 00 50 PM

See https://ddserverless.datadoghq.com/notebook/2987494/reys-purple-notebook?range=3600000&view=view-mode&start=1703617009962&live=false

# Additional Notes

Jira ticket: [PROJ-IDENT]

### DIFF
--- a/dd-trace-core/src/main/java/datadog/trace/lambda/LambdaHandler.java
+++ b/dd-trace-core/src/main/java/datadog/trace/lambda/LambdaHandler.java
@@ -15,6 +15,7 @@ import datadog.trace.core.propagation.PropagationTags;
 import java.io.ByteArrayInputStream;
 import java.nio.charset.StandardCharsets;
 import java.util.Base64;
+import okhttp3.ConnectionPool;
 import okhttp3.MediaType;
 import okhttp3.OkHttpClient;
 import okhttp3.Request;
@@ -49,6 +50,8 @@ public class LambdaHandler {
   private static final String END_INVOCATION = "/lambda/end-invocation";
 
   private static final Long REQUEST_TIMEOUT_IN_S = 1L;
+  private static final int MAX_IDLE_CONNECTIONS = 5;
+  private static final Long KEEP_ALIVE_DURATION = 300L;
 
   private static OkHttpClient HTTP_CLIENT =
       new OkHttpClient.Builder()
@@ -57,6 +60,7 @@ public class LambdaHandler {
           .writeTimeout(REQUEST_TIMEOUT_IN_S, SECONDS)
           .readTimeout(REQUEST_TIMEOUT_IN_S, SECONDS)
           .callTimeout(REQUEST_TIMEOUT_IN_S, SECONDS)
+          .connectionPool(new ConnectionPool(MAX_IDLE_CONNECTIONS, KEEP_ALIVE_DURATION, SECONDS))
           .build();
 
   private static final MediaType jsonMediaType = MediaType.parse("application/json");


dd-trace-core/src/main/java/datadog/trace/lambda/LambdaHandler.java
```diff
@@ -57,6 +58,7 @@ public class LambdaHandler {
           .writeTimeout(REQUEST_TIMEOUT_IN_S, SECONDS)
           .readTimeout(REQUEST_TIMEOUT_IN_S, SECONDS)
           .callTimeout(REQUEST_TIMEOUT_IN_S, SECONDS)
+          .connectionPool(new ConnectionPool(5, 300, SECONDS))
```
purple4reina, line 61: 5 connections, 300 second keep-alive.

duncanista: Where do the magic numbers come from?

purple4reina: Thin air. I mostly just made them up. Very happy to take any suggestions for edits though.

I figured we'll only be creating two connections, one for start and one for end. I wanted the number of connections in this pool to be a bit larger than the number we're expecting to need. This is why I chose 5.

The keep alive value of 5 minutes also just felt like it was a good amount of time to keep a connection alive. I wanna say that lambda only keeps the container alive for like 6 minutes, though I could very well be wrong.

duncanista: Maybe we could define a constant for them?

Sounds good, I'd verify with Darcy or AJ about the max time a lambda container will be alive, AFAIK, it can be longer than that (15 minutes(?) I'm not pretty sure.

joeyzhao2018: If this client is `static`, does it mean it's reused for multiple invocation? Can we test on how many connections it need by firing a bunch a requests (> 5) within a short period of time?

purple4reina: @duncanista Good idea about adding the constants. And we can't merge this until next week anyway, so I'll be sure to ask @DarcyRaynerDD what he thinks before we merge.

purple4reina: @joeyzhao2018 the test I performed yesterday (with the image and link to notebook above) was using a simple hello world java function executed 10x/sec. I looked to see if there was a way to log when a new connection is made, but while this functionality exists in later versions of `okhttp`, it doesn't in the one we're using. (We actually vendor the package, see https://github.com/DataDog/okhttp/blob/java7/okhttp/src/main/java/okhttp3/ConnectionPool.java)

So, I don't know how else to confirm this. But, I supposed that if we're seeing all the data appear as expected and that the average runtime duration is less than it was, then this seems like a good thing.

joeyzhao2018 (approved):
:shipit:

dd-trace-core/src/main/java/datadog/trace/lambda/LambdaHandler.java
```diff
@@ -57,6 +60,7 @@ public class LambdaHandler {
           .writeTimeout(REQUEST_TIMEOUT_IN_S, SECONDS)
           .readTimeout(REQUEST_TIMEOUT_IN_S, SECONDS)
           .callTimeout(REQUEST_TIMEOUT_IN_S, SECONDS)
+          .connectionPool(new ConnectionPool(MAX_IDLE_CONNECTIONS, KEEP_ALIVE_DURATION, SECONDS))
```
purple4reina, line 63: @DarcyRaynerDD do you have any opinions on the values we choose here?

dd-trace-core/src/main/java/datadog/trace/lambda/LambdaHandler.java
dougqh: Unfortunately, we're stuck on an old version of okhttp because the new version depends on kotlin.
But sounds like you've done your diligence.
Thanks

PR #6418, opened 2023-12-27
Add install_signature to app-started

ygree:
# What Does This Do

Adds `install_signature` to the `app-started` event

# Motivation

# Additional Notes

Jira ticket: [AIT-9228]

[AIT-9228]: https://datadoghq.atlassian.net/browse/AIT-9228?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ

### DIFF
--- a/telemetry/src/main/java/datadog/telemetry/TelemetryRequestBody.java
+++ b/telemetry/src/main/java/datadog/telemetry/TelemetryRequestBody.java
@@ -280,6 +280,19 @@ public class TelemetryRequestBody extends RequestBody {
     bodyWriter.endObject();
   }
 
+  public void writeInstallSignature(String installId, String installType, String installTime)
+      throws IOException {
+    if (installId == null && installType == null && installTime == null) {
+      return;
+    }
+    bodyWriter.name("install_signature");
+    bodyWriter.beginObject();
+    bodyWriter.name("install_id").value(installId);
+    bodyWriter.name("install_type").value(installType);
+    bodyWriter.name("install_time").value(installTime);
+    bodyWriter.endObject();
+  }
+
   @Nullable
   @Override
   public MediaType contentType() {
--- a/telemetry/src/main/java/datadog/telemetry/TelemetryRequest.java
+++ b/telemetry/src/main/java/datadog/telemetry/TelemetryRequest.java
@@ -27,7 +27,7 @@ public class TelemetryRequest {
   private final boolean debug;
   private final TelemetryRequestBody requestBody;
 
-  public TelemetryRequest(
+  TelemetryRequest(
       EventSource eventSource,
       EventSink eventSink,
       long messageBytesSoftLimit,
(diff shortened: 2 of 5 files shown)


telemetry/src/main/java/datadog/telemetry/TelemetryRequestBody.java
```diff
@@ -280,6 +280,19 @@ public void writeProducts(boolean appsecEnabled, boolean profilerEnabled) throws
...
+  public void writeInstallSignature(String installId, String installType, String installTime)
+      throws IOException {
+    if (installId == null && installType == null && installTime == null) {
+      return;
+    }
+    bodyWriter.name("install_signature");
+    bodyWriter.beginObject();
+    bodyWriter.name("install_id").value(installId);
+    bodyWriter.name("install_type").value(installType);
+    bodyWriter.name("install_time").value(installTime);
```
bantonsson, line 292: It's not obvious for me from the doc what the protocol expects and from the tests if this will serialize a `null` value as skipping the field, the empty `""` string, or the string `"null"`.

yshapiro-57: @bantonsson from the backend perspective, `{"install_id":null}`, `{"install_id":""}` and `{}` (no field in the JSON at all) will all be handled the same way and correctly. The only thing we do not want is `{"install_id":"null"}`, but looking at the tests, I think this is not going to be an issue.

yshapiro-57 (approved):

telemetry/src/main/java/datadog/telemetry/TelemetryRequestBody.java
ygree: It will skip `null`s during serialization. The tests verifies it with this assertion:

```groovy
testHttpClient.assertRequestBody(RequestType.APP_STARTED).assertPayload().installSignature(installId, installType, installTime)
```

It builds a map of expected non-null fields to compare against the deserialized `install_signature':

```groovy
    void installSignature(String installId, String installType, String installTime) {
      if (installId == null && installType == null && installTime == null) {
        assert this.payload['install_signature'] == null
        return
      }
      LinkedHashMap<String, String> expected = [:]
      if (installId != null) {
        expected.put("install_id", installId)
      }
      if (installType != null) {
        expected.put("install_type", installType)
      }
      if (installTime != null) {
        expected.put("install_time", installTime)
      }
      assert this.payload['install_signature'] == expected
    }
```

bantonsson: Indeed the test code builds a map only containing non-null values, but using the same library to serialize to and then deserialize from JSON really doesn't say much about how the JSON looks. I'm just being difficult I know. Let's assume that moshi does _the right thing™_

bantonsson (approved):

PR #6419, opened 2023-12-27
Introduce special toString for well known classes

jpbempel:
# What Does This Do
`toString` for `Class` add prefix of class or interface while we expect just he name of the class. When serializing a class name we use a special toString in this case.

# Motivation

# Additional Notes

Jira ticket: [DEBUG-1947]

[DEBUG-1947]: https://datadoghq.atlassian.net/browse/DEBUG-1947?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ


dd-java-agent/agent-debugger/src/main/java/com/datadog/debugger/util/MoshiSnapshotHelper.java
```diff
@@ -328,8 +329,12 @@ public void string(String value, boolean isComplete, int originalLength) throws
       @Override
       public void primitiveValue(Object value) throws Exception {
         jsonWriter.name(VALUE);
-        if (WellKnownClasses.isToStringSafe(value.getClass().getTypeName())) {
-          jsonWriter.value(String.valueOf(value));
+        String typeName = value.getClass().getTypeName();
+        if (WellKnownClasses.isToStringSafe(typeName)) {
```
shatzi, line 333: Nit - to simply this logic and ensure our toString works everywhere the same is to merge isToStringSafe and getSpecialToString.

For example, add all primitive classes to specialToString map and then converting to string is always the same: try to fetch the class from this map and call that function to convert it into string. then we can rename it to safeToStringFunctions to achieve both.

shatzi (approved):

dd-java-agent/agent-debugger/debugger-bootstrap/src/main/java/datadog/trace/bootstrap/debugger/util/WellKnownClasses.java
```diff
@@ -125,8 +131,20 @@ public static Function<Object, CapturedContext.CapturedValue> getSpecialFieldAcc
...
+  public static Function<Object, String> getSpecialToString(String type) {
+    return specialToString.get(type);
+  }
+
   private static CapturedContext.CapturedValue optionalSpecialField(Object o) {
     return CapturedContext.CapturedValue.of(
         "value", Object.class.getTypeName(), ((Optional<?>) o).orElse(null));
   }
+
+  private static String classToString(Object o) {
```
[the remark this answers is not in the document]
jpbempel: fixed

dd-java-agent/agent-debugger/src/main/java/com/datadog/debugger/util/MoshiSnapshotHelper.java
jpbempel: done

dd-java-agent/agent-debugger/debugger-el/src/main/java/com/datadog/debugger/el/Value.java
```diff
@@ -43,6 +46,10 @@ default boolean isNull() {
   }
 
   static Value<?> of(Object value) {
+    return of(value, null);
+  }
+
+  static Value<?> of(Object value, Expression<?> fromExpr) {
```
shatzi, line 52: maybe we should split this PR into just toString and fix the redaction refactor on another PR.

shatzi (approved):
Nice catch on redacting of Value.of expressions

PR #6423, opened 2023-12-29
Remove retransformation on failed resolution

jpbempel:
# What Does This Do
Previously, in case of failed resolution from probe id we retransformed the current class to make sure the code is in sync with probe definitions. if a loop was still executing instrumented code, this can lead to a storm of retransformation which can hurt significantly performance of the application (retransform is STW). we now do nothing except log a warning and send it as telemetry.

# Motivation
avoid performance issue in rare cases

# Additional Notes

Jira ticket: [DEBUG-2022]

[DEBUG-2022]: https://datadoghq.atlassian.net/browse/DEBUG-2022?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ

### DIFF
--- a/dd-java-agent/agent-debugger/src/main/java/com/datadog/debugger/agent/ConfigurationUpdater.java
+++ b/dd-java-agent/agent-debugger/src/main/java/com/datadog/debugger/agent/ConfigurationUpdater.java
@@ -1,5 +1,7 @@
 package com.datadog.debugger.agent;
 
+import static datadog.trace.api.telemetry.LogCollector.SEND_TELEMETRY;
+
 import com.datadog.debugger.instrumentation.InstrumentationResult;
 import com.datadog.debugger.probe.LogProbe;
 import com.datadog.debugger.probe.MetricProbe;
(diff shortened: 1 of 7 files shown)


dd-java-agent/agent-debugger/src/main/java/com/datadog/debugger/agent/ConfigurationUpdater.java
```diff
@@ -240,15 +242,10 @@ private void storeDebuggerDefinitions(ConfigurationComparer changes) {
...
     ProbeDefinition definition = appliedDefinitions.get(encodedProbeId);
-    if (definition == null && callingClass != null) {
-      LOGGER.info(
-          "Cannot resolve probe id={}, re-transforming calling class: {}",
-          encodedProbeId,
-          callingClass.getName());
-      retransformClasses(Collections.singletonList(callingClass));
-      return null;
+    if (definition == null) {
+      LOGGER.warn(SEND_TELEMETRY, "Cannot resolve probe id=" + encodedProbeId);
```
shatzi, line 248: does LOGGER.warn is sent to instrumentation telemetry? or that would be on a different PR

dd-java-agent/agent-debugger/src/main/java/com/datadog/debugger/agent/ConfigurationUpdater.java
```diff
@@ -240,15 +242,10 @@ private void storeDebuggerDefinitions(ConfigurationComparer changes) {
 
   // /!\ This is called potentially by multiple threads from the instrumented code /!\
   @Override
-  public ProbeImplementation resolve(String encodedProbeId, Class<?> callingClass) {
+  public ProbeImplementation resolve(String encodedProbeId) {
```
shatzi, line 245: nice cleanup.

shatzi (approved):

dd-java-agent/agent-debugger/src/main/java/com/datadog/debugger/agent/ConfigurationUpdater.java
jpbempel: yes it is sent only if `DD_TELEMETRY_LOG_COLLECTION_ENABLED` is set to true for the moment.
it will be turned to true by default once they assess the volume of log sent with that

PR #6425, opened 2023-12-29
Add `instanceof` operator in Expression Language

jpbempel:
# What Does This Do
Add comparison operator `instanceof` to test object value against a class type provided as string.
class type is looked up using `Class::forName` and the classloader of the left value.

# Motivation

# Additional Notes

Jira ticket: [DEBUG-1979]

[DEBUG-1979]: https://datadoghq.atlassian.net/browse/DEBUG-1979?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ

### DIFF
--- a/dd-java-agent/agent-debugger/debugger-el/src/main/java/com/datadog/debugger/el/expressions/ComparisonOperator.java
+++ b/dd-java-agent/agent-debugger/debugger-el/src/main/java/com/datadog/debugger/el/expressions/ComparisonOperator.java
@@ -1,5 +1,6 @@
 package com.datadog.debugger.el.expressions;
 
+import com.datadog.debugger.el.EvaluationException;
 import com.datadog.debugger.el.Value;
 import com.datadog.debugger.el.Visitor;
 import com.datadog.debugger.el.values.NumericValue;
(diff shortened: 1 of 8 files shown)


dd-java-agent/agent-debugger/debugger-el/src/main/java/com/datadog/debugger/el/expressions/ComparisonOperator.java
```diff
@@ -65,6 +66,23 @@ public Boolean apply(Value<?> left, Value<?> right) {
...
     }
+  },
+  INSTANCEOF("instanceof") {
+    @Override
+    public Boolean apply(Value<?> left, Value<?> right) {
+      if (right.getValue() instanceof String) {
+        String typeStr = (String) right.getValue();
+        Class<?> clazz;
+        try {
+          clazz = Class.forName(typeStr, false, left.getValue().getClass().getClassLoader());
```
shatzi, line 77: very cool idea.

One thing that I'm worried about this is side effects of loading the class on the right based on their name. 
This can cause the system to load arbitrary classes and run static ctors. 

Is that right? is that something we need to check? do we want to limit this to only loaded classes?

shatzi (approved):
just one comment about side effects.

dd-java-agent/agent-debugger/debugger-el/src/main/java/com/datadog/debugger/el/expressions/ComparisonOperator.java
jpbempel: Clearly this could be a problem that's why I mention it in the description of this PR.
if we want to limit to loaded classes, it's not just the list of loaded classes but also per classloader. not that trivial to put in place this limitation

jpbempel: Could use `Instrumentation::getInitiatedClasses(ClassLoader)` for that but need to do a linear scanning to look for the already loaded class.

jpbempel: Will merge like this but will explore how to leverage `Instrumentation::getInititedClasses` in a next PR

jpbempel: Update: I have carefully use the initialize=false parameter when calling `Class.forName` to avoid this. Not initializing the class will not call static initializers (static blocks)

PR #6429, opened 2024-01-02
Fix support of string primitives for index in EL

jpbempel:
# What Does This Do
add unit tests for redacted values and types

# Motivation

# Additional Notes


shatzi (approved):
one line fix with 100 lines of tests coverage :D love it.

PR #6436, opened 2024-01-03
Add finally block for `@duration` metric

jpbempel:
# What Does This Do
To make sure we measure the duration of method using `@duration` metric value in case of exception, we add a finally block with the metric call

# Motivation

# Additional Notes

Jira ticket: [DEBUG-1946]

[DEBUG-1946]: https://datadoghq.atlassian.net/browse/DEBUG-1946?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ


shatzi (commented):
I think we need to refactor this little more without side effects.

shatzi (approved):

PR #6444, opened 2024-01-05
Enforce ordering of probe instrumentation

jpbempel:
# What Does This Do
To make sure on a same method all probe defined on this location are instrumented in right order to interact each other. Order of instrumentation is:
 - metric probes
 - log probes
 - span decoration probes
 - span probes 
 
This way the last probe wraps method body and all other instrumented code.
Span should be first to be executed to create a span for the current method then span decoration should add tags for this new span. Logs should be decorated with trace/span ids for the current method. Finally metric are last because no dependency.

# Motivation

# Additional Notes

Jira ticket: [DEBUG-1752]

[DEBUG-1752]: https://datadoghq.atlassian.net/browse/DEBUG-1752?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ

### DIFF
--- a/dd-java-agent/agent-debugger/src/main/java/com/datadog/debugger/agent/DebuggerTransformer.java
+++ b/dd-java-agent/agent-debugger/src/main/java/com/datadog/debugger/agent/DebuggerTransformer.java
@@ -6,8 +6,10 @@ import static java.util.stream.Collectors.toList;
 import com.datadog.debugger.instrumentation.DiagnosticMessage;
 import com.datadog.debugger.instrumentation.InstrumentationResult;
 import com.datadog.debugger.probe.LogProbe;
+import com.datadog.debugger.probe.MetricProbe;
 import com.datadog.debugger.probe.ProbeDefinition;
 import com.datadog.debugger.probe.SpanDecorationProbe;
+import com.datadog.debugger.probe.SpanProbe;
 import com.datadog.debugger.probe.Where;
 import com.datadog.debugger.sink.DebuggerSink;
 import com.datadog.debugger.sink.ProbeStatusSink;
@@ -30,6 +32,7 @@ import java.nio.file.StandardOpenOption;
 import java.security.CodeSource;
 import java.security.ProtectionDomain;
 import java.util.ArrayList;
+import java.util.Arrays;
 import java.util.Collections;
 import java.util.HashMap;
 import java.util.HashSet;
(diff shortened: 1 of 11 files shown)


dd-java-agent/agent-debugger/src/main/java/com/datadog/debugger/agent/DebuggerTransformer.java
```diff
@@ -536,6 +528,44 @@ private InstrumentationResult applyInstrumentation(
...
+    final ProbeDefinition definition;
+    final List<ProbeId> probeIds;
+
+    ToInstrumentInfo(ProbeDefinition definition, List<ProbeId> probeIds) {
+      this.definition = definition;
+      this.probeIds = probeIds;
+    }
+  }
+
+  private List<ToInstrumentInfo> filterAndSortDefinitions(List<ProbeDefinition> definitions) {
```
shatzi, line 541: nit, instead of adding everything into a list and then sort. You can add them all into Map<class-name,List<Probedefinition>> and then add them in the right order.

dd-java-agent/agent-debugger/src/main/java/com/datadog/debugger/agent/DebuggerTransformer.java
```diff
@@ -536,6 +528,44 @@ private InstrumentationResult applyInstrumentation(
...
+        capturedContextProbes.add(definition);
+      } else {
+        toInstrument.add(new ToInstrumentInfo(definition, singletonList(definition.getProbeId())));
+      }
+    }
+    if (!capturedContextProbes.isEmpty()) {
+      List<ProbeId> probesIds =
+          capturedContextProbes.stream().map(ProbeDefinition::getProbeId).collect(toList());
+      ProbeDefinition referenceDefinition = selectReferenceDefinition(capturedContextProbes);
+      toInstrument.add(new ToInstrumentInfo(referenceDefinition, probesIds));
```
shatzi, line 557: minor nit. we are not ordering log/spanDecoration probes in this list. but I don't think that matter.

dd-java-agent/agent-debugger/src/test/java/com/datadog/debugger/agent/CapturedSnapshotTest.java
```diff
@@ -2037,6 +2045,75 @@ public void lineRecord() throws IOException, URISyntaxException {
     assertCaptureFields(capturedContext, "age", Integer.TYPE.getTypeName(), "0");
   }
 
+  @Test
+  public void allProbesSameMethod() throws IOException, URISyntaxException {
```
shatzi, line 2049: very cool

shatzi (approved):

PR #6455, opened 2024-01-08
Fix issue for parsing classes from debugger.symbol

jpbempel:
# What Does This Do
Hard code exclusion of `com.datadog.debugger.symbol` package to avoid duplicate class definition when loading a class from this package that will be transform by SymbolExtractionTransformer

# Motivation

# Additional Notes


ojung (approved):
Thanks!

PR #6460, opened 2024-01-09
Use spark.app.name as default service if not set by the user

paul-laffon-dd:
# What Does This Do

Use `spark.app.name` as the default service if not set by the user using the flag `dd.spark.app-name-as-service` 

It default to `false` and can be can be activated using
- the java option `-Ddd.spark.app-name-as-service=true`
- the env variable `DD_SPARK_APP_NAME_AS_SERVICE=true` 

# Motivation

Better default service name for spark applications. It currently default to `spark` when the service name is not set

# Additional Notes

When [running a spark application](https://spark.apache.org/docs/latest/running-on-kubernetes.html#cluster-mode), it is good practice to provide the application name, which will then be displayed by the cluster manager. Using this as the default service name

```
spark-submit --master k8s://k8s_url --name some_app_name
```


mcculls (approved):

paul-laffon-dd:
Discussing with @raphaelgavache, it will be safer to put it behind a flag to let users choose if they want to have this behaviour instead of having it by default 

@mcculls could you take a look at the updated PR ?

mcculls (approved):

PR #6461, opened 2024-01-10
Hardcoded secrets detection

jandro996:
# What Does This Do

- Use Java constant pool to get each loaded class constant 
- Check if the constants match secrets regular expressions
- When we found a secret use asm classReader to get the constant line and report the vulnerability

# Motivation

Detect hardcoded secrets in custom code

# Additional Notes

Jira ticket: [APPSEC-11890](https://datadoghq.atlassian.net/browse/APPSEC-11890)

[APPSEC-11890]: https://datadoghq.atlassian.net/browse/APPSEC-11890?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ


dd-java-agent/agent-iast/src/main/java/com/datadog/iast/sink/HardcodedSecretModuleImpl.java
```diff
@@ -0,0 +1,73 @@
...
+  @Nullable
+  private Set<Secret> getSecrets(final Set<String> literals, final String clazz) {
+    Set<Secret> secrets = null;
+    for (String literal : literals) {
+      for (HardcodedSecretMatcher secretMatcher : HARDCODED_SECRET_MATCHERS) {
+        if (secretMatcher.matches(literal)) {
+          if (secrets == null) {
+            secrets = new HashSet<>();
+          }
+          secrets.add(new Secret(literal, secretMatcher.getRedactedEvidence()));
```
manuel-alvarez-alvarez, line 49: Any reason not to short-circuit here?

benchmark/startup/petclinic/benchmark.json
```diff
@@ -29,6 +29,12 @@
         "VARIANT": "iast",
         "JAVA_OPTS": "-javaagent:${TRACER} -Ddd.benchmark.enabled=true -Ddd.benchmark.output.dir=${OUTPUT_DIR}/iast -Ddd.iast.enabled=true"
       }
+    },
```
manuel-alvarez-alvarez, line 32: Not sure if this provides much value, as there are not many hardoded strings in the code

dd-java-agent/agent-iast/src/main/java/com/datadog/iast/util/IastClassVisitor.java
```diff
@@ -0,0 +1,79 @@
...
+    private final String method;
+
+    private int currentLine;
+
+    public IastMethodVisitor(
+        final Set<Secret> secrets,
+        final String clazz,
+        final String method,
+        final Reporter reporter) {
+      super(Opcodes.ASM9);
```
manuel-alvarez-alvarez, line 49: Use `net.bytebuddy.utility.OpenedClassReader.ASM_API` instead

dd-java-agent/agent-iast/src/main/java/com/datadog/iast/sink/HardcodedSecretModuleImpl.java
```diff
@@ -0,0 +1,73 @@
...
+
+  private void reportVulnerability(
+      final Set<Secret> secrets, final String clazz, final @Nonnull byte[] classFile) {
+    ClassReader classReader = new ClassReader(classFile);
+    IastClassVisitor classVisitor = new IastClassVisitor(secrets, clazz, reporter);
+    classReader.accept(classVisitor, 0);
+  }
+
+  @Nullable
+  private Set<Secret> getSecrets(final Set<String> literals, final String clazz) {
```
manuel-alvarez-alvarez, line 41: If you use `Map<String, String>` you can simplify your code later and drop the secret class

dd-java-agent/agent-iast/src/main/java/com/datadog/iast/util/IastClassVisitor.java
```diff
@@ -0,0 +1,79 @@
...
+import com.datadog.iast.model.Vulnerability;
+import com.datadog.iast.model.VulnerabilityType;
+import com.datadog.iast.sink.HardcodedSecretModuleImpl.Secret;
+import java.util.Set;
+import org.objectweb.asm.ClassVisitor;
+import org.objectweb.asm.Label;
+import org.objectweb.asm.MethodVisitor;
+import org.objectweb.asm.Opcodes;
+
+public class IastClassVisitor extends ClassVisitor {
```
manuel-alvarez-alvarez, line 15: Maybe `IastSecretVisitor` is a better name

dd-java-agent/agent-iast/src/main/java/com/datadog/iast/util/IastClassVisitor.java
```diff
@@ -0,0 +1,79 @@
...
+    @Override
+    public void visitLineNumber(int line, Label start) {
+      currentLine = line;
+    }
+
+    @Override
+    public void visitLdcInsn(Object cst) {
+      if (cst instanceof String) {
+        String literal = ((String) cst);
+        for (Secret secret : secrets) {
```
manuel-alvarez-alvarez, line 65: If you use a Map<String, String> the coder is way shorter and you can do something like (remember to check for min length to speed up processing):

```java
if (literal.length >= MIN_SECRET_LENGTH) {
    String redacted = secrets.get(literal);
    if (redacted != null) {
        reporter.report(
                    null,
                    new Vulnerability(
                        VulnerabilityType.HARDCODED_SECRET,
                        Location.forClassAndMethodAndLine(clazz, method, currentLine),
                        new Evidence(redacted)));
   }
}
```

dd-java-agent/instrumentation/iast-instrumenter/src/main/java/datadog/trace/instrumentation/iastinstrumenter/IastInstrumentation.java
```diff
@@ -36,6 +37,11 @@ protected CallSiteSupplier callSites() {
     return IastCallSiteSupplier.INSTANCE;
   }
 
+  @Override
+  protected Advices buildAdvices(final Iterable<CallSites> callSites) {
+    return Advices.fromCallSites(callSites, IastHardcodedSecretListener.INSTANCE);
```
manuel-alvarez-alvarez, line 42: The short-circuit makes more sense here `Config.get().isIastHardcodedSecretEnabled()` (this way we remove all overhead)

dd-java-agent/instrumentation/iast-instrumenter/src/main/java/datadog/trace/instrumentation/iastinstrumenter/IastHardcodedSecretListener.java
```diff
@@ -0,0 +1,44 @@
...
+    if (!Config.get().isIastHardcodedSecretEnabled()) {
+      return;
+    }
+    final HardcodedSecretModule iastModule = InstrumentationBridge.HARDCODED_SECRET;
+    if (iastModule != null) {
+      Set<String> literals = new HashSet<>();
+      for (int index = 1; index < pool.getCount(); index++) {
+        if (pool.getType(index) == ConstantPool.CONSTANT_STRING_TAG) {
+          final int literalIndex = pool.readUnsignedShort(pool.getOffset(index));
+          final String literal = pool.readUTF8(pool.getOffset(literalIndex));
```
manuel-alvarez-alvarez, line 33: If you call `pool.readUnsignedShort(pool.getOffset(literalIndex))` it will give the size in bytes of the string. Since UTF-8 ranges between 1 and 4 bytes, we might be able to do some pre-filtering and not read small strings. 

What do you think @smola? (it might be too much optimization :smile: )

internal-api/src/main/java/datadog/trace/api/iast/secrets/HardcodedSecretMatcher.java
```diff
@@ -0,0 +1,326 @@
...
+
+  private final Pattern pattern;
+
+  private HardcodedSecretMatcher(final String redactedEvidence, final Pattern pattern) {
+    this.redactedEvidence = redactedEvidence;
+    this.pattern = pattern;
+  }
+
+  public String getRedactedEvidence() {
+    return redactedEvidence;
```
manuel-alvarez-alvarez, line 320: As a future improvement it will be super nice if the redacted evidence could be the original secret but masked, as an example, something like `glpat-<redacted>`

dd-java-agent/agent-tooling/src/main/java/datadog/trace/agent/tooling/iast/IastSecretClassReader.java
```diff
@@ -0,0 +1,20 @@
+package datadog.trace.agent.tooling.iast;
+
+import datadog.trace.api.function.TriConsumer;
+import java.util.Map;
+import javax.annotation.Nonnull;
+import org.objectweb.asm.ClassReader;
+
+public class IastSecretClassReader {
```
manuel-alvarez-alvarez, line 8: This class does not provide much value, can we add an static method to `IastSecretVisitor` like 

```java
  public static void readClass(
      final Map<String, String> secrets,
      final @Nonnull byte[] classFile,
      @Nonnull TriConsumer consumer) {
    ClassReader classReader = new ClassReader(classFile);
    IastSecretVisitor classVisitor = new IastSecretVisitor(secrets, consumer);
    classReader.accept(classVisitor, 0);
  }
```

manuel-alvarez-alvarez (approved):

internal-api/src/main/java/datadog/trace/api/iast/secrets/HardcodedSecretMatcher.java
jandro996: I will add it to the hardcode secrets RFC to discuss it with all the group

dd-java-agent/instrumentation/iast-instrumenter/src/main/java/datadog/trace/instrumentation/iastinstrumenter/IastHardcodedSecretListener.java
```diff
@@ -0,0 +1,96 @@
...
+
+    private final String clazz;
+
+    public ReportSecretConsumer(final String clazz) {
+      this.clazz = clazz;
+    }
+
+    @Override
+    public void accept(String method, String value, Integer currentLine) {
+      InstrumentationBridge.HARDCODED_SECRET.onHardcodedSecret(value, method, clazz, currentLine);
```
manuel-alvarez-alvarez, line 93: The module can be null at this point.

internal-api/src/main/java/datadog/trace/api/iast/secrets/HardcodedSecretMatcher.java
```diff
@@ -0,0 +1,326 @@
+package datadog.trace.api.iast.secrets;
+
+import java.util.regex.Pattern;
+
+public class HardcodedSecretMatcher {
+
+  public static final int MIN_SECRET_LENGTH = 10;
+
+  public static final HardcodedSecretMatcher ADOBE_CLIENT_SECRET =
```
smola, line 9: We could try optimizing this in the future to do a single pass for all regular expressions.

smola (approved):

dd-java-agent/agent-iast/src/main/java/com/datadog/iast/sink/HardcodedSecretModuleImpl.java
```diff
@@ -0,0 +1,31 @@
...
+  }
+
+  @Override
+  public void onHardcodedSecret(
+      @Nonnull final String value,
+      @Nonnull final String method,
+      @Nonnull final String clazz,
+      final int currentLine) {
+
+    reporter.report(
```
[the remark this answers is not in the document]
jandro996: It was intended, as the secrets are detected once the class is loaded if we don't report it due to the overhead they won't be detected until the app restarts. We have several approaches:

- Keep the PR as it is (we can disable the feature if something goes wild)
- Add the OverheadController check as usual
-  Wrap all the secrets per Class and send it in one vulnerability (It's better to do in the future as we have to change the RFC and discuss it with other language teams to be aligned) -> @manuel-alvarez-alvarez  suggestion

What do you think, @smola?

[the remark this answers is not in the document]
smola: I think would merge as is, since they should be rare and there is no chance to detect them later. If we add a quota (which I think should be pretty large), I wouldn't do it until we have separate quotas per vulnerability types.

internal-api/src/main/java/datadog/trace/api/iast/secrets/HardcodedSecretMatcher.java
jandro996: I will register it as an improvement in JIRA probably we need to implement it with @manuel-alvarez-alvarez  suggestion "As a future improvement it will be super nice if the redacted evidence could be the original secret but masked, as an example, something like glpat-<redacted> (can easily be done with capturing groups)".

With the current implementation we need a matcher per secret to know what type of secret is and set the redacted evidence. With both suggestions we can use the regexp to detect if there is any secret and then redact the value.

PR #6463, opened 2024-01-10
add operation label to endpoint events

richardstartin:
# What Does This Do

Adds the operation name to endpoint events which allows for more profiling metrics

# Motivation

# Additional Notes

Jira ticket: [PROF-8874]

[PROF-8874]: https://datadoghq.atlassian.net/browse/PROF-8874?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ

### DIFF
--- a/dd-java-agent/agent-profiling/profiling-controller-openjdk/src/main/java/com/datadog/profiling/controller/openjdk/events/EndpointEvent.java
+++ b/dd-java-agent/agent-profiling/profiling-controller-openjdk/src/main/java/com/datadog/profiling/controller/openjdk/events/EndpointEvent.java
@@ -22,6 +22,9 @@ public class EndpointEvent extends Event implements EndpointTracker {
   @Label("Endpoint")
   private String endpoint = "unknown";
 
+  @Label("Operation")
+  private String operation = "unknown";
+
   @Label("Local Root Span Id")
   private final long localRootSpanId;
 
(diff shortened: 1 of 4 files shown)


dd-java-agent/agent-profiling/profiling-controller-openjdk/src/main/java/com/datadog/profiling/controller/openjdk/events/EndpointEvent.java
```diff
@@ -35,6 +38,7 @@ public void endpointWritten(AgentSpan span) {
     if (span.getDurationNano() >= TEN_MILLISECONDS && shouldCommit()) {
       end();
       this.endpoint = span.getResourceName().toString();
+      this.operation = span.getOperationName().toString();
```
jbachorik, line 41: Just a wondering - at https://github.com/DataDog/dd-trace-java/pull/6463/files#diff-70a78e046b30b2a33d3762d6281c9157a6daa3e77664b04fa6ac14d60891faafR33 the `operationName` and `resourceName` are checked for `null`. 
Should they be checked here as well? I see that the `resourceName` was already doing `toString()` without null-check so it may be actually safe to do it here ...

jbachorik (approved):

dd-java-agent/agent-profiling/profiling-controller-openjdk/src/main/java/com/datadog/profiling/controller/openjdk/events/EndpointEvent.java
richardstartin: The values can't be null, and if these events ever came through with null values it would indicate a tracing bug, but the other implementation is paranoid because it's JNI.

PR #6465, opened 2024-01-10
Kafka support as IAST source

manuel-alvarez-alvarez:
# What Does This Do
Instruments Kafka serializers in order to taint values coming over the wire.

# Motivation
Data coming from Kafka topics can be considered as untrusted from the point of view of a single JVM.

# Additional Notes

Jira ticket: [APPSEC-10440]

[APPSEC-10440]: https://datadoghq.atlassian.net/browse/APPSEC-10440?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ


dd-trace-api/src/main/java/datadog/trace/api/internal/TraceSegment.java
```diff
@@ -55,6 +55,14 @@ default void setTagCurrent(String key, Object value) {
    */
   void setDataTop(String key, Object value);
 
+  /**
```
manuel-alvarez-alvarez, line 58: Maybe it makes sense to make this changes to the internal api in a separate PR? What do you think @smola?

smola (approved):

PR #6480, opened 2024-01-12
Add `@exception` as local var in snapshots

jpbempel:
# What Does This Do
When uncaught exceptions are captured we put them as local variables like return values. This way we can explore fields from custom exceptions from snapshots

# Motivation

# Additional Notes

Jira ticket: [DEBUG-2042]

[DEBUG-2042]: https://datadoghq.atlassian.net/browse/DEBUG-2042?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ

### DIFF
--- a/dd-java-agent/agent-debugger/debugger-bootstrap/src/main/java/datadog/trace/bootstrap/debugger/CapturedContext.java
+++ b/dd-java-agent/agent-debugger/debugger-bootstrap/src/main/java/datadog/trace/bootstrap/debugger/CapturedContext.java
@@ -193,11 +193,8 @@ public class CapturedContext implements ValueReferenceResolver {
     if (values == null) {
       return;
     }
-    if (arguments == null) {
-      arguments = new HashMap<>();
-    }
     for (CapturedValue value : values) {
-      arguments.put(value.name, value);
+      putInArguments(value.name, value);
     }
   }
 
@@ -205,11 +202,8 @@ public class CapturedContext implements ValueReferenceResolver {
     if (values == null) {
       return;
     }
-    if (locals == null) {
-      locals = new HashMap<>();
-    }
     for (CapturedValue value : values) {
-      locals.put(value.name, value);
+      putInLocals(value.name, value);
     }
   }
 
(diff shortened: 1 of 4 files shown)


dd-java-agent/agent-debugger/debugger-bootstrap/src/main/java/datadog/trace/bootstrap/debugger/CapturedContext.java
```diff
@@ -226,6 +226,11 @@ public void addReturn(CapturedValue retValue) {
 
   public void addThrowable(Throwable t) {
     addThrowable(new CapturedThrowable(t));
+    if (locals == null) {
+      locals = new HashMap<>();
+    }
```
cimi, line 231: Wouldn't it be cleaner to just initialise the map every time we create a capture context and get rid of the null checks in the methods?

I guess we're not doing it for performance, to avoid useless work when there are no locals.

dd-java-agent/agent-debugger/src/test/resources/CapturedSnapshot05.java
```diff
@@ -4,7 +4,7 @@
 public class CapturedSnapshot05 {
 
   void triggerUncaughtException() {
-    throw new IllegalStateException("oops");
+    throw new CustomException("oops", "I did it again");
```
cimi, line 7: 😂

cimi (approved):
Thanks!

dd-java-agent/agent-debugger/debugger-bootstrap/src/main/java/datadog/trace/bootstrap/debugger/CapturedContext.java
jpbempel: yes trying to minimise the impact of allocations done by user code execution.

cimi: Since all four maps use the same types, we could have a private helper method to get rid of the repeated null checks. 

Something like `addProperty(locals, ValueReferences.EXCEPTION_REF, CapturedValue.of(t.getClass().getTypeName(), t)`, wdyt?

PR #6481, opened 2024-01-12
Attach command execution spans to current span if it exists

smola:
# What Does This Do
If an active spans exists, append the span for subprocess execution to it. This may delay sending the parent trace if the command runs for longer. Although this has not been a problem for other languages so far, and it seems to be a more general problem also applying to HTTP client requests and others.

# Motivation

Application security research team uses this to detect certain attacks in the backend, and for it to be effective, it needs to be correlated with the request.

# Additional Notes

* Jira ticket: [APPSEC-10243](https://datadoghq.atlassian.net/browse/APPSEC-10243)
* There is an existing system test at https://github.com/DataDog/system-tests/blob/529b9b01217803bb9dcae62e29f5e3b16ece0352/tests/appsec/test_shell_execution.py and its Java implementation at https://github.com/DataDog/system-tests/pull/2024

[APPSEC-10243]: https://datadoghq.atlassian.net/browse/APPSEC-10243?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ


cataphract:
So what happens if the command keeps running after the parent finishes? Won't this delay the trace for a long time?

smola:
> So what happens if the command keeps running after the parent finishes? Won't this delay the trace for a long time?

Unfortunately, it will. As far as I can tell, the same would happen today with HTTP client calls. This has not been an issue so far in Dotnet (available for longer) and Node. If this becomes a problem in practice, we might need to add a mechanism to close the subprocess span earlier.

ValentinZakharov (approved):

PR #6491, opened 2024-01-15
Don't run debugger tests twice in CI

bantonsson:
# What Does This Do

Skips the `debugger` tests in the `base` configuration to avoid running them twice.

# Motivation

While looking at upgrading some test dependencies I noticed that it seem like the `debugger` tests run as part of the `base` tests https://app.circleci.com/pipelines/github/DataDog/dd-trace-java/33016/workflows/da5dc872-792f-4c99-ba27-27a2c6e22726/jobs/1149959/tests


PerfectSlayer (approved):
💸

PR #6501, opened 2024-01-16
Reduce the memory footprint of IAST sources

manuel-alvarez-alvarez:
# What Does This Do
Removes the string copies used for the `Source`'s name and values and replaces them with weak references to the original value. Once a vulnerability is found, the values are consolidated using copies.

# Motivation
The number of tainted objects present at runtime can vary quite a lot so we have to ensure we consume the minimum amount of memory.

### DIFF
--- a/dd-java-agent/agent-iast/src/main/java/com/datadog/iast/model/Source.java
+++ b/dd-java-agent/agent-iast/src/main/java/com/datadog/iast/model/Source.java
@@ -3,21 +3,21 @@ package com.datadog.iast.model;
 import com.datadog.iast.model.json.SourceTypeString;
 import datadog.trace.api.iast.SourceTypes;
 import datadog.trace.api.iast.Taintable;
+import java.lang.ref.Reference;
 import java.util.Objects;
 import java.util.StringJoiner;
 import javax.annotation.Nullable;
 
 public final class Source implements Taintable.Source {
+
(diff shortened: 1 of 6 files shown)


dd-java-agent/agent-iast/src/main/java/com/datadog/iast/model/Source.java
```diff
@@ -3,24 +3,21 @@
...
-      final byte origin, @Nullable final CharSequence name, @Nullable final CharSequence value) {
-    this(origin, name == null ? null : name.toString(), value == null ? null : value.toString());
-  }
-
-  public Source(final byte origin, @Nullable final String name, @Nullable final String value) {
+  public Source(final byte origin, @Nullable final Object name, @Nullable final Object value) {
     this.origin = origin;
-    this.name = name;
-    this.value = value;
+    this.name = name instanceof Reference<?> ? name : new WeakReference<>(name);
```
smola, line 19: Doesn't this mean we can loose the name and value before serialization?

manuel-alvarez-alvarez: No, the `Evidence` constructor consolidates the ranges for the vulnerability and enforces a copy of the source using strings

jandro996 (approved):

jandro996 (approved):

smola (approved):

PR #6502, opened 2024-01-16
Add component and fix cmd.exit_code tag in subprocess instrumentation

smola:
# What Does This Do

# Motivation
* cmd.exit_code was present but as an integer (metric).
* component tag was missing.
* 
# Additional Notes

Jira ticket: [PROJ-IDENT]


ValentinZakharov:
I'm sorry, didn't get what was the motivation for change exit code type. As for me, it should stay as integer.

smola:
@ValentinZakharov String tag values are sent as tags in the `meta` field in spans, while integer values are sent as metrics (`metrics` field). The spec here (and accompanying system-tests) specify that `cmd.exit_code` is a tag in `meta`, so this PR aligns it with that.

mcculls (approved):

PR #6507, opened 2024-01-17
Support special case of empty catch block

jpbempel:
# What Does This Do
We declare the missing local var if not present and store the exception present on the stack in it. then we collect as usual the local vars and the exception will be then present in the snapshot

# Motivation
if you want to put a line probe in an empty catch block, exception variable is not declared in that special case, so to capture it as local var we need to explicitly add it.
```
try {
  throw new IllegalArgumentException("oops");
} catch (IllegalArgumentException ex) {
}
```
# Additional Notes

Jira ticket: [DEBUG-2045]

[DEBUG-2045]: https://datadoghq.atlassian.net/browse/DEBUG-2045?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ

### DIFF
--- a/dd-java-agent/agent-debugger/src/main/java/com/datadog/debugger/instrumentation/CapturedContextInstrumentor.java
+++ b/dd-java-agent/agent-debugger/src/main/java/com/datadog/debugger/instrumentation/CapturedContextInstrumentor.java
@@ -39,6 +39,8 @@ import datadog.trace.util.Strings;
 import java.lang.reflect.Field;
 import java.util.ArrayList;
 import java.util.List;
+import java.util.Set;
+import java.util.stream.Collectors;
 import org.objectweb.asm.Opcodes;
 import org.objectweb.asm.Type;
 import org.objectweb.asm.tree.AbstractInsnNode;
@@ -120,6 +122,19 @@ public class CapturedContextInstrumentor extends Instrumentor {
       }
       if (beforeLabel != null) {
         InsnList insnList = new InsnList();
+        if (isSingleLine) {
+          TryCatchBlockNode catchHandler = findCatchHandler(beforeLabel);
+          if (catchHandler != null && !isExceptionLocalDeclared(catchHandler, methodNode)) {
+            // empty catch block does not declare exception local variable - we add it
+            int idx = addExceptionLocal(catchHandler, methodNode);
+            // store exception in the new local variable
+            // stack [exception]
+            insnList.add(new InsnNode(Opcodes.DUP));
+            // stack [exception, exception]
+            insnList.add(new VarInsnNode(Opcodes.ASTORE, idx));
+            // stack [exception]
+          }
+        }
         ldc(insnList, Type.getObjectType(classNode.name));
         // stack [class, array]
         pushProbesIds(insnList);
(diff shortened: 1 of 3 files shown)


dd-java-agent/agent-debugger/src/main/java/com/datadog/debugger/instrumentation/CapturedContextInstrumentor.java
```diff
@@ -170,6 +185,71 @@ private boolean addLineCaptures(LineMap lineMap) {
...
+    }
+    return null;
+  }
+
+  private boolean isExceptionLocalDeclared(TryCatchBlockNode catchHandler, MethodNode methodNode) {
+    if (methodNode.localVariables == null || methodNode.localVariables.isEmpty()) {
+      return false;
+    }
+    String catchDesc =
+        catchHandler.type != null ? "L" + catchHandler.type + ";" : "Ljava/lang/Throwable;";
```
ojung, line 244: When can `catchHandler.type` be null?

dd-java-agent/agent-debugger/src/main/java/com/datadog/debugger/instrumentation/CapturedContextInstrumentor.java
```diff
@@ -170,6 +185,71 @@ private boolean addLineCaptures(LineMap lineMap) {
...
+    // find next label assume this is the end of the handler
+    while (current != null && current.getType() != AbstractInsnNode.LABEL) {
+      current = current.getNext();
+    }
+    if (current == null) {
+      reportWarning("Cannot add exception local variable to catch block - no end of handler.");
+      return -1;
+    }
+    LabelNode end = (LabelNode) current;
+    for (int i = 0; i < 100; i++) {
```
ojung, line 213: Is the 100 arbitrary or an actual limit?

dd-java-agent/agent-debugger/src/main/java/com/datadog/debugger/instrumentation/CapturedContextInstrumentor.java
```diff
@@ -170,6 +185,71 @@ private boolean addLineCaptures(LineMap lineMap) {
...
+      String exceptionLocalName = "ex" + i;
```
ojung, line 214: Why do we generate these names? In the test fixture class, the catch handler has `ex`. We don't have access to that?

ojung (approved):
Thanks, just added a couple of questions for my understanding.

dd-java-agent/agent-debugger/src/main/java/com/datadog/debugger/instrumentation/CapturedContextInstrumentor.java
jpbempel: arbitrary.
100 catch blocks sounds arbitrary enough 😄

dd-java-agent/agent-debugger/src/main/java/com/datadog/debugger/instrumentation/CapturedContextInstrumentor.java
jpbempel: nope, unfortunately for this very special case (empty catch block), the variable ex is not translated by the javac compiler as bytecode symbol, not even in the constant pool or local variable table.
Example:
```
$ cat EmptyCatch.java
public class EmptyCatch {
    public static void main(String[] args) {
	try {
	    System.out.println("foo");
	} catch (Throwable ex) {
	}
    }
}
$ javac -g EmptyCatch.java
$ javap -v -p EmptyCatch
Classfile /Users/jean-philippe.bempel/projects/tmp/EmptyCatch.class
  Last modified Jan 17, 2024; size 602 bytes
  MD5 checksum b73fdc276fa40570eb1441ddab238d98
  Compiled from "EmptyCatch.java"
public class EmptyCatch
  minor version: 0
  major version: 52
  flags: ACC_PUBLIC, ACC_SUPER
Constant pool:
   #1 = Methodref          #7.#23         // java/lang/Object."<init>":()V
   #2 = Fieldref           #24.#25        // java/lang/System.out:Ljava/io/PrintStream;
   #3 = String             #26            // foo
   #4 = Methodref          #27.#28        // java/io/PrintStream.println:(Ljava/lang/String;)V
   #5 = Class              #29            // java/lang/Throwable
   #6 = Class              #30            // EmptyCatch
   #7 = Class              #31            // java/lang/Object
   #8 = Utf8               <init>
   #9 = Utf8               ()V
  #10 = Utf8               Code
  #11 = Utf8               LineNumberTable
  #12 = Utf8               LocalVariableTable
  #13 = Utf8               this
  #14 = Utf8               LEmptyCatch;
  #15 = Utf8               main
  #16 = Utf8               ([Ljava/lang/String;)V
  #17 = Utf8               args
  #18 = Utf8               [Ljava/lang/String;
  #19 = Utf8               StackMapTable
  #20 = Class              #29            // java/lang/Throwable
  #21 = Utf8               SourceFile
  #22 = Utf8               EmptyCatch.java
  #23 = NameAndType        #8:#9          // "<init>":()V
  #24 = Class              #32            // java/lang/System
  #25 = NameAndType        #33:#34        // out:Ljava/io/PrintStream;
  #26 = Utf8               foo
  #27 = Class              #35            // java/io/PrintStream
  #28 = NameAndType        #36:#37        // println:(Ljava/lang/String;)V
  #29 = Utf8               java/lang/Throwable
  #30 = Utf8               EmptyCatch
  #31 = Utf8               java/lang/Object
  #32 = Utf8               java/lang/System
  #33 = Utf8               out
  #34 = Utf8               Ljava/io/PrintStream;
  #35 = Utf8               java/io/PrintStream
  #36 = Utf8               println
  #37 = Utf8               (Ljava/lang/String;)V
{
  public EmptyCatch();
    descriptor: ()V
    flags: ACC_PUBLIC
    Code:
      stack=1, locals=1, args_size=1
         0: aload_0
         1: invokespecial #1                  // Method java/lang/Object."<init>":()V
         4: return
      LineNumberTable:
        line 1: 0
      LocalVariableTable:
        Start  Length  Slot  Name   Signature
            0       5     0  this   LEmptyCatch;

  public static void main(java.lang.String[]);
    descriptor: ([Ljava/lang/String;)V
    flags: ACC_PUBLIC, ACC_STATIC
    Code:
      stack=2, locals=2, args_size=1
         0: getstatic     #2                  // Field java/lang/System.out:Ljava/io/PrintStream;
         3: ldc           #3                  // String foo
         5: invokevirtual #4                  // Method java/io/PrintStream.println:(Ljava/lang/String;)V
         8: goto          12
        11: astore_1
        12: return
      Exception table:
         from    to  target type
             0     8    11   Class java/lang/Throwable
      LineNumberTable:
        line 4: 0
        line 6: 8
        line 5: 11
        line 7: 12
      LocalVariableTable:
        Start  Length  Slot  Name   Signature
            0      13     0  args   [Ljava/lang/String;
      StackMapTable: number_of_entries = 2
        frame_type = 75 /* same_locals_1_stack_item */
          stack = [ class java/lang/Throwable ]
        frame_type = 0 /* same */
}
SourceFile: "EmptyCatch.java"
```

ojung: Ok, thanks for the explanation :)

dd-java-agent/agent-debugger/src/main/java/com/datadog/debugger/instrumentation/CapturedContextInstrumentor.java
jpbempel: for finally block see https://docs.oracle.com/javase/specs/jvms/se8/html/jvms-3.html#jvms-3.13
we are creating also finally block this way here: https://github.com/DataDog/dd-trace-java/blob/79aae57a1cef8a610f0768c7da88fed559ddc9c0/dd-java-agent/agent-debugger/src/main/java/com/datadog/debugger/instrumentation/Instrumentor.java#L297 with type = null

PR #6512, opened 2024-01-17
Skip milestone and issues gh workflows for Release Candidates

am312:
# What Does This Do

Skips the GitHub workflow milestone update when tags contain "-RC".  This allows multiple Release Candidates to be published with the same issues in the release notes.  Also does not update & close issues for releases of type `prerelease`.

### DIFF
--- a/.github/workflows/increment-milestones-on-tag.yaml
+++ b/.github/workflows/increment-milestones-on-tag.yaml
@@ -4,7 +4,7 @@ on:
 
 jobs:
   increment_milestone:
-    if: github.event.ref_type == 'tag' && contains(github.ref,'-RC') == false && github.event.master_branch == 'master'
+    if: github.event.ref_type == 'tag' && contains(github.ref_name,'-RC') == false && github.event.master_branch == 'master'
     runs-on: ubuntu-latest
     steps:
       - name: Get milestone title
--- a/.github/workflows/increment-milestones-on-tag.yaml
+++ b/.github/workflows/increment-milestones-on-tag.yaml
@@ -4,7 +4,7 @@ on:
 
 jobs:
   increment_milestone:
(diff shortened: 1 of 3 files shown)


mcculls (commented):
looks reasonable, but I'd like @PerfectSlayer to approve before merging, as he has more experience with GH actions

.github/workflows/increment-milestones-on-tag.yaml
```diff
@@ -4,7 +4,7 @@ on:
 
 jobs:
   increment_milestone:
-    if: github.event.ref_type == 'tag' && github.event.master_branch == 'master'
+    if: github.event.ref_type == 'tag' && contains(github.ref,'-RC') == false && github.event.master_branch == 'master'
```
PerfectSlayer, line 7: I wonder if `ref_name` is not better suited than `ref`? [doc](https://docs.github.com/en/actions/learn-github-actions/contexts)
What about ignoring case? I feel we might end up with `-rc` prefix (as `v` is already in lower case) 🤷

PerfectSlayer (commented):
Good. Can you also update the related documentation file to state it won't skip RC on purpose?  Thanks 🙏 
https://github.com/DataDog/dd-trace-java/tree/master/.github/workflows#increment-milestones-on-tag-

.github/workflows/increment-milestones-on-tag.yaml
am312: yes `ref_name` is better. Thanks.  `contains()` is case insensitive so `RC` and `rC` etc all skip the workflow.

PerfectSlayer (approved):
Looks good.
As any CI change, we might need to keep an eye on the first run to check everything runs smoothly 😇

PR #6520, opened 2024-01-18
Track high watermark offsets

piochelepiotr:
# What Does This Do

Track high watermark offsets along with produce and commit offsets.
This information can be used to determine Kafka lag of consumers.
So we can now get the Kafka lag by only instrumenting the consumer service, with no instrumentation on the producer side.

# Motivation

Jira ticket: [PROJ-IDENT]


dd-smoke-tests/datastreams/kafkaschemaregistry/src/main/java/datadog/smoketest/datastreamskafka/KafkaProducerWithSchemaRegistry.java
```diff
@@ -80,11 +104,63 @@ public static void main(String[] args) {
...
+        System.out.println("produced message");
       }
     } catch (Exception e) {
       e.printStackTrace();
     } finally {
       producer.close();
     }
   }
+
+  public static void consume() {
```
piochelepiotr, line 116: pretty useful for testing, since confluent cloud is always up (and free if we don't have traffic)

piochelepiotr:
> > But I didn't find a way to hook into a place that is updated regularly.
> 
> Did you find a place, but were not able to hook into? Or were not able to find the right place?

Ah, @kr-igor suggested to use reflection to access the high watermark offset.
So I did that, and added instrumentation in the same place we capture commit offsets. The benefit is that to compute lag, we need both commit offsets and high watermark offsets, and they are now captured in the same place.

PerfectSlayer (approved):

dd-java-agent/instrumentation/kafka-clients-0.11/src/main/java/datadog/trace/instrumentation/kafka_clients/ConsumerCoordinatorInstrumentationHelper.java
```diff
@@ -0,0 +1,40 @@
...
+    }
+    Object state = partitionStates.stateValue(tp);
+    if (state == null) {
+      return null;
+    }
+    try {
+      Class<?> clazz = state.getClass();
+      Field field = clazz.getDeclaredField("highWatermark");
+      field.setAccessible(true);
+      return (Long) field.get(state);
```
kr-igor, line 35: I believe the field itself should be cached as a static field in this class.

dd-java-agent/instrumentation/kafka-clients-0.11/src/main/java/datadog/trace/instrumentation/kafka_clients/ConsumerCoordinatorInstrumentationHelper.java
```diff
@@ -0,0 +1,40 @@
...
+import org.apache.kafka.common.internals.PartitionStates;
+
+public class ConsumerCoordinatorInstrumentationHelper {
+
+  public static PartitionStates getPartitionStates(SubscriptionState state) {
+    if (state == null) {
+      return null;
+    }
+    try {
+      Field field = SubscriptionState.class.getDeclaredField("assignment");
```
kr-igor, line 15: The same as the comment below

dd-java-agent/instrumentation/kafka-clients-0.11/src/main/java/datadog/trace/instrumentation/kafka_clients/ConsumerCoordinatorInstrumentationHelper.java
amarziali: As a nit, for better performance, you can also use methodHandles like here: https://github.com/DataDog/dd-trace-java/blob/6ca92f88759e9556e5af71ab326cf94da8c18de0/internal-api/src/main/java/datadog/trace/util/MethodHandles.java#L31

piochelepiotr: Why doesn't the MethodHandless class cache the getters?

dougqh: That would be a system wide cache that would be hard to size appropriately.
Rather than doing that, the core library let's you decide when it is and isn't appropriate to cache.

And there are other advantages to "caching" closer to the usage.  By storing the handle into a static final, that allows the optimizer to optimize away the MethodHandle.

piochelepiotr:
Closing for now in favor of: https://github.com/DataDog/integrations-core/pulls?q=is%3Apr+author%3Apiochelepiotr+

PR #6521, opened 2024-01-18
remove suffix from tag in gh workflow

am312:
# What Does This Do

Retains only the portion of a release tag before any "-" in the tag.  Gets "1.2.3" from "1.2.3-RC" to lookup the milestone.


PerfectSlayer:
Just pushed my changes here: #6523 I went to a stricter version based on pattern to avoid any future parsing issue. 
It was already implemented by needed time to test it this morning.

PR #6525, opened 2024-01-19
Improve telemetry log deduplication

smola:
# What Does This Do
* Fix deduplication for log messages with exceptions (throwable and usual implementations do not override equals/hashCode).
* Use the new `count` field for deduplicated messages.

# Motivation
We're getting several duplicated logs when exceptions are present.

# Additional Notes

Jira ticket: [APPSEC-51229](https://datadoghq.atlassian.net/browse/APPSEC-51229)

[APPSEC-51229]: https://datadoghq.atlassian.net/browse/APPSEC-51229?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ

### DIFF
--- a/internal-api/src/main/java/datadog/trace/api/telemetry/LogCollector.java
+++ b/internal-api/src/main/java/datadog/trace/api/telemetry/LogCollector.java
@@ -34,13 +34,14 @@ public class LogCollector {
   }
 
   public void addLogMessage(String logLevel, String message, Throwable throwable) {
+    if (rawLogMessages.size() >= maxCapacity) {
+      // TODO: We could emit a metric for dropped logs.
+      return;
+    }
     RawLogMessage rawLogMessage =
         new RawLogMessage(logLevel, message, throwable, System.currentTimeMillis());
-
-    if (rawLogMessages.size() < maxCapacity) {
-      AtomicInteger count = rawLogMessages.computeIfAbsent(rawLogMessage, k -> new AtomicInteger());
-      count.incrementAndGet();
-    }
+    AtomicInteger count = rawLogMessages.computeIfAbsent(rawLogMessage, k -> new AtomicInteger());
+    count.incrementAndGet();
   }
 
   public Collection<RawLogMessage> drain() {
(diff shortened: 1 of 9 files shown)


internal-api/src/main/java/datadog/trace/api/telemetry/LogCollector.java
```diff
@@ -49,58 +50,56 @@ public Collection<RawLogMessage> drain() {
...
     List<RawLogMessage> list = new ArrayList<>();
-
     Iterator<Map.Entry<RawLogMessage, AtomicInteger>> iterator =
         rawLogMessages.entrySet().iterator();
 
     while (iterator.hasNext()) {
       Map.Entry<RawLogMessage, AtomicInteger> entry = iterator.next();
-
       RawLogMessage logMessage = entry.getKey();
+      // XXX: There might be lost writers to the counters under concurrency if another thread
```
smola, line 59:
```suggestion
      // XXX: There might be lost writes to the counters under concurrency if another thread
```

jandro996 (approved):

internal-api/src/main/java/datadog/trace/api/telemetry/LogCollector.java
```diff
@@ -49,58 +50,56 @@ public Collection<RawLogMessage> drain() {
...
+              that.throwable == null ? null : that.throwable.getClass())
+          && Objects.deepEquals(
+              throwable == null ? null : throwable.getStackTrace(),
+              that.throwable == null ? null : that.throwable.getStackTrace());
     }
 
     @Override
     public int hashCode() {
-      return Objects.hash(logLevel, messageOriginal, throwable);
+      return Objects.hash(logLevel, message, throwable == null ? null : throwable.getClass());
```
manuel-alvarez-alvarez, line 102: I understand why not using the stack here, just wonder how it will affect performance (probably not an issue)

manuel-alvarez-alvarez (approved):

internal-api/src/main/java/datadog/trace/api/telemetry/LogCollector.java
smola: The assumption is we would have not many logs in the same minute with the same message and exception type but different stack trace... hopefully I'm right, but I'll keep an eye to the results we get during dogfooding (this is currently disabled by default).

PR #6533, opened 2024-01-22
Enable Data Streams without adding context into Kafka headers

piochelepiotr:
# What Does This Do

This PR allows customers to enable a limited version of Data Streams monitoring, but without passing context through headers.
This can be useful is really high throughput pipelines with small message size, in which the header size would hugely impact the total throughput through Kafka.

# Motivation

# Additional Notes

Jira ticket: [PROJ-IDENT]

### DIFF
--- a/dd-java-agent/instrumentation/kafka-clients-0.11/src/main/java/datadog/trace/instrumentation/kafka_clients/KafkaProducerInstrumentation.java
+++ b/dd-java-agent/instrumentation/kafka-clients-0.11/src/main/java/datadog/trace/instrumentation/kafka_clients/KafkaProducerInstrumentation.java
@@ -140,8 +140,7 @@ public final class KafkaProducerInstrumentation extends InstrumenterModule.Traci
           // message size.
           // The stats are saved in the pathway context and sent in PayloadSizeAdvice.
           propagate()
-              .injectPathwayContextWithoutSendingStats(
-                  span, record.headers(), setter, sortedTags);
+              .injectPathwayContextWithoutSendingStats(span, record.headers(), setter, sortedTags);
           AvroSchemaExtractor.tryExtractProducer(record, span);
         }
       } catch (final IllegalStateException e) {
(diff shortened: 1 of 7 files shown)


dd-java-agent/instrumentation/kafka-clients-0.11/src/main/java/datadog/trace/instrumentation/kafka_clients/TextMapInjectAdapter.java
```diff
@@ -8,20 +8,35 @@
...
+  public static final TextMapInjectAdapter SETTER = new TextMapInjectAdapter(false);
+  public static final TextMapInjectAdapter NOOP_SETTER = new TextMapInjectAdapter(true);
+
+  public TextMapInjectAdapter(boolean isNoOp) {
+    this.isNoOp = isNoOp;
+  }
 
   @Override
   public void set(final Headers headers, final String key, final String value) {
+    if (isNoOp) {
```
ygree, line 22: You could have a separate NoOp implementation to avoid checking in each method.

dd-java-agent/instrumentation/kafka-clients-0.11/src/main/java/datadog/trace/instrumentation/kafka_clients/KafkaProducerInstrumentation.java
```diff
@@ -119,50 +119,50 @@ public static AgentScope onEnter(
...
-                  record.headers());
-
-          propagate().inject(span, record.headers(), SETTER);
-          if (STREAMING_CONTEXT.empty() || STREAMING_CONTEXT.isSinkTopic(record.topic())) {
-            propagate()
-                .injectPathwayContextWithoutSendingStats(
-                    span, record.headers(), SETTER, sortedTags);
-            AvroSchemaExtractor.tryExtractProducer(record, span);
-          }
+        setter = SETTER;
```
ygree, line 126: For the sake of clarity, perhaps it should be `TextMapInjectAdapter.SETTER`?

nayeem-kamal: +1

ygree (changes requested):
Looks good, just a few suggestions, but they are minor.

nayeem-kamal:
LGTM. @ygree 's suggestions are good for improving the readability of the changes.

PR #6536, opened 2024-01-23
Reduce long running default flush interval to 120 seconds

paul-laffon-dd:
# What Does This Do

Reduce long running default flush interval to 120 seconds

# Motivation

User feedback: 5 minutes not being reactive enough

# Additional Notes

All internal spark jobs are already using 120 seconds for their flush interval

### DIFF
--- a/dd-trace-api/src/main/java/datadog/trace/api/ConfigDefaults.java
+++ b/dd-trace-api/src/main/java/datadog/trace/api/ConfigDefaults.java
@@ -206,7 +206,7 @@ public final class ConfigDefaults {
 
   static final boolean DEFAULT_TRACE_HTTP_RESOURCE_REMOVE_TRAILING_SLASH = false;
   static final boolean DEFAULT_TRACE_LONG_RUNNING_ENABLED = false;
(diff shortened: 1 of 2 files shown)


dd-trace-api/src/main/java/datadog/trace/api/ConfigDefaults.java
```diff
@@ -206,7 +206,7 @@ public final class ConfigDefaults {
 
   static final boolean DEFAULT_TRACE_HTTP_RESOURCE_REMOVE_TRAILING_SLASH = false;
   static final boolean DEFAULT_TRACE_LONG_RUNNING_ENABLED = false;
-  static final long DEFAULT_TRACE_LONG_RUNNING_FLUSH_INTERVAL = 300; // seconds -> 5 minutes
+  static final long DEFAULT_TRACE_LONG_RUNNING_FLUSH_INTERVAL = 120; // seconds -> 2 minutes
```
amarziali, line 209: For all users or should it be configurable?

raphaelgavache (approved):

PerfectSlayer (approved):

PR #6537, opened 2024-01-23
Allow axis server span to set local root resource name

amarziali:
# What Does This Do

When using Axis2 to deploy jaxrpc services, this PR allows to use the axis2 span resource name as the entry point resource name. 
In order to activate this the either the system property `dd.trace.axis.promote.resource-name` either the env `DD_TRACE_AXIS_PROMOTE_RESOURCE_NAME` should be explicitly set to true

As an additional change (not opt-in) this PR marks the axis2 spans as measured. In fact, lacking of a span kind, those spans are never used to aggregate statistics, unlike other span of type server or client.

# Motivation

This will allow seeing that soap call as a dependency on the service catalog. (otherwise the servlet URL will be used as a resource)

# Additional Notes

A "better" approach would have been to instrument the AxisServlet and have this logic only applied at that level. This however required more efforts and I estimate that the proposed approach, being opt-in, is good enough.

Jira ticket: [PROJ-IDENT]

### DIFF
--- a/dd-java-agent/instrumentation/axis-2/src/main/java/datadog/trace/instrumentation/axis2/AxisMessageDecorator.java
+++ b/dd-java-agent/instrumentation/axis-2/src/main/java/datadog/trace/instrumentation/axis2/AxisMessageDecorator.java
@@ -3,7 +3,9 @@ package datadog.trace.instrumentation.axis2;
 import static datadog.trace.bootstrap.instrumentation.api.AgentTracer.activeSpan;
 import static datadog.trace.bootstrap.instrumentation.api.InternalSpanTypes.SOAP;
 
+import datadog.trace.api.Config;
 import datadog.trace.bootstrap.instrumentation.api.AgentSpan;
+import datadog.trace.bootstrap.instrumentation.api.ResourceNamePriorities;
 import datadog.trace.bootstrap.instrumentation.api.UTF8BytesString;
 import datadog.trace.bootstrap.instrumentation.decorator.BaseDecorator;
 import org.apache.axis2.context.MessageContext;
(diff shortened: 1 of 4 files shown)


nayeem-kamal (approved):
LGTM!

dd-java-agent/instrumentation/axis-2/src/main/java/datadog/trace/instrumentation/axis2/AxisMessageDecorator.java
```diff
@@ -41,11 +43,24 @@ public boolean shouldTrace(final MessageContext message) {
 
   public boolean sameTrace(final AgentSpan span, final MessageContext message) {
     return AXIS2_MESSAGE.equals(span.getSpanName())
```
ygree, line 45: Looks like the first comparison should either use `compareTo` or both parts should be converted to a string because `getSpanName()` returns `CharSequence`.

dd-java-agent/instrumentation/axis-2/src/main/java/datadog/trace/instrumentation/axis2/AxisMessageDecorator.java
```diff
@@ -41,11 +43,24 @@ public boolean shouldTrace(final MessageContext message) {
...
-        && span.getResourceName().equals(soapAction(message));
+        && span.getResourceName().toString().equals(soapAction(message));
```
ygree, line 46: Alternatively, `compareTo` could be used to avoid the `toString` call.

dd-java-agent/instrumentation/axis-2/src/main/java/datadog/trace/instrumentation/axis2/AxisMessageDecorator.java
amarziali: While it is not part of the changeset of this PR, what you remarked is semantically true. However this simply works today since AXIS2_MESSAGE is a UTF8ByteString that implements some kind of smart equals method and this method just want to see that span has been created by this decorator. Probably a == also would have worked

dd-java-agent/instrumentation/axis-2/src/main/java/datadog/trace/instrumentation/axis2/AxisMessageDecorator.java
ygree: Oh, I see. Then I guess the only option is to use toString as you did.

dd-java-agent/instrumentation/axis-2/src/main/java/datadog/trace/instrumentation/axis2/AxisMessageDecorator.java
ygree: You are right, if `span.getSpanName()` is also `UTF8BytesString` then it will work because of the way the equals method is implemented. Thanks!

ygree (approved):
Looks good! Thanks for responding to my concerns, which don't seem to be blockers for this PR.

PR #6540, opened 2024-01-23
Initial flush interval for long running spans

paul-laffon-dd:
# What Does This Do

Add the parameter `trace.experimental.long-running.initial.flush.interval` defaulting to 20s to control how long before flushing long running traces for the first time

# Motivation

Reduce the delay to see long running spans in datadog, without having to set a frequent flush interval 

# Additional Notes


dougqh (approved):
If I recall correctly, this interval is currently only used by data jobs monitoring, so I'm fine with the change.

PR #6547, opened 2024-01-24
Added Build version locking

nayeem-kamal:
# What Does This Do
Adds a new `resolveAndLockAll` task in the build. This step allows us to generate lockfiles based on the current project state so that we can use locked dependency versions for testing and building.

# Motivation
Locking dependency versions should help reduce flakiness during testing as well as documenting dependencies included with each build.

# Additional Notes
To generate updated lockfiles following new version releases or new `build.gradle` files, run `./gradlew resolveAndLockAll --write-locks`

Jira ticket: [APMJAVA-1179]

[APMJAVA-1179]: https://datadoghq.atlassian.net/browse/APMJAVA-1179?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ


bm1549 (approved):
Since @bantonsson approved this code in #6466 and this is identical, LGTM

It seems like there is some flakiness in gradle that needs future handling

PR #6549, opened 2024-01-24
Do not compute DBInfo on every statement call

manuel-alvarez-alvarez:
# What Does This Do
Uses context stores to cache the `DBInfo` instances computed for JDBC connections

# Motivation
The computation of the DBInfo is not cheap, so we should use the cached instances included in the context stores.


jandro996 (approved):
It would be great to add some test for the getDBInfo method

PR #6550, opened 2024-01-25
Fix header injection exclusions

jandro996:
# What Does This Do

Fix IAST Header Injection exclusions:

- access-control-allow-*: when the header starts with access-control-allow- and the source of the tainted range is a request header 
- set-cookie: We should ignore set-cookie header if the source of all the tainted ranges are cookies
- We should exclude the injection when the tainted string only has one range which comes from a request header with the same name that the header that we are checking in the response.

# Motivation

Discard false positives related with Header Injection

# Additional Notes

Extract test related with Header Injection to a new class

Jira ticket: [[PROJ-IDENT]](https://datadoghq.atlassian.net/browse/APPSEC-45949)

### DIFF
--- a/dd-java-agent/agent-iast/src/main/java/com/datadog/iast/sink/HeaderInjectionModuleImpl.java
+++ b/dd-java-agent/agent-iast/src/main/java/com/datadog/iast/sink/HeaderInjectionModuleImpl.java
@@ -1,10 +1,12 @@
 package com.datadog.iast.sink;
 
+import static com.datadog.iast.taint.Ranges.allRangesFromAnyHeader;
 import static com.datadog.iast.taint.Ranges.allRangesFromHeader;
+import static com.datadog.iast.taint.Ranges.rangeFromHeader;
 import static com.datadog.iast.util.HttpHeader.ACCESS_CONTROL_ALLOW_ORIGIN;
 import static com.datadog.iast.util.HttpHeader.CONNECTION;
+import static com.datadog.iast.util.HttpHeader.COOKIE;
 import static com.datadog.iast.util.HttpHeader.LOCATION;
-import static com.datadog.iast.util.HttpHeader.ORIGIN;
 import static com.datadog.iast.util.HttpHeader.SEC_WEBSOCKET_ACCEPT;
 import static com.datadog.iast.util.HttpHeader.SEC_WEBSOCKET_LOCATION;
 import static com.datadog.iast.util.HttpHeader.SET_COOKIE;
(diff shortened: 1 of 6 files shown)


dd-java-agent/agent-iast/src/main/java/com/datadog/iast/sink/HeaderInjectionModuleImpl.java
```diff
@@ -64,17 +66,21 @@ public void onHeader(@Nonnull final String name, @Nullable final String value) {
       return;
     }
 
-    // TODO (spec) we should fix this as it's not a 100% accurate
-    if (header == ACCESS_CONTROL_ALLOW_ORIGIN && allRangesFromHeader(ORIGIN, ranges)) {
+    // Exclude access-control-allow-*: when the header starts with access-control-allow- and the
+    // source of the tainted range is a request header
+    if (header == ACCESS_CONTROL_ALLOW_ORIGIN && allRangesFromAnyHeader(ranges)) {
```
manuel-alvarez-alvarez, line 71: This only checks `Access-Control-Allow-Origin` but should be all headers following the pattern `Access-Control-Allow-*`

jandro996: I've talked about this topic with @uurien and we keep ip open to any header that starts with access-control-allow-

dd-java-agent/agent-iast/src/main/java/com/datadog/iast/sink/HeaderInjectionModuleImpl.java
```diff
@@ -64,17 +68,22 @@ public void onHeader(@Nonnull final String name, @Nullable final String value) {
       return;
     }
 
-    // TODO (spec) we should fix this as it's not a 100% accurate
-    if (header == ACCESS_CONTROL_ALLOW_ORIGIN && allRangesFromHeader(ORIGIN, ranges)) {
+    // Exclude access-control-allow-*: when the header starts with access-control-allow- and the
+    // source of the tainted range is a request header
+    if (name.toUpperCase(Locale.ROOT).startsWith(ACCESS_CONTROL_ALLOW_PREFIX)
```
manuel-alvarez-alvarez, line 73:
```suggestion
    if (name.regionMatches(true, 0, ACCESS_CONTROL_ALLOW_PREFIX, ACCESS_CONTROL_ALLOW_PREFIX.length())
```

manuel-alvarez-alvarez (approved):

PR #6554, opened 2024-01-25
Cache IAST metric tag values

manuel-alvarez-alvarez:
# What Does This Do
Caches the tag values to be used in telemetry metrics and also the values to be used in the span tags in order to troubleshoot IAST.

# Motivation
There's quite a performance overhead of having to compute those tags every time, having them cached reduced the number of string allocation and the time spent while generating the metrics.


manuel-alvarez-alvarez:
Closed as too many people where added to the reviewers, sorry for the noise.

PR #6555, opened 2024-01-25
Add `aws.object.key` tag for aws-sdk-1

paul-laffon-dd:
# What Does This Do

Add `aws.object.key` for aws-sdk-1 to know which key is targeted by S3 get/put requests

# Motivation

More information on S3 request by having the key of the targeted object. It also unlocks object-level data lineage for S3

# Additional Notes

Jira ticket: [PROJ-IDENT]

### DIFF
--- a/dd-java-agent/instrumentation/aws-java-sdk-1.11.0/src/main/java/datadog/trace/instrumentation/aws/v0/AwsSdkClientDecorator.java
+++ b/dd-java-agent/instrumentation/aws-java-sdk-1.11.0/src/main/java/datadog/trace/instrumentation/aws/v0/AwsSdkClientDecorator.java
@@ -113,6 +113,10 @@ public class AwsSdkClientDecorator extends HttpClientDecorator<Request, Response
     String bestPrecursor = null;
     String bestPeerService = null;
 
+    String key = access.getKey(originalRequest);
+    if (null != key) {
+      span.setTag(InstrumentationTags.AWS_OBJECT_KEY, key);
+    }
     String bucketName = access.getBucketName(originalRequest);
     if (null != bucketName) {
       span.setTag(InstrumentationTags.AWS_BUCKET_NAME, bucketName);
--- a/dd-java-agent/instrumentation/aws-java-sdk-1.11.0/src/main/java/datadog/trace/instrumentation/aws/v0/GetterAccess.java
+++ b/dd-java-agent/instrumentation/aws-java-sdk-1.11.0/src/main/java/datadog/trace/instrumentation/aws/v0/GetterAccess.java
@@ -28,6 +28,7 @@ final class GetterAccess {
       Pattern.compile("Request", Pattern.LITERAL);
 
   private final String operationName;
+  private final MethodHandle getKey;
   private final MethodHandle getBucketName;
   private final MethodHandle getQueueUrl;
   private final MethodHandle getQueueName;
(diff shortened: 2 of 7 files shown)


dd-java-agent/instrumentation/aws-java-sdk-1.11.0/src/main/java/datadog/trace/instrumentation/aws/v0/AwsSdkClientDecorator.java
```diff
@@ -113,6 +113,10 @@ public AgentSpan onRequest(final AgentSpan span, final Request request) {
     String bestPrecursor = null;
     String bestPeerService = null;
 
+    String key = access.getKey(originalRequest);
```
dougqh, line 116: Is it safe to assume key is a String?

paul-laffon-dd: Depending on the request type, the key is most often a String, but not always ([like for dynamoDB where it is a Map](https://github.com/aws/aws-sdk-java/blob/master/aws-java-sdk-dynamodb/src/main/java/com/amazonaws/services/dynamodbv2/model/GetItemRequest.java#L261-L263))

The [findStringGetter](https://github.com/DataDog/dd-trace-java/blob/master/dd-java-agent/instrumentation/aws-java-sdk-1.11.0/src/main/java/datadog/trace/instrumentation/aws/v0/GetterAccess.java#L133-L140) used to [find the getKey](https://github.com/DataDog/dd-trace-java/pull/6555/files#diff-e0c40d0291e27fd55e328409b018ab3691a42908348798fc3c4f14d1238d5fb0R45) method will only match for methods returning a String, returning null otherwise (and if the methodHandle is null, [invokeForString](https://github.com/DataDog/dd-trace-java/blob/master/dd-java-agent/instrumentation/aws-java-sdk-1.11.0/src/main/java/datadog/trace/instrumentation/aws/v0/GetterAccess.java#L96-L105) will return null)
```
private static MethodHandle findStringGetter(
    final Class<?> requestType, final String methodName) {
  try {
    return PUBLIC_LOOKUP.findVirtual(requestType, methodName, STRING_RETURN_TYPE);
  } catch (Throwable e) {
    return null;
  }
}
```

But I did not find any tests for requests where the getKey is not a String, will add one

dougqh (approved):

dd-java-agent/instrumentation/aws-java-sdk-1.11.0/src/main/java/datadog/trace/instrumentation/aws/v0/GetterAccess.java
kr-igor, line 1: @paul-laffon-dd what about AWS SDK 2.2?

kr-igor (approved):

dd-java-agent/instrumentation/aws-java-sdk-1.11.0/src/main/java/datadog/trace/instrumentation/aws/v0/GetterAccess.java
paul-laffon-dd: Yes, it can be added to AWS SDK 2 in a subsequent PR. Started with version 1 since it is the one still used by spark

PR #6558, opened 2024-01-25
Improve the performance of the IAST cookie parser

manuel-alvarez-alvarez:
# What Does This Do
Improves the performance of the cookie parser by reducing string allocations and avoiding going over the string more than once, and only creating new strings for what is needed

# Motivation
In applications making heavy usage of cookies the parse might become effectively a bottleneck.

### DIFF
--- a/dd-java-agent/agent-iast/src/main/java/com/datadog/iast/util/CookieSecurityParser.java
+++ b/dd-java-agent/agent-iast/src/main/java/com/datadog/iast/util/CookieSecurityParser.java
@@ -1,130 +1,146 @@
 package com.datadog.iast.util;
 
+import static com.datadog.iast.util.HttpHeader.SET_COOKIE2;
 import static java.util.Collections.emptyList;
 
 import datadog.trace.api.iast.util.Cookie;
 import java.util.ArrayList;
+import java.util.Collections;
 import java.util.List;
-import java.util.Locale;
-import java.util.NoSuchElementException;
-import java.util.StringTokenizer;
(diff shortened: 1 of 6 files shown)


dd-java-agent/agent-iast/src/main/java/com/datadog/iast/util/CookieSecurityParser.java
```diff
@@ -1,130 +1,126 @@
...
-      }
-    }
-
-    cookies.add(header.substring(q));
-
-    return cookies;
-  }
-
-  private static int guessCookieVersion(String header) {
-    header = header.toLowerCase(Locale.ROOT);
```
manuel-alvarez-alvarez, line 117: This is not good in case the cookie does not contain any of the strings :cry:

dd-java-agent/agent-iast/src/main/java/com/datadog/iast/util/CookieSecurityParser.java
```diff
@@ -1,130 +1,126 @@
...
-      }
-      if (SAME_SITE_ATTR.equalsIgnoreCase(name)) {
-        sameSite = value;
-      }
-    }
-    return new Cookie(cookieName, secure, httpOnly, sameSite);
-  }
-
-  private static List<String> splitMultiCookies(final String header) {
-    List<String> cookies = new java.util.ArrayList<String>();
```
manuel-alvarez-alvarez, line 97: New version tries to tokenize multiple cookies at once

PR #6560, opened 2024-01-26
Get access to tag values from the top of TraceSegments

manuel-alvarez-alvarez:
# What Does This Do
Add getters to query for data from the top of the current trace segment.

# Motivation
From IAST we want to be able to add vulnerabilities even when there is no active request (e.g. Kafka)

# Additional Notes

Jira ticket: [APPSEC-10440]

This is required to make IAST work in Kafka consumers: https://github.com/DataDog/dd-trace-java/pull/6465

[APPSEC-10440]: https://datadoghq.atlassian.net/browse/APPSEC-10440?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ

### DIFF
--- a/dd-java-agent/agent-iast/src/main/java/com/datadog/iast/Reporter.java
+++ b/dd-java-agent/agent-iast/src/main/java/com/datadog/iast/Reporter.java
@@ -1,6 +1,7 @@
 package com.datadog.iast;
 
 import static com.datadog.iast.IastTag.ANALYZED;
+import static datadog.trace.api.telemetry.LogCollector.SEND_TELEMETRY;
 
 import com.datadog.iast.model.Vulnerability;
 import com.datadog.iast.model.VulnerabilityBatch;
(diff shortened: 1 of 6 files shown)


dd-java-agent/agent-iast/src/main/java/com/datadog/iast/Reporter.java
```diff
@@ -60,24 +66,44 @@ public void report(@Nullable final AgentSpan span, @Nonnull final Vulnerability
...
+      segment.setDataTop(IAST_TAG, batch);
       // Once we have added a vulnerability, try to override sampling and keep the trace.
       // TODO: We need to check if we can have an API with more fine-grained semantics on why traces
       // are kept.
       segment.setTagTop(DDTags.MANUAL_KEEP, true);
+      return batch;
     }
+
+    // TODO: can we do better here? (maybe create a new span)
+    LOG.error("Failed to attach vulnerability to the span, current object {}", current.getClass());
```
smola, line 105: Maybe debug level? Definitely not error. If this is happening, I don't think we want to unnecessarily worry users. We might want to use the `SEND_TELEMETRY` marker to send it to telemetry though (or attach an exception).

smola (commented):
Looks good to me other than the log level I mentioned.

It'd be good for APM core to review TracerSegment/DDSpanContext changes.

dd-java-agent/agent-iast/src/main/java/com/datadog/iast/Reporter.java
manuel-alvarez-alvarez: Changed to use the telemetry marker.

jandro996 (approved):

PR #6567, opened 2024-01-26
Simplify testing of telemetry log redaction

smola:
# What Does This Do

# Motivation
Relying on exceptions thrown from tests is hard if we want exhaustive tests, since they can vary a lot per JDK version and vendor.

# Additional Notes


telemetry/src/test/groovy/datadog/telemetry/log/LogPeriodicActionTest.groovy
```diff
@@ -99,4 +108,54 @@ class LogPeriodicActionTest extends DDSpecification {
...
+
+  static StackTraceElement[] stacktrace(StackTraceElement... frames) {
+    return frames
+  }
+
+  static StackTraceElement frame(String className) {
+    return new StackTraceElement(className, "method", "file", 42)
+  }
+
+  static StackTraceElement frame(String className, String methodName, String fileName, int lineNumber) {
```
manuel-alvarez-alvarez, line 158: You can have default parameter values in groovy and join both methods:

```groovy
static StackTraceElement frame(String className, String methodName = 'method', String fileName = 'file', int lineNumber = 42)
  return new StackTraceElement(className, methodName, fileName, lineNumber)
}
```

manuel-alvarez-alvarez (approved):

telemetry/src/test/groovy/datadog/telemetry/log/LogPeriodicActionTest.groovy
smola: Thanks, done!

PR #6574, opened 2024-01-29
Avoid concurrent modification failures in tests

amarziali:
# What Does This Do

Currently some tests are failing because:
```
02:07:13.435 [jetty-http-server-7] DEBUG datadog.trace.bootstrap.ExceptionLogger - Failed to handle exception in instrumentation for org.eclipse.jetty.server.HttpChannel
java.util.ConcurrentModificationException: null
	at java.base/java.util.ArrayList$Itr.checkForComodification(ArrayList.java:1013)
	at java.base/java.util.ArrayList$Itr.next(ArrayList.java:967)
	at datadog.trace.common.writer.TraceStructureWriter.write(TraceStructureWriter.java:103)
	at datadog.trace.common.writer.ListWriter.write(ListWriter.java:61)
	at datadog.trace.core.CoreTracer.write(CoreTracer.java:942)
	at datadog.trace.core.PendingTrace.write(PendingTrace.java:373)
	at datadog.trace.core.PendingTrace.write(PendingTrace.java:339)
	at datadog.trace.core.PendingTrace.decrementRefAndMaybeWrite(PendingTrace.java:309)
	at datadog.trace.core.PendingTrace.onPublish(PendingTrace.java:257)
	at datadog.trace.core.DDSpan.finishAndAddToTrace(DDSpan.java:154)
	at datadog.trace.core.DDSpan.finish(DDSpan.java:165)
	at org.eclipse.jetty.server.HttpChannel.recycle(HttpChannel.java:445)
	at org.eclipse.jetty.server.HttpChannelOverHttp.recycle(HttpChannelOverHttp.java:569)
	at org.eclipse.jetty.server.HttpConnection.onCompleted(HttpConnection.java:473)
	at org.eclipse.jetty.server.HttpChannel.onCompleted(HttpChannel.java:968)
	at org.eclipse.jetty.server.HttpChannel.handle(HttpChannel.java:485)
	at org.eclipse.jetty.server.HttpConnection.onFillable(HttpConnection.java:282)
	at org.eclipse.jetty.io.AbstractConnection$ReadCallback.succeeded(AbstractConnection.java:314)
	at org.eclipse.jetty.io.FillInterest.fillable(FillInterest.java:100)
	at org.eclipse.jetty.io.SelectableChannelEndPoint$1.run(SelectableChannelEndPoint.java:53)
	at org.eclipse.jetty.util.thread.QueuedThreadPool.runJob(QueuedThreadPool.java:969)
	at org.eclipse.jetty.util.thread.QueuedThreadPool$Runner.doRunJob(QueuedThreadPool.java:1194)
	at org.eclipse.jetty.util.thread.QueuedThreadPool$Runner.run(QueuedThreadPool.java:1149)
	at java.base/java.lang.Thread.run(Thread.java:840)
```

This PR just add a shallow copy before passing the list to the structure writer. Probably the clash is because the list assert is also sorting the list?
This should fix anyway the issue and the overhead of shallow copying is OK since structure writer is mostly used for testing

# Motivation

# Additional Notes

Jira ticket: [PROJ-IDENT]


bantonsson:
@amarziali I think the burden of copying should be on the trace asserts. There should be some type of guarantee for the writers that they can assume that no modifications should happen to the lists while they are writing.

amarziali:
> @amarziali I think the burden of copying should be on the trace asserts. There should be some type of guarantee for the writers that they can assume that no modifications should happen to the lists while they are writing.

@bantonsson I agree but to be honest I cannot understand where this modification is coming. The list assert seems already copying things. Today we can see this failure thanks to the fact we fail tests on unhandled instrumentation exceptions .
Despite it does not fix the root cause, what  I proposed is the best tradeoff I had in terms of time investment to fix this

bantonsson:
> @bantonsson I agree but to be honest I cannot understand where this modification is coming. The list assert seems already copying things. Today we can see this failure thanks to the fact we fail tests on unhandled instrumentation exceptions .
Despite it does not fix the root cause, what I proposed is the best tradeoff I had in terms of time investment to fix this

@amarziali I understand that this is the fastest way to make the tests pass, but if we don't understand the root cause, then maybe we have a real problem that could break the real writer?

dd-trace-core/src/main/java/datadog/trace/common/writer/ListWriter.java
```diff
@@ -58,7 +56,7 @@ public void write(List<DDSpan> trace) {
         }
       }
     }
-    structureWriter.write(trace);
+    structureWriter.write(new ArrayList<>(trace));
```
bantonsson, line 59: The problem is in `AppSecInactiveHttpServerTest` on this line https://github.com/DataDog/dd-trace-java/blob/61cc2465df7bc8a3e2076716bf1158ad44295e61/dd-java-agent/appsec/src/testFixtures/groovy/com/datadog/appsec/AppSecInactiveHttpServerTest.groovy#L50

amarziali: thanks a lot for having narrowed the issue. I used the mutate = false version of the sort

bantonsson (approved):

PR #6577, opened 2024-01-29
Improve header injection redaction

jandro996:
# What Does This Do

Headers could store sensitive information, we should redact whole <header_value> if:

- <header_name> matches with configured sensitive name pattern
- <header_value> matches with configured sensitive value pattern

We should redact the sensitive information from the evidence when tainted range is considered sensitive value

# Motivation

# Additional Notes

Jira ticket: [[PROJ-IDENT]](https://datadoghq.atlassian.net/browse/APPSEC-45956)

### DIFF
--- a/dd-java-agent/agent-iast/src/main/java/com/datadog/iast/sensitive/HeaderRegexpTokenizer.java
+++ b/dd-java-agent/agent-iast/src/main/java/com/datadog/iast/sensitive/HeaderRegexpTokenizer.java
@@ -42,9 +42,10 @@ public class HeaderRegexpTokenizer implements SensitiveHandler.Tokenizer {
   private Ranged buildNext() {
     if (!checked) {
       checked = true;
-      String[] split = evidenceValue.split(": ");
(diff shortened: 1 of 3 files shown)


dd-java-agent/agent-iast/src/main/java/com/datadog/iast/sensitive/HeaderRegexpTokenizer.java
```diff
@@ -0,0 +1,55 @@
...
+  }
+
+  @Nullable
+  private Ranged buildNext() {
+    if (!checked) {
+      checked = true;
+      // Header evidence format is <headerName>: <headerValue>
+      int index = evidenceValue.indexOf(":");
+      String name = evidenceValue.substring(0, index);
+      String value = evidenceValue.substring(index + 1);
```
smola, line 48: Same here if the evidence turns out to be `<headerName>:`, we'd have an out of bounds error here.

smola (approved):

PR #6580, opened 2024-01-29
Overhauling Matchers to support non-String types

dougqh:
# What Does This Do

Add matcher support for non-String tag types

# Motivation

Previously TagsMatcher would raise a ClassCastException if a tag had a non-String value

# Additional Notes

This change introduces matches methods for Object & primitive types. For numeric types, the matches routines only support integer values even for floating point types. This was done for consistency with languages that only support floating point, but to avoid matching complexities posed by floating point impression. The bulk of the handling of other types is done by the new BaseMatcher class from which PatternMatcher now extends.

To handle the "*" case, a new AnyMatcher is introduced which accepts all values including non-integral floating point values.  This changeset also adds support for equivalent patterns like "**" and "***".

New tests have been added to RuleSamplingTest & MatcherTests to cover a wide range of tag types.
Jira ticket: [PROJ-IDENT]


dd-trace-core/src/test/groovy/datadog/trace/common/sampling/RuleBasedSamplingTest.groovy
```diff
@@ -271,6 +271,74 @@ class RuleBasedSamplingTest extends DDCoreSpecification {
...
+    "foo"       | "not-foo"                 | false
+    "ba?"       | "bar"                     | true
+    "20"        | 20                        | true
+    "20"        | Integer.valueOf(20)       | true
+    "20"        | 20L                       | true
+    "20"        | Long.valueOf(20)          | true
+    "20"        | 20F                       | true
+    "20"        | 20.1F                     | false
+    "20.*"      | 20.1F                     | false
+    "20"        | 20F                       | true
```
ygree, line 334: This case already in line 331. Should it be `"20.1" | 20 F | true`?

ygree (approved):
LGTM

dd-trace-core/src/test/groovy/datadog/trace/common/sampling/RuleBasedSamplingTest.groovy
dougqh: Good catch, I'll clean that up.

dougqh:
Abandoning this branch, so I clean-up the history

dougqh:
Sorted out history - trying to salvage this PR

dougqh:
Sorted out history - trying to salvage this PR

dougqh:
Sorted out history - trying to salvage this PR

dougqh:
.

PR #6583, opened 2024-01-29
Report default null config values with Telemetry events

ygree:
# What Does This Do

Report default null config values with Telemetry events

# Motivation

All default config values must be reported even if it's a null

# Additional Notes

Jira ticket: [PROJ-IDENT]


smola (approved):

smola (commented):
Ideally there should be at least 1 test?

ygree:
Close in favor of #6694

PR #6584, opened 2024-01-30
Mute tracing for subprocesses executed in by the tracer

smola:
# What Does This Do
The tracer sometimes runs external processes in some cases. Currently, `hostname` and `getenforce`. The subprocess instrumentation creates spans for these, leading to visible `command_execution` spans for `hostname` and/or `enforce` which may be confusing to users.

This commit uses the bew blackhole span API (#6326) to mute tracing during when `hostname` and `getenforce` are run, preventing the creation of these spans.

Also reverts #6478, which would be redundant with the current cases.

# Motivation

# Additional Notes

Jira ticket: [PROJ-IDENT]


jbachorik (approved):
Looking good for the profiler part

mcculls (approved):

amarziali (approved):

PR #6586, opened 2024-01-31
Fix insecure cookies false positives when unsetting cookies

jandro996:
# What Does This Do

Add value, maxAge and expiresYear to Cookie class.
Update CookieParser with the new fields.
Modify InsecureCookieModuleImpl#isVulnearable to avoid report insecure cookie vulnerability when:

- value is empty
- maxAge is 0
- expires year is older than 2000

# Motivation

We have a customer with detection in a method called unsetCookies, this suggest we’re not supporting this part, or that there’s some additional case.

After some research we noticed that there are several ways to invalidate cookies, for instance:

[https://github.com/akka/akka-http/blob/595cc45d0581a501333b64951ca0381bcfc197aa/ak[…]ala/akka/http/scaladsl/server/directives/CookieDirectives.scala](https://github.com/akka/akka-http/blob/595cc45d0581a501333b64951ca0381bcfc197aa/akka-http/src/main/scala/akka/http/scaladsl/server/directives/CookieDirectives.scala#L57-L60)
`Set-Cookie`(c.withValue("deleted").withExpires(DateTime.MinValue))

# Additional Notes

Jira ticket: [APPSEC-46961](https://datadoghq.atlassian.net/browse/APPSEC-46961)

[APPSEC-46961]: https://datadoghq.atlassian.net/browse/APPSEC-46961?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ


dd-java-agent/agent-iast/src/main/java/com/datadog/iast/util/CookieSecurityParser.java
```diff
@@ -73,9 +73,17 @@ public static List<Cookie> parse(final HttpHeader headerName, String headerValue
           switch (state) {
             case COOKIE_NAME:
               cookieName = headerValue.substring(start, end).trim();
```
manuel-alvarez-alvarez, line 75:
```suggestion
              cookieName = headerValue.substring(start, next == '=' ? end - 1 : end).trim();
```
Then you can drop the extra check

dd-java-agent/agent-iast/src/main/java/com/datadog/iast/util/CookieSecurityParser.java
```diff
@@ -109,23 +132,49 @@ public static List<Cookie> parse(final HttpHeader headerName, String headerValue
...
+    } catch (final NumberFormatException e) {
+      LOG.debug(SEND_TELEMETRY, "Failed to parse the max-age {}", headerValue);
+      return null;
+    }
+  }
+
+  @Nullable
+  private static Integer parseExpires(final String value, final String headerValue) {
+    try {
+      String[] tokens = value.split(" ");
```
manuel-alvarez-alvarez, line 170: Don't use split if you only care about on of the tokens.

dd-java-agent/agent-iast/src/main/java/com/datadog/iast/util/CookieSecurityParser.java
```diff
@@ -77,30 +84,46 @@ public static List<Cookie> parse(final HttpHeader headerName, String headerValue
               state = next == '=' ? COOKIE_VALUE : COOKIE_ATTR_NAME;
               break;
             case COOKIE_VALUE:
+              cookieValue = headerValue.substring(start, end).trim();
```
manuel-alvarez-alvarez, line 87: There is an small issue here, cookies by the [RFC](https://datatracker.ietf.org/doc/html/rfc6265#section-4.1.1) can optionally be wrapped around double quotes, if first and last character are `"` we should remove them, as `""` is effectively an empty value.

manuel-alvarez-alvarez (changes requested):

PR #6589, opened 2024-01-31
Add CI Static Analysis integration

smola:
# What Does This Do
Add a GitHub Workflow to run DataDog's Static Analyzer.

# Motivation

# Additional Notes

* See [docs](https://docs.datadoghq.com/static_analysis/github_actions)
* And https://github.com/DataDog/datadog-static-analyzer-github-action

### DIFF
--- a/.github/workflows/ci-static-analysis.yml
+++ b/.github/workflows/ci-static-analysis.yml
@@ -1,0 +1,25 @@
+name: "DataDog CI Static Analysis"
+on:
+  push:
+    branches: [ master ]
+  pull_request:
+    # The branches below must be a subset of the branches above
+    branches: [ master ]
+jobs:
+  check-quality:
+    runs-on: ubuntu-latest
+    name: Datadog Static Analyzer
+    steps:
+      - name: Checkout repository
+        uses: actions/checkout@5a4ac9002d0be2fb38bd78e4b4dbde5606d7042f # 2.3.4
+      - name: Check code meets quality standards
+        id: datadog-static-analysis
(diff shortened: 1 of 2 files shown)


.github/workflows/ci-static-analysis.yml
```diff
@@ -0,0 +1,25 @@
...
+jobs:
+  check-quality:
+    runs-on: ubuntu-latest
+    name: Datadog Static Analyzer
+    steps:
+      - name: Checkout repository
+        uses: actions/checkout@5a4ac9002d0be2fb38bd78e4b4dbde5606d7042f # 2.3.4
+      - name: Check code meets quality standards
+        id: datadog-static-analysis
+        uses: DataDog/datadog-static-analyzer-github-action@8f2b6213cadc5fef47befda2dc3f1d7c68f33ddd # 1.0.7
```
markazerdd, line 17: @smola I haven't seen anyone specify this before, I'd recommend following whatever is in the onboarding UI to make sure you always get the latest version of the analyzer instead of the same version forever

PerfectSlayer (approved):
LGTM. 

Do we have to store the config at the root to the project?
I read at the action doc stating we have and found no configuration to move it elsewhere 😞

.github/workflows/ci-static-analysis.yml
PerfectSlayer: This is mandatory for non-entreprise actions but our organisation settings allows us to use unpinned enterprise action.

smola:
@PerfectSlayer I think it is a limitation of the static analyzer, not just for the GitHub Action, but also as used by the IntelliJ plugin. It would definitely be nice to have some sort of lookup path list (e.g. `.ci`, `.github`, and such).

I'm merging atm, but we can change this if this limitation is removed.

PR #6591, opened 2024-01-31
Adding support for tag matching of non-String types

dougqh:
# What Does This Do

This change introduces matching support for non-String tag values.

# Motivation

Additional supported types include CharSequences, booleans, integer types, and limited support for floating types.

Floating point support is limited to integer values.  This was done for consistency with languages that only support floating point while avoiding matching complexities posed by floating point impression.

# Additional Notes

The bulk of the handling of other types is done by the new BaseMatcher class from which PatternMatcher now extends.

This change introduces matches methods for Object & primitive types. For numeric types, the matches routines only support integer values even for floating point types.

To handle the any star case, a new AnyMatcher is introduced which accepts all values including non-integral floating point values. This changeset also adds support for any all stars pattern.

Jira ticket: [APMJAVA-1207]

[APMJAVA-1207]: https://datadoghq.atlassian.net/browse/APMJAVA-1207?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ


dd-trace-core/src/test/groovy/datadog/trace/core/util/MatchersTest.groovy
```diff
@@ -4,12 +4,12 @@ import datadog.trace.test.util.DDSpecification
 
 class MatchersTest extends DDSpecification {
 
-  def "match-all scenarios must return a null matcher"() {
+  def "match-all scenarios must return an any matcher"() {
     expect:
-    Matchers.compileGlob(glob) == null
+    Matchers.compileGlob(glob).isAny()
```
ygree, line 9: Looks like this is the only use of the isAny method. Maybe remove it altogether and just check the type of the Matcher returned by the compileGlob method?

dougqh: Fair enough, I'll do that

PR #6597, opened 2024-02-01
Add _dd.appsec.waf.timeouts tag

smola:
# What Does This Do

This tag counts number of timeouts in a request.

# Motivation

# Additional Notes

Jira ticket: [APPSEC-51507](https://datadoghq.atlassian.net/browse/APPSEC-51507)

[APPSEC-51507]: https://datadoghq.atlassian.net/browse/APPSEC-51507?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ

### DIFF
--- a/dd-java-agent/appsec/src/main/java/com/datadog/appsec/gateway/AppSecRequestContext.java
+++ b/dd-java-agent/appsec/src/main/java/com/datadog/appsec/gateway/AppSecRequestContext.java
@@ -80,6 +80,7 @@ public class AppSecRequestContext implements DataBundle, Closeable {
   // set after additive is set
   private volatile PowerwafMetrics wafMetrics;
   private volatile boolean blocked;
+  private volatile long timeouts;
 
   // to be called by the Event Dispatcher
   public void addAll(DataBundle newData) {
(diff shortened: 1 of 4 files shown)


manuel-alvarez-alvarez (approved):

dd-java-agent/appsec/src/main/java/com/datadog/appsec/gateway/AppSecRequestContext.java
```diff
@@ -114,6 +115,14 @@ public boolean isBlocked() {
     return blocked;
   }
 
+  public void increaseTimeouts() {
+    this.timeouts++;
```
cataphract, line 119: this is not atomic. You need to use AtomicIntegerFieldUpdater.

jandro996 (approved):

dd-java-agent/appsec/src/main/java/com/datadog/appsec/gateway/AppSecRequestContext.java
smola: @cataphract Good point. I have not used `AtomicIntegerFieldUpdater` before. Is there any caveat compared to using `AtomicInteger`?

cataphract: you can also use `AtomicInteger`, but it will have a penalty in memory/indirection because then you have an object rather than an int. Of course, there's an extra `AtomicIntegerFieldUpdater` object, but that can be static.

cataphract (approved):
lgtm

PR #6484, opened 2024-01-15
Upload SARIF from CodeQL to CI Static Analysis

smola:
# What Does This Do

# Motivation

# Additional Notes

To Do:

* [ ] Check we have a proper API token in GitHub Workflows.

PR #6600, opened 2024-02-01
Run CodeQL in master only

smola:
# What Does This Do

Moves CodeQL job to master only. The job takes 30min to run, and results are generally not very interesting (we'll still get them through CI Static Analysis (see https://github.com/DataDog/dd-trace-java/pull/6484)). We'll still have DataDog Static Analyzer in PRs, which is more comprehensive and takes just 2 minutes to run (see https://github.com/DataDog/dd-trace-java/pull/6589).

# Motivation

# Additional Notes

### DIFF
--- a/.github/workflows/codeql-analysis.yml
+++ b/.github/workflows/codeql-analysis.yml
@@ -3,9 +3,6 @@ name: "CodeQL"
 on:
   push:
     branches: [ master ]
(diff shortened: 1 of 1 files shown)


PerfectSlayer (approved):
Seems fair. Thanks!

PR #6603, opened 2024-02-01
Fix dependency resolution for nested JARs

smola:
# What Does This Do
When finding nested jar (e.g. in Spring Boot), we were only applying a
subset of the dependency resolution logic. Now the same heuristics are
applied to nested jars.

In practice, this means that for nested JARs we will:
* Identify the dependency more accurately (e.g. extract group id).
* Find shaded dependencies.

# Motivation

# Additional Notes

Jira ticket: [APPSEC-11822](https://datadoghq.atlassian.net/browse/APPSEC-11822)

[APPSEC-11822]: https://datadoghq.atlassian.net/browse/APPSEC-11822?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ

### DIFF
--- a/telemetry/src/main/java/datadog/telemetry/dependency/Dependency.java
+++ b/telemetry/src/main/java/datadog/telemetry/dependency/Dependency.java
@@ -128,6 +128,11 @@ public final class Dependency {
 
   public static synchronized Dependency guessFallbackNoPom(
       Manifest manifest, String source, InputStream is) throws IOException {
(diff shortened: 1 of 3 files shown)


PerfectSlayer (approved):
Looks good

PR #6605, opened 2024-02-01
Do not send agent API errors to telemetry logs

smola:
# What Does This Do
* Introduce a logging marker (NO_SEND_TELEMETRY) to exclude a log message from telemetry even if other criteria is met. This is meant to be used for very frequent errors that are better handled by metrics or other reporting mechanisms.
* Use it for agent API errors (which may use trace_api.responses/trace_api.errors telemetry metrics, eventually).

# Motivation
Reduce the volume of telemetry messages we receive.

# Additional Notes


PerfectSlayer (approved):
Nice addition.

About the marker itself, I would prefer having an action verb rather than a negative form.
I am not really sure about how you call it in English but `exclude`, `drop`, `filter`, `block` rather than `no_send`.

smola:
@PerfectSlayer Makes sense. I renamed it to `EXCLUDE_TELEMETRY`.

PR #6609, opened 2024-02-02
Add instrumentation of exception stacktraces

jpbempel:
# What Does This Do
When instrumenting exception on first occurence, we get the stacktrace of the exception and for each frame we will generate a special method probe (ExceptionProbe) based on LogProbe doing a snapshot at Exit. to place correctly the probes we are basing on StackTrace element that gives class, method and line number.
for overloaded methods (with same name) the only way to disambiguate which method to instrument is to use the line number. We introduce then the ClassFileLines class to generalize the LineNumber information extracted from the parsing of a classfile by the ASM library and be reused for line instrumentation and method matching. previous LineMap class is then replaced by it.
Created exception probe are kept by ExceptionProbeManager and must be merged with the Configuration coming from Remote Config. Therefore the ConfigurationUpdater class must now take into account that and also be able to be triggered not only by the Remote Config thread but also the thread coming from the exception captured added to the span. Applying configuration require now to be atomic between Configuration object, new transformer and retransformation of the classes impacted.

# Motivation
Exception debugging feature

# Additional Notes

Jira ticket: [DEBUG-2068]


dd-java-agent/agent-debugger/src/main/java/com/datadog/debugger/exception/DefaultExceptionDebugger.java
```diff
@@ -27,6 +31,7 @@ public void handleException(Throwable t) {
       // TODO trigger send snapshots already captured
     } else {
       exceptionProbeManager.instrument(fingerprint, t.getStackTrace());
```
shatzi, line 33: I would suggest rename "instrument" to "createProbesForException" - it represent better what this function is doing as it not really instrument.

dd-java-agent/agent-debugger/src/main/java/com/datadog/debugger/exception/DefaultExceptionDebugger.java
```diff
@@ -27,6 +31,7 @@ public void handleException(Throwable t) {
...
+      configurationUpdater.reapplyCurrentConfig();
```
shatzi, line 34: like we talked. I think that because "reapplyCurrentConfig" is blocking it should not be called inside the hot path of span.attachError.

Maybe this can stay here is configurationUpdater would reapply the configuration async.

dd-java-agent/agent-debugger/src/main/java/com/datadog/debugger/agent/ConfigurationUpdater.java
```diff
@@ -102,46 +92,62 @@ public void accept(Configuration configuration) {
...
-        new ConfigurationComparer(currentConfiguration, newConfiguration, instrumentationResults);
-    currentConfiguration = newConfiguration;
-    if (changes.hasRateLimitRelatedChanged()) {
-      // apply rate limit config first to avoid racing with execution/instrumentation of log probes
-      applyRateLimiter(changes);
-    }
-    if (changes.hasProbeRelatedChanges()) {
-      LOGGER.info("Applying new probe configuration, changes: {}", changes);
-      handleProbesChanges(changes);
+    configurationLock.lock();
```
shatzi, line 109: one things we can get benefits from making is async is "batching".
assume we got multiple exception triggered and we calling applyConfiguration 3 time each time with few more probes.
This code can check if we are currently applying configuration, and if we do, just mark that the latestConfig needed to be applied. this could happen multiple times if retransformation taking along time, and every time it happen we basically skipping applying a old configuration that will no longer relevant or if lock is not ordered we could be in a case where we apply a configuration that is not the latest.

shatzi (commented):
leaving on comment until we fix the hot-path issue.

dd-java-agent/agent-debugger/src/main/java/com/datadog/debugger/agent/ConfigurationUpdater.java
```diff
@@ -51,46 +54,33 @@ DebuggerTransformer supply(
...
       DebuggerSink sink,
-      ClassesToRetransformFinder finder) {
+      ClassesToRetransformFinder finder,
+      ExceptionProbeManager exceptionProbeManager) {
     this.instrumentation = instrumentation;
     this.transformerSupplier = transformerSupplier;
     this.serviceName = TagsHelper.sanitize(config.getServiceName());
     this.sink = sink;
     this.finder = finder;
+    this.exceptionProbeManager = exceptionProbeManager;
```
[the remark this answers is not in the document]
jpbempel: I understand the idea, but let's have another source before make an abstraction out of it.

dd-java-agent/agent-debugger/src/main/java/com/datadog/debugger/exception/DefaultExceptionDebugger.java
jpbempel: yep will do later

dd-java-agent/agent-debugger/src/main/java/com/datadog/debugger/agent/ConfigurationUpdater.java
jpbempel: yes with async and a queue we could conflate to discard intermediate configs

dd-java-agent/agent-debugger/src/main/java/com/datadog/debugger/instrumentation/ClassFileInfo.java
```diff
@@ -0,0 +1,40 @@
+package com.datadog.debugger.instrumentation;
+
+import com.datadog.debugger.util.ClassFileLines;
+import org.objectweb.asm.tree.ClassNode;
+import org.objectweb.asm.tree.MethodNode;
+
+/** Data class to store all information related to a classfile */
+public class ClassFileInfo {
```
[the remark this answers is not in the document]
jpbempel: done

dd-java-agent/agent-debugger/src/main/java/com/datadog/debugger/exception/DefaultExceptionDebugger.java
jpbempel: done

shatzi (approved):
looks good

PR #6610, opened 2024-02-02
Run CI only for changed modules in pull requests

smola:
# What Does This Do

* Use [AffectedModuleDetector](https://github.com/dropbox/AffectedModuleDetector) gradle plugin to skip projects that are unaffected by git changes.
* Only applied to PRs. Master/release branches run all tests. Using the label `run-tests: all` will also skip this logic and run all tests.
* The current configuration considers that a project is affected if that project, or any dependency project (transitively) has changed.
* Git changes are computed by comparing the merge checkout (that is, the code we actually test) against the latest commit in the base branch (by default, `master`).
* Previous logic to skip instrumentation tests for profiler-only changes is kept intact, since there is an actual dependency across these modules which would always trigger all instrumentation tests. Ideally we would do some further changes to use a single skip mechanism for both cases.
* Results: a PR pipeline currently takes 30-35min. A PR only touching a smoke test (one of the best cases), takes ~18min. Example PR with partial run: https://github.com/DataDog/dd-trace-java/pull/6620

# Motivation

Reduce PR turnaround time, and reduce the chance of hitting a flaky test in a PR.

# Additional Notes

## To Do

- [x] Correctly select comparison base (master, release/).

## Future work

- [ ] Exclude irrelevant files (CI conf, README, etc)
- [ ] Replace current profiler/debugging skip mechanisms
- [ ] Avoid creating Circle CI jobs if none if its tasks will run (e.g. if no smoke test is run, setup should avoid creating `smoke` jobs at all, this might require a lockfile of project dependencies).
- [ ] Add lockfile for project dependencies? Replace current profiler/debugging skip mechanism.


.circleci/render_config.py
```diff
@@ -52,7 +52,7 @@
             )
             resp.raise_for_status()
         except Exception as e:
-            print(f"Request filed: {e}")
```
bantonsson, line 55: 🤣

build.gradle
```diff
@@ -1,6 +1,7 @@
 buildscript {
   dependencies {
     classpath "pl.allegro.tech.build:axion-release-plugin:1.14.4"
+    classpath "com.dropbox.affectedmoduledetector:affectedmoduledetector:0.2.2"
```
bantonsson, line 4: I had to look, and it doesn't use `jgit` 🎉

bantonsson (approved):

PR #6619, opened 2024-02-05
Refactor checkout merge commit in CI

smola:
# What Does This Do
* Move logic to a standalone script (easier to test).
* Use it during pipeline setup. This ensures changes to Circle CI config itself are tested in the post-merge state.

# Motivation

# Additional Notes

Jira ticket: [PROJ-IDENT]


PerfectSlayer (commented):
Nice change. I think we could extract even more scripts to their own files 👍

PR #6625, opened 2024-02-06
Partition build caches

smola:
# What Does This Do

Partition builds, matching test partitions. This should reduce workflow time by reducing both build time as well as cache save/restore times.

# Motivation

# Additional Notes

Jira ticket: [PROJ-IDENT]


smola:
With our current setup, there is always one instrumentation partition that picks up the build for `:dd-java-agent:instrumentation`, and that partition takes almost as much time as the old one. Not worth pursuing this change without addressing that first.

PR #6628, opened 2024-02-06
Fixes to trace agent service rates feedback - e.g. MAX TPS support

dougqh:
This change set fixes the MAX TPS support in the tracer. 

# What Does This Do

Previously, the tracer treated service name and env in a case sensitive fashion. Additionally, the tracer didn't properly handle the fallback case of an empty service and empty env.

While the empty service and empty env case was parsed properly, getSampler would return a static DEFAULT sampler that always used a rate 1.0.  This wasn't obvious from the tests because the tests also use a rate of 1.0.

# Motivation

Bring Java tracer into compliance with trace agent feedback specification

Jira ticket: [APMS-11270]

[APMS-11270]: https://datadoghq.atlassian.net/browse/APMS-11270?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ


dd-trace-core/src/main/java/datadog/trace/common/sampling/RateByServiceTraceSampler.java
```diff
@@ -65,24 +65,37 @@ private <T extends CoreSpan<T>> String getSpanEnv(final T span) {
   public void onResponse(
       final String endpoint, final Map<String, Map<String, Number>> responseJson) {
     final Map<String, Number> newServiceRates = responseJson.get("rate_by_service");
-    if (null != newServiceRates) {
```
dougqh, line 68: I refactored the code a fair amount.

The biggest change is introducing a fallbackSampler field rather than storing the fallbackSampler into the Map.

dd-trace-core/src/main/java/datadog/trace/common/sampling/RateByServiceTraceSampler.java
```diff
@@ -140,29 +163,42 @@ public EnvAndService apply(String key) {
...
-    static final EnvAndService DEFAULT = new EnvAndService("", "");
+    static final EnvAndService FALLBACK = new EnvAndService("", "");
 
     public static EnvAndService fromString(String key) {
       return CACHE.computeIfAbsent(key, PARSE);
     }
 
-    private final String env;
-    private final String service;
+    private final String lowerEnv;
```
dougqh, line 188: Renamed so that it is clear that the values have been transformed

dd-trace-core/src/test/groovy/datadog/trace/common/sampling/RateByServiceTraceSamplerTest.groovy
```diff
@@ -17,8 +17,10 @@ class RateByServiceTraceSamplerTest extends DDCoreSpecification {
     RateByServiceTraceSampler serviceSampler = new RateByServiceTraceSampler()
     String response = '{"rate_by_service": {"service:,env:":' + rate + '}}'
     serviceSampler.onResponse("traces", serializer.fromJson(response))
+
     expect:
-    serviceSampler.serviceRates.getSampler(RateByServiceTraceSampler.EnvAndService.DEFAULT).sampleRate == expectedRate
+    serviceSampler.serviceRates.getSampler(RateByServiceTraceSampler.EnvAndService.FALLBACK).sampleRate == expectedRate
+    serviceSampler.serviceRates.getSampler("foo", "bar").sampleRate == expectedRate
```
dougqh, line 23: This is the key check that was missing previously -- that would have caught the broken fallback handling

dd-trace-core/src/test/groovy/datadog/trace/common/sampling/RateByServiceTraceSamplerTest.groovy
```diff
@@ -66,6 +69,29 @@ class RateByServiceTraceSamplerTest extends DDCoreSpecification {
     tracer.close()
   }
 
+  def "rate by service name - case-insensitive"() {
```
dougqh, line 72: New test to cover case insensitive matching

dd-trace-core/src/main/java/datadog/trace/common/sampling/RateByServiceTraceSampler.java
```diff
@@ -100,35 +113,45 @@ private static RateSampler createRateSampler(final double sampleRate) {
...
+    }
+
+    private RateSampler getSamplerImpl(String lowerEnv, String lowerService) {
+      if (EnvAndService.isFallback(lowerEnv, lowerService)) {
+        return fallbackSampler;
+      }
+
+      Map<String, RateSampler> serviceRates = envServiceRates.get(lowerEnv);
       if (serviceRates == null) {
-        return DEFAULT;
```
dougqh, line 123: These lines returning DEFAULT are the bugs in the original implementation.

dd-trace-core/src/test/groovy/datadog/trace/common/sampling/RateByServiceTraceSamplerTest.groovy
```diff
@@ -30,6 +33,29 @@ class RateByServiceTraceSamplerTest extends DDCoreSpecification {
...
+    when:
+    def sampler = serviceSampler.serviceRates.getSampler(env, service)
+
+    then:
+    sampler.sampleRate > expectedRate - 0.01
+    sampler.sampleRate < expectedRate + 0.01
+
+    where:
+    service | env     | expectedRate
+    "foo"   | "bar"   | 0.8
```
dougqh, line 51: New tests to verify case insensitivity
And check that partial matches uses the fallback

dd-trace-core/src/test/groovy/datadog/trace/common/sampling/RateByServiceTraceSamplerTest.groovy
```diff
@@ -66,6 +93,29 @@ class RateByServiceTraceSamplerTest extends DDCoreSpecification {
     tracer.close()
   }
 
+  def "rate by service name - case-insensitive"() {
+    setup:
+    RateByServiceTraceSampler serviceSampler = new RateByServiceTraceSampler()
+    def tracer = tracerBuilder().writer(new ListWriter()).build()
+
+    when:
+    def response = '{"rate_by_service": {"service:spock,env:test":1.0}}'
```
ygree, line 102: I'd add one more rule `,"service:SPOCK,env:Test":0.0` to make sure we verify case-insensitivity otherwise this test will pass even if it's case-sensitive b/o the default case.

ygree (approved):
Looks good! Left one proposal for the test to make sure we are testing case-insensitivity and not the default case

dd-trace-core/src/test/groovy/datadog/trace/common/sampling/RateByServiceTraceSamplerTest.groovy
dougqh: Added another commit with that test in addition to some others that check case insensitive equivalent service/env pairs

PerfectSlayer (approved):
I didn't had the time to submit my review before Yury approved it but I wanted to share I appreciated the inline comments about the refactoring part, thanks for taking time to add them 👍

PR #6636, opened 2024-02-08
Remote Config. Add target service+env check.

ygree:
# What Does This Do
Add the target service and env checks.

# Motivation
Prevent remote config update if service+env doesn't match.

# Additional Notes

Jira ticket: [APMJAVA-1222]
- [ ] Enable [system test](https://github.com/DataDog/system-tests/blob/main/tests/parametric/test_dynamic_configuration.py#L265) once this PR merged
- [ ] Extends system tests to make sure that the check is case-insensitive

[APMJAVA-1222]: https://datadoghq.atlassian.net/browse/APMJAVA-1222?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ

### DIFF
--- a/dd-trace-core/src/test/groovy/datadog/trace/core/CoreTracerTest.groovy
+++ b/dd-trace-core/src/test/groovy/datadog/trace/core/CoreTracerTest.groovy
@@ -438,6 +438,134 @@ class CoreTracerTest extends DDCoreSpecification {
     null      | "test"
     "some"    | "some"
   }
+
+  def "reject configuration when target service+env mismatch"() {
+    setup:
+    injectSysConfig(SERVICE_NAME, service)
+    injectSysConfig(ENV, env)
+
+    def key = ParsedConfigKey.parse("datadog/2/APM_TRACING/config_overrides/config")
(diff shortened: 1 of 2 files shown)


dd-trace-core/src/test/groovy/datadog/trace/core/CoreTracerTest.groovy
```diff
@@ -353,8 +353,12 @@ class CoreTracerTest extends DDCoreSpecification {
     tracer.captureTraceConfig().traceSampleRate == null
 
     when:
-    updater.accept(key, '''
+    updater.accept(key, """
       {
+        "service_target": {
+          "service": "${Config.get().getServiceName()}",
+          "env": "${Config.get().getEnv()}"
```
PerfectSlayer, line 360: Should we add a test case to check the config is not apply when one of those values do not match?

PerfectSlayer (approved):

PR #6643, opened 2024-02-08
Add support for periodic submission of telemetry distribution series

nikita-tkachenko-datadog:
# What Does This Do
Adds support for periodical polling and submission of distribution series telemetry.

# Motivation
Distribution series will be used by future CI Visibility telemetry logic.

# Additional Notes
Most of the logic for sending distribution series data was implemented previously.
This PR adds the missing part - polling periodic metric collectors, aggregating drained data and passing it over to telemetry service.

Jira ticket: [CIVIS-2427]

### DIFF
--- a/internal-api/src/main/java/datadog/trace/api/telemetry/MetricCollector.java
+++ b/internal-api/src/main/java/datadog/trace/api/telemetry/MetricCollector.java
@@ -120,11 +120,11 @@ public interface MetricCollector<M extends Metric> {
     public final String metricName;
     public final boolean common;
     public final String namespace;
-    public final Integer value;
+    public final int value;
     public final List<String> tags;
 
     public DistributionSeriesPoint(
-        String metricName, boolean common, String namespace, Integer value, List<String> tags) {
+        String metricName, boolean common, String namespace, int value, List<String> tags) {
       this.metricName = metricName;
       this.common = common;
       this.namespace = namespace;
--- a/internal-api/src/main/java/datadog/trace/api/telemetry/MetricCollector.java
+++ b/internal-api/src/main/java/datadog/trace/api/telemetry/MetricCollector.java
@@ -6,6 +6,7 @@ import static java.util.Collections.singletonList;
 
 import java.util.Arrays;
 import java.util.Collection;
+import java.util.Collections;
 import java.util.List;
 import java.util.Objects;
 
@@ -19,6 +20,10 @@ public interface MetricCollector<M extends Metric> {
   void prepareMetrics();
 
   Collection<M> drain();
+
+  default Collection<DistributionSeriesPoint> drainDistributionSeries() {
+    return Collections.emptySet();
+  }
 
   class Metric {
     public final String metricName;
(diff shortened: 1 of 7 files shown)


internal-api/src/main/java/datadog/trace/api/telemetry/MetricCollector.java
```diff
@@ -110,4 +115,55 @@ public String toString() {
...
+      DistributionSeriesPoint that = (DistributionSeriesPoint) o;
+      return common == that.common
+          && Objects.equals(metricName, that.metricName)
+          && Objects.equals(namespace, that.namespace)
+          && Objects.equals(tags, that.tags);
+    }
+
+    @Override
+    public int hashCode() {
+      return Objects.hash(metricName, common, namespace, tags);
```
dougqh, line 148: Objects.hash doesn't always optimize well, but that's probably fine in this location.

telemetry/src/main/java/datadog/telemetry/api/DistributionSeries.java
```diff
@@ -28,6 +28,10 @@ public DistributionSeries points(List<Integer> points) {
     return this;
   }
 
+  public void addPoint(Integer point) {
```
dougqh, line 31: Why Integer - not int?

internal-api/src/main/java/datadog/trace/api/telemetry/MetricCollector.java
```diff
@@ -110,4 +115,55 @@ public String toString() {
...
+
+  class DistributionSeriesPoint {
+    public final String metricName;
+    public final boolean common;
+    public final String namespace;
+    public final Integer value;
+    public final List<String> tags;
+
+    public DistributionSeriesPoint(
+        String metricName, boolean common, String namespace, Integer value, List<String> tags) {
```
dougqh, line 127: Why Integer?  And not int?

internal-api/src/main/java/datadog/trace/api/telemetry/MetricCollector.java
```diff
@@ -20,6 +21,10 @@ public interface MetricCollector<M extends Metric> {
 
   Collection<M> drain();
 
+  default Collection<DistributionSeriesPoint> drainDistributionSeries() {
```
dougqh, line 24: Having a default implementation that returns an empty collection seems odd.
Won't every implementor of this interface need to implement this method anyway?

internal-api/src/main/java/datadog/trace/api/telemetry/MetricCollector.java
nikita-tkachenko-datadog: My initial thought was that either way it'll have to be boxed at some point since it ends up being added to a `List<Integer>`. But then it still makes sense to have an `int` in order to prevent `null` from being added. Changed to `int`.

telemetry/src/main/java/datadog/telemetry/api/DistributionSeries.java
nikita-tkachenko-datadog: My initial thought was that either way it'll have to be boxed at some point since it ends up being added to a `List<Integer>`. But then it still makes sense to use the primitive in order to prevent `null` from being added. Changed to `int`.

internal-api/src/main/java/datadog/trace/api/telemetry/MetricCollector.java
nikita-tkachenko-datadog: Just out of curiosity, does it have fewer chances of being inlined compared to a "manually" constructed hash code because there're two extra method calls and a cycle?

internal-api/src/main/java/datadog/trace/api/telemetry/MetricCollector.java
nikita-tkachenko-datadog: > Won't every implementor of this interface need to implement this method anyway?
Since none of the current implementations use distribution series, returning an empty collection seemed reasonable.
The future CI Visibility collector implementation will be the only one that returns a non empty collection.
I could implement this method in every existing implementation, but it'll look the same - `return Collections.emptyList();`. Let me know if this is preferable to having a default impl :)

dougqh (approved):

PR #6644, opened 2024-02-08
Do not switch to agent telemetry client on detecting agent telemetry endpoint availability when intake client is preferred

nikita-tkachenko-datadog:
# What Does This Do

Updates telemetry client switch logic.
Prior to the change the logic could switch from intake client to agent client in the following circumstances:
- if a telemetry request fails
- if agent is available and has telemetry proxy endpoint

With the change the logic is to do the switch if:
- if a telemetry request fails
- if agent is available, has telemetry proxy endpoint and _the tracer is not configured for "agentless" mode_.

# Motivation
If the tracer is explicitly configured to work in agentless mode, the intake telemetry client should be the preferred one.

Jira ticket: [CIVIS-2427]

### DIFF
--- a/telemetry/src/main/java/datadog/telemetry/TelemetryRouter.java
+++ b/telemetry/src/main/java/datadog/telemetry/TelemetryRouter.java
@@ -12,9 +12,8 @@ public class TelemetryRouter {
 
   private final DDAgentFeaturesDiscovery ddAgentFeaturesDiscovery;
   private final TelemetryClient agentClient;
-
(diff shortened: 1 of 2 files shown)


PerfectSlayer (approved):
LGTM
Should we have more test about it?

nikita-tkachenko-datadog:
> LGTM Should we have more test about it?

Sure, it won't hurt to have a couple of dedicated test cases for it. Added.

PR #6649, opened 2024-02-08
Custom Tags via Remote Config for Tracing

nayeem-kamal:
# What Does This Do
This change allows for the new `tracing_tags` field of the remote config payload to be applied to spans.
The capability bits for Remote config are updated to reflect the current capabilities of the tracer.
# Motivation
[RFC](https://docs.google.com/document/d/1V4ZBsTsRPv8pAVG5WCmONvl33Hy3gWdsulkYsE4UZgU/edit?usp=sharing)
# Additional Notes

Jira ticket: [APMJAVA-1215]

[APMJAVA-1215]: https://datadoghq.atlassian.net/browse/APMJAVA-1215?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ


dd-trace-core/src/test/groovy/datadog/trace/core/CoreTracerTest.groovy
```diff
@@ -422,6 +422,113 @@ class CoreTracerTest extends DDCoreSpecification {
...
+             "header": "C!!!ont_____ent----tYp!/!e",
+             "tag_name": ""
+          }, {
+             "header": "this.header",
+             "tag_name": "whatever.the.user.wants.this.header"
+          }]
+          ,
+          "tracing_sampling_rate": 0.5
+          ,
+          "tracing_tags": ["a:b", "c:d", "e:f"]
```
ygree, line 489: Do we need to test all other lib_config entries besides `tracing_tags`, or are they already covered by existing tests?

dd-trace-core/src/main/java/datadog/trace/core/TracingConfigPoller.java
```diff
@@ -163,6 +178,24 @@ private <T> void maybeOverride(Consumer<T> setter, T override) {
     }
   }
 
+  private Map<String, String> parseTagListToMap(List<String> input) {
+    Map<String, String> resultMap = Collections.emptyMap();
+    if (null != input && !input.isEmpty()) {
+      resultMap = new HashMap<>();
```
ygree, line 184: The initial capacity can be passed here since it is already known from the input parameter.

dd-trace-core/src/main/java/datadog/trace/core/TracingConfigPoller.java
```diff
@@ -37,8 +45,16 @@ public TracingConfigPoller(DynamicConfig<?> dynamicConfig) {
 
   public void start(Config config, SharedCommunicationObjects sco) {
     this.startupLogsEnabled = config.isStartupLogsEnabled();
-
-    stopPolling = new Updater().register(config, sco);
+    ConfigurationPoller ConfigPoller = sco.configurationPoller(config);
+    // TODO: Add CAPABILITY_APM_TRACING_TRACING_ENABLED flag when tracing_enabled is implemented for
+    // Remote config
+    ConfigPoller.addCapabilities(
```
ygree, line 51: Wow, I didn't know we use bits for encoding capabilities.

dd-trace-core/src/main/java/datadog/trace/core/TracingConfigPoller.java
```diff
@@ -163,6 +178,24 @@ private <T> void maybeOverride(Consumer<T> setter, T override) {
...
+  private Map<String, String> parseTagListToMap(List<String> input) {
+    Map<String, String> resultMap = Collections.emptyMap();
+    if (null != input && !input.isEmpty()) {
+      resultMap = new HashMap<>();
+      for (String s : input) {
+        int colonIndex = s.indexOf(":");
+        if (colonIndex > -1
+            && colonIndex
+                < s.length() - 1) { // ensure there's a colon that's not at the start or end
+          String key = s.substring(0, colonIndex).trim();
```
ygree, line 190: What type of validation or normalization (trim) necessary here? Does the RFC specify this?
I noticed a similar logic in `DynamicConfig.cleanMapping`, but I'm not sure if it applies here.

internal-api/src/main/java/datadog/trace/api/Config.java
```diff
@@ -3390,6 +3390,13 @@ public Map<String, String> getMergedSpanTags() {
     return Collections.unmodifiableMap(result);
   }
 
+  public Map<String, String> getSpanTags() {
+    // Do not include runtimeId into span tags: we only want that added to the root span
```
ygree, line 3394: I am uncertain about the meaning of this comment. It seems to be filtering out runtimeId, but it does not appear to be relevant in this context.

dd-trace-core/src/main/java/datadog/trace/core/TracingConfigPoller.java
```diff
@@ -163,6 +178,24 @@ private <T> void maybeOverride(Consumer<T> setter, T override) {
     }
   }
 
+  private Map<String, String> parseTagListToMap(List<String> input) {
+    Map<String, String> resultMap = Collections.emptyMap();
+    if (null != input && !input.isEmpty()) {
```
ygree, line 183: Could reduce nesting by checking and returning earlier

dd-trace-core/src/main/java/datadog/trace/core/TracingConfigPoller.java
nayeem-kamal: Yep, Spec is available [here](https://docs.google.com/document/d/1FM9Wz2CjtmfM405U8fxFwttpeWSFe4LwO1UwckZgGGI/edit?usp=sharing)

dd-trace-core/src/main/java/datadog/trace/core/TracingConfigPoller.java
nayeem-kamal: There is no specification for trimming/normalization. It looks like other libraries are not Trimming so I will remove the trim step ([dotnet](https://github.com/DataDog/dd-trace-dotnet/blob/622aa6db1d679ec233871661c75c1898d3b994c2/tracer/src/Datadog.Trace/Configuration/ConfigurationSources/DynamicConfigConfigurationSource.cs#L79), [go](https://github.com/DataDog/dd-trace-dotnet/blob/622aa6db1d679ec233871661c75c1898d3b994c2/tracer/src/Datadog.Trace/Configuration/ConfigurationSources/DynamicConfigConfigurationSource.cs#L79) )

nayeem-kamal: The cleanMapping method wan't useful here because it is used for cleaning existing JSON maps but I am parsing a JSON String[].

dd-trace-core/src/main/java/datadog/trace/core/TracingConfigPoller.java
```diff
@@ -163,6 +178,24 @@ private <T> void maybeOverride(Consumer<T> setter, T override) {
     }
   }
 
+  private Map<String, String> parseTagListToMap(List<String> input) {
```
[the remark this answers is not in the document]
nayeem-kamal: Added some logic to check for empty keys and values. Also changed the test to allow for a few different inputs to check this.

dd-trace-core/src/test/groovy/datadog/trace/core/CoreTracerTest.groovy
nayeem-kamal: Changed this test to only test `tracing_tags`

dd-trace-core/src/main/java/datadog/trace/core/TracingConfigPoller.java
nayeem-kamal: Fixed.

internal-api/src/main/java/datadog/trace/api/Config.java
nayeem-kamal: Cleaned up

dd-trace-core/src/main/java/datadog/trace/core/TracingConfigPoller.java
nayeem-kamal: Implemented this suggestion

dd-trace-core/src/main/java/datadog/trace/core/TracingConfigPoller.java
```diff
@@ -37,8 +45,16 @@ public TracingConfigPoller(DynamicConfig<?> dynamicConfig) {
 
   public void start(Config config, SharedCommunicationObjects sco) {
     this.startupLogsEnabled = config.isStartupLogsEnabled();
-
-    stopPolling = new Updater().register(config, sco);
+    ConfigurationPoller ConfigPoller = sco.configurationPoller(config);
```
ygree, line 48: I think `ConfigPoller` should be lowercased

dd-trace-core/src/main/java/datadog/trace/core/TracingConfigPoller.java
ygree, line 51: A null check is necessary here because `sco.configurationPoller` returns null when remote config is disabled.
This is the reason for the smoke test failure.

dd-trace-core/src/main/java/datadog/trace/core/TracingConfigPoller.java
nayeem-kamal: Good catch. Fixed!

dd-trace-core/src/main/java/datadog/trace/core/TracingConfigPoller.java
nayeem-kamal: Added the check ✅

ygree: Yeah, I know. I'm just surprised that it uses a bit set on the API level.

dd-trace-core/src/test/groovy/datadog/trace/core/CoreTracerTest.groovy
```diff
@@ -422,6 +422,60 @@ class CoreTracerTest extends DDCoreSpecification {
...
+    updater.commit()
+
+    then:
+    tracer.captureTraceConfig().tracingTags == [:]
+
+    cleanup:
+    tracer?.close()
+
+    where:
+    value | expectedValue
```
ygree, line 470: We typically align columns, although it is minor since there are only two columns

ygree (approved):
Look good to me!

PR #6653, opened 2024-02-08
Update telemetry intake client to use CI Visibility agentless URL, if configured

nikita-tkachenko-datadog:
# What Does This Do
Updates telemetry intake client to use CI Visibility agentless URL if CI Visibility agentless mode is enabled and the URL is configured.

# Motivation
Some users who run the tracer in agentless mode might need to use a non-standard URL.
A non-standard URL is also needed in smoke tests that start a mock intake server.

# Additional Notes
The custom URL is only used if CI Visibility is enabled.

Jira ticket: [CIVIS-2427]

### DIFF
--- a/telemetry/src/test/groovy/datadog/telemetry/TelemetryClientTest.groovy
+++ b/telemetry/src/test/groovy/datadog/telemetry/TelemetryClientTest.groovy
@@ -5,21 +5,27 @@ import spock.lang.Specification
 
 class TelemetryClientTest extends Specification {
 
-  def "Uses CI Visibility agentless URL if configured to do so"() {
+  def "Intake client uses CI Visibility agentless URL if configured to do so"() {
     setup:
     def config = Stub(Config)
     config.getApiKey() >> "dummy-key"
     config.getAgentTimeout() >> 123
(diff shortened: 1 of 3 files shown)


telemetry/src/test/groovy/datadog/telemetry/TelemetryClientTest.groovy
```diff
@@ -0,0 +1,25 @@
+package datadog.telemetry
+
+import datadog.trace.api.Config
+import spock.lang.Specification
+
+class TelemetryClientTest extends Specification {
+
+  def "Uses CI Visibility agentless URL if configured to do so"() {
```
ygree, line 8: For the sake of completeness, it would be great if it would check when ci visibility is disabled.

ygree (approved):
Looks good to me! Maybe an extra test case would be good to have, but definitely not mandatory

telemetry/src/test/groovy/datadog/telemetry/TelemetryClientTest.groovy
nikita-tkachenko-datadog: Good point, added some more test cases

PR #6657, opened 2024-02-09
Decouple ConfigurationUpdater and exception probes

jpbempel:
# What Does This Do
Introducing the concept of source for `ConfigurationUpdater` that is just a list of `ProbeDefinition`. As RemoteConfig configuration are only for probes for now, we are reducing it to this list. Hence the `accept`
method in `ConfigurationUpdater` is taking a source (`REMOTE_CONFIG` or `EXCEPTION`) and a list of `ProbeDefinition`. For RemoteConfig part, we are building this list instead of `Configuration` object and capping the number of probes per type at that level now instead of `ConfigurationUpdater`. The side effect of this is that now there is no capping for probe definitions coming from file. `Configuration::accept` method is now responsible to merge definitions from all source (`REMOTE_CONFIG` and `EXCEPTION`) and create a `Configuration` object that will be applied as before. As a consequence, `ExceptionProbeManager` can now be created directly into `DefaultExceptionDebugger` class.

# Motivation
Decouple `ConfigurationUpdater` and `ExceptionProbeManager`

# Additional Notes

Jira ticket: [DEBUG-2068]

### DIFF
--- a/dd-java-agent/agent-debugger/src/main/java/com/datadog/debugger/agent/DebuggerProductChangesListener.java
+++ b/dd-java-agent/agent-debugger/src/main/java/com/datadog/debugger/agent/DebuggerProductChangesListener.java
@@ -1,6 +1,6 @@
 package com.datadog.debugger.agent;
 
-import static com.datadog.debugger.agent.DebuggerProductChangesListener.ConfigurationAcceptor.Source.REMOTE_CONFIG;
+import static com.datadog.debugger.agent.ConfigurationAcceptor.Source.REMOTE_CONFIG;
 
 import com.datadog.debugger.probe.LogProbe;
 import com.datadog.debugger.probe.MetricProbe;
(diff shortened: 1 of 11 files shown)


dd-java-agent/agent-debugger/src/main/java/com/datadog/debugger/agent/DebuggerProductChangesListener.java
```diff
@@ -1,35 +1,47 @@
...
 
 public class DebuggerProductChangesListener implements ProductListener {
+  public static final int MAX_ALLOWED_METRIC_PROBES = 100;
+  public static final int MAX_ALLOWED_LOG_PROBES = 100;
+  public static final int MAX_ALLOWED_SPAN_PROBES = 100;
+  public static final int MAX_ALLOWED_SPAN_DECORATION_PROBES = 100;
   private static final Logger LOGGER =
       LoggerFactory.getLogger(DebuggerProductChangesListener.class);
 
   public interface ConfigurationAcceptor {
```
shatzi, line 38: should we move this outside of this file? doesn't look like this belongs here. include the Source

shatzi (approved):
nice.

dd-java-agent/agent-debugger/src/main/java/com/datadog/debugger/agent/DebuggerProductChangesListener.java
jpbempel: done

PR #6663, opened 2024-02-12
Spring boot: autodetect service name from environment

amarziali:
# What Does This Do

This PR allows automatically set the default service name for a spring boot standalone application from the spring property `spring.application.name`

It applies only if:
* The user did not provide a service name through `DD_SERVICE` or equivalents
* The application is deployed as a standalone jar (no war packages deployed on app servers)
* The user defined `spring.application.name` in the spring environment (args, properties, env variables, ...)

This PR modifies only the default service name generation replacing the use of the jar name with the spring application name that's more appropriate.

The instrumentation is still in beta and it has to be explicitly enabled. It can be done by setting:
* The env `DD_TRACE_INTEGRATION_SPRING_BOOT_ENABLED=true`
* The sysprop `-Ddd.trace.integration.spring-boot.enabled=true`

# Motivation

# Additional Notes

Jira ticket: [AIT-9541]

[AIT-9541]: https://datadoghq.atlassian.net/browse/AIT-9541?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ


amarziali:
> e: what’t the default service name when it’s not set by the customer?

Usually it's the jar name (otherwise `unnamed-java-service`). Right now I will set that instrumentation disabled by default (to be safe). I will enable only on single step.
I'll apply also your suggestion before merging thanks

PR #6669, opened 2024-02-12
Implement telemetry for CI Visibility events

nikita-tkachenko-datadog:
# What Does This Do
Implements sending telemetry metrics for CI Visibility events (test, suite, module, session start and finish).

# Motivation
Monitoring CI Visibility behaviour to detect unexpected errors/problems.

Jira ticket: [CIVIS-2427]


dd-java-agent/agent-ci-visibility/src/main/java/datadog/trace/civisibility/ci/CIProviderInfo.java
```diff
@@ -7,4 +7,8 @@ public interface CIProviderInfo {
   GitInfo buildCIGitInfo();
 
   CIInfo buildCIInfo();
+
+  default boolean isSupportedCiProvider() {
+    return true;
```
romainkomorndatadog, line 12: Should a CI provider just be unsupported by default?

romainkomorndatadog (approved):

dd-java-agent/agent-ci-visibility/src/main/java/datadog/trace/civisibility/ci/CIProviderInfo.java
nikita-tkachenko-datadog: I can see how this looks somewhat counterintuitive :) The thing is that if we're adding an implementation for a specific provider it means we're supporting it. So every single implementation of this interface would have to return `true` except the one for the `UnknownProvider`

romainkomorndatadog: Touché. :D

PR #6680, opened 2024-02-14
Update README.md due to disabled add-assets-to-release workflow

am312:
# What Does This Do
 
Updated README.md because the add-assets-to-release gh workflow was replaced with a GitLab job so JARs are not relayed through central unnecessarily.


smola (approved):
Sounds good, although I we could just remove the workflow.

mcculls (approved):

PerfectSlayer:
I think we can drop it completely as it is already running fine.
Worst case scenario, we can bring it back from versioning.

PR #6660, opened 2024-02-10
Implement a mechanism for adding ad-hoc listeners to OKHTTP requests

nikita-tkachenko-datadog:
# What Does This Do
Adds a mechanism for registering ad-hoc event listeners for okhttp requests.
A listener can be registered when building a request:
```
class MyListener extends OkHttpUtils.CustomListener {
    // override event handlers
} 

final Request request =
        new Request.Builder()
            .url(...)
            .post(...)
            .tag(OkHttpUtils.CustomListener.class, new MyListener())
            .build();
```

# Motivation
This is needed for CI Visibility telemetry. There are metrics like requests count, errors count, request size, etc for events/coverages intake, CI Visibility Git API and other endpoitns.

# Additional Notes
`EventListener.NONE` is what all the requests are using by default: it is a listener that has an empty handler for every event.
With the new mechanism this "empty" listener is replaced with a custom one for those requests that have the corresponding tag.

This PR only adds the means for registering custom listeners. Actual listeners will be added in a separate PR.

Jira ticket: [CIVIS-2427]

[CIVIS-2427]: https://datadoghq.atlassian.net/browse/CIVIS-2427?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ

PR #6682, opened 2024-02-14
Fix OKHTTP instrumentation tests

nikita-tkachenko-datadog:
# What Does This Do
Fixes OKHTTP tests that were broken when #6660 got merged.
The PR uses an OKHTTP feature (event listeners) available since v3.12.0
The tests for OKHTTP instrumentation use v3.0.0 (the oldest that is supported).

In "real-life" scenarios the OKHTTP classes used by the tracer core do not conflict with the classes that may be used by the application since the tracer classes are relocated.
However, this does not happen in instrumentation tests. As the result, the tracer ends up using old OKHTTP version - the one that is being tested.

Jira ticket: [CIVIS-2427]


mcculls (approved):
Approving this to un-break the build

PR #6686, opened 2024-02-15
Refactor ExceptionProbe to get one per frame

jpbempel:
# What Does This Do
Instead of having multiple probes for each frame when sharing same frame for different exception (and fingerprint) we have only one instance of ExceptionProbe per frame and using ExceptionProbeManager we make sure that:
1. we only instrument once.
2. we compute the fingerprint and assess if we need to capture the exception or not for this frame
When instrumenting, we store a normalized version of a Where instance with className + methodName + signature. The normalization can only happen at instrumentation time to resolve which line correspond to which method (in case of overloaded methods).
When retransformations happen for a class we need to remove the instrumented method from ExceptionProbeManager by introducing a listener mechanism in ConfigurationUpdater.

# Motivation
Following comment: https://github.com/DataDog/dd-trace-java/pull/6641#discussion_r1483071837

# Additional Notes

### DIFF
--- a/dd-java-agent/agent-debugger/src/main/java/com/datadog/debugger/exception/ExceptionProbeManager.java
+++ b/dd-java-agent/agent-debugger/src/main/java/com/datadog/debugger/exception/ExceptionProbeManager.java
@@ -15,13 +15,18 @@ public class ExceptionProbeManager {
   private final Set<String> fingerprints = ConcurrentHashMap.newKeySet();
   private final Map<String, ExceptionProbe> probes = new ConcurrentHashMap<>();
   private final ClassNameFiltering classNameFiltering;
+  private final Map<Where, ExceptionProbe> probesByLocation = new ConcurrentHashMap<>();
+  private final Map<Where, ExceptionProbe> instrumentedMethods = new ConcurrentHashMap<>();
 
   public ExceptionProbeManager(ClassNameFiltering classNameFiltering) {
     this.classNameFiltering = classNameFiltering;
   }
 
+  public ClassNameFiltering getClassNameFiltering() {
+    return classNameFiltering;
+  }
+
   public void createProbesForException(String fingerprint, StackTraceElement[] stackTraceElements) {
-    fingerprints.add(fingerprint);
     for (StackTraceElement stackTraceElement : stackTraceElements) {
       if (stackTraceElement.isNativeMethod() || stackTraceElement.getLineNumber() < 0) {
         // Skip native methods and lines without line numbers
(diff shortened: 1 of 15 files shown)


dd-java-agent/agent-debugger/src/main/java/com/datadog/debugger/exception/ExceptionProbeManager.java
```diff
@@ -66,4 +67,16 @@ public boolean isAlreadyInstrumented(String fingerprint) {
   public Collection<ExceptionProbe> getProbes() {
     return probes.values();
   }
+
+  public boolean shouldCaptureException(Where where, String fingerprint) {
+    return probesByLocation.containsKey(where) && fingerprints.contains(fingerprint);
```
shatzi, line 72: Why do we need to check the location? Doesn't the fingerprint already contain the line numbers? Therefore we don't need to check the where and maybe we can skip maintaining this map at all.

dd-java-agent/agent-debugger/src/main/java/com/datadog/debugger/probe/ExceptionProbe.java
```diff
@@ -102,6 +99,17 @@ public void buildLocation(InstrumentationResult result) {
     this.location = new ProbeLocation(type, method, where.getSourceFile(), emptyList());
   }
 
+  @Override
+  public InstrumentationResult.Status instrument(
```
shatzi, line 103: Maybe this logic can be moved to the transformer. Same as we one instrument a single log probe for the entire method. 

I think this map is only needed to be alive during instrumentation. Building a set of method info that needed to be instrumented and the call it once per method.

shatzi (commented):
I like the simplification here. And I think we can push this even more. 

I would suggest to change the transformer logic of grouping probes together so we would call instrument only one time per method. 

Then we no don't need to notification of transform logic (which could lead to race conditions I think). And we don't need to maintain this on the probes manager.

Let's chat about async

dd-java-agent/agent-debugger/src/main/java/com/datadog/debugger/exception/ExceptionProbeManager.java
jpbempel: true fingerprint is enough

dd-java-agent/agent-debugger/src/main/java/com/datadog/debugger/probe/ExceptionProbe.java
jpbempel: no the map need to be kept not just for instrumentation as any other exceptions that share the frame can happen at any time.
The modeling to have only one probe (ExceptionProbe) for multiple exceptions (fingerprints) does not fit quite nicely with the other probe instrumentation logic. hence the logic in instrument method and the listener

jpbempel:
Updated PR description to relfect the new way to handle the probe definitions

dd-java-agent/agent-debugger/src/main/java/com/datadog/debugger/probe/Where.java
```diff
@@ -60,6 +60,14 @@ protected static SourceLine[] sourceLines(String[] defs) {
     return lines;
   }
 
+  public static Where from(Where lineWhere, ClassFileLines classFileLines) {
```
shatzi, line 63: maybe convertLineToMethod ?

dd-java-agent/agent-debugger/src/main/java/com/datadog/debugger/probe/Where.java
```diff
@@ -60,6 +60,14 @@ protected static SourceLine[] sourceLines(String[] defs) {
...
+    if (lineWhere.methodName != null && lineWhere.lines != null) {
+      MethodNode method = classFileLines.getMethodByLine(lineWhere.lines[0].getFrom());
+      return new Where(lineWhere.typeName, method.name, method.desc, (SourceLine[]) null, null);
+    }
+    return lineWhere;
```
shatzi, line 68: throw an error instead? not sure what exception probe should do with line only...

shatzi (approved):
Thank you

dd-java-agent/agent-debugger/src/main/java/com/datadog/debugger/probe/Where.java
jpbempel: done

dd-java-agent/agent-debugger/src/main/java/com/datadog/debugger/probe/Where.java
jpbempel: done

PR #6375, opened 2023-12-15
Added GraphQL monitoring support

ValentinZakharov:
# What Does This Do
Implemented GraphQL monitoring support for ASM. Allows to detect and report attacks via GraphQL. 
Current implementation consist of extension of GraphQL instrumentation along with new smoke-test application for testing resolvers.
ASM GraphQL monitoring will be enabled if the AppSec is enabled.

# Motivation
The implementation support of GraphQL in our application security product is driven by the evolving landscape of customer needs and the growing adoption of GraphQL. GraphQL applications are susceptible to traditional types of attacks like Cross-Site Scripting (XSS) and SQL injection. Recognising potential security risks, we implemented GraphQL support in the ASM product to shielded from both novel and conventional threats.

# Additional Notes

Jira ticket: [APPSEC-11459]

[APPSEC-11459]: https://datadoghq.atlassian.net/browse/APPSEC-11459?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ

PR #6687, opened 2024-02-15
Replace ASM transient mechanism with WAF ephemeral addresses

ValentinZakharov:
# What Does This Do
This pull request signifies an important enhancement in ASM. Originally, we employed a transient mechanism to manage temporary data submissions to the WAF. The recent introduction of ephemeral addresses in `libddwaf` presents a native and more efficient implementation of this transient mechanism. Through this PR, we are transitioning from the older transient method to utilizing ephemeral addresses.

# Motivation
This improvement is required by #6375 

# Additional Notes
Upgrading WAF to version `1.16.0` made it possible to start using ephemeral addresses (see: #6658)

Jira ticket: [APPSEC-51774]

[APPSEC-51774]: https://datadoghq.atlassian.net/browse/APPSEC-51774?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ

### DIFF
--- a/dd-java-agent/appsec/src/main/java/com/datadog/appsec/gateway/AppSecRequestContext.java
+++ b/dd-java-agent/appsec/src/main/java/com/datadog/appsec/gateway/AppSecRequestContext.java
@@ -2,7 +2,6 @@ package com.datadog.appsec.gateway;
 
 import com.datadog.appsec.event.data.Address;
 import com.datadog.appsec.event.data.DataBundle;
-import com.datadog.appsec.event.data.KnownAddresses;
 import com.datadog.appsec.report.AppSecEvent;
 import com.datadog.appsec.util.StandardizedLogging;
 import datadog.trace.api.http.StoredBodySupplier;
@@ -74,6 +73,7 @@ public class AppSecRequestContext implements DataBundle, Closeable {
   private boolean rawReqBodyPublished;
   private boolean convertedReqBodyPublished;
   private boolean respDataPublished;
+  private boolean pathParamsPublished;
   private Map<String, String> apiSchemas;
 
   // should be guarded by this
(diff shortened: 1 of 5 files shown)


cataphract (commented):
I'm not a big fan of this solution. You're combining ephemeral and persistent objects into an object and then you have to separate them again so that you submit them separately to the waf. In the way, you end up creating more objects and having more conditional logic.

A minimally viable solution would be to just change `runPowerwafTransient` so that it does the correct call to the WAF. I think this would be enough for the foreseeable future, as we never submit transient and non-transient addresses at the same time.

An alternative would be to pass `publishDataEvent` two data bundles, one for persistent and another for ephemeral addresses.

None of these solutions require changing `Address` so that there are persistent and ephemeral addresses. I'm agnostic on whether that's a good change on itself or not.

cataphract (commented):
Just to add one thing. If you want to keep a solution in this line, I think it would be better to:

1) Keep the transient flag in addresses.
2) Keep the removal of isTransient on `onDataAvailable`.
3) Act on the distinction between persistent/ephemeral addresses just at the time of serialization in the PowerwafModule. Right now, you will have only persistent xor ephemeral addresses, so as an optimization you don't need to separate the bundles. If you want to add support for mixed addresses, you can just create two instances of `DataBundleWrapper`, as long as you previously separate `com.datadog.appsec.powerwaf.PowerWAFModule.CtxAndAddresses#addressesOfInterest` into two collections of addresses of interest (transient and persistent).

smola (approved):

cataphract (approved):

PR #6698, opened 2024-02-19
Add bounds to all range related operations

manuel-alvarez-alvarez:
# What Does This Do
This PR ensure that range related operations (propagation of strings, evidence construction in sinks) take an upper bound regarding the number of elements they can iterate and the number of spans they can generate. 

# Motivation
When dealing with huge string manipulation operations (e.g. joining a high number of strings), IAST might become a bottleneck due to the need to analyze each string separately.

### DIFF
--- a/dd-java-agent/agent-iast/src/main/java/com/datadog/iast/propagation/StringModuleImpl.java
+++ b/dd-java-agent/agent-iast/src/main/java/com/datadog/iast/propagation/StringModuleImpl.java
@@ -3,27 +3,23 @@ package com.datadog.iast.propagation;
 import static com.datadog.iast.taint.Ranges.EMPTY;
 import static com.datadog.iast.taint.Ranges.highestPriorityRange;
 import static com.datadog.iast.taint.Ranges.mergeRanges;
-import static com.datadog.iast.taint.Ranges.rangesProviderFor;
 import static com.datadog.iast.taint.Tainteds.canBeTainted;
 import static com.datadog.iast.taint.Tainteds.getTainted;
 
 import com.datadog.iast.model.Range;
 import com.datadog.iast.taint.Ranges;
-import com.datadog.iast.taint.Ranges.RangeList;
-import com.datadog.iast.taint.Ranges.RangesProvider;
 import com.datadog.iast.taint.TaintedObject;
 import com.datadog.iast.taint.TaintedObjects;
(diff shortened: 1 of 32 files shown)


dd-java-agent/agent-iast/src/main/java/com/datadog/iast/propagation/StringModuleImpl.java
```diff
@@ -153,38 +149,26 @@ public void onStringConcatFactory(
     }
 
     final TaintedObjects taintedObjects = ctx.getTaintedObjects();
-    final Map<Integer, Range[]> sourceRanges = new HashMap<>();
```
manuel-alvarez-alvarez, line 156: Range builder is a specialized data structure designed to work with ranges with low allocation overhead, it should provide better time/space complexity compared to do a previous lookup of tainted ranges + storage in a map.

dd-java-agent/agent-iast/src/main/java/com/datadog/iast/sink/LdapInjectionModuleImpl.java
```diff
@@ -26,16 +23,8 @@ public void onDirContextSearch(
...
-      LOG.debug("No IastRequestContext available");
-      return;
-    }
-    final TaintedObjects to = ctx.getTaintedObjects();
     checkInjection(
         VulnerabilityType.LDAP_INJECTION,
-        rangesProviderFor(to, name),
-        rangesProviderFor(to, filterExpr),
-        rangesProviderFor(to, filterArgs));
+        Iterators.join(Iterators.of(name, filterExpr), Iterators.of(filterArgs)));
```
jandro996, line 28: This should be like this?

`Iterators.join(Iterators.of(name), Iterators.of (filterExpr), Iterators.of(filterArgs)));`

jandro996 (approved):
Great Job! these changes will improve our delivery pace for new Vulnerabilities 👍

dd-java-agent/agent-iast/src/main/java/com/datadog/iast/sink/LdapInjectionModuleImpl.java
manuel-alvarez-alvarez: You save on `Iterator` allocation by joining `name` and `filterExpr` in a single one.

PR #6709, opened 2024-02-21
Add `aws.object.key` tag for aws-sdk-2

paul-laffon-dd:
# What Does This Do

Add `aws.object.key` for aws-sdk-2 to know which key is targeted by S3 get/put requests. Similar to this PR https://github.com/DataDog/dd-trace-java/pull/6555 where it was done for aws-sdk-1 

# Motivation

More information on S3 request by having the key of the targeted object. It also unlocks object-level data lineage for S3

# Additional Notes

### DIFF
--- a/dd-java-agent/instrumentation/aws-java-sdk-2.2/src/main/java/datadog/trace/instrumentation/aws/v2/AwsSdkClientDecorator.java
+++ b/dd-java-agent/instrumentation/aws-java-sdk-2.2/src/main/java/datadog/trace/instrumentation/aws/v2/AwsSdkClientDecorator.java
@@ -106,6 +106,7 @@ public class AwsSdkClientDecorator extends HttpClientDecorator<SdkHttpRequest, S
 
     // S3
     request.getValueForField("Bucket", String.class).ifPresent(name -> setBucketName(span, name));
+    getRequestKey(request).ifPresent(key -> setObjectKey(span, key));
     request
         .getValueForField("StorageClass", String.class)
         .ifPresent(
@@ -213,6 +214,21 @@ public class AwsSdkClientDecorator extends HttpClientDecorator<SdkHttpRequest, S
     span.setTag(InstrumentationTags.AWS_BUCKET_NAME, name);
     span.setTag(InstrumentationTags.BUCKET_NAME, name);
     setPeerService(span, InstrumentationTags.AWS_BUCKET_NAME, name);
+  }
+
+  private static Optional<String> getRequestKey(SdkRequest request) {
+    Optional<String> key = Optional.empty();
+    try {
+      key = request.getValueForField("Key", String.class);
+    } catch (ClassCastException ignored) {
+      // Key is not always a string, like for dynamodb GetItemRequest
+    }
+
+    return key;
+  }
+
+  private static void setObjectKey(AgentSpan span, String key) {
+    span.setTag(InstrumentationTags.AWS_OBJECT_KEY, key);
   }
 
   private static void setQueueName(AgentSpan span, String name) {
--- a/dd-java-agent/instrumentation/aws-java-sdk-2.2/src/test/groovy/Aws2ClientTest.groovy
+++ b/dd-java-agent/instrumentation/aws-java-sdk-2.2/src/test/groovy/Aws2ClientTest.groovy
@@ -17,7 +17,10 @@ import software.amazon.awssdk.http.apache.ApacheHttpClient
 import software.amazon.awssdk.regions.Region
 import software.amazon.awssdk.services.dynamodb.DynamoDbAsyncClient
 import software.amazon.awssdk.services.dynamodb.DynamoDbClient
+import software.amazon.awssdk.services.dynamodb.model.AttributeValue
 import software.amazon.awssdk.services.dynamodb.model.CreateTableRequest
+import software.amazon.awssdk.services.dynamodb.model.GetItemRequest
+import software.amazon.awssdk.services.dynamodb.model.UpdateItemRequest
 import software.amazon.awssdk.services.ec2.Ec2AsyncClient
 import software.amazon.awssdk.services.ec2.Ec2Client
 import software.amazon.awssdk.services.kinesis.KinesisClient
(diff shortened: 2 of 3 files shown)


dd-java-agent/instrumentation/aws-java-sdk-2.2/src/main/java/datadog/trace/instrumentation/aws/v2/AwsSdkClientDecorator.java
```diff
@@ -215,6 +216,21 @@ private static void setBucketName(AgentSpan span, String name) {
     setPeerService(span, InstrumentationTags.AWS_BUCKET_NAME, name);
   }
 
+  private static Optional<String> getRequestKey(SdkRequest request) {
```
dougqh, line 219: Please don't use Optional.  It typically isn't optimized away by the JIT.
Just use null instead.

dd-java-agent/instrumentation/aws-java-sdk-2.2/src/main/java/datadog/trace/instrumentation/aws/v2/AwsSdkClientDecorator.java
```diff
@@ -215,6 +216,21 @@ private static void setBucketName(AgentSpan span, String name) {
...
+    Optional<String> key = Optional.empty();
+    try {
+      key = request.getValueForField("Key", String.class);
+    } catch (ClassCastException ignored) {
+      // Key is not always a string, like for dynamodb GetItemRequest
```
dougqh, line 224: If we have the option, it would be better to do an instanceof check rather than catching a ClassCastException.

dd-java-agent/instrumentation/aws-java-sdk-2.2/src/test/groovy/Aws2ClientTest.groovy
```diff
@@ -270,6 +278,9 @@ abstract class Aws2ClientTest extends VersionedNamingTestBase {
             if (service == "S3") {
               "aws.bucket.name" "somebucket"
               "bucketname" "somebucket"
+              if (operation == "PutObject" || operation == "GetObject") {
```
dougqh, line 281: I don't see a test that covers the ClassCastException.
Is there a way to set a type other than String?

dougqh (approved):
For overhead reasons, I'm not keen on the use of Optional.
Similarly catching ClassCastException

I also don't see a test that exercises the ClassCastException case.

dd-java-agent/instrumentation/aws-java-sdk-2.2/src/main/java/datadog/trace/instrumentation/aws/v2/AwsSdkClientDecorator.java
paul-laffon-dd: Using an Optional<String> because the [getValueForField](https://github.com/aws/aws-sdk-java-v2/blob/master/services/s3/src/main/java/software/amazon/awssdk/services/s3/model/GetUrlRequest.java#L133-L144) used for all the other fields is returning an Optional. It look like this for most of s3 requests (with extra fields)

```java
public <T> Optional<T> getValueForField(String fieldName, Class<T> clazz) {
    switch (fieldName) {
        case "Bucket":
            return Optional.ofNullable(clazz.cast(bucket()));
        case "Key":
            return Optional.ofNullable(clazz.cast(key()));
        case "VersionId":
            return Optional.ofNullable(clazz.cast(versionId()));
        default:
            return Optional.empty();
    }
}
```

dd-java-agent/instrumentation/aws-java-sdk-2.2/src/test/groovy/Aws2ClientTest.groovy
paul-laffon-dd: The `ClassCastException` is covered by the DynamoDB `GetItemRequest` / `UpdateItemRequest` just above where the key is a [Map and not a String](https://sdk.amazonaws.com/java/api/latest/software/amazon/awssdk/services/dynamodb/model/GetItemRequest.html#key())

dd-java-agent/instrumentation/aws-java-sdk-2.2/src/main/java/datadog/trace/instrumentation/aws/v2/AwsSdkClientDecorator.java
paul-laffon-dd: [getValueForField](https://github.com/aws/aws-sdk-java-v2/blob/master/services/s3/src/main/java/software/amazon/awssdk/services/s3/model/GetUrlRequest.java#L133-L144) is casting the field to the provided type `Optional.ofNullable(clazz.cast(key()))`

It can be possible to do a instanceof using reflection, but I think we should keep catching the ClassCastException since getValueForField is doing a direct cast, so I am not sure if this is worth it

dd-java-agent/instrumentation/aws-java-sdk-2.2/src/test/groovy/Aws2ClientTest.groovy
dougqh: Okay.

dd-java-agent/instrumentation/aws-java-sdk-2.2/src/main/java/datadog/trace/instrumentation/aws/v2/AwsSdkClientDecorator.java
dougqh: I'm not familiar enough with the code in question to be certain myself.
It just looked a bit suspicious, so I wanted to ask.
I'll leave it to your discretion.

dd-java-agent/instrumentation/aws-java-sdk-2.2/src/main/java/datadog/trace/instrumentation/aws/v2/AwsSdkClientDecorator.java
dougqh: Okay

Issue #6704, opened 2024-02-20
Stream closed exception during Spring application startup


feser:
We started to get errors during Spring application startup when we upgraded `com.datadoghq:dd-java-agent` to `1.29.0`.

`org.springframework.asm.ClassReader.readStream` gets `java.io.IOException: Stream closed` randomly.

We have verified that we don't have this issue with `dd-java-agent:1.28.0`.

Spring boot version: `3.2.2`

Stack trace:
```
org.springframework.beans.factory.BeanDefinitionStoreException: I/O failure while processing configuration class [com.test.TestApplication]
    at org.springframework.context.annotation.ConfigurationClassParser.processConfigurationClass(ConfigurationClassParser.java:251)
    at org.springframework.context.annotation.ConfigurationClassParser.parse(ConfigurationClassParser.java:197)
    at org.springframework.context.annotation.ConfigurationClassParser.parse(ConfigurationClassParser.java:165)
    at org.springframework.context.annotation.ConfigurationClassPostProcessor.processConfigBeanDefinitions(ConfigurationClassPostProcessor.java:416)
    at org.springframework.context.annotation.ConfigurationClassPostProcessor.postProcessBeanDefinitionRegistry(ConfigurationClassPostProcessor.java:289)
    at org.springframework.context.support.PostProcessorRegistrationDelegate.invokeBeanDefinitionRegistryPostProcessors(PostProcessorRegistrationDelegate.java:349)
    at org.springframework.context.support.PostProcessorRegistrationDelegate.invokeBeanFactoryPostProcessors(PostProcessorRegistrationDelegate.java:118)
    at org.springframework.context.support.AbstractApplicationContext.invokeBeanFactoryPostProcessors(AbstractApplicationContext.java:788)
    at org.springframework.context.support.AbstractApplicationContext.refresh(AbstractApplicationContext.java:606)
    at org.springframework.boot.web.servlet.context.ServletWebServerApplicationContext.refresh(ServletWebServerApplicationContext.java:146)
    at org.springframework.boot.SpringApplication.refresh(SpringApplication.java:754)
    at org.springframework.boot.SpringApplication.refreshContext(SpringApplication.java:456)
    at org.springframework.boot.SpringApplication.run(SpringApplication.java:334)
    at org.springframework.boot.SpringApplication.run(SpringApplication.java:1354)
    at org.springframework.boot.SpringApplication.run(SpringApplication.java:1343)
[...]

PR #6720, opened 2024-02-22
Fix dependency resolution for nested JARs (with fix for Spring Boot)

smola:
# What Does This Do

* Fix dependency resolution for nested JARs, as originally implemented in #6603
* Now using our own JAR readers which avoid all interactions with Spring Boot loader (or whatever other esoteric URLConnection handlers we might find), and the end result is faster than re-using Spring Boot loader (see benchmarks below).

# Motivation

# Additional Notes

Jira ticket: [APPSEC-51864](https://datadoghq.atlassian.net/browse/APPSEC-51864)

## Original error reproduction

In the intermediate commits,  you will find a test case reproducing the original error as well as a naive fix:

* Introduce test by @manuel-alvarez-alvarez to reproduce the crash in Spring Boot 3 loader, as reported in #6704
* Apply a fix for the above issue. The issue was caused by `JarUrlConnection` caching `JarFile` instances. Under certain order of actions, we would close the same `JarFile` instance while the application is using it to read something from it, causing an exception. Now we force disable the use of caching on our side with `setUseCaches(false)`.

A test failure here would look like this (e.g. if you revert the last commit in this PR, which contains the actual fix):

```
java.io.IOException: Stream closed
	at java.util.zip.InflaterInputStream.ensureOpen(InflaterInputStream.java:68)
	at java.util.zip.InflaterInputStream.read(InflaterInputStream.java:122)
	at org.springframework.boot.loader.net.protocol.jar.LazyDelegatingInputStream.read(LazyDelegatingInputStream.java:33)
	at datadog.telemetry.dependency.DependencyResolverSpring3Specification.resolving nested dependency does not interfere with application reading the same jar(DependencyResolverSpring3Specification.groovy:110)
```

## Benchmarks

Master, no full scanning of nested dependencies:

```
DependencyResolverBenchmark.resolveNestedSpringBootJar  thrpt   25  288.602 ± 11.002  ops/ms
DependencyResolverBenchmark.resolveSimpleJar            thrpt   25   20.920 ±  2.308  ops/ms
```

Naive solution, re-using Spring's JAR URLConnection handler:

```
DependencyResolverBenchmark.resolveNestedSpringBootJar  thrpt   25   5.524 ± 0.162  ops/ms
DependencyResolverBenchmark.resolveSimpleJar            thrpt   25  20.918 ± 2.299  ops/ms
```

Final state, with our own JAR reader:

```
DependencyResolverBenchmark.resolveNestedSpringBootJar  thrpt   25  10.131 ± 0.364  ops/ms
DependencyResolverBenchmark.resolveSimpleJar            thrpt   25  22.093 ± 2.292  ops/ms
```

[APPSEC-51864]: https://datadoghq.atlassian.net/browse/APPSEC-51864?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ


manuel-alvarez-alvarez (approved):
LGTM

telemetry/src/main/java/datadog/telemetry/dependency/DependencyResolver.java
```diff
@@ -20,110 +15,38 @@ public class DependencyResolver {
...
-    String scheme = uri.getScheme();
-    List<Dependency> dependencies = Collections.emptyList();
+    final String scheme = uri.getScheme();
     try {
+      JarReader.Extracted metadata = null;
+      String path = null;
       if ("file".equals(scheme)) {
         File f;
         if (uri.isOpaque()) {
           f = new File(uri.getSchemeSpecificPart());
```
manuel-alvarez-alvarez, line 25: What is the reason of doing it like this instead of:

```java
File f = Paths.get(uri).toFile();
```

I think this part is not covered by any test :cry:

telemetry/src/main/java/datadog/telemetry/dependency/JarReader.java
```diff
@@ -0,0 +1,87 @@
...
+    final int sepIdx = jarPath.indexOf("!/");
+    if (sepIdx == -1) {
+      throw new IllegalArgumentException("Invalid nested jar path: " + jarPath);
+    }
+    final String outerJarPath = jarPath.substring(0, sepIdx);
+    String innerJarPath = jarPath.substring(sepIdx + 2);
+    if (innerJarPath.endsWith("!/")) {
+      innerJarPath = innerJarPath.substring(0, innerJarPath.length() - 2);
+    }
+    try (final JarFile outerJar = new JarFile(outerJarPath, false /* no verify */)) {
```
manuel-alvarez-alvarez, line 64: Out of curiosity, but is it a possibility to have more levels of nested jars?

smola: I have seen at least 1 user report at Spring Boot asking about a third level of nesting, but Spring Boot does not support that. It would definitely be possible to do it with a custom `URLConnection` handler. But until we find some real scenario, I would rather not implement it.

Also note that, while it should be relatively easy to add support for an arbitrary nesting depth, we would always set a limit to nesting for performance reasons since each additional level of nesting needs to decompress a stream.

telemetry/src/main/java/datadog/telemetry/dependency/DependencyResolver.java
smola: The background story here is tricky, so I'm not touching this logic until I can test it properly. We'll probably require Windows tests for it. See https://github.com/DataDog/dd-trace-java/pull/4820

PR #6721, opened 2024-02-22
Add DSM support to SNS

vandonr:
# What Does This Do

Adds DSM checkpointing to SNS.

 - only on send, because the receive end of SNS is either an SQS queue or an http endpoint
 - without injection, because doing so would require having a separate instrumentation for SNS, which would be a lot more work.

# Motivation

This is not a messaging system per say, but it's been implemented in other languages (python, node), and it makes sense to be able to see it in a data pipeline view.


dd-java-agent/instrumentation/aws-java-sdk-2.2/src/test/groovy/Aws2SnsDataStreamsTest.groovy
```diff
@@ -0,0 +1,306 @@
...
+    builder.addExecutionInterceptor(new ExecutionInterceptor() {
+      @Override
+      void afterExecution(Context.AfterExecution context, ExecutionAttributes executionAttributes) {
+        callback.call()
+      }
+    })
+  }
+
+  @Unroll
+  def "send #operation request with builder #builder.class.getSimpleName() mocked response"() {
```
amarziali, line 119: According to the PR description we're going to instrument the send side. However we also have the receive side with sqs. Would it make sense to have a more complete scenario (using testcontainer, e.g. localstack or others) to test what's happening if we publish a sns message and we receive it in a sqs queue? If not interesting, I would expect at least to have a test case for the batch

vandonr: I think sending via sns and receiving with sqs would be a pretty boring test because we don't do any injection, so we'd just get vanilla checkpoints.
I added tests for the batch method though, thanks for reminding me :)

dd-java-agent/instrumentation/aws-java-sdk-1.11.0/src/dsmTest/groovy/AWS1SnsClientTest.groovy
```diff
@@ -0,0 +1,198 @@
...
+  def timestamp = Instant.now().minusSeconds(60)
+
+  @Shared
+  def timestamp2 = timestamp.plusSeconds(1)
+
+  @Override
+  protected void configurePreAgent() {
+    super.configurePreAgent()
+    // the actual service returns cbor encoded json
+    System.setProperty("com.amazonaws.sdk.disableCbor", "true")
```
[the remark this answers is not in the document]
vandonr: ah ok. I copied the existing DSM test for kinesis, and edited from there, but this might not be needed. I'll try without, and I'll check if it's needed for kinesis as well.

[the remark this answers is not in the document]
vandonr: Yeah it's especially not needed for this particular test since we don't use the return value of the call

utils/test-utils/src/main/groovy/datadog/trace/test/util/DDSpecification.groovy
```diff
@@ -90,8 +90,6 @@ abstract class DDSpecification extends Specification {
 
   void setupSpec() {
     assert !configModificationFailed: "Config class modification failed.  Ensure all test classes extend DDSpecification"
-    assert System.getenv().findAll { it.key.startsWith("DD_") }.isEmpty()
```
amarziali, line 93: this file is probably changed by error (I do the same since I've DD_ env but you should not commit them)

vandonr: ah yes, my bad, I think I added a `-a` by reflex to my `git commit` by mistake

amarziali (approved):
You can rerun also the CI to see if tests pass a second time. I remember that the one for kinesis were flaky and I'd avoid merging flaky test.  LGTM for me

vandonr:
> I remember that the one for kinesis were flaky and I'd avoid merging flaky test.
I think you're right, they might be flaky 😬

amarziali:
> > I remember that the one for kinesis were flaky and I'd avoid merging flaky test.
> 
> I think you're right, they might be flaky 😬

I remember that I did some investigations and it might be due to the duration of collection of bucket stats since we assume to collect x events in one bucket. It would be great if you find a way to remove flakiness of those DSM tests

vandonr:
I was able to reproduce the issue by adding a `Thread.sleep()` in the for loop where the checkpoints are produced, and actually the solution had already been implemented in the sdk 1.11 tests, by querying al results and using a `conditions.eventually` to retry the assert.

PR #6729, opened 2024-02-23
Limit DependencyResolverQueue size with a config flag

nayeem-kamal:
# What Does This Do
Allows the user to set a limit to the size of the telemetry `DependencyResolverQueue` to prevent OOM errors when there are a large number of dependencies
Introduces a new config variable (`dd.telemetry.dependency-resolution.queue.size`) which takes an integer to set as the limit for the number of dependencies to store in the queue.
# Motivation
Resolve [APMS-11445](https://datadoghq.atlassian.net/browse/APMS-11445?atlOrigin=eyJpIjoiYWJjZmQzOGIyMzhhNDRlOTkyZDZmMWE1ZDc4YTM2YTMiLCJwIjoiaiJ9)
# Additional Notes

Jira ticket: [APMS-11445]

[APMS-11445]: https://datadoghq.atlassian.net/browse/APMS-11445?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ
[APMS-11445]: https://datadoghq.atlassian.net/browse/APMS-11445?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ

### DIFF
--- a/telemetry/src/main/java/datadog/telemetry/dependency/DependencyResolverQueue.java
+++ b/telemetry/src/main/java/datadog/telemetry/dependency/DependencyResolverQueue.java
@@ -49,7 +49,9 @@ public class DependencyResolverQueue {
 
       // this resolver will be disabled so we can clear the stored URIs from processedUrlsSet
       // since they will no longer be checked against for duplicates
-      processedUrlsSet.clear();
+      synchronized (this) {
+        processedUrlsSet.clear();
+      }
       return;
     }
 
--- a/telemetry/src/main/java/datadog/telemetry/dependency/DependencyResolverQueue.java
+++ b/telemetry/src/main/java/datadog/telemetry/dependency/DependencyResolverQueue.java
@@ -1,5 +1,6 @@
 package datadog.telemetry.dependency;
 
+import datadog.trace.api.Config;
 import java.net.URI;
 import java.util.Collections;
 import java.util.HashSet;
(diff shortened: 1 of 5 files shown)


smola (approved):
We could probably set some large but not `MAX_INTEGER` by default.

telemetry/src/main/java/datadog/telemetry/dependency/DependencyResolverQueue.java
```diff
@@ -16,17 +17,28 @@ public class DependencyResolverQueue {
...
+    newUrlsQueue = new ConcurrentLinkedQueue<>();
+    processedUrlsSet = new HashSet<>();
+  }
+
   public void queueURI(URI uri) {
     if (uri == null) {
       return;
     }
-
+    if (processedUrlsSet.size() >= MAX_QUEUE_SIZE || newUrlsQueue.size() >= MAX_QUEUE_SIZE) {
```
smola, line 38: Once we're past this threshold, we'll stop processing dependencies. So we could just set a boolean flag (e.g. `disabled`) and use that in later checks. This would allow:
* Emptying `processedUrlSet` once we're past the limit. There's no point in keeping it around. We should probablu check if the queue is empty before clearing it, to avoid duplicates.
* Print a single message, only the first time it becomes full, rather than on every enqueued dependency. That would make it far less noisy, and then we could increment it from debug to warning level.

nayeem-kamal: Thanks for the review @smola!! I think these are all good ideas. I have updated my implementation to reflect these changes. 
WRT "We should probably check if the queue is empty before clearing it, to avoid duplicates." : my new implementation should never reach the duplicate check again (where `processedUrlSet` is used) once the queuing is disabled so I didn't include an emptiness check for `newUrlsQueue` as even it is not empty, we won't need to check against the set. Let me know if the implementation looks good or if I may have missed something else.

PR #6738, opened 2024-02-26
collapse implementations of timer and checkpointer interfaces into profiling context

richardstartin:
# What Does This Do

Simplifies the way the tracer and profiler integrate. Currently there are 3 distinct services: labeling (callbacks per scope activation), endpoint events (one per trace), timing (one per continuation for a subset of continuations). This complexity is reflected in the way the profiler is set up, where each of these services needs to be instantiated by reflection. In the long run, we'd like to get rid of this reflective mechanism entirely, but by refactoring the three services into one the surface area is reduced and it will make it easier to change in the future.

# Motivation

# Additional Notes

Jira ticket: [PROJ-IDENT]


jbachorik (approved):
Looks good

PR #6741, opened 2024-02-27
Add support for precise tainting of byte array objects

manuel-alvarez-alvarez:
# What Does This Do
Adds support for accurate tainting of byte array instances.

# Motivation
We need to keep track of multiple ranges in order to add proper support for Kafka 3.x as they use byte buffers extensively.

### DIFF
--- a/dd-java-agent/agent-iast/src/main/java/com/datadog/iast/propagation/PropagationModuleImpl.java
+++ b/dd-java-agent/agent-iast/src/main/java/com/datadog/iast/propagation/PropagationModuleImpl.java
@@ -12,6 +12,7 @@ import com.datadog.iast.taint.TaintedObject;
 import com.datadog.iast.taint.TaintedObjects;
 import com.datadog.iast.taint.Tainteds;
 import com.datadog.iast.util.ObjectVisitor;
+import com.datadog.iast.util.Ranged;
 import datadog.trace.api.Config;
 import datadog.trace.api.iast.IastContext;
 import datadog.trace.api.iast.Taintable;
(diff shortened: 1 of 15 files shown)


dd-java-agent/agent-iast/src/main/java/com/datadog/iast/propagation/PropagationModuleImpl.java
```diff
@@ -118,6 +140,45 @@ public void taintIfTainted(
...
+  }
+
+  @Override
+  public void taintIfTainted(
+      @Nullable final IastContext ctx,
+      @Nullable final Object target,
+      @Nullable final Object input,
+      final int start,
+      final int length,
+      boolean keepRanges,
```
smola, line 161: I'm a bit concerned about the excessive overloading `taintIfTainted` functions, especially adding boolean parameters (not obvious from the call).

dd-java-agent/agent-iast/src/main/java/com/datadog/iast/taint/Ranges.java
```diff
@@ -217,6 +243,36 @@ public static Range[] newArray(final long size) {
     return new Range[size > MAX_RANGE_COUNT ? MAX_RANGE_COUNT : (int) size];
   }
 
+  /** Insert the new range maintaining order (it assumes that both arrays are already sorted) */
+  public static Range[] insert(final Range[] currentRanges, final Range... newRanges) {
```
smola, line 247:
```suggestion
  public static Range[] append(final Range[] currentRanges, final Range... newRanges) {
```

Can we call this `append` or `concat`? `insert` sounds like an in-place operation.

smola (approved):
LGTM. Just a couple of stylistic comments.

jandro996 (approved):

ValentinZakharov (approved):

dd-java-agent/agent-iast/src/main/java/com/datadog/iast/taint/Ranges.java
manuel-alvarez-alvarez: Changed into `mergeRangesSorted` as it makes more sense as in a merge sort algo (also rename the vars to make it clear what it does)

dd-java-agent/agent-iast/src/main/java/com/datadog/iast/propagation/PropagationModuleImpl.java
manuel-alvarez-alvarez: Agreed the API is too verbose and makes it difficult to reason about it and of course use it. Can we try to do a refactor in a separate PR?

PR #6214, opened 2023-11-14
Append DBM trace context comments for SQL Server & Postgres

jmeunier28:
# What Does This Do

Fix for reported customer issue around calling procedures for `postgresql` while having DBM-APM link enabled 

```
2023-11-10 13:30:52.870 [Thread-66184] [Error] [modifyJdbcCall] [api_call_unique_identifier-2026618091699623052870] [Client_Id-0] [span_id-99074af426ba7b60ccc2] - com.paymentology.payapi.v1.logging.ApiLogger
Malformed function or procedure escape syntax at offset 40.

org.postgresql.util.PSQLException: Malformed function or procedure escape syntax at offset 40.
at org.postgresql.core.Parser.modifyJdbcCall(Parser.java:1146)
at org.postgresql.core.CachedQueryCreateAction.create(CachedQueryCreateAction.java:48)
at org.postgresql.core.CachedQueryCreateAction.create(CachedQueryCreateAction.java:19)
at org.postgresql.util.LruCache.borrow(LruCache.java:123)
at org.postgresql.core.QueryExecutorBase.borrowCallableQuery(QueryExecutorBase.java:301)
at org.postgresql.jdbc.PgConnection.borrowCallableQuery(PgConnection.java:200)
at org.postgresql.jdbc.PgCallableStatement.<init>(PgCallableStatement.java:54)
at org.postgresql.jdbc.PgConnection.prepareCall(PgConnection.java:1333)
at org.postgresql.jdbc.PgConnection.prepareCall(PgConnection.java:1786)
at org.postgresql.jdbc.PgConnection.prepareCall(PgConnection.java:459)
at com.zaxxer.hikari.pool.ProxyConnection.prepareCall(ProxyConnection.java:295)
at com.zaxxer.hikari.pool.HikariProxyConnection.prepareCall(HikariProxyConnection.java)
at com.paymentology.payapi.v1.logging.ApiLogger.lambda$SaveError$7(ApiLogger.java:320)
at java.lang.Thread.run(Thread.java:750)
```

This happens when running prepared callable statements. E.g. This code:

```java
          final String storedProcedureCall = "{call dogshelterProc(?, ?)}";
          CallableStatement callableStmt = con.prepareCall(storedProcedureCall);
```

... results in the following exception 
```
exception running jdbc ERROR: syntax error at or near "{"
  Position: 95
 ```
 
[...]

PR #6743, opened 2024-02-27
Don't inject comment for callable JDBC statements

sethsamuel:
# What Does This Do
When parsing JDBC callable statements we now skip injecting the DBM tracer comment.

# Motivation
The JDBC parse has a very strict grammar that doesn't allow for SQL comments in any position. I attempted before the brace, after the brace, and inserting new lines. Examining https://github.com/pgjdbc/pgjdbc/blob/master/pgjdbc/src/main/java/org/postgresql/core/Parser.java#L1038 suggests that we can't put a comment in there anywhere.

# Additional Notes
This is an incremental fix that will at least stop users from crashing. Ideally we could inject our comment closer to the database call, i.e., after the JDBC parser has transformed JDBC callables into SQL. Attempting to simply call the parser ourselves earlier in the execution process failed because it interferes with Java's ability to capture the output variable.

Jira ticket: [SDBM-611](https://datadoghq.atlassian.net/browse/SDBM-611)

[SDBM-611]: https://datadoghq.atlassian.net/browse/SDBM-611?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ


dd-java-agent/instrumentation/jdbc/src/main/java/datadog/trace/instrumentation/jdbc/DBMCompatibleConnectionInstrumentation.java
```diff
@@ -100,7 +100,7 @@ public static String onEnter(
         final DBInfo dbInfo =
             JDBCDecorator.parseDBInfo(
                 connection, InstrumentationContext.get(Connection.class, DBInfo.class));
-        sql = SQLCommenter.append(sql, DECORATE.getDbService(dbInfo));
+        sql = SQLCommenter.prepend(sql, DECORATE.getDbService(dbInfo));
```
sethsamuel, line 103: Reverting this since it didn't help the callable case and @jmeunier28 mentioned appending can come with its own issues.

dd-java-agent/instrumentation/jdbc/src/test/groovy/RemoteJDBCInstrumentationTest.groovy
```diff
@@ -421,6 +422,34 @@ abstract class RemoteJDBCInstrumentationTest extends VersionedNamingTestBase {
...
+    TEST_WRITER.waitForTraces(1)
+
+    then:
+    assert upperProc.getString(2) == "HELLO WORLD"
+    cleanup:
+    upperProc?.close()
+    connection.close()
+
+    where:
+    driver       | connection                                              | query
```
dougqh, line 445: You should test this with extra leading whitespace as well.

sethsamuel:
> I'm curious if leading whitespace is allowed in the call SQL. If so, there should be a test case that covers that.

Good call, JDBC does support leading whitespace, so added regex handler and test cases.

dd-java-agent/instrumentation/jdbc/src/main/java/datadog/trace/instrumentation/jdbc/SQLCommenter.java
```diff
@@ -45,8 +45,9 @@ public static String inject(
       return sql;
     }
 
-    if (sql.matches("\\s*\\{.*")) {
-      //      The JDBC parser doesn't allow SQL comments anywhere in a callable statement
+    System.out.println("SERVICE " + dbService);
+    if (sql.matches("\\s*\\{.*") && dbService.startsWith("postgresql")) {
```
dougqh: Regexes especially uncompiled are very bad for performance.
You are better off writing a simple character scanner.

dd-java-agent/instrumentation/jdbc/src/main/java/datadog/trace/instrumentation/jdbc/SQLCommenter.java
```diff
@@ -45,8 +45,9 @@ public static String inject(
       return sql;
     }
 
-    if (sql.matches("\\s*\\{.*")) {
-      //      The JDBC parser doesn't allow SQL comments anywhere in a callable statement
+    System.out.println("SERVICE " + dbService);
```
dougqh: Debugging write should be removed

dd-java-agent/instrumentation/jdbc/src/main/java/datadog/trace/instrumentation/jdbc/DBMCompatibleConnectionInstrumentation.java
jmeunier28: you could just revert this PR https://github.com/DataDog/dd-trace-java/pull/6214 and work off of that. The change is a bit more involved than this

dd-java-agent/instrumentation/jdbc/src/main/java/datadog/trace/instrumentation/jdbc/SQLCommenter.java
```diff
@@ -44,6 +44,19 @@ public static String inject(
     if (hasDDComment(sql, appendComment)) {
       return sql;
     }
+
+    //      The Postgres JDBC parser doesn't allow SQL comments anywhere in a callable statement
+    //
```
jmeunier28, line 49: I'm surprised this wasn't caught by the existing system tests. IIRC we have one for this exact case

sethsamuel: I saw that in there (still struggling to get them running locally at all) and am equally confused. The crash is totally reproducible and looking at the parser there's no way it shouldn't crash when injecting the comment before the JDBC parse. Maybe a driver/version issue?

jmeunier28:
> This is an incremental fix that will at least stop users from crashing. Ideally we could inject our comment closer to the database call, i.e., after the JDBC parser has transformed JDBC callables into SQL. Attempting to simply call the parser ourselves earlier in the execution process failed because it interferes with Java's ability to capture the output variable.

makes sense, I think this means we will have to instrument at the driver level for postgres

dd-java-agent/instrumentation/jdbc/src/main/java/datadog/trace/instrumentation/jdbc/DBMCompatibleConnectionInstrumentation.java
```diff
@@ -100,7 +99,11 @@ public static String onEnter(
         final DBInfo dbInfo =
             JDBCDecorator.parseDBInfo(
                 connection, InstrumentationContext.get(Connection.class, DBInfo.class));
-        sql = SQLCommenter.append(sql, DECORATE.getDbService(dbInfo));
+        if (dbInfo.getType().equals("sqlserver")) {
+          sql = SQLCommenter.append(sql, DECORATE.getDbService(dbInfo), dbInfo.getType());
+        } else {
+          sql = SQLCommenter.prepend(sql, DECORATE.getDbService(dbInfo), dbInfo.getType());
```
jmeunier28, line 105: If we are going to do this, then we may as well add the classes from here https://github.com/DataDog/dd-trace-java/blob/master/dd-java-agent/instrumentation/jdbc/src/main/java/datadog/trace/instrumentation/jdbc/MySQLConnectionInstrumentation.java back to this one & remove that extra class

dd-java-agent/instrumentation/jdbc/src/main/java/datadog/trace/instrumentation/jdbc/DBMCompatibleConnectionInstrumentation.java
```diff
@@ -3,9 +3,7 @@
 import static datadog.trace.agent.tooling.bytebuddy.matcher.HierarchyMatchers.hasInterface;
 import static datadog.trace.agent.tooling.bytebuddy.matcher.NameMatchers.nameStartsWith;
 import static datadog.trace.agent.tooling.bytebuddy.matcher.NameMatchers.named;
-import static datadog.trace.instrumentation.jdbc.JDBCDecorator.DECORATE;
-import static datadog.trace.instrumentation.jdbc.JDBCDecorator.INJECT_COMMENT;
-import static datadog.trace.instrumentation.jdbc.JDBCDecorator.logQueryInfoInjection;
+import static datadog.trace.instrumentation.jdbc.JDBCDecorator.*;
```
jmeunier28, line 6: I think the preferred IDE settings for this are to import the full path, but I'll defer to the java team for that

jmeunier28 (approved):
LGTM for the database-monitoring side of things

dd-java-agent/instrumentation/jdbc/src/main/java/datadog/trace/instrumentation/jdbc/SQLCommenter.java
```diff
@@ -44,6 +45,21 @@ public static String inject(
...
       return sql;
     }
+
+    if (dbType != null && dbType.startsWith("postgres")) {
+      //      The Postgres JDBC parser doesn't allow SQL comments anywhere in a callable statement
+      //
+      // https://github.com/pgjdbc/pgjdbc/blob/master/pgjdbc/src/main/java/org/postgresql/core/Parser.java#L1038
+      //      TODO: Could we inject the comment after the JDBC has been converted to standard SQL?
+      int charIndex = 0;
+      while (charIndex < sql.length() && Character.isWhitespace(sql.charAt(charIndex))) {
```
dougqh, line 55: Stylistically, I'd prefer this in a separate helper function, and I'd prefer a for loop to a white loop.
But that's minor, so I'll approve and let you decide.

dougqh (approved):

PR #6748, opened 2024-02-28
remove unneeded @Unroll annotation

am312:
# What Does This Do

Removes the unnecessary `@Unroll` annotation from groovy files

# Motivation

https://spockframework.org/spock/docs/2.2-SNAPSHOT/migration_guide.html#_unroll_changes


smola (approved):
Looks good. Any idea why `PowerWAFModuleSpecification` needs it? (note this suite was failing yesterday in master, and also has a flaky test, I can review any failure)

amarziali (approved):

am312:
> Looks good. Any idea why `PowerWAFModuleSpecification` needs it?  

That's the only usage where it's `@Unroll` is used to define a custom format in addition to the name of the test.  Maybe it could also be removed in another PR if we're happy to rename the test to include the details from the `@Unroll` parameter instead.

PR #6758, opened 2024-02-28
Fix NPE on traceState() in DDSpanLink.java

cecile75:
# What Does This Do

Ensure `tracestate` from span links can't be null.

# Motivation
NPE observed.

# Additional Notes

Jira ticket: [[APMS-11554](https://datadoghq.atlassian.net/browse/APMS-11554)]

[APMS-11554]: https://datadoghq.atlassian.net/browse/APMS-11554?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ


PerfectSlayer (commented):
Before merging, I would like to check why there is `null` value in the first place.

PerfectSlayer (changes requested):
Hey, I would rather see the constructor changed in order to prevent any further issue.

To do this, you would need to check if `traceState` is `null` in `SpanLink` constructor and set an empty string `""` if it is.
Similarly, you can check for `null` `attributes` and set it to `EMPTY` from `SpanLinkAttributes`. 

The main idea is, if we fix the usage of the variable, we will need to fix everywhere is is (and will be) used. Whereas if you fix it where is it set, you fix it once for all (as the variable is `final`, once set it won’t change).

PerfectSlayer (approved):

PerfectSlayer:
Thanks for the fix! 💪  I update the PR description to match the new changes.

PR #6765, opened 2024-03-01
API Security schema computes at the end of request

ValentinZakharov:
# What Does This Do
Computation of API Security schema moved to the end of request (after submit response)

# Motivation
Computing schema imposed additional latency overhead.

# Additional Notes
This is preparation work for improvement of API Security Sampling mechanism.


dd-java-agent/appsec/src/main/java/com/datadog/appsec/gateway/GatewayBridge.java
```diff
@@ -588,6 +583,37 @@ private Flow<Void> maybePublishResponseData(AppSecRequestContext ctx) {
     }
   }
 
+  private void maybeExtractSchemas(AppSecRequestContext ctx) {
+    boolean extractSchema = false;
+    if (Config.get().isApiSecurityEnabled() && requestSampler != null) {
+      extractSchema = requestSampler.sampleRequest();
+    }
+
+    if (!extractSchema) return;
```
jandro996, line 592: IMHO use always {} in if statements is more readable

jandro996 (approved):

PR #6773, opened 2024-03-05
Minify AppSec rules

smola:
# What Does This Do
Minify AppSec rules.

# Motivation
Reduce JAR size and load time.

# Additional Notes

This partially offsets the startup time regression introduced by https://github.com/DataDog/dd-trace-java/pull/6754 when `DD_APPSEC_ENABLED=true`.

### DIFF
--- a/dd-java-agent/appsec/build.gradle
+++ b/dd-java-agent/appsec/build.gradle
@@ -1,3 +1,6 @@
+import groovy.json.JsonOutput
+import groovy.json.JsonSlurper
+
 plugins {
   id "com.github.johnrengelman.shadow"
   id "me.champeau.jmh"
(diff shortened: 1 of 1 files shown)


mcculls (approved):

manuel-alvarez-alvarez (commented):
Doesn't it make more sense to minify during at build time?, it might make it easier to spot diffs when upgrading the rules.

smola:
@manuel-alvarez-alvarez Yeah, I guess that would also help extending this to a few other JSONs we have around. We could try https://github.com/gradle-webtools/gradle-minify-plugin or whatever that does not require pulling in Node.

smola:
@manuel-alvarez-alvarez Here's an attempt at doing it at build time. I did not extend to other modules at the moment.

manuel-alvarez-alvarez (approved):

PR #6777, opened 2024-03-05
improve instrumentation doc

amarziali:
# What Does This Do

# Motivation

# Additional Notes

Jira ticket: [PROJ-IDENT]

### DIFF
--- a/docs/how_instrumentations_work.md
+++ b/docs/how_instrumentations_work.md
@@ -94,7 +94,10 @@ To run muzzle on your instrumentation, run:
   It checks that the types and methods used by the instrumentation are present in particular versions of libraries.
   It can be subverted with `MethodHandle` and reflection, so muzzle passing is not the end of the story.
 
-**TODO: discuss why 'name' is not always included.**
+By default, all the muzzle directives are checked against all the instrumentations included in a module.
+However, there can be situations in which it’s only needed to check one specific directive on an instrumentation.
+At this point the instrumentation should override the method `muzzleDirective()` by returning the name of the directive to execute.
+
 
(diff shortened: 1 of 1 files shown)


docs/how_instrumentations_work.md
```diff
@@ -527,6 +530,8 @@ It is important to understand that even though they look like maps, since the va
 retrieve a value if you use the exact same key object as when it was set. Using a different object that is “`.equals()`”
 to the first will yield nothing.
 
+Also, the ContextStore is not supporting null keys, hence null check must be enforced before using an object as a key.
```
bm1549, line 533:
```suggestion
Since `ContextStore` does not support null keys, null checks must be enforced _before_ using an object as a key.
```

docs/how_instrumentations_work.md
```diff
@@ -536,6 +541,8 @@ and [decremented](https://github.com/DataDog/dd-trace-java/blob/9d5c7ea524cfec98
 or [reset](https://github.com/DataDog/dd-trace-java/blob/9d5c7ea524cfec982176e687a489fc8c2865e445/dd-java-agent/instrumentation/java-http-client/src/main/java11/datadog/trace/instrumentation/httpclient/SendAdvice.java#L44))
 when exiting.
 
+Please note that it works only if the methods are called on the same thread (since the counter is a threadlocal variable)
```
bm1549, line 544:
```suggestion
This only works if the methods are called on the same thread since the counter is a ThreadLocal variable.
```

bm1549 (approved):
Added some minor suggestions - otherwise LGTM

PR #6785, opened 2024-03-06
Filter out bridge methods when instrumenting methods

evanchooly:
# What Does This Do

This PR will filter out bridge methods from instrumentation to reduce the number of duplicate events from a probe.

# Motivation
Bridge methods are generated by javac to support generics and are invisible to end users.  Currently they get instrumented along with user methods resulting in trace information that is duplicated by the actual method of interest.

Jira ticket: [DEBUG-2177]

[DEBUG-2177]: https://datadoghq.atlassian.net/browse/DEBUG-2177?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ


evanchooly:
This PR needs the circle CI run to complete but it's blocked because of a missing config file.  Where do I go from here?

PR #6787, opened 2024-03-06
Fixes muzzle reporting to be more accurate for telemetry

zarirhamza:
# What Does This Do
Remove `spring-aop` from being reported as supported due to the muzzle reports even though it is simply a dependency for `spring-data-commons`

# Motivation
Cleaning up reporting code to improve future endeavors to rely on version data from tracer telemetry data extracted from Muzzle reports

# Additional Notes

Jira ticket: [PROJ-IDENT]

### DIFF
--- a/buildSrc/src/main/groovy/MuzzlePlugin.groovy
+++ b/buildSrc/src/main/groovy/MuzzlePlugin.groovy
@@ -248,8 +248,6 @@ class MuzzlePlugin implements Plugin<Project> {
       }
     }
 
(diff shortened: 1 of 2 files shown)


mcculls (changes requested):
Rather than hard-code the exclusion in this generic plugin I would prefer to add a new boolean property to `MuzzleDirective` such as `skipFromReport` which defaults to `false`

You could then set `skipFromReport = true` in the `spring-aop` muzzle directive in the `spring-data` module, which would then also be self-documenting.

mcculls (approved):

PR #6789, opened 2024-03-08
create TaskRunnerInstrumentation

am312:
# What Does This Do

Adds TaskRunnerInstrumentation to java-concurrent

### DIFF
--- a/dd-java-agent/instrumentation/java-concurrent/build.gradle
+++ b/dd-java-agent/instrumentation/java-concurrent/build.gradle
@@ -1,3 +1,7 @@
+ext {
+  latestDepTestMinJavaVersionForTests = JavaVersion.VERSION_21
+}
+
 muzzle {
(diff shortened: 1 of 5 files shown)


richardstartin:
`TaskRunner` is used to propagate context when there isn't a future available. This is better than attempting to use the application's `Runnable` because the `TaskRunner`'s lifecycle is known to be contained within the `executor`, whereas instrumentation of the executor would be wrong if the same `Runnable` instance were reused. If the task is rejected, the `RejectedExecutionHandler` instrumentation already takes care of cancelling the continuation. When there is a future, no new instrumentation is required because `ThreadBoundFuture extends FutureTask` which means the `RunnableFutureInstrumentation` and `RejectedExecutionHandlerInstrumentation` take effect.

richardstartin (approved):

PR #6812, opened 2024-03-14
Add DSM checkpoint for S3 calls in aws-sdk-1

paul-laffon-dd:
# What Does This Do

Add DSM checkpoint when reading/writing to S3 using `aws-sdk-1`

# Motivation

Connect services based on which data they are reading/writing in S3

# Additional Notes

Using the [Content-Length](https://developer.mozilla.org/en-US/docs/Web/HTTP/Headers/Content-Length) header in order to get the payload size

Adding two new DSM tags:
- `ds.name`: full path to the object
- `ds.namespace` bucket of the object

### DIFF
--- a/dd-java-agent/instrumentation/aws-java-sdk-1.11.0/src/main/java/datadog/trace/instrumentation/aws/v0/AwsSdkClientDecorator.java
+++ b/dd-java-agent/instrumentation/aws-java-sdk-1.11.0/src/main/java/datadog/trace/instrumentation/aws/v0/AwsSdkClientDecorator.java
@@ -6,6 +6,7 @@ import com.amazonaws.AmazonWebServiceRequest;
 import com.amazonaws.AmazonWebServiceResponse;
 import com.amazonaws.Request;
 import com.amazonaws.Response;
+import com.amazonaws.http.HttpMethodName;
 import datadog.trace.api.Config;
 import datadog.trace.api.DDTags;
 import datadog.trace.api.cache.DDCache;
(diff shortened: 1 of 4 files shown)


dd-java-agent/instrumentation/aws-java-sdk-1.11.0/src/main/java/datadog/trace/instrumentation/aws/v0/AwsSdkClientDecorator.java
```diff
@@ -240,9 +244,61 @@ public AgentSpan onResponse(final AgentSpan span, final Response response) {
...
+        && response.getHttpResponse().getRequest() != null) {
+
+      Object key = span.getTag(InstrumentationTags.AWS_OBJECT_KEY);
+      Object bucket = span.getTag(InstrumentationTags.AWS_BUCKET_NAME);
+      HttpMethodName httpMethod = response.getHttpResponse().getRequest().getHttpMethod();
+
+      if (key != null && bucket != null && httpMethod != HttpMethodName.HEAD) {
+        long contentLength = 0;
+
+        LinkedHashMap<String, String> sortedTags = new LinkedHashMap<>();
```
dougqh, line 260: LinkedHashMap is problematic from an overhead perspective.  As I've said previously, DSM should use a custom object for this purpose rather than a LinkedHashMap.

Realistically, I don't expect DSM to stop using a LinkedHashMap this time, but you should at least set the initial capacity for LinkedHashMap.

dougqh (approved):

kr-igor:
This PR requires some additional work, as in the current implementation DSM stats are aggregated without taking ds tags into account. Please, do not merge yet.

paul-laffon-dd:
Implemented in https://github.com/DataDog/dd-trace-java/pull/6859

PR #6814, opened 2024-03-14
Update base image used for lib injection

randomanderson:
# What Does This Do
Updates the base image for lib injection.

# Motivation
The old image has reported vulnerabilities

# Additional Notes

Jira ticket: VULN-5886

### DIFF
--- a/lib-injection/Dockerfile
+++ b/lib-injection/Dockerfile
@@ -1,7 +1,7 @@
 FROM scratch AS java_agent
 COPY dd-java-agent.jar /
 
-FROM alpine:3.18.3
+FROM alpine:3.19.1
 
 ARG UID=10000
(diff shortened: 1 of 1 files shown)


Kyle-Verhoog (approved):

lib-injection/Dockerfile
```diff
@@ -1,7 +1,7 @@
 FROM scratch AS java_agent
 COPY dd-java-agent.jar /
 
-FROM alpine:3.18.3
+FROM alpine:3.19.1
```
bm1549, line 4: Does `alpine:latest` or `alpine:3` make sense here? The container is very simple so I wouldn't expect updates to cause issues here

bm1549 (approved):

PR #6817, opened 2024-03-15
Use JDK11 to build javadoc of the internal-api-9 module

manuel-alvarez-alvarez:
# What Does This Do
Removes the error during the assemble of the `internal-api-9` module due to using the wrong JDK to build the JavaDoc documentation.

image

### DIFF
--- a/internal-api/internal-api-9/build.gradle
+++ b/internal-api/internal-api-9/build.gradle
@@ -9,6 +9,16 @@ ext {
 apply from: "$rootDir/gradle/java.gradle"
 apply plugin: "idea"
 
+java {
+  toolchain {
+    languageVersion = JavaLanguageVersion.of(11)
+  }
+}
+
+tasks.withType(Javadoc) {
+  javadocTool.set(javaToolchains.javadocToolFor(java.toolchain))
+}
+
 [JavaCompile, GroovyCompile].each {
   tasks.withType(it).configureEach {
     setJavaVersion(it, 11)


internal-api/internal-api-9/build.gradle
```diff
@@ -9,6 +9,16 @@ ext {
 apply from: "$rootDir/gradle/java.gradle"
 apply plugin: "idea"
 
+java {
+  toolchain {
+    languageVersion = JavaLanguageVersion.of(11)
```
smola, line 14: I'm not sure what the effect of this is in the build. Will this still produce bytecode compatible with JDK 9?

smola (approved):

internal-api/internal-api-9/build.gradle
manuel-alvarez-alvarez: Actually it produces JDK8 byte code as per:
```
[JavaCompile, GroovyCompile].each {
  tasks.withType(it).configureEach {
    setJavaVersion(it, 11)
    sourceCompatibility = JavaVersion.VERSION_1_8
    targetCompatibility = JavaVersion.VERSION_1_8
  }
}
```

Just tested it locally (to be sure):
```
public class datadog/trace/util/stacktrace/JDK9StackWalker extends datadog/trace/util/stacktrace/AbstractStackWalker {
     <ClassVersion=52>
     <SourceFile=JDK9StackWalker.java>
```

PR #6828, opened 2024-03-20
ensure queue timer is started for RunnableFuture types

richardstartin:
# What Does This Do

This fixes a regression in an unreleased profiling feature (queue time) which times how long is spent scheduling tasks which run in thread pools. This was inadvertently broken by #6446 but there wasn't a test in place to prevent this. The root cause was that FutureTask is excluded as a `RUNNABLE` so the timer was no longer started after being moved into `TPEHelper`, but the original instrumentation was using the wrong context field within FutureTask (the one for `Runnable`, not `RunnableFuture`) in the first place. The solution is to move the starting the queue timer back into the instrumentation, and (hopefully, temporarily) complicating the instrumentation slightly to handle `RunnableFuture`s whenever excluded as `RUNNABLE`. A test is also added to verify the fix and prevent regression. 

In the longer term, this instrumentation could be simplified by entirely removing the `RunnableInstrumentation` and only instrumenting or wrapping instances we fully know the state lifecycle of.

# Motivation

# Additional Notes

Jira ticket: [PROJ-IDENT]

### DIFF
--- a/dd-java-agent/agent-bootstrap/src/main/java/datadog/trace/bootstrap/instrumentation/java/concurrent/TPEHelper.java
+++ b/dd-java-agent/agent-bootstrap/src/main/java/datadog/trace/bootstrap/instrumentation/java/concurrent/TPEHelper.java
@@ -72,11 +72,9 @@ public final class TPEHelper {
     return Boolean.TRUE.equals(contextStore.get(executor));
   }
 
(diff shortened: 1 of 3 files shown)


mcculls (approved):
LGTM, thanks for the additional test

PR #6832, opened 2024-03-21
Add CPU and memory metrics to benchmarks

manuel-alvarez-alvarez:
# What Does This Do
Use PS to track CPU and RSS usage during the run of the load benchmarks

# Motivation
From IAST we've noticed some regressions that do not affect latency but impacting CPU usage on the target applications.


jbachorik:
Hi, can't we use rather our bundled profiler instead of pulling in async-profiler?
It is based on the async-profiler but already available in dd-java-agent.jar.

jbachorik:
👌 If you need a hand when setting up the profiler, just ping me.

PR #6833, opened 2024-03-21
Enable queue time tracing by default

richardstartin:
# What Does This Do

Enables a feature which records the amount of time spent queueing in various thread pool mechanisms. This feature is conservative and will only report waits longer than 50ms.

# Motivation

# Additional Notes

Jira ticket: [PROJ-IDENT]


dd-trace-api/src/main/java/datadog/trace/api/config/ProfilingConfig.java
```diff
@@ -187,13 +187,12 @@ public final class ProfilingConfig {
...
-  public static final String PROFILING_QUEUEING_TIME_ENABLED =
-      "profiling.experimental.queueing.time.enabled";
+  public static final String PROFILING_QUEUEING_TIME_ENABLED = "profiling.queueing.time.enabled";
 
-  public static final boolean PROFILING_QUEUEING_TIME_ENABLED_DEFAULT = false;
+  public static final boolean PROFILING_QUEUEING_TIME_ENABLED_DEFAULT = true;
 
   public static final String PROFILING_QUEUEING_TIME_THRESHOLD_MILLIS =
-      "profiling.experimental.queueing.time.threshold.millis";
+      "profiling.queueing.time.threshold.millis";
```
jbachorik, line 195: Just a note - this will have to have a release note just in case there are some customer actually using the experimental config.

jbachorik (approved):

PR #6834, opened 2024-03-22
Fix AppSec without IAST petclinic benchmark variant

manuel-alvarez-alvarez:
# What Does This Do
Changes the variant name for the petclinic benchmark in the scenario where AppSec starts with IAST completely disabled.


jandro996 (approved):
nice! ;)

PR #6841, opened 2024-03-25
Optionally prevent undertow from setting http.route

amarziali:
# What Does This Do

Undertow  is today setting `http.route` from `HttpServletMapping#getMappingMatch()`. There few issues with that:
* Other servlet/application server instrumentations are not doing this hence we have an incoherent behaviour
* It's only happening on undertow < 2.2. For more recent versions of undertow based on jakarta servlet, that's not happening
* `http.route` should be set by more specific frameworks (i.e. jax-rs, spring mvc) and should not contain a part of the relative uri that can lead to high cardinality resources

This PR allows preventing undertow instrumentation from setting  the`http.route` tag by explicitly setting:
* The env `DD_UNDERTOW_LEGACY_TRACING_ENABLED=false`
* The sysprop `-Ddd.undertow.legacy.tracing.enabled=false`

This cannot be done as an easy change this it will break the resource name.

# Motivation

We received customer complaints about having servlet URI in http.route

# Additional Notes

Jira ticket: [AIT-10012]

[AIT-10012]: https://datadoghq.atlassian.net/browse/AIT-10012?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ

### DIFF
--- a/dd-java-agent/instrumentation/undertow/src/main/java/datadog/trace/instrumentation/undertow/UndertowDecorator.java
+++ b/dd-java-agent/instrumentation/undertow/src/main/java/datadog/trace/instrumentation/undertow/UndertowDecorator.java
@@ -1,5 +1,6 @@
 package datadog.trace.instrumentation.undertow;
 
+import datadog.trace.api.Config;
 import datadog.trace.api.gateway.BlockResponseFunction;
 import datadog.trace.api.naming.SpanNaming;
 import datadog.trace.bootstrap.ContextStore;
@@ -33,6 +34,8 @@ public class UndertowDecorator
   public static final UndertowDecorator DECORATE = new UndertowDecorator();
   public static final CharSequence UNDERTOW_REQUEST =
       UTF8BytesString.create(DECORATE.operationName());
+  public static final boolean UNDERTOW_LEGACY_TRACING =
+      Config.get().isLegacyTracingEnabled(true, "undertow");
 
   @Override
   protected String[] instrumentationNames() {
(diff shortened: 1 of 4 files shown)


dd-trace-api/src/main/java/datadog/trace/api/config/TraceInstrumentationConfig.java
```diff
@@ -143,6 +143,8 @@ public final class TraceInstrumentationConfig {
   public static final String JAX_RS_ADDITIONAL_ANNOTATIONS = "trace.jax-rs.additional.annotations";
   /** If set, the instrumentation will set its resource name on the local root too. */
   public static final String AXIS_PROMOTE_RESOURCE_NAME = "trace.axis.promote.resource-name";
+  /** If explicitly set to true, undertow won't set the `http.route` tag. */
+  public static final String UNDERTOW_IGNORE_HTTP_ROUTE = "trace.undertow.ignore.http.route";
```
nayeem-kamal, line 147: This is not blocking this PR, but I think you could have used `config.isLegacyInstrumentationEnabled(true,"undertow")` instead of creating a separate config variable with it's own unique name.

nayeem-kamal (approved):
Left a small comment but this LGTM

dd-trace-api/src/main/java/datadog/trace/api/config/TraceInstrumentationConfig.java
amarziali: It's a great suggestion. thanks

PR #6848, opened 2024-03-26
refactor isLegacyTracingEnabled usages

amarziali:
# What Does This Do

The legacy tracing option was coupled to the fact that the current schema version allows or not inferred services (to avoid using 'fake' services like kafka, sqs, jms).

While this is working for all the instrumentation using this setting to mutate the service name or not, it is not working when the an instrumentation is using a legacy tracing option to put a breaking change under a feature flag (i.e. httpasyncclient4 for url parsing).

In order to fix it I decoupled the two conditions and also factorized a bit the commonly used legacy tracing options (for sqs, jms, kafka, aws-sdk) to avoid code duplication

# Motivation

# Additional Notes

Jira ticket: [PROJ-IDENT]

### DIFF
--- a/dd-java-agent/instrumentation/aws-java-sqs-1.0/src/test/groovy/SqsClientTest.groovy
+++ b/dd-java-agent/instrumentation/aws-java-sqs-1.0/src/test/groovy/SqsClientTest.groovy
@@ -388,7 +388,7 @@ abstract class SqsClientTest extends VersionedNamingTestBase {
       }
       trace(1) {
         span {
-          serviceName SpanNaming.instance().namingSchema().messaging().inboundService("jms", Config.get().isSqsLegacyTracingEnabled()) ?: Config.get().getServiceName()
+          serviceName SpanNaming.instance().namingSchema().messaging().inboundService("jms", Config.get().isJmsLegacyTracingEnabled()) ?: Config.get().getServiceName()
           operationName SpanNaming.instance().namingSchema().messaging().inboundOperation("jms")
           resourceName "Consumed from Queue somequeue"
           spanType DDSpanTypes.MESSAGE_CONSUMER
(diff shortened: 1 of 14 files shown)


dd-java-agent/instrumentation/aws-java-sqs-1.0/src/test/groovy/SqsClientTest.groovy
```diff
@@ -388,7 +388,7 @@ abstract class SqsClientTest extends VersionedNamingTestBase {
       }
       trace(1) {
         span {
-          serviceName SpanNaming.instance().namingSchema().messaging().inboundService("jms", Config.get().isLegacyTracingEnabled(true, "jms")) ?: Config.get().getServiceName()
+          serviceName SpanNaming.instance().namingSchema().messaging().inboundService("jms", Config.get().isSqsLegacyTracingEnabled()) ?: Config.get().getServiceName()
```
nayeem-kamal, line 391: looks like this changed from checking `isLegacyTracingEnabled(true, "jms")` to `isLegacyTracingEnabled(true, "sqs")`. Not sure if this was intentional but it is a change in the logic.

nayeem-kamal (commented):
Noticed a small change in logic that may be unintentional. Otherwise LGTM.

dd-java-agent/instrumentation/aws-java-sqs-1.0/src/test/groovy/SqsClientTest.groovy
amarziali: it is not intentional. It's a typo (and will be fail the test) :) thanks for the catch

nayeem-kamal (approved):
LGTM

PR #6859, opened 2024-03-29
[DSM] - Set DSM checkpoints for S3 put / get operations

kr-igor:
# What Does This Do
* Adding DSM checkpoints for S3 `put/get` operations
* Adding request/response content length to span tags (only if DSM is enabled)
* Introduces concept of datasources for DSM checkpoints inline with OpenLineage spec (ds.name / ds.namespace). 
* Changes the way DSM stats are aggregated. Instead of hash-based aggregation stats are now aggregated per data source (datasource hash is generated using checkpoint hash as a base).

# Motivation
Setting "data source" checkpoints unlocks end-to-end pipeline lineage. For instance, this  particular change allows visualizing data flow between streaming and batch processing. The same mechanism can be used to support more types of data sources. 

# Additional Notes
[Jira ticket](https://datadoghq.atlassian.net/browse/DSMON-282)


dd-java-agent/instrumentation/aws-java-sdk-2.2/src/main/java/datadog/trace/instrumentation/aws/v2/AwsSdkClientDecorator.java
```diff
@@ -335,6 +350,51 @@ public AgentSpan onSdkResponse(
...
+        long responseSize = getResponseContentLength(httpResponse);
+        span.setTag(Tags.HTTP_RESPONSE_CONTENT_LENGTH, responseSize);
+
+        Object key = span.getTag(InstrumentationTags.AWS_OBJECT_KEY);
+        Object bucket = span.getTag(InstrumentationTags.AWS_BUCKET_NAME);
+        Object awsOperation = span.getTag(InstrumentationTags.AWS_OPERATION);
+
+        if (key != null && bucket != null && awsOperation != null) {
+          if ("GetObject".equalsIgnoreCase(awsOperation.toString())) {
+            LinkedHashMap<String, String> sortedTags = new LinkedHashMap<>();
```
kr-igor, line 364: LinkedHashMap can be replaced with a fixed size array. This will be done in a followup PR in Q2.

dd-java-agent/instrumentation/aws-java-sdk-1.11.0/src/main/java/datadog/trace/instrumentation/aws/v0/AwsSdkClientDecorator.java
```diff
@@ -88,6 +92,9 @@ public AgentSpan onRequest(final AgentSpan span, final Request request) {
     CharSequence awsRequestName = AwsNameCache.getQualifiedName(request);
 
     span.setResourceName(awsRequestName, RPC_COMMAND_NAME);
+    if ("Amazon S3".equalsIgnoreCase(awsServiceName) && span.traceConfig().isDataStreamsEnabled()) {
```
dougqh, line 95: Basing this on service name seems brittle to me, it would make more sense to base this check on the raw information -- e.g. the type of Request.

dd-java-agent/instrumentation/aws-java-sdk-1.11.0/src/main/java/datadog/trace/instrumentation/aws/v0/AwsSdkClientDecorator.java
```diff
@@ -233,12 +240,67 @@ public AgentSpan onRequest(final AgentSpan span, final Request request) {
...
 
+  public AgentSpan onServiceResponse(
+      final AgentSpan span, final String awsService, final Response response) {
+    if ("Amazon S3".equalsIgnoreCase(awsService) && span.traceConfig().isDataStreamsEnabled()) {
+      long responseSize = getResponseContentLength(response);
+      span.setTag(Tags.HTTP_RESPONSE_CONTENT_LENGTH, responseSize);
+
+      Object key = span.getTag(InstrumentationTags.AWS_OBJECT_KEY);
+      Object bucket = span.getTag(InstrumentationTags.AWS_BUCKET_NAME);
+      Object awsOperation = span.getTag(InstrumentationTags.AWS_OPERATION);
```
dougqh, line 251: There are a lot or repetitive toString calls here.
If these are Strings or CharSequences, I would type check and cast them sooner.

Or, just create variables to hold the String values immediately after calling getTag.
Something like the line below would be fine...
String keyStr = (key == nul) ? null : key.toString()

dd-java-agent/instrumentation/aws-java-sdk-1.11.0/src/main/java/datadog/trace/instrumentation/aws/v0/AwsSdkClientDecorator.java
```diff
@@ -233,12 +240,67 @@ public AgentSpan onRequest(final AgentSpan span, final Request request) {
...
+
+        if ("PutObjectRequest".equalsIgnoreCase(awsOperation.toString())
+            || "UploadPartRequest".equalsIgnoreCase(awsOperation.toString())) {
+          Object requestSize = span.getTag(Tags.HTTP_REQUEST_CONTENT_LENGTH);
+          long payloadSize = 0;
+          if (requestSize != null) {
+            payloadSize = (long) requestSize;
+          }
+
+          LinkedHashMap<String, String> sortedTags = new LinkedHashMap<>();
```
dougqh, line 279: I would like to see DSM replace the use of LinkedHashMap with a simple struct-like object or builder.
LinkedHashMap is creating a lot of unnecessary allocation.

dd-java-agent/instrumentation/aws-java-sdk-2.2/src/main/java/datadog/trace/instrumentation/aws/v2/AwsSdkClientDecorator.java
```diff
@@ -106,13 +108,23 @@ public CharSequence spanName(final ExecutionAttributes attributes) {
...
+      final SdkRequest request,
+      final SdkHttpRequest httpRequest,
+      final ExecutionAttributes attributes) {
     final String awsServiceName = attributes.getAttribute(SdkExecutionAttribute.SERVICE_NAME);
     final String awsOperationName = attributes.getAttribute(SdkExecutionAttribute.OPERATION_NAME);
     onOperation(span, awsServiceName, awsOperationName);
 
     // S3
     request.getValueForField("Bucket", String.class).ifPresent(name -> setBucketName(span, name));
+    if ("s3".equalsIgnoreCase(awsServiceName) && span.traceConfig().isDataStreamsEnabled()) {
```
dougqh, line 121: Why is service name different here?

dd-java-agent/instrumentation/aws-java-sdk-2.2/src/main/java/datadog/trace/instrumentation/aws/v2/AwsSdkClientDecorator.java
dougqh: I think a simple struct-like object would be better than an array.

dd-trace-core/src/main/java/datadog/trace/core/datastreams/DefaultPathwayContext.java
```diff
@@ -388,6 +398,25 @@ private static DefaultPathwayContext decode(
         hash);
   }
 
+  static class DataSetHashBuilder {
+    private final StringBuilder builder;
+
+    public DataSetHashBuilder() {
+      builder = new StringBuilder();
```
dougqh, line 405: Builder should be given an initial size if possible.

dd-trace-core/src/main/java/datadog/trace/core/datastreams/DefaultPathwayContext.java
```diff
@@ -388,6 +398,25 @@ private static DefaultPathwayContext decode(
...
+
+    public void addTag(String tag) {
+      if (dataSetTagKeys.contains(tag)) {
+        builder.append(tag);
+      }
+    }
+
+    public long generateDataSourceHash(long parentHash) {
+      builder.append(parentHash);
+      return FNV64Hash.generateHash(builder.toString(), FNV64Hash.Version.v1);
```
dougqh, line 416: It seems like the builder & string aren't actually needed for the hash calculation.
I presume the FNV64Hash computation could be done in a streaming fashion as the tags are added.

dougqh (approved):
I think there's some opportunity for clean-up and optimization.

dd-java-agent/instrumentation/aws-java-sdk-2.2/src/main/java/datadog/trace/instrumentation/aws/v2/AwsSdkClientDecorator.java
kr-igor: SDK v1 and V2 have different naming for services. I'll check if this can be unified.

dd-trace-core/src/main/java/datadog/trace/core/datastreams/DefaultPathwayContext.java
kr-igor: Makes sense, I'll update the code.

dd-java-agent/instrumentation/aws-java-sdk-1.11.0/src/main/java/datadog/trace/instrumentation/aws/v0/AwsSdkClientDecorator.java
kr-igor: I've switched to "s3" service name for both versions of AWS SDK. I didn't use the request type to avoid adding new dependencies.

dd-java-agent/instrumentation/aws-java-sdk-2.2/src/main/java/datadog/trace/instrumentation/aws/v2/AwsSdkClientDecorator.java
kr-igor: Both implementations now use 's3'

PR #6865, opened 2024-04-01
Instrument Google protobuf

piochelepiotr:
# What Does This Do

Add protobuf instrumentation for serialize / deserialize operations.
If data streams monitoring is enabled, capture schemas of messages.

# Motivation

# Additional Notes

[It impacts the GRPC instrumentation](https://ddstaging.datadoghq.com/apm/traces?query=%40_top_level%3A1%20env%3Apiotr&cols=core_service%2Ccore_resource_name%2Clog_duration%2Clog_http.method%2Clog_http.status_code&fromUser=false&graphType=flamegraph&historicalData=false&messageDisplay=inline&query_translation_version=v0&shouldShowLegend=true&sort=time&spanType=service-entry&traceQuery=&view=spans&start=1711130875088&end=1711131775088&paused=false)

since it adds spans for serialize / deserialize

Jira ticket: [PROJ-IDENT]


dd-java-agent/instrumentation/protobuf-3.0.0/src/main/java/datadog/trace/instrumentation/protobuf_java/AbstractParserInstrumentation.java
```diff
@@ -0,0 +1,85 @@
...
+        @Advice.Enter final AgentScope scope,
+        @Advice.Thrown final Throwable throwable,
+        @Advice.Return MessageLite message) {
+      if (scope == null) {
+        return;
+      }
+      AgentSpan span = scope.span();
+      DESERIALIZER_DECORATOR.onError(span, throwable);
+      if (message instanceof AbstractMessage) {
+        DESERIALIZER_DECORATOR.attachSchemaOnSpan((AbstractMessage) message, span);
```
dougqh, line 79: This seems like work that would probably be better done in a background thread.
There is a mechanism for doing that, but I think it could use some improvement.

For now, I'm fine with this as is - as long as it is not on by default.

dd-java-agent/instrumentation/protobuf-3.0.0/src/main/java/datadog/trace/instrumentation/protobuf_java/Decorator.java
```diff
@@ -0,0 +1,75 @@
...
+  }
+
+  public void attachSchemaOnSpan(Descriptors.Descriptor descriptor, AgentSpan span) {
+    if (descriptor == null) {
+      return;
+    }
+    span.setTag(DDTags.SCHEMA_TYPE, PROTOBUF);
+    span.setTag(DDTags.SCHEMA_NAME, descriptor.getName());
+    span.setTag(DDTags.SCHEMA_OPERATION, operation);
+    Integer prio = span.forceSamplingDecision();
```
dougqh, line 59: There probably needs to a product specific code for why the decision was forced.
You should check with APM ingestion.

dd-java-agent/instrumentation/protobuf-3.0.0/src/main/java/datadog/trace/instrumentation/protobuf_java/OpenAPIFormatExtractor.java
```diff
@@ -0,0 +1,134 @@
...
+  private static final int TYPE_SINT64 = 18;
+
+  private static String wrapInArray(String serializedType, boolean isArray) {
+    return isArray ? "{\"type\": \"array\", \"items\":" + serializedType + "}" : serializedType;
+  }
+
+  public static String extractSchema(Descriptor descriptor) {
+    List<String> schemas = new ArrayList<>();
+    extractDescriptorSchema(descriptor, schemas);
+    return "{"
```
dougqh, line 37: This is not properly escaped

dd-java-agent/instrumentation/protobuf-3.0.0/src/main/java/datadog/trace/instrumentation/protobuf_java/OpenAPIFormatExtractor.java
```diff
@@ -0,0 +1,134 @@
...
+        + "\"openapi\": \"3.0.0\","
+        + "\"components\": {\"schemas\": {"
+        + String.join(", ", schemas)
+        + "}}}";
+  }
+
+  private static void extractDescriptorSchema(Descriptor descriptor, List<String> schemas) {
+    StringBuilder schema = new StringBuilder();
```
dougqh, line 45: This is not properly escaped - you should probably use a JSON API instead

piochelepiotr: Do you think it's ok to add the dependency to JSON API?
It would make the code easier indeed. But add a dependency

dougqh (changes requested):
I have concerns that JSON is not done using JSON API and is not properly escaped.
I also have concerns with the altering of the sampling decision.
When altering the sampling decision, I believe we should be adding information about which product made that decision.  We should check that with APM ingestion.

dd-java-agent/instrumentation/protobuf-3.0.0/src/main/java/datadog/trace/instrumentation/protobuf_java/Decorator.java
dumontg: Indeed, if a customer pays for the ingestion of a span, we want to be able to tell why it was ingested (it’s also pretty useful for troubleshooting). The tracer set a specific attribute called `_dd.p.dm` depending on the sampling mechanism ([see all potential values that exist](https://datadoghq.atlassian.net/wiki/spaces/APM/pages/2564915820/Trace+Ingestion+Mechanisms#Propagated-tags%3A-_dd.p.dm-(sampling-mechanism-aka-decision-maker))).

Questions:
* When you override the sampling decision, do you just want the span or the whole trace?
* Do you expect a lot of spans coming in? Once per tracer process maybe?

dd-java-agent/instrumentation/protobuf-3.0.0/src/main/java/datadog/trace/instrumentation/protobuf_java/Decorator.java
```diff
@@ -0,0 +1,76 @@
...
+    if (prio == null || prio <= 0) {
+      // don't extract schema if span is not sampled
+      return;
+    }
+    int weight = AgentTracer.get().getDataStreamsMonitoring().shouldSampleSchema(operation);
+    if (weight == 0) {
+      return;
+    }
+    String schema = SchemaExtractor.extractSchemas(descriptor);
+    System.out.println("schema is: " + schema);
```
dumontg, line 69: Unsure you wanted to leave this

dd-java-agent/instrumentation/protobuf-3.0.0/src/main/java/datadog/trace/instrumentation/protobuf_java/Decorator.java
piochelepiotr: I checked a bit more the code, it doesn’t look like forceSamplingPriority is going to increase sampling rate, is that correct?
It looks like it’s going to force the sampling decision to be taken earlier.
What I can do, is reverse the two samplings.
Today, the code extracts only 1 schema every 30 seconds, so I can do that sampling first. That way, the change would only affect the sampling decision of 1 trace every 30 seconds, what do you think?

dd-java-agent/instrumentation/protobuf-3.0.0/src/main/java/datadog/trace/instrumentation/protobuf_java/SchemaExtractor.java
```diff
@@ -0,0 +1,137 @@
...
+      case TYPE_GROUP:
+        // Groups are deprecated and usually represented as nested messages in OpenAPI
+        type = "object";
+        description = "Group type";
+        break;
+      case TYPE_MESSAGE:
+        ref = "#/components/schemas/" + field.getMessageType().getName();
+        // Recursively add nested message schemas
+        extractSchema(field.getMessageType(), builder);
+        break;
```
manuel-alvarez-alvarez, line 90: This code will end up in a `StackOverflowError` in case there are cycles in the schema, see the example:

```
message DummyMessage {
    int32 id = 1;
    string name = 2;
    DummyMessage nested = 3;
}
```

A cache of already processed types will help here and the `SchemaBuilder` should be able to handle cycles. Optionally, we might want to include a limit in the recursion depth and also number of elements visited to deal with problematic/big schemas.

piochelepiotr: Good catch! Will update the code

dd-java-agent/instrumentation/protobuf-3.0.0/src/main/java/datadog/trace/instrumentation/protobuf_java/Decorator.java
piochelepiotr: I switched the two sampling calls.
Now the forceSamplingDecision is only rarely called (a few times every 30 seconds, 1 time in the best case depending on race conditions).
Also, forceSamplingDecision doesn't force the trace to be sampled. It forces the sampling decision to be made.
So this should not increase sampling rate.

dd-java-agent/instrumentation/protobuf-3.0.0/src/main/java/datadog/trace/instrumentation/protobuf_java/SchemaExtractor.java
piochelepiotr: I updated the code & added limits. Thanks for the comment!

dd-java-agent/instrumentation/protobuf-3.0.0/src/main/java/datadog/trace/instrumentation/protobuf_java/Decorator.java
dumontg: Ah that makes sense, I had not understood that initially. If you don't influence sampling I don't think we need to do anything at intake. For accomplishing what you want I am not knowledgeable enough with the tracer itself.

dd-java-agent/instrumentation/protobuf-3.0.0/build.gradle
```diff
@@ -0,0 +1,15 @@
+muzzle {
+  pass {
+    group = "com.google.protobuf"
+    module = "protobuf-java"
+    versions = "[3.0.0,)"
+    assertInverse = false
```
amarziali, line 6: any particular reason not to test that the instrumentation is disabled on versions < 3?

dd-smoke-tests/datastreams/kafkaschemaregistry/src/main/java/datadog/smoketest/datastreamskafka/KafkaProducerWithSchemaRegistry.java
```diff
@@ -37,49 +40,23 @@ public static void main(String[] args) {
...
-
-    // Avro record
-    Schema schema = new Schema.Parser().parse(avroSchema);
-
-    Producer<String, GenericRecord> producer =
+    Producer<String, MyMessage> producer =
         new org.apache.kafka.clients.producer.KafkaProducer<>(properties);
 
+    Duration duration = Duration.newBuilder().setSeconds(10).build();
+    System.out.println("duration is " + duration.getSeconds());
```
amarziali, line 47: a logger should be used to print out this

internal-api/src/main/java/datadog/trace/bootstrap/instrumentation/api/InternalSpanTypes.java
```diff
@@ -44,4 +44,6 @@ public class InternalSpanTypes {
       UTF8BytesString.create(DDSpanTypes.TEST_SESSION_END);
   public static final UTF8BytesString VULNERABILITY =
       UTF8BytesString.create(DDSpanTypes.VULNERABILITY);
+  public static final UTF8BytesString SERIALIZE = UTF8BytesString.create(DDSpanTypes.SERIALIZE);
```
amarziali, line 47: Perhaps I would have been used `protobuf` as span type since serialize/deserialize are rather operations

dd-java-agent/agent-tooling/build.gradle
```diff
@@ -38,6 +38,7 @@ compileJava.dependsOn 'generateClassNameTries'
 packageSources.dependsOn 'generateClassNameTries'
 sourcesJar.dependsOn 'generateClassNameTries'
 
+
```
amarziali, line 41: nit: remove this file from the changeset

dd-java-agent/instrumentation/protobuf-3.0.0/src/main/java/datadog/trace/instrumentation/protobuf_java/AbstractParserInstrumentation.java
```diff
@@ -0,0 +1,84 @@
...
+import net.bytebuddy.description.type.TypeDescription;
+import net.bytebuddy.matcher.ElementMatcher;
+
+@AutoService(InstrumenterModule.class)
+public final class AbstractParserInstrumentation extends InstrumenterModule.Tracing
+    implements Instrumenter.ForTypeHierarchy {
+
+  static final String instrumentationName = "protobuf";
+  static final String TARGET_TYPE = "com.google.protobuf.AbstractParser";
+  static final String DESERIALIZE = "deserialize";
```
amarziali, line 29: I have same naming concern than "serialize"

dd-java-agent/instrumentation/protobuf-3.0.0/src/main/java/datadog/trace/instrumentation/protobuf_java/Decorator.java
```diff
@@ -0,0 +1,79 @@
...
+    Integer prio = span.forceSamplingDecision();
+    if (prio == null || prio <= 0) {
+      // don't extract schema if span is not sampled
+      return;
+    }
+    int weight = AgentTracer.get().getDataStreamsMonitoring().trySampleSchema(operation);
+    if (weight == 0) {
+      return;
+    }
+    String schema = SchemaExtractor.extractSchemas(descriptor);
```
amarziali, line 72: This seems to me a quite expensive operation. Can we cache this? The same for the schemaId being a hash depending on the schema name

dd-java-agent/instrumentation/protobuf-3.0.0/src/main/java/datadog/trace/instrumentation/protobuf_java/AbstractMessageInstrumentation.java
```diff
@@ -0,0 +1,79 @@
...
+import net.bytebuddy.description.type.TypeDescription;
+import net.bytebuddy.matcher.ElementMatcher;
+
+@AutoService(InstrumenterModule.class)
+public final class AbstractMessageInstrumentation extends InstrumenterModule.Tracing
+    implements Instrumenter.ForTypeHierarchy {
+
+  static final String instrumentationName = "protobuf";
+  static final String TARGET_TYPE = "com.google.protobuf.AbstractMessage";
+  static final String SERIALIZE = "serialize";
```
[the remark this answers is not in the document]
piochelepiotr: Updated the name, for some reason I can't explain, it doesn't work if I switch to Utf8ByteString 🤷

dd-java-agent/instrumentation/protobuf-3.0.0/src/main/java/datadog/trace/instrumentation/protobuf_java/Decorator.java
piochelepiotr: I cached based on the schema name, but I'm worried the schema changes, and the schema name doesn't 🤔 
Do you have a better idea on how to cache?

dd-java-agent/instrumentation/protobuf-3.0.0/src/main/java/datadog/trace/instrumentation/protobuf_java/AbstractMessageInstrumentation.java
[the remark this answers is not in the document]
amarziali: If you have tests failing is probably because they are comparing String with CharSequence that's not the same thing.
So 
```
String s = "test"
CharSequence cs = UTF8ByteString.create("test")

s.equals(cs) -> false
s.contentEquals(cs) -> true
s.equals(cs.toString()) -> true
```

dd-java-agent/instrumentation/protobuf-3.0.0/src/main/java/datadog/trace/instrumentation/protobuf_java/Decorator.java
amarziali: Can a schema be dynamically changed at runtime and what is the use case?

dd-java-agent/instrumentation/protobuf-3.0.0/src/main/java/datadog/trace/instrumentation/protobuf_java/AbstractMessageInstrumentation.java
```diff
@@ -0,0 +1,79 @@
...
+import datadog.trace.agent.tooling.Instrumenter;
+import datadog.trace.agent.tooling.InstrumenterModule;
+import datadog.trace.bootstrap.instrumentation.api.AgentScope;
+import datadog.trace.bootstrap.instrumentation.api.AgentSpan;
+import net.bytebuddy.asm.Advice;
+import net.bytebuddy.description.type.TypeDescription;
+import net.bytebuddy.matcher.ElementMatcher;
+
+@AutoService(InstrumenterModule.class)
+public final class AbstractMessageInstrumentation extends InstrumenterModule.Tracing
```
amarziali: Concerning the instrumentation, do we want to systematically trace all the ser/de operations or having this couple of instrumentations enabled by default only when datastreams is enabled? I'm wondering if we're adding too many details for the non dsm users

dougqh (approved):

PR #6883, opened 2024-04-03
Fix inaccessible fields capture

jpbempel:
# What Does This Do
To fix this we need to catch eventual exception thrown by accessing field values. For that we continue to try to access directly to field when possible because it will be faster, but when we detect fields are not accessible we use another method that try to access it through reflection but could catch exception if not and returns a CapturedValue instance filled with the reason of this exception.

# Motivation
NoSuchFieldException can be thrown if we try to capture, through reflection, (static) fields from JDK classes or from another module. 

# Additional Notes

Jira ticket: [DEBUG-2270]

[DEBUG-2270]: https://datadoghq.atlassian.net/browse/DEBUG-2270?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ

### DIFF
--- a/dd-java-agent/agent-debugger/src/main/java/com/datadog/debugger/instrumentation/CapturedContextInstrumentor.java
+++ b/dd-java-agent/agent-debugger/src/main/java/com/datadog/debugger/instrumentation/CapturedContextInstrumentor.java
@@ -1,6 +1,5 @@
 package com.datadog.debugger.instrumentation;
 
-import static com.datadog.debugger.instrumentation.ASMHelper.emitReflectiveCall;
 import static com.datadog.debugger.instrumentation.ASMHelper.getStatic;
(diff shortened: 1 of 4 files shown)


dd-java-agent/agent-debugger/src/main/java/com/datadog/debugger/instrumentation/CapturedContextInstrumentor.java
```diff
@@ -793,20 +793,30 @@ private void collectStaticFields(InsnList insnList) {
       // stack: [capturedcontext, capturedcontext, array, array]
```
evanchooly, line 793: i like the tracking of the stack values.  :)

evanchooly (approved):
this looks good to me. a rather straightforward solution.

PR #6885, opened 2024-04-04
Detect a vulnerability when a default application is deployed

jandro996:
# What Does This Do

Add new default deployed vulnerability 
Give support for Tomcat and Jetty default applications checking display-name tags into web.xml application file

# Motivation
The applications supplied by default with application servers are mostly not intended to be deployed in secure production environments as they may be vulnerable, or even if they are not, their discovery by an attacker could encourage them to seek security flaws in that service. 

# Additional Notes

Jira ticket: [PROJ-IDENT]


dd-java-agent/agent-iast/src/main/java/com/datadog/iast/sink/ApplicationModuleImpl.java
```diff
@@ -212,6 +240,27 @@ private void checkWebXmlVulnerabilities(@Nonnull final Path path, final AgentSpa
         case TOMCAT_HOST_MANAGER_APP_PATTERN:
           reportAdminConsoleActive(span, TOMCAT_HOST_MANAGER_APP);
           break;
+        case TOMCAT_SAMPLES_APP_PATTERN:
```
manuel-alvarez-alvarez, line 243: Since this is growing quite a bit, wouldn't it make more sense to look for `<display-name>` and then use a list of default apps to match?

jandro996: Potentially It could grow even more, so I will go for your approach

manuel-alvarez-alvarez (approved):

PR #6893, opened 2024-04-08
Add Onboarding test on gitlab pipelines

robertomonteromiguel:
# What Does This Do

Add the onboarding tests on gitlab pipeline. These tests will be executed after deb and rpm packages generation.
# Motivation
System-tests nightly build pipeline is launching the onboarding tests agains latest released packages.
We need to launch the onboarding tests for current branch generated packages.
These tests are adding for auto-inject and dd-trace-rb repositories. We are trying to expand to all dd-trace-lang repos.
# Additional Notes

Jira ticket: [PROJ-IDENT]

### DIFF
--- a/.gitlab-ci.yml
+++ b/.gitlab-ci.yml
@@ -4,6 +4,7 @@ include:
   - local: ".gitlab/benchmarks.yml"
   - local: ".gitlab/exploration-tests.yml"
   - local: ".gitlab/ci-visibility-tests.yml"
+  - local: ".gitlab/single-step-instrumentation-tests.yml"
 
 stages:
   - build
@@ -12,6 +13,7 @@ stages:
   - benchmarks
   - exploration-tests
   - ci-visibility-tests
+  - onboarding-tests
   - generate-signing-key
 
 variables:
--- a/.gitlab-ci.yml
+++ b/.gitlab-ci.yml
@@ -13,7 +13,7 @@ stages:
   - benchmarks
   - exploration-tests
   - ci-visibility-tests
-  - onboarding-tests
+  - single-step-instrumentation-tests
   - generate-signing-key
 
 variables:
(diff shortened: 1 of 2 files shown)


PerfectSlayer:
Hey @robertomonteromiguel 
I had a quick look at the changes but I struggle to find when the tests will be run? Is it for all commits to any branch?

robertomonteromiguel:
> Hey @robertomonteromiguel I had a quick look at the changes but I struggle to find when the tests will be run? Is it for any commit to any branch?

@PerfectSlayer The tests will be launched whenever deb and rpm packages are generated. 
These packages are being generated for the PRs as well.
I agree, that it can be hard to run the onboarding tests for which PR. I can make them run only on master commits. What do you think?

PerfectSlayer:
> I agree, that it can be hard to run the onboarding tests for which PR. I can make them run only on master commits. What do you think?

Are onboarding tests candidate to fail due to a tracer / instrumentation change? Or even some other product like profiling/ASM/etc...? Or is it more to ensure that the whole packaging things are still working as expected?

Because if such tests are not expected to break on feature update, I might be best to run them on `master` only.
If anything goes wrong, we can check which PR merge introduce the issue and call out the contributor (and revert the PR if it’s blocking a release).

robertomonteromiguel:
> > I agree, that it can be hard to run the onboarding tests for which PR. I can make them run only on master commits. What do you think?
> 
> Are onboarding tests candidate to fail due to a tracer / instrumentation change? Or even some other product like profiling/ASM/etc...? Or is it more to ensure that the whole packaging things are still working as expected?
> 
> Because if such tests are not expected to break on feature update, I might be best to run them on `master` only. If anything goes wrong, we can check which PR merge introduce the issue and call out the contributor (and revert the PR if it’s blocking a release).

I added the condition, to run only for master branch pushes

PerfectSlayer (approved):
Thanks for the change. Looks good to me 👍

Issue #6903, opened 2024-04-12
RabbitMQ instrumentation tests is flaky 


PerfectSlayer:
Test suite:  `test rabbit publish/get with given disabled queue (consumer side) [type: deliver, exchangeName: some-exchange, routingKey: some-routing-key, queueName: queueNameTest, config: some-exchange, noParent: false, #4](test rabbit publish/get with given disabled queue`

Error:
```
Target of 'verifyAll' block must not be null
	at spock.lang.Specification.verifyAll(Specification.java:272)
	at RabbitMQTestBase.test rabbit publish/get with given disabled queue (consumer side)(RabbitMQTest.groovy:683)
```

Cause:
The second `StatGroup` could not be find and is `null`: 
https://github.com/DataDog/dd-trace-java/blob/16b8e15674a6d1d481a03f43c2362b310801692d/dd-java-agent/instrumentation/rabbitmq-amqp-2.7/src/test/groovy/RabbitMQTest.groovy#L682

PR #6907, opened 2024-04-12
Add some retries around flaky RabbitMQ DSM test

vandonr:
closes https://github.com/DataDog/dd-trace-java/issues/6903

What I suppose is happening is that the DSM data can come in 3 groups, and since we only wait for 2, we can run the asserts before everything is there. 
https://github.com/DataDog/dd-trace-java/blob/bd114faddf0fa79e10cac21769ba7121b8e761de/dd-java-agent/instrumentation/rabbitmq-amqp-2.7/src/test/groovy/RabbitMQTest.groovy#L638
Adding retries around it allows more time for the data to arrive.


PerfectSlayer (approved):
Thanks for the quick fix! 💪

PR #6916, opened 2024-04-15
WIP. will need clean up.

evanchooly:
# What Does This Do

# Motivation

# Additional Notes

Jira ticket: [PROJ-IDENT]


dd-java-agent/instrumentation/span-origin/src/main/java/datadog/trace/instrumentation/span_origin/EntrySpanOriginAdvice.java
```diff
@@ -0,0 +1,51 @@
...
+    StackTraceElement[] stackTrace =
+        new Exception("\"EntrySpanOriginAdvice.onEnter\" trace").getStackTrace();
+    AgentSpan span = AgentTracer.get().activeScope().span();
+    StackTraceElement stackTraceElement = stackTrace[0];
+
+    span.setTag(DDTags.DD_ENTRY_LOCATION_FILE, stackTraceElement.getClassName());
+    span.setTag(DDTags.DD_ENTRY_METHOD, stackTraceElement.getMethodName());
+    span.setTag(DDTags.DD_ENTRY_START_LINE, stackTraceElement.getLineNumber());
+
+    List<String> params = new ArrayList();
```
dougqh, line 29: You could initial size based on parameterTypes.  Or better yet, just forego the ArrayList and use a StringBuilder to build up the method signature.

If we're doing this often, it might even be worth caching.  There's probably already a Util for that for Method.

evanchooly: actually memoization is on the list of clean ups to do once the basic functionality is in place.

PR #6921, opened 2024-04-16
Fix call-site-plugin build script to avoid rebuilding plugin every time

PerfectSlayer:
# What Does This Do

This PR fixes the CSI (Call Site Plugin) build script which is part of the project build script. 

There are two main fixes:

## Avoid rebuilding CSI plugin jar for every Gradle command

CSI plugin shadowed jar was built for every gradle command runned as part of the project configuration:
image

With the changes, all plugin gradle tasks are properly cached (`UP-TO-DATE` state):
image

## Avoid generating call site sources file for every build

CSI plugin drops generated call site source files (from `build/generated/sources/csi/`) for every build, breaking the task cache and making the plugin needed to re-run.

Running `gradlew :dd-java-agent:shadowJar` multiple times shows CSI plugin tasks are not cached and are the slowest taks to run:
image

After the fix, all `generateCallSiteJava` generated tasks are properly cached:
image

The generated source files will still be cleared as part of the `clean` gradle task and will still be updated if the sources files (build/classes/raw) are updated.

## Other changes

* Upgrade deprecated API usage,
* Avoid cleaning every temporary files twice

# Motivation

This should help to run gradle project configuration step faster. 

# Additional Notes

Fixes and upgrades are split in multiple commits to help reviewing the PR.

Jira ticket: [PROJ-IDENT]

### DIFF
--- a/buildSrc/call-site-instrumentation-plugin/build.gradle.kts
+++ b/buildSrc/call-site-instrumentation-plugin/build.gradle.kts
@@ -4,7 +4,7 @@ plugins {
   java
   groovy
   id("com.diffplug.spotless") version "6.13.0"
-  id("com.github.johnrengelman.shadow") version "7.1.2"
+  id("com.github.johnrengelman.shadow") version "8.1.1"
 }
 
 java {
@@ -49,7 +49,7 @@ dependencies {
 sourceSets {
   test {
     java {
-      srcDirs("src/test/java", "$buildDir/generated/sources/csi")
+      srcDirs("src/test/java", "${layout.buildDirectory.get()}/generated/sources/csi")
     }
   }
 }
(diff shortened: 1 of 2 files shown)


smola (approved):

manuel-alvarez-alvarez:
One of the reasons why the plugin was dropping the generated source files is that under very particular circumstances (e.g. after certain incompatible changes in the call site definition), the generated sources ended up in a bad state that forced us to delete them by hand.

It's very possible that this is no longer an issue (e.g. gradle has been updated a few times) and since we don't update call sites so often nowadays I think that shouldn't be a problem any more.

Many thanks for the change @PerfectSlayer

manuel-alvarez-alvarez (approved):

PerfectSlayer:
> One of the reasons why the plugin was dropping the generated source files is that under very particular circumstances (e.g. after certain incompatible changes in the call site definition), the generated sources ended up in a bad state that forced us to delete them by hand.

I checked the following things:
- If you edit any file from the instrumentation source folder, it will generate CSI files,
- If you manually edited the generated sources, they are well overridden with the new fresh generated file,

The only issue I found is if you delete a source file related to CSI, you might have build issue.
But it is also required to clean the build folder to remove the related class file so it should not be an issue.

PR #6926, opened 2024-04-17
Do not abort CI Visibility spans dispatch on interrupt

nikita-tkachenko-datadog:
# What Does This Do
Updates utility method that executes (and retries, if necessary) an HTTP request.
The method is used exclusively by CI Visibility logic that sends spans to the backend.

# Motivation
When a child JVM that was forked for tests execution finishes its work, the build system shuts the process down.
As part of this shutdown the tracer threads receive interrupts.
It is possible that when these interrupts are received some tests spans are still being dispatched.
Acting on interrupt results in these spans being lost.

# Additional Notes
The logic is updated in a way that an interrupted request will be reattempted.
If an interrupt is received at any point during request execution, the thread's interrupted flag is restored on method exit.

Jira ticket: [CIVIS-9801]

[CIVIS-9801]: https://datadoghq.atlassian.net/browse/CIVIS-9801?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ

### DIFF
--- a/communication/src/main/java/datadog/communication/http/HttpRetryPolicy.java
+++ b/communication/src/main/java/datadog/communication/http/HttpRetryPolicy.java
@@ -1,5 +1,8 @@
 package datadog.communication.http;
 
+import java.io.IOException;
+import java.io.InterruptedIOException;
+import java.net.ConnectException;
 import java.util.concurrent.ThreadLocalRandom;
 import java.util.concurrent.TimeUnit;
 import javax.annotation.Nullable;
@@ -34,7 +37,7 @@ import org.slf4j.LoggerFactory;
  * instance.
  */
 @NotThreadSafe
-public class HttpRetryPolicy {
+public class HttpRetryPolicy implements AutoCloseable {
 
   private static final Logger log = LoggerFactory.getLogger(HttpRetryPolicy.class);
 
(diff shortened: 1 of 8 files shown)


romainkomorndatadog (approved):

PerfectSlayer:
Even if the `sendWithRetries` method is only use for CI visibility for now, I found it dangerous to be the default behavior.

What about evolving the retry policy object to handle the exception?
Rather than calling `shouldRetry(null)` on error, we can have a dedicated method `boolean shouldRetryWhen(Throwable)` where your retry policy can safely ignore interrupted exception (the boolean result can define if the error must be rethrown).

It should allow to have a more generic approach for this kind of behavior. What do you think about it?

nikita-tkachenko-datadog:
> Even if the `sendWithRetries` method is only use for CI visibility for now, I found it dangerous to be the default behavior.
> 
> What about evolving the retry policy object to handle the exception? Rather than calling `shouldRetry(null)` on error, we can have a dedicated method `boolean shouldRetryWhen(Throwable)` where your retry policy can safely ignore interrupted exception (the boolean result can define if the error must be rethrown).
> 
> It should allow to have a more generic approach for this kind of behavior. What do you think about it?

Yes, that makes a lot of sense. I had to change a couple of things to encapsulate interrupts suppression inside the policy

PR #6931, opened 2024-04-22
Update buildscan plugin to latest version

randomanderson:
# What Does This Do

Updates the build scan plugin to the latest artifact coodinates.

# Motivation
The previous version, `3.0`, was 17 minor version old. It failed constantly in Gitlab, likely due to incompatibilities with the latest Gradle versions. Gradle has moved the plugin from `com.gradle.enterprise` to `com.gradle.develocity`. See [Migration guide](https://docs.gradle.com/develocity/gradle-plugin/legacy/#develocity_migration)

PR #6939, opened 2024-04-23
Fix folder permission interaction with Gitlab and gradle

randomanderson:
# What Does This Do

https://github.com/DataDog/dd-trace-java/pull/6931 inadvertently broke Gitlab builds because of the interactions between the Gitlab cache and gradle:

* Gradle always attempts to change permissions of the `.gradle` folder. See [here](https://github.com/gradle/gradle/blob/master/platforms/core-execution/persistent-cache/src/main/java/org/gradle/cache/internal/FileBackedObjectHolder.java#L114)
* With Gitlab caching, `.gradle` is owned by root
* Gradle's attempt fails and the build fails

This PR adds a workaround by copying moving back the cache folder so that it is owned by the gitlab runner user.

Related Gradle issues:
https://github.com/gradle/gradle/issues/14200
https://github.com/gradle/gradle/issues/8673

### DIFF
--- a/.gitlab-ci.yml
+++ b/.gitlab-ci.yml
@@ -48,6 +48,14 @@ cache: &default_cache
   image: ghcr.io/datadog/dd-trace-java-docker-build:v23.10-base
   before_script:
     - export GRADLE_USER_HOME=`pwd`/.gradle
+    # for weird reasons, gradle will always "chmod 700" the .gradle folder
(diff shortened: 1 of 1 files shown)


PerfectSlayer (approved):
Thanks for documenting the workaround in the issue 🙏 
We will stick with it until we find a better solution (if ever).

PR #6941, opened 2024-04-24
Ensure time waiting for `CyclicBarrier` and `CountDownLatch` is eligible for wallclock profiling

richardstartin:
# What Does This Do

* add instrumentation to make sure the wallclock profiler can sample `CyclicBarrier` and `CountDownLatch` waiting time, regardless of whether there is an active trace
* Consolidate existing similar instrumentation for netty event loops into new instrumentation

# Motivation

More complete wallclock profiling

# Additional Notes

Jira ticket: [PROF-9658]

[PROF-9658]: https://datadoghq.atlassian.net/browse/PROF-9658?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ


dd-java-agent/instrumentation/enable-wallclock-profiling/src/main/java/datadog/trace/instrumentation/wallclock/EnableWallclockProfilingInstrumentation.java
```diff
@@ -0,0 +1,104 @@
...
+    "io.grpc.netty.shaded.io.netty.channel.ThreadPerChannelEventLoop",
+    "io.grpc.netty.shaded.io.netty.channel.nio.NioEventLoop",
+    "io.grpc.netty.shaded.io.netty.channel.epoll.EPollEventLoop",
+    "io.grpc.netty.shaded.io.netty.channel.kqueue.KQueueEventLoop"
+  };
+
+  @Override
+  public boolean isEnabled() {
+    // only needed if wallclock profiling is enabled, which requires tracing
+    return super.isEnabled()
```
jbachorik, line 46: I assume `super.isEnabled()` will return `true` by default? Eg. it is not necessary to enable the integration by adding a config option?

richardstartin: Exactly, we want to bypass the tracer based enablement mechanism and use our own. Maybe we can refactor this at a later date so we don't need to do this, but this was just carried over from the instrumentation being generalised.

PR #6955, opened 2024-04-26
Only capture stacktrace for snapshot probes

jpbempel:
# What Does This Do

# Motivation
For any log probes we recorded stacktrace while this is useless and wasteful for template log only

# Additional Notes

Jira ticket: [PROJ-IDENT]


evanchooly (approved):
hard to get much simpler.  i might add a test for the inverse, though.

jpbempel:
> hard to get much simpler. i might add a test for the inverse, though.

Largely covered by tests in `CapturedSnapshotTest`

PR #6956, opened 2024-04-26
Finish pyspark-shell trace using SparkListenerApplicationEnd

paul-laffon-dd:
# What Does This Do

Finish pyspark-shell application span when the spark listener is receiving the `SparkListenerApplicationEnd` event

The application span is currently finished [when exiting the runMain](https://github.com/DataDog/dd-trace-java/blob/master/dd-java-agent/instrumentation/spark/src/main/java/datadog/trace/instrumentation/spark/AbstractSparkInstrumentation.java#L56-L63) method to capture JVM errors ([from this PR](https://github.com/DataDog/dd-trace-java/pull/5505)), however this function is not exited as expected (System.exit(0)) when executing as pyspark-shell

# Motivation

Pyspark application can be launched [using multiple method](https://spark.apache.org/docs/latest/quick-start.html#self-contained-applications), mainly:
- `spark-submit script.py` 
- `python script.py` 

The spark instrumentation is working as expected in the `spark-submit` case, because this is java/scala code launched behind the scene

However, the `python` case, the listener is being injected as expected, but the `spark.application` span is not being closed properly because the JVM is [launched as PythonGatewayServer](https://github.com/apache/spark/blob/v3.5.1/core/src/main/scala/org/apache/spark/deploy/SparkSubmit.scala#L540-L542) which is [exited using System.exit(0)](https://github.com/apache/spark/blob/v3.5.1/core/src/main/scala/org/apache/spark/api/python/PythonGatewayServer.scala#L74), leading to the exit advice not being called

# Additional Notes

### DIFF
--- a/dd-java-agent/instrumentation/spark/src/main/java/datadog/trace/instrumentation/spark/AbstractSparkInstrumentation.java
+++ b/dd-java-agent/instrumentation/spark/src/main/java/datadog/trace/instrumentation/spark/AbstractSparkInstrumentation.java
@@ -1,13 +1,12 @@
 package datadog.trace.instrumentation.spark;
 
 import static datadog.trace.agent.tooling.bytebuddy.matcher.NameMatchers.named;
-import static net.bytebuddy.matcher.ElementMatchers.isDeclaredBy;
-import static net.bytebuddy.matcher.ElementMatchers.isMethod;
-import static net.bytebuddy.matcher.ElementMatchers.nameEndsWith;
+import static net.bytebuddy.matcher.ElementMatchers.*;
 
 import datadog.trace.agent.tooling.Instrumenter;
 import datadog.trace.agent.tooling.InstrumenterModule;
 import net.bytebuddy.asm.Advice;
+import org.apache.spark.deploy.SparkSubmitArguments;
 
 public abstract class AbstractSparkInstrumentation extends InstrumenterModule.Tracing
     implements Instrumenter.ForKnownTypes {
@@ -32,6 +31,14 @@ public abstract class AbstractSparkInstrumentation extends InstrumenterModule.Tr
 
   @Override
   public void methodAdvice(MethodTransformer transformer) {
+    // Capture spark submit arguments
+    transformer.applyAdvice(
+        isMethod()
+            .and(named("prepareSubmitEnvironment"))
+            .and(takesArgument(0, named("org.apache.spark.deploy.SparkSubmitArguments")))
+            .and(isDeclaredBy(named("org.apache.spark.deploy.SparkSubmit"))),
+        AbstractSparkInstrumentation.class.getName() + "$PrepareSubmitEnvAdvice");
+
     // SparkSubmit class used for non YARN/Mesos environment
     transformer.applyAdvice(
         isMethod()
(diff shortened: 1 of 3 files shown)


tabgok (approved):

dd-java-agent/instrumentation/spark/src/main/java/datadog/trace/instrumentation/spark/AbstractSparkInstrumentation.java
```diff
@@ -1,13 +1,12 @@
 package datadog.trace.instrumentation.spark;
 
 import static datadog.trace.agent.tooling.bytebuddy.matcher.NameMatchers.named;
-import static net.bytebuddy.matcher.ElementMatchers.isDeclaredBy;
-import static net.bytebuddy.matcher.ElementMatchers.isMethod;
-import static net.bytebuddy.matcher.ElementMatchers.nameEndsWith;
+import static net.bytebuddy.matcher.ElementMatchers.*;
```
amarziali, line 4: The star imports are discouraged. Could you please reconfigure the IDE according to the [contributing](https://github.com/DataDog/dd-trace-java/blob/2b0502cacf7715840d83a23c49580b6dc19931e8/CONTRIBUTING.md#intellij-idea) doc and format again?

dd-java-agent/instrumentation/spark/src/main/java/datadog/trace/instrumentation/spark/AbstractSparkInstrumentation.java
```diff
@@ -47,10 +54,29 @@ public void methodAdvice(MethodTransformer transformer) {
...
+        AbstractDatadogSparkListener.finishTraceOnApplicationEnd = true;
+      }
+    }
+  }
+
   public static class RunMainAdvice {
     @Advice.OnMethodEnter(suppress = Throwable.class)
     public static void enter() {
-      AbstractDatadogSparkListener.finishTraceOnApplicationEnd = false;
+      if (!AbstractDatadogSparkListener.isPysparkShell) {
```
amarziali, line 77: This is probably not needed since it's false by default

amarziali (approved):
LGTM. I left a couple of comments. I strongly encourage to apply the right formatting before merging

dd-java-agent/instrumentation/spark/src/main/java/datadog/trace/instrumentation/spark/AbstractSparkInstrumentation.java
paul-laffon-dd: It is false by default, but if `runMain` is called after `prepareSubmitEnvironment`, we still want to have when using pyspark-shell `finishTraceOnApplicationEnd=true`

dd-java-agent/instrumentation/spark/src/main/java/datadog/trace/instrumentation/spark/AbstractSparkInstrumentation.java
paul-laffon-dd: Yes, thanks, fixed the imports

PR #6960, opened 2024-04-26
Only enable byte[] or char[] call sites in full detection mode

manuel-alvarez-alvarez:
# What Does This Do
Moves the `String` call sites dealing with `byte[]` and `char[]` to a separate class that is only enabled on full mode.

# Motivation
Those call sites can be specially harmful for performance so we will left them out of the default mode and only include them full detection mode is activated.

# Additional Notes

Jira ticket: [PROJ-IDENT]


ValentinZakharov (approved):

ygree (approved):
LGTM

smola (approved):

PR #6963, opened 2024-04-29
Fix CI broken sources artifact

PerfectSlayer:
# What Does This Do

This PR removes the broken `packageSources` task.
This task creates "-sources.jar" like the `withSourcesJar()` from the [Java gradle plugin](https://docs.gradle.org/current/kotlin-dsl/gradle/org.gradle.api.plugins/-java-plugin-extension/with-sources-jar.html).

The artifact produced by the `packageSources` tasks were overridden by the Java plugin ones at every build and broke the build cache.

I checked the artifact that should be produced by the `packageSources` task are the same (in size and in file count) that the ones the Java plugin builds. We should be able to safely remove the task without trying to fix it.

# Motivation

This task breaks the build cache by creating artifacts that are always overridden.

image

Fixing it finally makes the all the `gradle assemble` command tasks _up-to-date_:
```bash
BUILD SUCCESSFUL in 10s
2114 actionable tasks: 2114 up-to-date
``` 

# Additional Notes

I also group the two `jar` task configuration into one as clean up.

Jira ticket: [PROJ-IDENT]

### DIFF
--- a/dd-java-agent/agent-tooling/build.gradle
+++ b/dd-java-agent/agent-tooling/build.gradle
@@ -35,7 +35,6 @@ configurations {
 }
 
 compileJava.dependsOn 'generateClassNameTries'
-packageSources.dependsOn 'generateClassNameTries'
 sourcesJar.dependsOn 'generateClassNameTries'
 
 dependencies {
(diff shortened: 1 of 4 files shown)


mcculls (approved):
Makes sense

PR #6967, opened 2024-04-29
Add enabledProducts tag to all spans (APM / DSM / DJM)

kr-igor:
# What Does This Do
This PR adds product tags for DSM and DJM.

# Motivation
Because customers may use APM / DJM / DSM in any configuration we need an easy way of distinguishing spans produced by the tracer (to avoid double charging in some cases).

# Additional Notes
[Jira](https://datadoghq.atlassian.net/browse/DSMON-356)


richardstartin (changes requested):
There is already a pattern of adding product enablement tags to root spans for appsec and profiling, I suggest following this pattern. Let's discuss offline.

richardstartin (approved):

internal-api/src/main/java/datadog/trace/api/InstrumenterConfig.java
```diff
@@ -262,6 +263,12 @@ public boolean isIntegrationShortcutMatchingEnabled(
         integrationNames, "integration.", ".matching.shortcut.enabled", defaultEnabled);
   }
 
+  public boolean isDataJobsEnabled() {
+    // there's no dedicated flag to enabled DJM, it's enough to just enable
+    // spark instrumentation
+    return isIntegrationEnabled(Collections.singletonList("spark"), false);
+  }
```
gabe-ochoa, line 270: @paul-laffon-dd, we should update this as apart of #6972. Now that this PR is merged we can do both.

PR #6969, opened 2024-04-30
kafka: Add poll span when DSM is enabled

piochelepiotr:
# What Does This Do

Adds a span for Kafka polls when Data Streams is enabled. A few benefits of doing that:
1. Debug slow polls, which can sometimes be an issue with the Kafka brokers
2. We get visibility into Kafka schema registry calls (and extracting schemas) when a schema registry is used. This can cause slow Kafka polls, so it's useful debugging information

# Motivation

Track schemas on Kafka consume (follow up to https://github.com/DataDog/dd-trace-java/pull/6865)

# Additional Notes

Jira ticket: [PROJ-IDENT]


dd-java-agent/instrumentation/kafka-clients-0.11/src/main/java/datadog/trace/instrumentation/kafka_clients/KafkaConsumerInfoInstrumentation.java
```diff
@@ -130,10 +136,28 @@ public static void muzzleCheck(ConsumerRecord record) {
    * KafkaConsumer class.
    */
   public static class RecordsAdvice {
+    @Advice.OnMethodEnter(suppress = Throwable.class)
+    public static AgentScope onEnter() {
+      if (Config.get().isDataStreamsEnabled()) {
```
dougqh, line 141: Extra spans have the potential to create noisy traces.
But I guess this is okay, since it is under "data streams enabled".

dd-java-agent/instrumentation/kafka-clients-0.11/src/test/groovy/KafkaClientTestBase.groovy
```diff
@@ -53,13 +55,57 @@ abstract class KafkaClientTestBase extends VersionedNamingTestBase {
...
+    @Override
+    boolean accept(List<DDSpan> trace) {
+      return !(trace.size() == 1 &&
+        trace.get(0).getResourceName().toString().equals("kafka.poll") &&
+        trace.get(0).getTag(InstrumentationTags.KAFKA_RECORDS_COUNT).equals(0))
+    }
+  }
+
+  // TraceID, start times & names changed based on the configuration, so overriding the sort to give consistent test results
+  private static class SortKafkaTraces implements Comparator<List<DDSpan>> {
```
dougqh, line 77: Are the usual test sorting mechanisms not sufficient?

piochelepiotr: Unfortunately, it doesn't look like it 😢 
trace ID are inconsistent because poll can start before produce (and finish after)
names are not good because the kafka consume span has different names depending on the configs
start times are not good either, same reason as trace IDs

dd-java-agent/instrumentation/kafka-clients-0.11/src/main/java/datadog/trace/instrumentation/kafka_clients/KafkaConsumerInfoInstrumentation.java
amarziali: You can also try to see if you have an active span and take the dsm config from it since using `Config.get()` won't let you access to the dynamic config

dd-java-agent/instrumentation/kafka-clients-0.11/src/main/java/datadog/trace/instrumentation/kafka_clients/KafkaConsumerInfoInstrumentation.java
```diff
@@ -130,10 +136,28 @@ public static void muzzleCheck(ConsumerRecord record) {
...
     @Advice.OnMethodExit(suppress = Throwable.class)
     public static void captureGroup(
-        @Advice.This KafkaConsumer consumer, @Advice.Return ConsumerRecords records) {
+        @Advice.Enter final AgentScope scope,
+        @Advice.This KafkaConsumer consumer,
+        @Advice.Return ConsumerRecords records) {
       if (records == null) {
+        if (scope != null) {
+          AgentSpan span = scope.span();
+          if (span != null) {
```
ygree, line 156: Is this scope span null check needed at all?

dd-java-agent/instrumentation/kafka-clients-0.11/src/main/java/datadog/trace/instrumentation/kafka_clients/KafkaConsumerInfoInstrumentation.java
```diff
@@ -142,6 +166,15 @@ public static void captureGroup(
         InstrumentationContext.get(ConsumerRecords.class, KafkaConsumerInfo.class)
             .put(records, kafkaConsumerInfo);
       }
+      if (scope == null) {
```
ygree, line 169: Maybe do this scope null check once at the beginning of the method?

dd-java-agent/instrumentation/kafka-clients-0.11/src/main/java/datadog/trace/instrumentation/kafka_clients/KafkaConsumerInfoInstrumentation.java
```diff
@@ -130,10 +136,28 @@ public static void muzzleCheck(ConsumerRecord record) {
...
+    }
+
     @Advice.OnMethodExit(suppress = Throwable.class)
     public static void captureGroup(
-        @Advice.This KafkaConsumer consumer, @Advice.Return ConsumerRecords records) {
+        @Advice.Enter final AgentScope scope,
+        @Advice.This KafkaConsumer consumer,
+        @Advice.Return ConsumerRecords records) {
       if (records == null) {
+        if (scope != null) {
```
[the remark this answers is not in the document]
piochelepiotr: I updated the code, please let me know what you think of the new version.
The tricky part is that both records & scope can be null, but in both cases, I still want to take an action.

dd-java-agent/instrumentation/kafka-clients-0.11/src/main/java/datadog/trace/instrumentation/kafka_clients/KafkaConsumerInfoInstrumentation.java
piochelepiotr: You're right, is scope is not null, there must be a span (as far as I understand)

dd-java-agent/instrumentation/kafka-clients-0.11/src/main/java/datadog/trace/instrumentation/kafka_clients/KafkaConsumerInfoInstrumentation.java
piochelepiotr: I updated a bit the code, the hard part is that I want to attach the kafka consumer info to records even if the scope is null.

ygree (approved):
LGTM

PR #6972, opened 2024-04-30
Add single parameter `dd.data.jobs.enabled` to enable DJM

paul-laffon-dd:
# What Does This Do

Add single parameter `dd.data.jobs.enabled=true` to enable data jobs monitoring. 

This parameter will change the default to have
- `dd.trace.experimental.long-running.enabled=true`
- `dd.integration.spark.enabled=true`
- `dd.integration.spark-executor.enabled=true`

Only the default are changed, those parameters can still be changed by explicitly providing them

# Motivation

Simpler way to enabled DJM

# Additional Notes


dd-java-agent/agent-bootstrap/src/main/java/datadog/trace/bootstrap/Agent.java
```diff
@@ -197,6 +198,22 @@ public static void start(final Instrumentation inst, final URL agentJarURL, Stri
...
+      // If data jobs is enabled, changing default to only enable spark instrumentations
+      setSystemPropertyDefault(
+          propertyNameToSystemPropertyName(TraceInstrumentationConfig.INTEGRATIONS_ENABLED),
+          "false");
+      setSystemPropertyDefault(
+          propertyNameToSystemPropertyName(TracerConfig.TRACE_LONG_RUNNING_ENABLED), "true");
+      setSystemPropertyDefault(
+          propertyNameToSystemPropertyName("integration.spark.enabled"), "true");
+      setSystemPropertyDefault(
+          propertyNameToSystemPropertyName("integration.spark-executor.enabled"), "true");
```
gabe-ochoa, line 214: Once the apm intake work is done for #6967, can we just turn on all integrations?

kr-igor (changes requested):
I think this change may cause some issues for customers who tries to use DSM in combination with DJM, as instrumentations required by DSM will be forcibly disabled. Could you also add "force enable" for DSM instrumentations if DSM_ENABLED = true? You can find these instrumentations by searching setDatastreamsCheckpoints.

dd-java-agent/agent-bootstrap/src/main/java/datadog/trace/bootstrap/Agent.java
kr-igor: Do we want to disable sampling as well? It may be a big issue for some hight throughput systems with millions of operations per second.

paul-laffon-dd: Changed it so that `dd.data.jobs.enabled=true` does not change the default for all integrations (enabled by default), only enables spark/long running spans

dd-java-agent/agent-bootstrap/src/main/java/datadog/trace/bootstrap/Agent.java
```diff
@@ -197,6 +198,20 @@ public static void start(final Instrumentation inst, final URL agentJarURL, Stri
       }
     }
 
+    boolean dataJobsEnabled = isFeatureEnabled(AgentFeature.DATA_JOBS);
+    if (dataJobsEnabled) {
+      log.info("Data Jobs Monitoring enabled, changing default integrations enabled");
+      setSystemPropertyDefault(
+          propertyNameToSystemPropertyName(TraceInstrumentationConfig.INTEGRATIONS_ENABLED),
```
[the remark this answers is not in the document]
paul-laffon-dd: The `setSystemPropertyDefault` is only changing the default, if the user explicitly set a parameter it will take precedence over the default

kr-igor (approved):

PR #6974, opened 2024-04-30
Add coordinated sampling for Exception Debugging

jpbempel:
# What Does This Do
if the first stack level generates a snapshot after sampling
 we keep continuing to generate snapshot for next levels by forcing
sampling (not calling the sampler).
For each evaluation of exception probe we query the state by throwable
 instance and based on this state (list of snapshots) we take the
decision keep sampling or drop)

# Motivation
Having complete stacks with all snpashots without depending on the sampling for each stack level

# Additional Notes

Jira ticket: [DEBUG-2068]


ojung (approved):
Thanks 👍

PR #6980, opened 2024-05-02
[Dynamic Instrumentation] Consolidate PII redaction keys for all libraries.

shurivich:
# What Does This Do
Consolidate PII redaction keys for all libraries.

# Additional Notes
Related PRs:
`dotnet` - https://github.com/DataDog/dd-trace-dotnet/pull/5522
`python` - https://github.com/DataDog/dd-trace-py/pull/9148


dd-java-agent/agent-debugger/debugger-bootstrap/src/main/java/datadog/trace/bootstrap/debugger/util/Redaction.java
```diff
@@ -20,22 +20,24 @@ public class Redaction {
       Arrays.asList(
           "2fa",
           "accesstoken",
+          "address",
```
jpbempel, line 23: I still think `address` is too generic and should not be part of this list

shurivich: I think we spoke about `ipadress`? 🤔 
Anyway, I don't have a strong opinion here. 
@OmerRaviv, @cimi ?

cimi: IMO there's no right answer here - e.g. if it's a customer address then we should redact it because it's PII; if it's a public institution address or some logistics thing the we should not redact it.

Given we redact other generic words that could potentially be useful, like `config`, I think we can be conservative and keep it on the redaction list.

dd-java-agent/agent-debugger/debugger-bootstrap/src/main/java/datadog/trace/bootstrap/debugger/util/Redaction.java
```diff
@@ -47,15 +49,17 @@ public class Redaction {
...
+          "ipaddress",
           "jti",
           "jwt",
           "licensekey",
           "masterkey",
           "mysqlpwd",
           "nonce",
           "oauth",
-          "oauth_token",
+          "oauthtoken",
```
cimi, line 62: :+1: :+1:

cimi (approved):

dd-java-agent/agent-debugger/debugger-bootstrap/src/main/java/datadog/trace/bootstrap/debugger/util/Redaction.java
jpbempel: ah yes `config` is redacted and this is problematic too!
we can debate a lot about this list, but the fact that we don't have a mechanism for opting out a keyword from that list is the most problematic one, specially if we add generic word like `address` and `config`

shurivich: I'm going to leave it as it is for now and merge it like this. 

IMO, the best scenario is that we shouldn't have these things hardcoded, but rather dynamically updated by configuration. Another option is to provide a feature that allows customers to ignore hardcoded redacted keys.

PR #6990, opened 2024-05-03
[DO] - Add product tags to each span if products are enabled

kr-igor:
# What Does This Do
This PR is a followup to https://github.com/DataDog/dd-trace-java/pull/6967. 

# Motivation
Adding product tags only to local root is not enough, as there's no guarantee that every trace segment received by the backend will have this tag (due to partial flushing, trace segment splitting etc). We need a simple and robust way of marking all data coming from DSM / DJM products. Adding the tag only when the product is enabled will add the overhead only for customers of this product.

# Additional Notes
[Jira](https://datadoghq.atlassian.net/browse/DSMON-356)


internal-api/src/main/java/datadog/trace/api/Config.java
```diff
@@ -3535,6 +3533,12 @@ public Map<String, String> getMergedSpanTags() {
   public Map<String, String> getSpanTags() {
     final Map<String, String> result = newHashMap(spanTags.size());
```
ygree, line 3534: Maybe add enough space for extra tags?

amarziali:
We had the same issue in the past for the sampling priority. If you want to minimize the data overhead but still ensuring that the intake will read that value, you can just write on first and last span of each chunk. We already do it [here](https://github.com/DataDog/dd-trace-java/blob/146516473928794b177fbdb3e517b03ece851f02/dd-trace-core/src/main/java/datadog/trace/common/writer/ddagent/TraceMapperV0_4.java#L74) (same for V05) for the sampling priority. If you would like to adopt that way, you can exploit the `writeSamplePriority` boolean which will be true for first and last span of each chunk. It will also work with partial flush

kr-igor:
> writeSamplePriority

@amarziali thank you for your comment! I'm a bit worried about cases when the trace gets split on the backed, can't really tell if tagging first and last will work. I'll need some time to investigate and can optimize in a follow up PR(s).

ygree (approved):
LGTM

PR #6992, opened 2024-05-06
Add support of Set in Expression Language

jpbempel:
# What Does This Do
When having a value with a Set interface creates a SetValue to hold that Set instance
Allow to call size() on it if it safe list for lists and maps

# Motivation
Support `count(set)`

# Additional Notes

Jira ticket: [DEBUG-2329]

[DEBUG-2329]: https://datadoghq.atlassian.net/browse/DEBUG-2329?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ

### DIFF
--- a/dd-java-agent/agent-debugger/debugger-el/src/main/java/com/datadog/debugger/el/values/SetValue.java
+++ b/dd-java-agent/agent-debugger/debugger-el/src/main/java/com/datadog/debugger/el/values/SetValue.java
@@ -1,0 +1,86 @@
+package com.datadog.debugger.el.values;
+
+import com.datadog.debugger.el.Value;
+import com.datadog.debugger.el.Visitor;
+import com.datadog.debugger.el.expressions.ValueExpression;
+import datadog.trace.bootstrap.debugger.el.ValueReferenceResolver;
+import datadog.trace.bootstrap.debugger.el.Values;
+import datadog.trace.bootstrap.debugger.util.WellKnownClasses;
+import java.util.Set;
+
+public class SetValue implements CollectionValue<Object>, ValueExpression<SetValue> {
+
+  private final Object setHolder;
+
+  public SetValue(Object object) {
+    if (object instanceof Set) {
+      setHolder = object;
+    } else if (object == null || object == Values.NULL_OBJECT) {
+      setHolder = Value.nullValue();
+    } else {
+      setHolder = Value.undefinedValue();
+    }
+  }
+
+  @Override
+  public SetValue evaluate(ValueReferenceResolver valueRefResolver) {
+    return this;
+  }
+
+  @Override
+  public Object getValue() {
+    return setHolder;
+  }
+
+  @Override
+  public boolean isEmpty() {
+    if (setHolder instanceof Set) {
+      return ((Set<?>) setHolder).isEmpty();
+    } else if (setHolder instanceof Value) {
+      Value<?> val = (Value<?>) setHolder;
+      return val.isNull() || val.isUndefined();
+    }
+    return true;
+  }
+
+  @Override
+  public int count() {
+    if (setHolder instanceof Set) {
+      if (WellKnownClasses.isSizeSafe((Set<?>) setHolder)) {
+        return ((Set<?>) setHolder).size();
+      }
+      throw new RuntimeException("Unsupported Set class: " + setHolder.getClass().getTypeName());
+    } else if (setHolder == Value.nullValue()) {
+      return 0;
+    }
+    return -1;
+  }
+
+  @Override
+  public Value<?> get(Object key) {
+    if (key == Value.undefinedValue() || key == Values.UNDEFINED_OBJECT) {
+      return Value.undefinedValue();
+    }
+    if (key == null || key == Value.nullValue() || key == Values.NULL_OBJECT) {
+      return Value.nullValue();
+    }
+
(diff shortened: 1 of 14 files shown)


dd-java-agent/agent-debugger/debugger-el/src/main/java/com/datadog/debugger/el/values/SetValue.java
```diff
@@ -0,0 +1,86 @@
...
+  @Override
+  public Value<?> get(Object key) {
+    if (key == Value.undefinedValue() || key == Values.UNDEFINED_OBJECT) {
+      return Value.undefinedValue();
+    }
+    if (key == null || key == Value.nullValue() || key == Values.NULL_OBJECT) {
+      return Value.nullValue();
+    }
+
+    if (setHolder instanceof Set) {
```
shatzi, line 68: why we don't check WellKnownClasses here?

dd-java-agent/agent-debugger/debugger-el/src/main/java/com/datadog/debugger/el/values/SetValue.java
```diff
@@ -0,0 +1,86 @@
...
+  public int count() {
+    if (setHolder instanceof Set) {
+      if (WellKnownClasses.isSizeSafe((Set<?>) setHolder)) {
+        return ((Set<?>) setHolder).size();
+      }
+      throw new RuntimeException("Unsupported Set class: " + setHolder.getClass().getTypeName());
+    } else if (setHolder == Value.nullValue()) {
+      return 0;
+    }
+    return -1;
```
shatzi, line 56: why -1? why not undefined? is that in the spec ?
I think count on String also return -1 for null. not sure why we do that.

dd-java-agent/agent-debugger/debugger-el/src/test/resources/test_conditional_13.json
```diff
@@ -0,0 +1,13 @@
+{
+  "dsl": "",
+  "json": {
+    "and": [
+      {"eq": [{"count": {"ref": "intArray"}}, 3]},
+      {"eq": [{"len": {"ref": "intArray"}}, 3]},
```
shatzi, line 6: why we have both len and count?

dd-java-agent/agent-debugger/debugger-el/src/test/resources/test_conditional_13.json
```diff
@@ -0,0 +1,13 @@
...
+      {"eq": [{"count": {"ref": "strArray"}}, 2]},
+      {"eq": [{"count": {"ref": "strMap"}}, 2]},
+      {"eq": [{"count": {"ref": "strSet"}}, 1]},
```
shatzi, line 9: do we have tests like that for isEmpty?

dd-smoke-tests/debugger-integration-tests/src/test/java/datadog/smoketest/MoshiConfigTestHelper.java
```diff
@@ -333,5 +334,10 @@ public Void visit(ListValue listValue) {
     public Void visit(MapValue mapValue) {
       throw new UnsupportedOperationException("mapValue");
     }
+
+    @Override
+    public Void visit(SetValue setValue) {
+      throw new UnsupportedOperationException("mapValue");
```
shatzi, line 340: setValue

shatzi (approved):

dd-java-agent/agent-debugger/debugger-el/src/main/java/com/datadog/debugger/el/values/SetValue.java
jpbempel: legacy from the initial implementation. should never reached this code.
> why not undefined?
because method should return an int! otherwise we need to throw an exception

dd-java-agent/agent-debugger/debugger-el/src/main/java/com/datadog/debugger/el/values/SetValue.java
jpbempel: Same for Maps.
Need to revisit this in fact in a next PR

dd-java-agent/agent-debugger/debugger-el/src/test/resources/test_conditional_13.json
jpbempel: Initial implementation supported both. 
I think it's because you prefer doing `len` over a string and `count` on a collection

dd-java-agent/agent-debugger/debugger-el/src/test/resources/test_conditional_13.json
jpbempel: ah yes missed that. we have the test but forgot to add Set for them

dd-smoke-tests/debugger-integration-tests/src/test/java/datadog/smoketest/MoshiConfigTestHelper.java
jpbempel: done

PR #7020, opened 2024-05-13
Improve onboarding documentation

PerfectSlayer:
# What Does This Do

This PR improves the project onboarding and documentation.

The README now has a list of documentation to improve discoverability,
The setup documentation now has MacOS, Linux and Windows guidelines,
The test document is introduced. 

# Motivation

This should improve the overall onboarding experience.

# Additional Notes

The test document could include an _how to write tests_ section but it will require some more time.
The next documentation should be about how to setup IntelliJ IDEA and how to troubleshoot build/gradle issues.
 
Jira ticket: [PROJ-IDENT]

### DIFF
--- a/setup.sh
+++ b/setup.sh
@@ -18,7 +18,7 @@ function check-jvm() {
     fi
 }
 
-echo "ℹ️  Checking required JVM:"
+echo "ℹ️ Checking required JVM:"
 if [ -e "$JAVA_HOME" ]; then
     check-jvm "JAVA_HOME" "1.8"
 fi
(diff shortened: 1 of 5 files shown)


setup.sh
```diff
@@ -105,7 +95,23 @@ function check-docker-server() {
     fi
 }
 
-echo "ℹ️  Checking shell configuration:"
-check-ulimit 1024
+echo "ℹ️  Checking Docker environment:"
```
smola, line 98: Just a minor nitpick: these info headers have a two-space indent, while the nested items have one-space indent. So the nesting looks backwards. It should probably be 1 space for the info headers, and 2 spaces for each item.

PerfectSlayer: I set two spaces as my shell failed to properly render the ℹ️ (it is swallowing the first space and I don’t know why 🤷 ).
I removed it but I am curious to know how it renders on your side if you could run it?

mcculls (approved):
Great work!

PR #7047, opened 2024-05-17
Add Throwable capturing fields support for JDK16+

jpbempel:
# What Does This Do
We introduce in `WellKnownClasses` the mechanism to fetch fields through dedicated call to getter for registered classes.
In the special case of `Throwable` we are allowing any inherited classes to capture the `Throwable` fields as well as custom fields. Though, to capture `Throwable` fields (like `detailMessage`) we need to call getter that could be overridden. Hence ,we are checking before calling them that they are not overridden by checking the declaring class of the getter.
Add option in build to not allow reflection access for debugger tests to assess that we can access correctly even for recent JDKs

# Motivation
Since JDK16 it's not possible to extract values by reflection for JDK modules without adding `--add-opens=java.base/java.lang=ALL-UNNAMED` so when capturing an exception (like in `@exception`) we are not allowed to capture fields like `detailMessage`, `cause`, `stacktrace` which can be valuable to troubleshoot.

# Additional Notes
Results:
Screenshot 2024-05-17 at 14 00 48

Screenshot 2024-05-17 at 14 01 23

Jira ticket: [DEBUG-2389]

[DEBUG-2389]: https://datadoghq.atlassian.net/browse/DEBUG-2389?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ

### DIFF
--- a/dd-java-agent/agent-debugger/debugger-bootstrap/src/main/java/datadog/trace/bootstrap/debugger/util/WellKnownClasses.java
+++ b/dd-java-agent/agent-debugger/debugger-bootstrap/src/main/java/datadog/trace/bootstrap/debugger/util/WellKnownClasses.java
@@ -1,6 +1,8 @@
 package datadog.trace.bootstrap.debugger.util;
 
 import datadog.trace.bootstrap.debugger.CapturedContext;
+import java.lang.reflect.InvocationTargetException;
+import java.lang.reflect.Method;
 import java.util.Arrays;
 import java.util.Collection;
 import java.util.HashMap;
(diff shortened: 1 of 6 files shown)


dd-java-agent/agent-debugger/debugger-bootstrap/src/main/java/datadog/trace/bootstrap/debugger/util/WellKnownClasses.java
```diff
@@ -141,4 +204,87 @@ private static String classToString(Object o) {
...
+      LOGGER.debug("Failed to get declaring class for Throwable::getMessage", e);
+    }
+    return declaringClass != originalDeclaringClass;
+  }
+
+  private static class ThrowableFields {
+    public static final String BECAUSE_OVERRIDDEN =
+        "Special access method not safe to be called because overridden";
+
+    public static CapturedContext.CapturedValue detailMessage(Object o) {
```
shatzi, line 223: maybe refactor this.
```
private static captureValueIfNotOverridden(Object o, string method, string field, string type, callback callback) {
   if (isOverridden(o, method, Throwable.class)) {
        return CapturedContext.CapturedValue.notCapturedReason(field, type, BECAUSE_OVERRIDDEN);
      }
      return CapturedContext.CapturedValue.of(field, type, callback());
    }
}
```

then every function can be a single line:
```
captureValueIfNotOverridden(o, "getMessage", "detailMessage", String.class.getTypeName(), () => ((Throwable) o).getMessage());
```

shatzi (approved):
can we please add tests to verify that all this works?

dougqh (approved):

dd-java-agent/agent-debugger/debugger-bootstrap/src/main/java/datadog/trace/bootstrap/debugger/util/WellKnownClasses.java
jpbempel: done. a little more complex than what you suggest :)

dd-java-agent/agent-debugger/debugger-bootstrap/src/main/java/datadog/trace/bootstrap/debugger/util/WellKnownClasses.java
```diff
@@ -110,23 +153,43 @@ public static boolean isSafe(Map<?, ?> map) {
...
+      Object value) {
+    if (value == null) {
+      return null;
+    }
+    Map<String, Function<Object, CapturedContext.CapturedValue>> specialTypeAccess =
+        SPECIAL_TYPE_ACCESS.get(value.getClass());
+    if (specialTypeAccess != null) {
+      return specialTypeAccess;
+    }
+    if (value instanceof Throwable) {
```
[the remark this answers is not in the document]
jpbempel: done

PR #7055, opened 2024-05-21
[DO] - Extra connection params as tags for JDBC statements

kr-igor:
…bility in Git config parsing logic (#7053)

# What Does This Do

# Motivation

# Additional Notes

Jira ticket: [PROJ-IDENT]

### DIFF
--- a/dd-java-agent/agent-bootstrap/src/main/java/datadog/trace/bootstrap/instrumentation/jdbc/DBInfo.java
+++ b/dd-java-agent/agent-bootstrap/src/main/java/datadog/trace/bootstrap/instrumentation/jdbc/DBInfo.java
@@ -1,6 +1,9 @@
 package datadog.trace.bootstrap.instrumentation.jdbc;
 
+import java.util.Arrays;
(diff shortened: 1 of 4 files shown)


dd-java-agent/instrumentation/jdbc/src/main/java/datadog/trace/instrumentation/jdbc/JDBCDecorator.java
```diff
@@ -122,6 +124,8 @@ protected String dbHostname(final DBInfo info) {
   public AgentSpan onConnection(final AgentSpan span, DBInfo dbInfo) {
     if (dbInfo != null) {
       processDatabaseType(span, dbInfo.getType());
+      span.setTag(DB_WAREHOUSE, dbInfo.getWarehouse());
```
amarziali, line 127: nit: a null check on the value can be done before setting the tag (same for the schema) to avoid synchronizing the map of tags at span level for nothing.

amarziali (approved):
lgtm

PR #7056, opened 2024-05-22
Fix error handling on RabbitMQ channel cleanup

PerfectSlayer:
# What Does This Do

Closing channels may raise wider type of error.

# Motivation

Uncaught errors make the CI flaky.

# Additional Notes

Jira ticket: [PROJ-IDENT]

### DIFF
--- a/dd-java-agent/instrumentation/rabbitmq-amqp-2.7/src/test/groovy/RabbitMQTest.groovy
+++ b/dd-java-agent/instrumentation/rabbitmq-amqp-2.7/src/test/groovy/RabbitMQTest.groovy
@@ -1,11 +1,11 @@
 import com.rabbitmq.client.AMQP
-import com.rabbitmq.client.AlreadyClosedException
 import com.rabbitmq.client.Channel
 import com.rabbitmq.client.ConnectionFactory
 import com.rabbitmq.client.Consumer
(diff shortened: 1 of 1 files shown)


dd-java-agent/instrumentation/rabbitmq-amqp-2.7/src/test/groovy/RabbitMQTest.groovy
```diff
@@ -89,7 +89,7 @@ abstract class RabbitMQTestBase extends VersionedNamingTestBase {
       channel?.close()
       conn?.close()
       TEST_DATA_STREAMS_WRITER?.clear()
-    } catch (AlreadyClosedException e) {
+    } catch (ShutdownSignalException e) {
```
amarziali, line 92: we might catch also both

amarziali (approved):

dd-java-agent/instrumentation/rabbitmq-amqp-2.7/src/test/groovy/RabbitMQTest.groovy
PerfectSlayer: `AlreadyClosedException` will be covered by `ShutdownSignalException` as it's a one of its sub-classes.

PR #7057, opened 2024-05-22
Update lib-injection docker image tags

andrewlock:
# What Does This Do

Adds `vMAJOR` and `vMAJOR.MINOR` lib-injection images (in addition to `vMAJOR.MINOR.PATCH` and `latest`)

# Motivation

We want to enable customers to be able to pin to a major version. We were also incorrectly tagging images as `latest` when they weren't (only would have happened on hotfixes, so hasn't _actually_ occurred). 

We are following ([this doc's suggestions](https://docs.google.com/document/d/1xYvWBz8OXeavZWxBmIGheZmcZkF7Cx9tXqT6mxgHkgs/edit?pli=1#heading=h.nud45hbcwu7p)), but in summary:

- Prerelease versions are _not_ tagged
- Every non-prerelease release gets the `vMajor.Minor.Patch` version tag
    - `2.5.0` gets `v2.5.0`
    - `2.2.3` gets `v2.2.3` 
    - `1.2.1` gets `v1.2.1`
- Every non-prerelease release gets the `vMajor.Minor` version tag initially (which assumes we never "go back" in release values)
    - `2.5.0` gets `v2.5`
    - `2.2.3` gets `v2.2`
    - `1.2.1` gets `v1.2`
- _Some_ releases get the `vMajor` version tag. Only releases for which this is the _highest_ version in the major get the tag.
    - `2.5.0` gets `v2` (_if_ there's no higher `2.x.x` release)
    - `1.2.1` gets `v1` (_if_ there's no higher `1.x.x` release)
- _Some_ releases get the latest tag. Only releases for which this is the highest version **ever** get the tag.
    - `2.5.0` gets `latest` if it's the highest release so far
    - `1.2.1` will not get `latest` if there's already `2.x.x` releases

The logic is now more complicated and requires knowing the state of the git repository. The script shown here mirrors [the one added for .NET](https://github.com/DataDog/dd-trace-dotnet/pull/5544).

> Note we're taking advantage of the [Datadog/public-images](https://github.com/DataDog/public-images) support for passing multiple csv values in the `IMG_DESTINATIONS` variable. You can see an example of this used in "the wild" [here](https://github.com/DataDog/datadog-ci/blob/a4041c4fca346122fa54dd67dad3d93d4f1731b9/.gitlab-ci.yml#L34).

# Additional Notes

The generation stage is quite verbose about printing out all the variables, but overall this is obviously very hard to test so I set up [a dummy GitHub repository and GitLab YAML](https://github.com/DataDog/dd-trace-dotnet-gitlab-test/blob/main/.gitlab-ci.yml), which just echoes the values it receives, to confirm they're sent across to the child pipeline correctly. 

If it is safe to do so, we can test this by reverting [the `revert "TESTING"` commit](https://github.com/DataDog/dd-trace-js/commit/11f9216ea4a5eb8610ea6439c2128bf756ecfe96)

I don't know how backporting or the extra `-rc` commit tags work, so hopefully someone else could pick that up 😅


.gitlab-ci.yml
```diff
@@ -213,47 +213,41 @@ deploy_artifacts_to_github:
...
-    project: DataDog/public-images
-    branch: main
-    strategy: depend
   variables:
-    IMG_SOURCES: ghcr.io/datadog/dd-trace-java/dd-lib-java-init:$CI_COMMIT_SHA
-    IMG_DESTINATIONS: dd-lib-java-init:$CI_COMMIT_TAG
-    IMG_SIGNING: "false"
+    IMG_DESTINATION_BASE: dd-lib-java-init
+  script:
+    - ./.gitlab/build-lib-init.sh
```
PerfectSlayer, line 232:
```suggestion
    - .gitlab/build-lib-init.sh
```

.gitlab-ci.yml
```diff
@@ -213,47 +213,41 @@ deploy_artifacts_to_github:
...
-      allow_failure: true
-    - if: '$POPULATE_CACHE'
-      when: never
-    - if: '$CI_COMMIT_TAG =~ /^v1\..*/'
-      when: on_success
-    - when: manual
-      allow_failure: true
+  # needs the version from the generate-tag-values job
+  needs:
+    - job: generate-lib-init-tag-values
```
PerfectSlayer, line 241: Why split `generate-lib-init-tag-values` and `deploy-lib-init-trigger` into two separate jobs?

.gitlab-ci.yml
```diff
@@ -213,47 +213,41 @@ deploy_artifacts_to_github:
...
   stage: deploy
   rules:
-    - if: '$CI_COMMIT_TAG =~ /.*-[Rr][Cc].*/'
-      when: manual
-      allow_failure: true
     - if: '$POPULATE_CACHE'
       when: never
-    - if: '$CI_COMMIT_TAG =~ /^v.*/'
+    # We don't tag prerelease versions
+    - if: '$CI_COMMIT_TAG =~ /^v[0-9]+\.[0-9]+\.[0-9]+$/'
```
PerfectSlayer, line 225: About RC, I don't know who are using them, and where they are getting them from?
@am312 Can you provide more info about them?

PerfectSlayer (commented):
Hey @andrewlock 👋 

Thanks for taking care of this.
As you could guess, our CI is somewhat messy but we will try to get it working.

I left a bunch of comments in your PRs and I still have few quick questions for you though:

The docker image is build on GitHub using [this action](https://github.com/DataDog/dd-trace-java/tree/master/.github/workflows#lib-injection-).
This means the tracer that is inside the docker image is not signed by our GitLab CI, but rather a dev build for the system tests to run.
Don't we want to have the right agent release binary (the one built on GitLab) to be distributed to our customers instead?

By the way, how to do expect the CI to coordinate themselves? (that GitHub will have built and published GitHub the docker image to GHCR before your publishing task to be executed)?

I also wonder if we will end up duplicating the `build-lib-init.sh` script in every repository or if we should communalise it (and if it is not something that could be related to [the existing script that builds the original image/tag](https://github.com/DataDog/system-tests/blob/main/lib-injection/docker-tags/action.yml)).

Pinging @randomanderson if he has more context about it 🙏

.gitlab-ci.yml
randomanderson: Trigger jobs can't have any `script` blocks

andrewlock: Because I think we have to due to the way GitLab works with trigger jobs I think. I could be wrong though, I suck at gitlab 😅 FWIW, other languages are using this script, so it _works_ even if it's not optimal

PerfectSlayer:
Thanks for the comments @andrewlock 🙏 
I understand you're not the one to blame for the current design / implementation of our delivery pipeline and you are just trying to get the SSI working as expected.

I will approve the PR but in the mean time, I want to make sure we start address all the pipeline issues as we keep adding complexity and features without addressing the root issues.

I hope it went better with the other languages 😅

PerfectSlayer (approved):

.gitlab-ci.yml
```diff
@@ -213,47 +213,41 @@ deploy_artifacts_to_github:
...
+      artifacts: true
   trigger:
+#    project: DataDog/dd-trace-dotnet-gitlab-test # can be used for testing
     project: DataDog/public-images
     branch: main
     strategy: depend
   variables:
-    IMG_SOURCES: ghcr.io/datadog/dd-trace-java/dd-lib-java-init:$CI_COMMIT_SHA
-    IMG_DESTINATIONS: dd-lib-java-init:latest
+    IMG_SOURCES: ghcr.io/datadog/dd-trace-java/dd-lib-java-init:$CI_COMMIT_TAG
```
andrewlock, line 249: I think this was a typo, the _source_ of the image shouldn't have changed 
```suggestion
    IMG_SOURCES: ghcr.io/datadog/dd-trace-java/dd-lib-java-init:$CI_COMMIT_TAG
```
```suggestion
    IMG_SOURCES: ghcr.io/datadog/dd-trace-java/dd-lib-java-init:$CI_COMMIT_SHA
```

PR #7059, opened 2024-05-22
Prevent AppSec context from being closed more than once on partial flush

smola:
# What Does This Do

Ensure that AppSec hooks for root span finished run only once.

# Motivation

We had some hooks in `CoreTracer.write` meant to be run whenever a root span is finished. However, they were effectively called not just when the root span finished, but also whenever a partial flush on a child span was performed. This ended up calling the hooks multiple teams, and earlier than expected.

When `DD_APPSEC_ENABLED=true`, this led to `WAF object had not been closed` warnings whenever a partial flush was triggered and the root span was not finished yet.

Reproducing the issue is easier when setting `-Ddd.trace.partial.flush.min.spans=1`.

# Additional Notes

Jira ticket: [APPSEC-53203](https://datadoghq.atlassian.net/browse/APPSEC-53203)

[APPSEC-53203]: https://datadoghq.atlassian.net/browse/APPSEC-53203?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ

### DIFF
--- a/dd-trace-core/src/main/java/datadog/trace/core/CoreTracer.java
+++ b/dd-trace-core/src/main/java/datadog/trace/core/CoreTracer.java
@@ -240,9 +240,31 @@ public class CoreTracer implements AgentTracer.TracerAPI {
     return propagationTagsFactory;
   }
 
+  /**
+   * Called when a root span is finished before it is serialized. This is might be called multiple
+   * times per root span. If a child span is part of a partial flush, this method will be called for
+   * its root even if not finished.
+   */
   @Override
   public void onRootSpanFinished(AgentSpan root, EndpointTracker tracker) {
     profilingContextIntegration.onRootSpanFinished(root, tracker);
+  }
+
+  /**
+   * Called when a root span is finished before it is serialized. This is guaranteed to be called
+   * exactly once per root span.
+   */
+  void onRootSpanPublished(final AgentSpan root) {
+    // Request context is propagated to contexts in child spans.
+    // Assume here that if present it will be so starting in the top span.
+    RequestContext requestContext = root.getRequestContext();
+    if (requestContext != null) {
+      try {
+        requestContext.close();
+      } catch (IOException e) {
+        log.warn("Error closing request context data", e);
+      }
+    }
   }
 
   @Override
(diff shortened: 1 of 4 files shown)


dd-trace-core/src/main/java/datadog/trace/core/CoreTracer.java
```diff
@@ -240,11 +240,33 @@ PropagationTags.Factory getPropagationTagsFactory() {
...
     profilingContextIntegration.onRootSpanFinished(root, tracker);
   }
 
+  /**
+   * Called when a root span is finished before it is serialized. This is guaranteed to be called
+   * exactly once per root span.
+   */
+  void onRootSpanPublished(final AgentSpan root) {
+    // Request context is propagated to contexts in child spans.
+    // Assume here that if present it will be so starting inthe top span.
```
manuel-alvarez-alvarez, line 259:
```suggestion
    // Assume here that if present it will be so starting in the top span.
```

manuel-alvarez-alvarez (approved):

jandro996 (approved):

dd-trace-core/src/main/java/datadog/trace/core/CoreTracer.java
```diff
@@ -240,11 +240,33 @@ PropagationTags.Factory getPropagationTagsFactory() {
     return propagationTagsFactory;
   }
 
+  /**
+   * Called when a root span is finished before it is serialized. This is might be called multiple
+   * times per root span. If a child span is part of a partial flush, this method will be called for
+   * its root even if not finished.
+   */
```
PerfectSlayer, line 247: Should this be part of the document in `EndpointCheckpointer` instead if the behavior differs?

dd-trace-core/src/main/java/datadog/trace/core/CoreTracer.java
```diff
@@ -240,11 +240,33 @@ PropagationTags.Factory getPropagationTagsFactory() {
...
   @Override
   public void onRootSpanFinished(AgentSpan root, EndpointTracker tracker) {
     profilingContextIntegration.onRootSpanFinished(root, tracker);
   }
 
+  /**
+   * Called when a root span is finished before it is serialized. This is guaranteed to be called
+   * exactly once per root span.
+   */
+  void onRootSpanPublished(final AgentSpan root) {
```
PerfectSlayer, line 257: Should it be added in `EndpointCheckpointer`interface as it is playing with AppSec context like the two others `OnRootSpanFinished` and `OnRootSpanStarted`?

PerfectSlayer (approved):

ValentinZakharov (approved):

dd-trace-core/src/main/java/datadog/trace/core/CoreTracer.java
smola: @PerfectSlayer Added some docs to the `EndpointCheckpointer` to clarify it will be called multiple times.

PR #7063, opened 2024-05-22
Add config param for local root exception handling

jpbempel:
# What Does This Do
Add `DD_INTERNAL_EXCEPTION_REPLAY_ONLY_LOCAL_ROOT` config option to handle exception by exception debugging only for local root span (Service entry span) or all spans

# Motivation

# Additional Notes

### DIFF
--- a/dd-trace-core/src/main/java/datadog/trace/core/DDSpan.java
+++ b/dd-trace-core/src/main/java/datadog/trace/core/DDSpan.java
@@ -365,11 +365,21 @@ public class DDSpan
 
       setTag(DDTags.ERROR_MSG, message);
       setTag(DDTags.ERROR_TYPE, error.getClass().getName());
-      if (Config.get().isDebuggerEnabled() && isLocalRootSpan()) {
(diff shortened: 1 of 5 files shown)


dd-trace-core/src/main/java/datadog/trace/core/DDSpan.java
```diff
@@ -365,13 +365,23 @@ public DDSpan addThrowable(Throwable error, byte errorPriority) {
...
       setTag(DDTags.ERROR_TYPE, error.getClass().getName());
-      if (Config.get().isDebuggerEnabled() && isLocalRootSpan()) {
+      if (isExceptionDebuggingEnabled()) {
         DebuggerContext.handleException(error, this);
       }
     }
     return this;
   }
 
+  private boolean isExceptionDebuggingEnabled() {
```
shatzi, line 375: nit: I would rename to something like shouldCaptureExceptionOnSpan

shatzi (approved):
nice

PR #7064, opened 2024-05-22
apply configured service name mapping to DBM-injected `dddbs`

### LINKED ISSUE #7034: JDBC connection service mapping not being used, and traceparent missing, in SQL comment
Hello Team!

We've been trying to get our DBM propagation to work in our JVM service. What we've got so far is this:

Environment variables:
`DD_DBM_PROPAGATION_MODE=full`
`DD_INTEGRATION_JDBC_DATASOURCE_ENABLED=true`
`DD_SERVICE_MAPPING=java-aws-sdk:service2-aws,kafka:service2-kafka,postgresql:service1-postgres,redis:service2-redis`

I've changed the name of our services to service1 and service2, but otherwise it's 1:1. 

We're using version `1.34.0` of the `dd-java-agent`. 

In our AWS RDS monitoring I can see that a comment is injected into the SQL that looks like this:
```
/*ddps='service2',dddbs='postgresql',dde='product-production',ddpv='4a388a64e335e580b0a009bde65e5b8ca98c2123'*/
```

As you can see, the `dddbs` is `postgresql` but I am expecting it to be `service1-postgres`. Also there is no `traceparent`

In SQL statements from service1, which is a node service, we can see the following comments injected. 

```
/*dddbs='service1-postgres',dde='product-production',ddps='service1',ddpv='33e69dcfc8b2fedef92a0d5c6b02770245912787',traceparent='00-00000000000000006a12fad2e813a38c-28d6e919161bf07e-01'*/
```

Which works as expected. 

I've tried running the the service with the variable `DD_TRACE_DEBUG=true` and confirmed that the service mapping was picked up: 
```
[dd.trace 2024-05-15 13:26:46:971 +0000] [main] DEBUG datadog.trace.api.Config - New instance: Config{instrumenterConfig=InstrumenterConfig{ [...] serviceMapping={java-aws-sdk=service2-aws, postgresql=service1-postgres, redis=service2-redis, kafka=service2-kafka} [...] }
```

1.  What do I need to do to get the service mapping applied to the JDBC connection? Is it a bug?
2. What do I need to do to get a `tranceparent`? 

Thank you for your time 🙏

vandonr:
# What Does This Do

The injected dddbs tag should respect service mapping.

For single statements, we create the span and inject the comment in the same instrumentation, so everything is well ([code](https://github.com/DataDog/dd-trace-java/blob/ee9c0f803a0e9e33ecc2a8cd250625bf1792c8ad/dd-java-agent/instrumentation/jdbc/src/main/java/datadog/trace/instrumentation/jdbc/StatementInstrumentation.java#L107))
However, for prepared statements, it's more complicated, because it happens in 2 steps:
 - step 1: the call to prepare is instrumented, this is where we inject the comment, it's done in DBMCompatibleConnectionInstrumentation
 - step 2: the call to execute is instrumented, this is when the command is actually run, so this is where we create the span (this is done in AbstractPreparedStatementInstrumentation)
Thus, we don't have the span yet when we do the injection for prepared statements.

Service mapping logic was only applied when a service name is set in a span:
https://github.com/DataDog/dd-trace-java/blob/ad66ae831ff35f39b40065180db609c003dda470/dd-trace-core/src/main/java/datadog/trace/core/DDSpanContext.java#L398-L401
so if we get a service name _not_ from a span, the mapping wouldn't be applied to it.

To fix this, I'm repeating the logic that is found here:
https://github.com/DataDog/dd-trace-java/blob/0779b79b4acb05a10acf47b1ac1570a27f2c1df7/dd-trace-core/src/main/java/datadog/trace/core/PendingTrace.java#L173

@mcculls helped with the code to retrieve the current trace's config snapshot.

Jira ticket: [APMS-12191](https://datadoghq.atlassian.net/browse/APMS-12191)

[APMS-12191]: https://datadoghq.atlassian.net/browse/APMS-12191?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ

fixes #7034


amarziali:
Would it be possible to add tests?

vandonr:
> Would it be possible to add tests?

yes I'll look into it. For now, it seems there _might_ be some tests failing, that I cannot reproduce locally :(
I was hoping to draw inspiration from there.

dd-java-agent/instrumentation/jdbc/src/test/groovy/DBMInjectionForkedTest.groovy
```diff
@@ -0,0 +1,57 @@
...
+  @Override
+  void cleanupAfterAgent() {
+    removeSysConfig(TraceInstrumentationConfig.DB_DBM_PROPAGATION_MODE_MODE)
+    removeSysConfig(TracerConfig.SERVICE_MAPPING)
+    removeSysConfig("dd.trace.jdbc.prepared.statement.class.name")
+    removeSysConfig("dd.trace.jdbc.connection.class.name")
+  }
+
+  static query = "SELECT 1"
+  static serviceInjection = "ddps='worker.org.gradle.process.internal.worker.GradleWorkerMain',dddbs='remapped_testdb'"
```
amarziali, line 30: I would set a service name (`injectSysConfig("service.name", "something") and rely on this name rather using the Gradle canonical class name (that can change in the future)

dd-java-agent/instrumentation/jdbc/src/main/java/datadog/trace/instrumentation/jdbc/DBMCompatibleConnectionInstrumentation.java
```diff
@@ -116,22 +120,21 @@ public static String onEnter(
         final DBInfo dbInfo =
             JDBCDecorator.parseDBInfo(
                 connection, InstrumentationContext.get(Connection.class, DBInfo.class));
+        String dbService = DECORATE.getDbService(dbInfo);
+        if (dbService != null) {
+          dbService =
+              AgentTracer.traceConfig(activeSpan())
```
amarziali, line 126: `traceConfig` can be statically imported for readability

dd-java-agent/instrumentation/jdbc/src/test/groovy/DBMInjectionForkedTest.groovy
```diff
@@ -0,0 +1,57 @@
...
+
+    injectSysConfig(TraceInstrumentationConfig.DB_DBM_PROPAGATION_MODE_MODE, "full")
+    // to check that we use the remapped service name in dddbs
+    injectSysConfig(TracerConfig.SERVICE_MAPPING, "testdb:remapped_testdb")
+    injectSysConfig("dd.trace.jdbc.prepared.statement.class.name", "test.TestPreparedStatement")
+    injectSysConfig("dd.trace.jdbc.connection.class.name", "test.TestConnection")
+  }
+
+  @Override
+  void cleanupAfterAgent() {
```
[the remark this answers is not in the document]
vandonr: ah yes I added it was not a forked test yet and I was trying to avoid side effects

PR #7066, opened 2024-05-23
Fix the cws tracer on systems using musl libc (like alpine)

spikat:
# What Does This Do

This fixes the CWS on systems using musl, which doesn't expose the `gettid()` syscall. We replace it instead by directly calling syscall with the related syscall ID.

# Motivation

Be able to retrieve trace spans on docker images build upon alpine.

# Additional Notes

Only AMD64 and ARM64 archs are handled.

Jira ticket: [PROJ-IDENT]

### DIFF
--- a/dd-java-agent/cws-tls/src/main/java/datadog/cws/tls/ErpcTls.java
+++ b/dd-java-agent/cws-tls/src/main/java/datadog/cws/tls/ErpcTls.java
@@ -39,7 +39,7 @@ public class ErpcTls implements Tls {
   static int getGettidSyscallId() {
     String arch = System.getProperty("os.arch");
     if (arch.equals("amd64")) {
-        return 186; // 186 is the syscall ID for "gettid" on amd64
+      return 186; // 186 is the syscall ID for "gettid" on amd64
     } else if (arch.equals("arm64")) {
       return 178; // 78 is the syscall ID for "gettid" on arm64
     }
@@ -52,7 +52,7 @@ public class ErpcTls implements Tls {
       return false;
     }
     try {
-        CLibrary.Instance.syscall(new NativeLong(syscallID));
+      CLibrary.Instance.syscall(new NativeLong(syscallID));
     } catch (UnsatisfiedLinkError error) {
       return false;
     }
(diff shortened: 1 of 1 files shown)


dd-java-agent/cws-tls/src/main/java/datadog/cws/tls/ErpcTls.java
```diff
@@ -26,18 +26,27 @@ public class ErpcTls implements Tls {
...
+    if (!arch.equals("amd64") && !arch.equals("arm64")) {
+      return false;
+    }
     try {
-      CLibrary.Instance.gettid();
+      if (arch.equals("amd64")) {
+        CLibrary.Instance.syscall(new NativeLong(186L));
+      } else {
+        CLibrary.Instance.syscall(new NativeLong(178L));
+      }
```
PerfectSlayer, line 49: Would you mind having a dedicated function to get the syscall according the arch? It's a duplicated logic on lines 70+ too.

And could you define/document the magic numbers too for maintainability. Thanks!

PerfectSlayer (commented):
Hey @spikat  👋 
Only few minor comments

By the way, how CWS relates to IAST and WAF (if ever)? Because I have no proper label for it.
It stands for "Cloud Workload Security", right?

spikat:
> By the way, how CWS relates to IAST and WAF (if ever)? Because I have no proper label for it. It stands for "Cloud Workload Security", right?

Hi @PerfectSlayer. Exactly, CWS is "Cloud Workload Security", and is part of the CSM (Cloud Security Management). And AFAIK, for instance the only bridge we want to build is bringing APM traces (if available) to enrich CWS logs/signals

PerfectSlayer (approved):
LGTM!
You have formatting issue so you will need to run `./gradlew :dd-java-agent:cws-tls:spotlessApply`.
You can install the hook that runs it automatically: https://github.com/DataDog/dd-trace-java/blob/master/BUILDING.md#clone-the-repository-and-setup-git

spikat:
> LGTM! You have formatting issue so you will need to run `./gradlew :dd-java-agent:cws-tls:spotlessApply`. You can install the hook that runs it automatically: https://github.com/DataDog/dd-trace-java/blob/master/BUILDING.md#clone-the-repository-and-setup-git

Thanks for your review! And sorry for the formating issue, it's fixed now. :smiley: Thanks for the links :bow: 

> For the label, I will create the CWS one. We can edit it later if more features become available 😉

That's perfect! :+1:

Issue #7024, opened 2024-05-14
Gradle smoke tests are flaky


PerfectSlayer:
### Test suite

1. `test test-succeed-multi-module-legacy-instrumentation, v7.6.4, configCache: false`
2. `test test-succeed-multi-forks-legacy-instrumentation, v7.6.4, configCache: false`
3. `test test-failed-legacy-instrumentation, v7.6.4, configCache: false`
4. `test test-succeed-legacy-instrumentation, v7.6.4, configCache: false`

From `datadog.smoketest.GradleDaemonSmokeTest`

### Error

```
Condition not satisfied after 15.00 seconds and 40 attempts

Condition not satisfied after 15.00 seconds and 40 attempts
	at spock.util.concurrent.PollingConditions.within(PollingConditions.java:185)
	at spock.util.concurrent.PollingConditions.eventually(PollingConditions.java:140)
	at datadog.trace.civisibility.CiVisibilitySmokeTest.waitForCoverages(CiVisibilitySmokeTest.groovy:236)
	at datadog.trace.civisibility.CiVisibilitySmokeTest.verifyEventsAndCoverages(CiVisibilitySmokeTest.groovy:177)
	at datadog.smoketest.GradleDaemonSmokeTest.test #projectName, v#gradleVersion, configCache: #configurationCache(GradleDaemonSmokeTest.groovy:78)
Caused by: Condition not satisfied:

receivedCoverages.size() == traceSize
|                 |      |  |
[]                0      |  2
                         false

	at datadog.trace.civisibility.CiVisibilitySmokeTest.waitForCoverages_closure2(CiVisibilitySmokeTest.groovy:237)
	at datadog.trace.civisibility.CiVisibilitySmokeTest.waitForCoverages_closure2(CiVisibilitySmokeTest.groovy)
	at groovy.lang.Closure.call(Closure.java:412)
	at spock.util.concurrent.PollingConditions.within(PollingConditions.java:168)
	... 4 more
```

### Environment

It seems to happens exclusively when testing gradle v7.6.4 on Semuru 17 JVM.

PR #7087, opened 2024-05-29
Fix flakiness in Maven and Gradle smoke tests

nikita-tkachenko-datadog:
# What Does This Do
Fixes logic that waits for coverage messages in Gradle and Maven smoke tests: now coverage records are counted instead of coverage requests, so asserts are not affected by batching in the tracer.

Disables smoke tests for older Gradle versions when testing with OpenJ9: Gradle Jacoco plugin does not work with this combination, and neither does the coverage solution built on top of Jacoco.

Gradle and Maven builds triggered by the smoke tests now run with tracer debug logging to simplify debugging any future failures.

"Flaky" annotation is removed from Gradle and Maven smoke tests.

# Motivation
Re-enabling Gradle and Maven instrumentations testing.

Jira ticket: [SDTEST-384]

[SDTEST-384]: https://datadoghq.atlassian.net/browse/SDTEST-384?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ

### DIFF
--- a/dd-smoke-tests/gradle/src/test/groovy/datadog/smoketest/GradleDaemonSmokeTest.groovy
+++ b/dd-smoke-tests/gradle/src/test/groovy/datadog/smoketest/GradleDaemonSmokeTest.groovy
@@ -7,7 +7,6 @@ import datadog.trace.api.Platform
 import datadog.trace.api.config.CiVisibilityConfig
 import datadog.trace.api.config.GeneralConfig
 import datadog.trace.civisibility.CiVisibilitySmokeTest
-import datadog.trace.test.util.Flaky
 import datadog.trace.util.Strings
 import okhttp3.OkHttpClient
 import okhttp3.Request
@@ -73,7 +72,6 @@ class GradleDaemonSmokeTest extends CiVisibilitySmokeTest {
     mockBackend.reset()
   }
 
-  @Flaky("https://github.com/DataDog/dd-trace-java/issues/7024")
   def "test #projectName, v#gradleVersion, configCache: #configurationCache"() {
     givenGradleVersionIsCompatibleWithCurrentJvm(gradleVersion)
     givenConfigurationCacheIsCompatibleWithCurrentPlatform(configurationCache)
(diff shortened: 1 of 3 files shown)


dd-smoke-tests/gradle/src/test/groovy/datadog/smoketest/GradleDaemonSmokeTest.groovy
```diff
@@ -140,6 +150,7 @@ class GradleDaemonSmokeTest extends CiVisibilitySmokeTest {
       "-javaagent:${agentShadowJar}=" +
       // for convenience when debugging locally
       (System.getenv("DD_CIVISIBILITY_SMOKETEST_DEBUG_CHILD") != null ? "${Strings.propertyNameToSystemPropertyName(CiVisibilityConfig.CIVISIBILITY_DEBUG_PORT)}=5055," : "") +
+      "dd.trace.debug=true," +
```
PerfectSlayer, line 153:
```suggestion
     "${Strings.propertyNameToSystemPropertyName(GeneralConfig.TRACE_DEBUG)}=true," +
```

dd-smoke-tests/maven/src/test/groovy/datadog/smoketest/MavenSmokeTest.groovy
```diff
@@ -234,6 +231,7 @@ class MavenSmokeTest extends CiVisibilitySmokeTest {
       def agentArgument = "-javaagent:${agentShadowJar}=" +
         // for convenience when debugging locally
         (System.getenv("DD_CIVISIBILITY_SMOKETEST_DEBUG_CHILD") != null ? "${Strings.propertyNameToSystemPropertyName(CiVisibilityConfig.CIVISIBILITY_DEBUG_PORT)}=5055," : "") +
+        "dd.trace.debug=true," +
```
PerfectSlayer, line 234:
```suggestion
        "${Strings.propertyNameToSystemPropertyName(GeneralConfig.TRACE_DEBUG)}=true," +
```

PerfectSlayer (approved):
Looks good to me! Thanks for the improvements 👍 
Only minor comments about debug flag.

I will also tag #7024 for the next release then.

PR #7092, opened 2024-05-30
Report exception when deserializing config

jpbempel:
# What Does This Do
Send probe diagnostic messages when an error occurred during config deserialization (compatibility issue, unsupported feature, malformed json, parsing bug)

# Motivation
Report error to the user

# Additional Notes

Jira ticket: [DEBUG-2418]

[DEBUG-2418]: https://datadoghq.atlassian.net/browse/DEBUG-2418?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ


shatzi (approved):
should we also want to write error this to telemetry?

jpbempel:
> should we also want to write error this to telemetry?

because we are logging the exception with:
`LOGGER.warn("Error handling probe configuration: {}", configId, ex);` it is also send to telemetry

PR #7117, opened 2024-06-03
Remove too generic redaction keywords

jpbempel:
# What Does This Do

# Motivation
config and address are too generic and can redact useful and not sensible variables

# Additional Notes


OmerRaviv (approved):
LGTM

PR #7132, opened 2024-06-06
SpanId in reported AppSec events

ValentinZakharov:
# What Does This Do
Added `spanId` in AppSec events. It points to the span where the AppSec detected malicious data.

# Motivation
This is preparation work for stack trace reporting in Exploit prevention initiative (RASP)

# Additional Notes
image

Jira ticket: [APPSEC-46818]

[APPSEC-46818]: https://datadoghq.atlassian.net/browse/APPSEC-46818?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ

### DIFF
--- a/dd-java-agent/appsec/src/main/java/com/datadog/appsec/powerwaf/PowerWAFModule.java
+++ b/dd-java-agent/appsec/src/main/java/com/datadog/appsec/powerwaf/PowerWAFModule.java
@@ -27,6 +27,8 @@ import datadog.trace.api.ProductActivation;
 import datadog.trace.api.gateway.Flow;
 import datadog.trace.api.telemetry.LogCollector;
 import datadog.trace.api.telemetry.WafMetricCollector;
+import datadog.trace.bootstrap.instrumentation.api.AgentSpan;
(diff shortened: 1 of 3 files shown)


dd-java-agent/appsec/src/main/java/com/datadog/appsec/powerwaf/PowerWAFModule.java
```diff
@@ -580,6 +582,12 @@ private AppSecEvent buildEvent(PowerWAFResultData wafResult) {
       ruleMatchList.add(ruleMatch);
     }
 
+    Long spanId = null;
+    AgentSpan agentSpan = AgentTracer.get().activeSpan();
```
smola, line 586: Not for this PR, because it's a far wider issue, but the fact that we need this call here in a context where we have an `AppSecRequestContext` in scope is not a good sign. That means that we had the active span just a bit higher in the call stack, and it could have been propagated.

smola (approved):

PR #7135, opened 2024-06-07
Use three modes for auto user id collection: identification (default), anonymization and disabled

manuel-alvarez-alvarez:
# What Does This Do
Applies the new erratum/addendum of automated user events trancking. The new RFC defines three different modes based on the env `DD_APPSEC_AUTO_USER_INSTRUMENTATION_MODE`

- `identification` (by default) user ids are sent the backend without any modification
- `anonymization` user ids are hashed with SHA256
- `disabled` auto user event tracking is disabled

This new option deprecates `DD_APPSEC_AUTOMATED_USER_EVENTS_TRACKING`, the deprecated option will be automatically converted to the new schema following the next table:

| DD_APPSEC_AUTOMATED_USER_EVENTS_TRACKING |  DD_APPSEC_AUTO_USER_INSTRUMENTATION_MODE |
|---|---|
|`extended`|`identification`|
|`safe`|`anonymization`|
|`disabled`|`disabled`|

# Motivation
Tracking of real user ids is very important for the detection of account take over campaigns, this PR ensures that the backend has all required information in order to provide adequate protection.

# Additional Notes

Jira ticket: [APPSEC-53547]

RFC: [link](https://docs.google.com/document/d/19VHLdJLVFwRb_JrE87fmlIM5CL5LdOBv4AmLxgdo9qI/edit?pli=1#heading=h.hidcz5iyug2k)

[APPSEC-53547]: https://datadoghq.atlassian.net/browse/APPSEC-53547?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ

### DIFF
--- a/dd-java-agent/agent-bootstrap/src/main/java/datadog/trace/bootstrap/instrumentation/decorator/AppSecUserEventDecorator.java
+++ b/dd-java-agent/agent-bootstrap/src/main/java/datadog/trace/bootstrap/instrumentation/decorator/AppSecUserEventDecorator.java
@@ -23,7 +23,7 @@ public class AppSecUserEventDecorator {
   private static final Logger LOGGER = LoggerFactory.getLogger(AppSecUserEventDecorator.class);
   private static final int HASH_SIZE_BYTES = 16; // 128 bits
   private static final String ANONYM_PREFIX = "anon_";
-  private static AtomicBoolean SHA_MISSING_REPORTED = new AtomicBoolean(false);
+  private static final AtomicBoolean SHA_MISSING_REPORTED = new AtomicBoolean(false);
 
   public boolean isEnabled() {
     if (!ActiveSubsystems.APPSEC_ACTIVE) {
@@ -36,10 +36,7 @@ public class AppSecUserEventDecorator {
   }
 
   public void onUserNotFound() {
-    if (!isEnabled()) {
-      return;
-    }
-    TraceSegment segment = getSegment();
+    TraceSegment segment = beforeEvent();
     if (segment == null) {
       return;
     }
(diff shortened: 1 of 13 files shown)


dd-trace-api/src/main/java/datadog/trace/api/internal/util/StringCodecUtils.java
```diff
@@ -0,0 +1,20 @@
+package datadog.trace.api.internal.util;
+
+import static datadog.trace.api.internal.util.LongStringUtils.HEX_DIGITS;
+import static java.nio.charset.StandardCharsets.US_ASCII;
+
+public class StringCodecUtils {
```
smola, line 6: We already have `./internal-api/src/main/java/datadog/trace/util/Strings.java`. 

That one is not visible from the required modules?

manuel-alvarez-alvarez: I've moved it there, it really does make more sense. I was trying to reuse the hex chars from another class but not a problem.

dd-trace-api/src/main/java/datadog/trace/api/config/AppSecConfig.java
```diff
@@ -21,6 +21,8 @@ public final class AppSecConfig {
       "appsec.http.blocked.template.json";
   public static final String APPSEC_AUTOMATED_USER_EVENTS_TRACKING =
       "appsec.automated-user-events-tracking";
+  public static final String APPSEC_AUTO_USER_INSTRUMENTATION_MODE =
```
smola, line 24: `appsec.automated-user-events-tracking` vs `appsec.auto-user-instrumentation-mode`... a bit unfortunate the usage of different prefix (`automated` vs `auto` here). Not blocking the PR since I assume this comes from the spec, but worth discussing it in the relevant working group.

smola (approved):

jandro996 (approved):

dd-trace-api/src/main/java/datadog/trace/api/config/AppSecConfig.java
manuel-alvarez-alvarez: Yep, this is coming straight from the [RFC](https://docs.google.com/document/d/19VHLdJLVFwRb_JrE87fmlIM5CL5LdOBv4AmLxgdo9qI/edit?pli=1)

dd-java-agent/agent-bootstrap/src/main/java/datadog/trace/bootstrap/instrumentation/decorator/AppSecUserEventDecorator.java
```diff
@@ -89,19 +107,56 @@ private void onEvent(@Nonnull TraceSegment segment, String eventName, Map<String
...
+    if (segment.getTagTop(tagName) != null) {
+      // do not override user ids set by the SDK
       return;
     }
-    UserEventTrackingMode mode = Config.get().getAppSecUserEventsTrackingMode();
-    if (mode == SAFE && !SAFE_USER_ID_PATTERN.matcher(userId).matches()) {
-      // do not set the user id if not numeric or UUID
-      return;
+    String finalUserId =
+        getUserIdCollectionMode() == UserIdCollectionMode.ANONYMIZATION
```
ValentinZakharov, line 117: Shouldn't we also check for disabled mode at the beginning of this function? 🤔

ValentinZakharov: Or something like:
```
String finalUserId;
switch (getUserIdCollectionMode()) {
    case ANONYMIZATION:
        finalUserId = anonymize(userId);
        segment.setTagTop(tagName, finalUserId);
        break;
    case IDENTIFICATION:
        finalUserId = userId;
        segment.setTagTop(tagName, finalUserId);
        break;
    case DISABLED:
        break;
    default:
        LOGGER.error("Unsupported UserIdCollectionMode " + getUserIdCollectionMode());
}
```

internal-api/src/main/java/datadog/trace/api/telemetry/WafMetricCollector.java
```diff
@@ -174,6 +179,11 @@ public void prepareMetrics() {
         }
       }
     }
+
+    // Missing user id
+    if (missingUserIdCounter.get() > 0) {
+      rawMetricsQueue.offer(new MissingUserIdMetric(missingUserIdCounter.getAndReset()));
```
ValentinZakharov, line 185: Can we, check the result of `offer(...)` call to interrupt method in case of queue overflow?

dd-java-agent/agent-bootstrap/src/main/java/datadog/trace/bootstrap/instrumentation/decorator/AppSecUserEventDecorator.java
manuel-alvarez-alvarez: Nope, all methods from the class are short circuit via:
```java
  public boolean isEnabled() {
    if (!ActiveSubsystems.APPSEC_ACTIVE) {
      return false;
    }
    if (getUserIdCollectionMode() == UserIdCollectionMode.DISABLED) {
      return false;
    }
    return true;
  }
```

internal-api/src/main/java/datadog/trace/api/telemetry/WafMetricCollector.java
manuel-alvarez-alvarez: Yep, done, didn't do it before as it was the last statement, thanks!

ValentinZakharov (approved):

PR #7148, opened 2024-06-10
Avoid publishing RC to Sonatype and upload the right assets to GitHub release

PerfectSlayer:
# What Does This Do

This PR tries to fix two issues of the GitLab CI:
* Avoid publishing release candidates to Sonatype,
* Upload the same artifacts we send to Sonatype as GitHub release assets

# Motivation

Our release process is broken. We ship three different binaires for the same version.
If this works, Maven central and GitHub release should be the same. Only lib-injection will be wrong.

# Additional Notes

I could not test the pipeline, so it will be test as part of the next release.

Added @randomanderson for review as he might have a better understanding of the whole pipeline and [the GitLab workflow syntax](https://docs.gitlab.com/ee/ci/yaml/#needs).

Jira ticket: [PROJ-IDENT]

### DIFF
--- a/.gitlab-ci.yml
+++ b/.gitlab-ci.yml
@@ -184,7 +184,8 @@ deploy_to_sonatype:
   rules:
     - if: '$CI_COMMIT_BRANCH == "master"'
       when: on_success
-    - if: '$CI_COMMIT_TAG =~ /^v.*/'
+    # Do not deploy release candidate versions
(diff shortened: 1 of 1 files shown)


randomanderson (approved):

PerfectSlayer:
Thanks for the review Laplie 👍

PR #7150, opened 2024-06-10
Do not modify readonly message attributes map on SNS integration

vandonr:
# What Does This Do

Fix a bug where the integration would crash in the sdk v1 integration if the message attributes were a readonly map
Also fix a minor bug in the sdk v2 batch requests where we'd write more than 10 message attributes, which is not supported in SQS.

# Motivation

#7147


mcculls (approved):

amarziali (approved):

ygree:
@codex review

vandonr:
> @codex review

did you intend to call for a review on your PR ? Or was it to see if codex would have caught the bug you fixed ?

PR #7151, opened 2024-06-10
Use unified Gitlab pipeline for APM libraries

randomanderson:
# What Does This Do

Converts the Gitlab workflow to use the unified pipeline.

# Motivation

By using the unified pipeline across all APM libraries, there is less maintenance hassle, fewer mistakes, and a more uniform creation of artifacts.

# Additional Notes

* Requires:  https://github.com/DataDog/libdatadog-build/pull/5
* "Gitlab Stages" as a concept just doesn't work anymore. Everything needs explicit dependencies and job execution order is wholly specified by those dependencies

Jira ticket: APMSP-1184


PerfectSlayer (approved):
It looks good (hard to tell how if everything will click together right the first time just by looking at the diff).

Quick question about lib-injection: if the build is made by the unified pipeline, can this same pipeline handle the pruning too? 
I had to put one in place to keep things manageable and documented it here:
https://github.com/DataDog/dd-trace-java/tree/master/.github/workflows#lib-injection-prune-registry-

randomanderson:
@PerfectSlayer agreed. So much is removed from `.gitlab-ci.yml` that the diff is mostly useless.

For the pruning workflow, I'm going to take a second look once most libraries are onboarded to see if we can do something uniformly later. I updated the Readme to remove the two workflows removed with this PR

PR #7160, opened 2024-06-11
Start Datadog appender when doing agentless log submission for Log4j2

nikita-tkachenko-datadog:
# What Does This Do
Fixes agentless log submission for Log4j2: some Log4j versions require appenders to be started explicitly before they can be used. Otherwise log messages forwarded to an appender that is not started are ignored.

Also adds a check that prevents multiple Datadog appenders from being added to the same logging config.

Jira ticket: [CIVIS-10104]

[CIVIS-10104]: https://datadoghq.atlassian.net/browse/CIVIS-10104?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ


amarziali (approved):
looks good

PR #7186, opened 2024-06-13
Full mode for SQL Server

nenadnoveljic:
# What Does This Do

Add full APM/DBM mode for SQL Server.

# Motivation

A large demand.

# Additional Notes

Jira ticket: [PROJ-IDENT]


dd-java-agent/instrumentation/jdbc/src/main/java/datadog/trace/instrumentation/jdbc/StatementInstrumentation.java
```diff
@@ -112,6 +113,15 @@ public static AgentScope onEnter(
                   appendComment);
         }
         DECORATE.onStatement(span, copy);
+
+        Statement instrumentationStatement = connection.createStatement();
+        instrumentationStatement.execute(
+            "set context_info 0x"
+                + span.getTraceId().toHexString()
+                + DDSpanId.toHexStringPadded(span.getSpanId())
+                + "0");
```
jmeunier28, line 122: is the appened `0` suppose to be the sampling decision? we actually force a sampling decision here https://github.com/DataDog/dd-trace-java/pull/7186/files#diff-7b6077517113067ace610d82b5399250c601c78d97da30e2993868c3fad647d8L96 so this value will be correct

nenadnoveljic: `0` in the end is the formatting version. We assume that the format might change over time, so it makes sense to reserve 4 bits for it.

jmeunier28: ah gotcha, are you planning to pass the sampling decision? we currently do that for all of the other apm-dbm link implementations

nenadnoveljic: I don't now what that is - I haven't looked into that yet. What do you think? If we know that we are enforcing it on the tracer, why do we need to transmit it?

.githooks/pre-commit
vandonr, line 1: if you removed this because you didn't want to run the long pre-commit check, know that you can also use the `git commit -n` option to skip it :)

dd-java-agent/instrumentation/jdbc/src/main/java/datadog/trace/instrumentation/jdbc/SQLServerPreparedStatementExcecution.java
```diff
@@ -0,0 +1,96 @@
...
+  }
+
+  @Override
+  public String hierarchyMarkerType() {
+    return null; // bootstrap type
+  }
+
+  @Override
+  public ElementMatcher<TypeDescription> hierarchyMatcher() {
+    return implementsInterface(named("java.sql.PreparedStatement"));
```
vandonr, line 46: I there a way we can scope this instrumentation to SQL Server Prepared statements only ?

dd-java-agent/instrumentation/jdbc/src/main/java/datadog/trace/instrumentation/jdbc/SQLServerPreparedStatementExcecution.java
```diff
@@ -0,0 +1,96 @@
...
+                + span.getTraceId().toHexString()
+                + DDSpanId.toHexStringPadded(span.getSpanId())
+                + "0");
+        instrumentationStatement.close();
+
+        return activateSpan(span);
+      } catch (SQLException e) {
+        // if we can't get the connection for any reason
+        return null;
+      }
```
vandonr, line 80: I don't think you need this, because of the `@Advice.OnMethodEnter(suppress = Throwable.class)` annotation which is going to catch all exceptions anyway.

dd-java-agent/instrumentation/jdbc/src/main/java/datadog/trace/instrumentation/jdbc/StatementInstrumentation.java
vandonr: The version is usually at the beginning, which allows to parse the data linearly, instead of having to go at the end of the buffer to lookup the version, and then come back to the beginning to read the data.
Also, I don't think it helps a lot to know the sampling decision in this injection, since there is no use for it currently.

dd-java-agent/instrumentation/jdbc/src/main/java/datadog/trace/instrumentation/jdbc/SQLServerPreparedStatementExcecution.java
nenadnoveljic: Yes. Currently we are testing the interface name `PreparedStatement`, which is the same for all DBMSs. But we could test for the name of the SQL Server concrete class, which is `SQLServerPreparedStatement`.

dd-java-agent/instrumentation/jdbc/src/main/java/datadog/trace/instrumentation/jdbc/StatementInstrumentation.java
nenadnoveljic: > The version is usually at the beginning, which allows to parse the data linearly, instead of having to go at the end of the buffer to lookup the version, and then come back to the beginning to read the data. 

OK. Let's revert it then. Starting from the most significant bits:
1. version (4 bits)
2. span_id (64 bits)
3. trace_id (64 or 128 bits).

Placing `trace_id` at the end is beneficial because we get additional 64 bits for storing another information, such as `environment` or `calling_service`, if `trace_id` is only 64 bits.

> Also, I don't think it helps a lot to know the sampling decision in this injection, since there is no use for it currently.

Let's skip it then.

dd-java-agent/instrumentation/jdbc/src/main/java/datadog/trace/instrumentation/jdbc/AbstractPreparedStatementInstrumentation.java
```diff
@@ -65,30 +66,31 @@ public static AgentScope onEnter(@Advice.This final Statement statement) {
...
             InstrumentationContext.get(Statement.class, DBQueryInfo.class).get(statement);
         if (null == queryInfo) {
           logMissingQueryInfo(statement);
           return null;
         }
+        final AgentSpan span;
+        final DBInfo dbInfo =
+                JDBCDecorator.parseDBInfo(
+                        connection, InstrumentationContext.get(Connection.class, DBInfo.class));
+        final boolean isSqlServer = dbInfo.getType().equals("sqlserver");
```
PerfectSlayer, line 80: Can you do this instead to avoid possible NPE?
```java
"sqlserver".equals(dbInfo.getType());
```

dd-java-agent/instrumentation/jdbc/src/main/java/datadog/trace/instrumentation/jdbc/JDBCDecorator.java
```diff
@@ -251,7 +251,8 @@ public String traceParent(AgentSpan span, int samplingPriority) {
    * @param connection The same connection as the one that will be used for the actual statement
    * @param spanID The ID of the span covering the actual statement
    */
-  public void setContextInfo(Connection connection, long spanID, DBInfo dbInfo) {
+  public long setContextInfo(Connection connection, DBInfo dbInfo) {
```
PerfectSlayer, line 254: It miss the `@return` documentation

dd-java-agent/instrumentation/jdbc/src/main/java/datadog/trace/instrumentation/jdbc/StatementInstrumentation.java
```diff
@@ -86,8 +87,17 @@ public static AgentScope onEnter(
                 connection, InstrumentationContext.get(Connection.class, DBInfo.class));
         boolean injectTraceContext = DECORATE.shouldInjectTraceContext(dbInfo);
         boolean isSqlServer = dbInfo.getType().equals("sqlserver");
+        final long spanID;
```
PerfectSlayer, line 90: `spanID` scope can be reduced to the if statement

dd-java-agent/instrumentation/jdbc/src/main/java/datadog/trace/instrumentation/jdbc/AbstractPreparedStatementInstrumentation.java
nenadnoveljic: done, thanks!

dd-java-agent/instrumentation/jdbc/src/main/java/datadog/trace/instrumentation/jdbc/StatementInstrumentation.java
nenadnoveljic: done, thanks!

dd-java-agent/instrumentation/jdbc/src/main/java/datadog/trace/instrumentation/jdbc/JDBCDecorator.java
```diff
@@ -258,7 +259,10 @@ public boolean isSqlServer(final DBInfo dbInfo) {
    */
   public long setContextInfo(Connection connection, DBInfo dbInfo) {
     final long spanID = Config.get().getIdGenerationStrategy().generateSpanId();
-    AgentSpan instrumentationSpan = startSpan("set context_info");
+    AgentSpan instrumentationSpan = AgentTracer.get()
+            .buildSpan("set context_info")
+            .withTag("_dd.instrumentation", "true")
```
vandonr, line 264: I'll let you check in your demo env, but I think that tags that are in `_dd` are hidden by default, so this wouldn't be very helpful as "documentation"

nenadnoveljic: Changed now, thanks!

dd-java-agent/instrumentation/jdbc/src/main/java/datadog/trace/instrumentation/jdbc/AbstractPreparedStatementInstrumentation.java
```diff
@@ -64,21 +66,29 @@ public static AgentScope onEnter(@Advice.This final Statement statement) {
...
-        final AgentSpan span = startSpan(DATABASE_QUERY);
-        DECORATE.afterStart(span);
-        DBInfo dbInfo =
+        final AgentSpan span;
+        final DBInfo dbInfo =
             JDBCDecorator.parseDBInfo(
                 connection, InstrumentationContext.get(Connection.class, DBInfo.class));
+        final boolean injectTraceContext = DECORATE.shouldInjectTraceContext(dbInfo);
+
+        if (DECORATE.isSqlServer(dbInfo) && INJECT_COMMENT && injectTraceContext) {
```
amarziali, line 82: nit: you can do `INJECT_COMMENT && injectTraceContext && DECORATE.isSqlServer(dbInfo)` since the first two are already calculated. This will probably save a string compare most of the time

dd-java-agent/instrumentation/jdbc/src/main/java/datadog/trace/instrumentation/jdbc/JDBCDecorator.java
```diff
@@ -235,6 +238,59 @@ public String traceParent(AgentSpan span, int samplingPriority) {
...
+      String instrumentationSql =
+          "set context_info 0x"
+              + samplingDecision
+              + DDSpanId.toHexStringPadded(spanID)
+              + instrumentationSpan.getTraceId().toHexString();
+      DECORATE.onStatement(instrumentationSpan, instrumentationSql);
+      instrumentationStatement.execute(instrumentationSql);
+      instrumentationStatement.close();
+    } catch (SQLException e) {
+      log.warn(
```
amarziali, line 281: I would avoid to set it as warn. Instead I'd would log as debug and rate limited

dd-java-agent/instrumentation/jdbc/src/main/java/datadog/trace/instrumentation/jdbc/JDBCDecorator.java
```diff
@@ -235,6 +238,59 @@ public String traceParent(AgentSpan span, int samplingPriority) {
...
+          "Failed to set extra DBM data in context info for trace {}. "
+              + "To disable this behavior, set DBM_PROPAGATION_MODE to 'service' mode. "
+              + "See https://docs.datadoghq.com/database_monitoring/connect_dbm_and_apm/ for more info.\n{}",
```
amarziali, line 284: the last `{}` is not needed

dd-java-agent/instrumentation/jdbc/src/test/groovy/SQLServerTest.groovy
```diff
@@ -0,0 +1,34 @@
...
+    def metadata = new TestDatabaseMetaData()
+    metadata.setURL("jdbc:microsoft:sqlserver://;")
+    connection.setMetaData(metadata)
+
+    when:
+    def statement = connection.createStatement() as TestStatement
+    statement.executeQuery(query)
+
+    then:
+    assert statement.sql == "/*${serviceInjection}*/ ${query}"
```
amarziali, line 32: Would it be possible to add assertions for the spans created as well?

dd-java-agent/instrumentation/jdbc/src/main/java/datadog/trace/instrumentation/jdbc/AbstractPreparedStatementInstrumentation.java
nenadnoveljic: done, thanks!

dd-java-agent/instrumentation/jdbc/src/main/java/datadog/trace/instrumentation/jdbc/JDBCDecorator.java
nenadnoveljic: Converted to `debug` now. What technique do you recommend for rate limiting?

dd-java-agent/instrumentation/jdbc/src/main/java/datadog/trace/instrumentation/jdbc/JDBCDecorator.java
nenadnoveljic: removed.

dd-java-agent/instrumentation/jdbc/src/main/java/datadog/trace/instrumentation/jdbc/JDBCDecorator.java
amarziali: There is something that's doing it `datadog.trace.relocate.api.RatelimitedLogger` but it works for the warn level. It can be perhaps changed to rate limit debugging also. My reason suggesting to avoid warn as level is because it might be interpret by the customer as an issue of their application while it's a message internal of our tracer. Also, an application might do sql query very often. If there is an issue, logging each time will plummet the performances a lot

dd-java-agent/instrumentation/jdbc/src/test/groovy/SQLServerTest.groovy
nenadnoveljic: `span_id` is created in [setContextInfo](https://github.com/DataDog/dd-trace-java/blob/f21c7dfc84c18e5257549d359201023ca062de95/dd-java-agent/instrumentation/jdbc/src/main/java/datadog/trace/instrumentation/jdbc/JDBCDecorator.java#L259) and then started in [StatementAdvisor](https://github.com/DataDog/dd-trace-java/blob/f21c7dfc84c18e5257549d359201023ca062de95/dd-java-agent/instrumentation/jdbc/src/main/java/datadog/trace/instrumentation/jdbc/StatementInstrumentation.java#L94). I don't know how to intercept the call and get the value in any of those places. Unlike in other DBs, `span_id/trace_id` aren't in the comment.

amarziali: I was thinking at using trace assert to test that the set context span is created in the expected order and carries tags you need. A bit like other tests are doing: https://github.com/DataDog/dd-trace-java/blob/db1f4c300aca5abc983e6f6c974c5812b5b7bb8b/dd-java-agent/instrumentation/jdbc/src/test/groovy/JDBCInstrumentationTestBase.groovy#L181

nenadnoveljic: The tests you pointed to in your previous message seem to run queries on local databases of type `h2`, `derby`, etc. We don't have a SQL Server running locally. Consequently, if we want to make the same end-to-end test with SQL Server, we would need to integrate spinning up a SQL Server docker container in the CI. Do you think that's a good way to proceed?

dd-java-agent/instrumentation/jdbc/src/main/java/datadog/trace/instrumentation/jdbc/AbstractPreparedStatementInstrumentation.java
```diff
@@ -64,21 +66,29 @@ public static AgentScope onEnter(@Advice.This final Statement statement) {
...
-        DBInfo dbInfo =
+        final AgentSpan span;
+        final DBInfo dbInfo =
             JDBCDecorator.parseDBInfo(
                 connection, InstrumentationContext.get(Connection.class, DBInfo.class));
+        final boolean injectTraceContext = DECORATE.shouldInjectTraceContext(dbInfo);
+
+        if (INJECT_COMMENT && injectTraceContext && DECORATE.isSqlServer(dbInfo)) {
+          final long spanID = DECORATE.setContextInfo(connection, dbInfo);
+          span = AgentTracer.get().buildSpan(DATABASE_QUERY).withSpanId(spanID).start();
```
vandonr, line 84: I know there is historically few comments in the tracer code, but I'd like to see that change :D

```suggestion
          // The span ID is pre-determined so that we can reference it when setting the context
          final long spanID = DECORATE.setContextInfo(connection, dbInfo);
          // we then force that pre-determined span ID for the span covering the actual query
          span = AgentTracer.get().buildSpan(DATABASE_QUERY).withSpanId(spanID).start();
```

dd-java-agent/instrumentation/jdbc/src/main/java/datadog/trace/instrumentation/jdbc/StatementInstrumentation.java
```diff
@@ -81,19 +82,31 @@ public static AgentScope onEnter(
...
             JDBCDecorator.parseDBInfo(
                 connection, InstrumentationContext.get(Connection.class, DBInfo.class));
+        boolean injectTraceContext = DECORATE.shouldInjectTraceContext(dbInfo);
+        final AgentSpan span;
+        final boolean isSqlServer = DECORATE.isSqlServer(dbInfo);
+
+        if (isSqlServer && INJECT_COMMENT && injectTraceContext) {
+          final long spanID = DECORATE.setContextInfo(connection, dbInfo);
+          span = AgentTracer.get().buildSpan(DATABASE_QUERY).withSpanId(spanID).start();
+
```
vandonr, line 95:
```suggestion
          // The span ID is pre-determined so that we can reference it when setting the context
          final long spanID = DECORATE.setContextInfo(connection, dbInfo);
          // we then force that pre-determined span ID for the span covering the actual query
          span = AgentTracer.get().buildSpan(DATABASE_QUERY).withSpanId(spanID).start();
```

dd-java-agent/instrumentation/jdbc/src/main/java/datadog/trace/instrumentation/jdbc/StatementInstrumentation.java
```diff
@@ -81,19 +82,31 @@ public static AgentScope onEnter(
...
+        } else {
+          span = startSpan(DATABASE_QUERY);
+        }
+
+        DECORATE.afterStart(span);
         DECORATE.onConnection(span, dbInfo);
         final String copy = sql;
+        Integer priority = null;
```
vandonr, line 103: we don't need to allocate this here, we can leave it where it was before

dd-java-agent/instrumentation/jdbc/src/main/java/datadog/trace/instrumentation/jdbc/StatementInstrumentation.java
```diff
@@ -81,19 +82,31 @@ public static AgentScope onEnter(
...
+        final AgentScope scope = activateSpan(span);
```
vandonr, line 104: any reason to activate the span here rather on return as it was done before ?

dd-java-agent/instrumentation/jdbc/src/main/java/datadog/trace/instrumentation/jdbc/StatementInstrumentation.java
nenadnoveljic: changed now.

dd-java-agent/instrumentation/jdbc/src/main/java/datadog/trace/instrumentation/jdbc/StatementInstrumentation.java
nenadnoveljic: i had to do it when implementing the solution with hierarchical spans. changed now.

dd-java-agent/instrumentation/jdbc/src/main/java/datadog/trace/instrumentation/jdbc/JDBCDecorator.java
```diff
@@ -235,6 +238,56 @@ public String traceParent(AgentSpan span, int samplingPriority) {
...
+   * <p>Upsides: still "visible" in sub-queries, does not bust caches based on full query text
+   * Downsides: takes time.
+   *
+   * @param connection The same connection as the one that will be used for the actual statement
+   * @param dbInfo dbInfo of the instrumented database
+   * @return spanID pre-created spanID
+   */
+  public long setContextInfo(Connection connection, DBInfo dbInfo) {
+    final long spanID = Config.get().getIdGenerationStrategy().generateSpanId();
+    AgentSpan instrumentationSpan =
```
dougqh, line 261: As discussed before, I don't think we should create a span for this, but I'll let DBM make the final call.

dd-java-agent/instrumentation/jdbc/src/main/java/datadog/trace/instrumentation/jdbc/JDBCDecorator.java
```diff
@@ -235,6 +238,56 @@ public String traceParent(AgentSpan span, int samplingPriority) {
...
+      String samplingDecision = instrumentationSpan.forceSamplingDecision() > 0 ? "1" : "0";
+      Statement instrumentationStatement = connection.createStatement();
+      String instrumentationSql =
+          "set context_info 0x"
+              + samplingDecision
+              + DDSpanId.toHexStringPadded(spanID)
+              + instrumentationSpan.getTraceId().toHexString();
+      DECORATE.onStatement(instrumentationSpan, instrumentationSql);
+      instrumentationStatement.execute(instrumentationSql);
+      instrumentationStatement.close();
```
dougqh, line 276: This is a resource leak.  instrumentationStatement.close should be in finally block by itself.
This MUST be fixed before this is merged.

dd-java-agent/instrumentation/jdbc/src/main/java/datadog/trace/instrumentation/jdbc/JDBCDecorator.java
```diff
@@ -235,6 +238,56 @@ public String traceParent(AgentSpan span, int samplingPriority) {
...
+  public long setContextInfo(Connection connection, DBInfo dbInfo) {
+    final long spanID = Config.get().getIdGenerationStrategy().generateSpanId();
+    AgentSpan instrumentationSpan =
+        AgentTracer.get().buildSpan("set context_info").withTag("dd.instrumentation", true).start();
+    DECORATE.afterStart(instrumentationSpan);
+    DECORATE.onConnection(instrumentationSpan, dbInfo);
+
+    try (AgentScope scope = activateSpan(instrumentationSpan)) {
+      String samplingDecision = instrumentationSpan.forceSamplingDecision() > 0 ? "1" : "0";
+      Statement instrumentationStatement = connection.createStatement();
```
dougqh, line 268: This should used a prepared statement.

dd-java-agent/instrumentation/jdbc/src/main/java/datadog/trace/instrumentation/jdbc/JDBCDecorator.java
nenadnoveljic: If we remove the span, will we be able to see the duration of `set context_info` for each call? I can try to do it, and see what we get. However, we use [instrumentationSpan to retrieve the current traceID](https://github.com/DataDog/dd-trace-java/blob/25dab25632e88b93ff15f0580a116acfe206e1ba/dd-java-agent/instrumentation/jdbc/src/main/java/datadog/trace/instrumentation/jdbc/JDBCDecorator.java#L273). If we eliminate the `instrumentationSpan`, how can we retrieve the current `traceID`?

dougqh: I think customers will find the extra span to be noise, but maybe I'm wrong.  As I said, I'll let DBM decide for now, and we'll see what the customer feedback is.

vandonr: If we remove this span, we have to include the "set context" operation in the span of the query (and we can measure how long the "set context" took and write this as a tag on the span), but we decided against it to avoid having the duration of the actual operation change when the propagation mode changes.

dd-java-agent/instrumentation/jdbc/src/main/java/datadog/trace/instrumentation/jdbc/JDBCDecorator.java
nenadnoveljic: The instrumentation is executed as a prepared statement now.

dd-java-agent/instrumentation/jdbc/src/main/java/datadog/trace/instrumentation/jdbc/JDBCDecorator.java
nenadnoveljic: `close` is executed in the `finally` block now.

nenadnoveljic:
> Overall, design is okay. However, there are injection vulnerabilities and resource leaks that need to be addressed.

Fixed now.

dd-java-agent/instrumentation/jdbc/src/main/java/datadog/trace/instrumentation/jdbc/JDBCDecorator.java
```diff
@@ -235,6 +238,56 @@ public String traceParent(AgentSpan span, int samplingPriority) {
...
+      String instrumentationSql =
```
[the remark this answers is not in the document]
nenadnoveljic: Prepared statements are used now.

vandonr:
note that when you force-push, it makes it hard to review what was changed since the last comment.
If you need to update the branch to the latest master, you can merge master to it.

dd-java-agent/instrumentation/jdbc/src/main/java/datadog/trace/instrumentation/jdbc/JDBCDecorator.java
```diff
@@ -267,16 +268,18 @@ public long setContextInfo(Connection connection, DBInfo dbInfo) {
     try (AgentScope scope = activateSpan(instrumentationSpan)) {
       String samplingDecision = instrumentationSpan.forceSamplingDecision() > 0 ? "1" : "0";
       String contextInfo =
-          "0x"
+          "0"
```
vandonr, line 280: can we introduce a constant named `version` for this ? To make it obvious what this represents.

dd-java-agent/instrumentation/jdbc/src/main/java/datadog/trace/instrumentation/jdbc/JDBCDecorator.java
```diff
@@ -267,16 +268,18 @@ public long setContextInfo(Connection connection, DBInfo dbInfo) {
...
               + samplingDecision
               + DDSpanId.toHexStringPadded(spanID)
               + instrumentationSpan.getTraceId().toHexString();
 
-      String instrumentationSql = "set context_info " + contextInfo;
+      String instrumentationSql = "set context_info ?";
       instrumentationStatement = connection.prepareStatement(instrumentationSql);
+      instrumentationStatement.setBytes(1, DatatypeConverter.parseHexBinary(contextInfo));
```
vandonr: it is a bit wasteful to create a string only to convert it to bytes later. Maybe on such an instrumentation where we are going to run a sql request we don't care about small optimizations like this, idk.

nenadnoveljic:
> note that when you force-push, it makes it hard to review what was changed since the last comment. If you need to update the branch to the latest master, you can merge master to it.

Got it & will do it next time.

dd-java-agent/instrumentation/jdbc/src/main/java/datadog/trace/instrumentation/jdbc/JDBCDecorator.java
nenadnoveljic: done

dd-java-agent/instrumentation/jdbc/src/main/java/datadog/trace/instrumentation/jdbc/JDBCDecorator.java
nenadnoveljic: implemented `context_info` as `byte []`; concatenation more efficient with stream

vandonr:
Ok, there are still strings being converted to byte arrays, but honestly I think it's OK.
On the dotnet version, I started working on some code that would generate the bytes directly, you can look at this temp implem:
https://github.com/DataDog/dd-trace-dotnet/blob/5fb6e2e0e6fc6b4b3ada35ca25fd71b650d99020/tracer/src/Datadog.Trace/DatabaseMonitoring/DatabaseMonitoringPropagator.cs#L133-L150

Doug still has a blocking "change requested" btw

nenadnoveljic:
> Ok, there are still strings being converted to byte arrays, but honestly I think it's OK. On the dotnet version, I started working on some code that would generate the bytes directly, you can look at this temp implem: https://github.com/DataDog/dd-trace-dotnet/blob/5fb6e2e0e6fc6b4b3ada35ca25fd71b650d99020/tracer/src/Datadog.Trace/DatabaseMonitoring/DatabaseMonitoringPropagator.cs#L133-L150
> 
> Doug still has a blocking "change requested" btw

ok, i'm writing all values directly to the stream. thus, i completely eliminated string conversion.

dd-java-agent/instrumentation/jdbc/src/main/java/datadog/trace/instrumentation/jdbc/JDBCDecorator.java
```diff
@@ -235,6 +242,73 @@ public String traceParent(AgentSpan span, int samplingPriority) {
...
+    } catch (Exception e) {
+      log.debug(
+          "Failed to set extra DBM data in context info for trace {}. "
+              + "To disable this behavior, set DBM_PROPAGATION_MODE to 'service' mode. "
+              + "See https://docs.datadoghq.com/database_monitoring/connect_dbm_and_apm/ for more info.{}",
+          instrumentationSpan.getTraceId().toHexString(),
+          e);
+      DECORATE.onError(instrumentationSpan, e);
+    } finally {
+      if (instrumentationStatement != null) {
```
dougqh, line 301: If you create a smaller try / finally just around instrumentationStatement, then you won't need the null handling.
But that's a bit of a matter of preference

dd-java-agent/instrumentation/jdbc/src/main/java/datadog/trace/instrumentation/jdbc/JDBCDecorator.java
```diff
@@ -235,6 +242,73 @@ public String traceParent(AgentSpan span, int samplingPriority) {
...
+    DECORATE.afterStart(instrumentationSpan);
+    DECORATE.onConnection(instrumentationSpan, dbInfo);
+    PreparedStatement instrumentationStatement = null;
+    try (AgentScope scope = activateSpan(instrumentationSpan)) {
+      final byte samplingDecision =
+          (byte) (instrumentationSpan.forceSamplingDecision() > 0 ? 1 : 0);
+      final byte versionAndSamplingDecision =
+          (byte) ((VERSION << 4) & 0b11110000 | samplingDecision & 0b00000001);
+
+      ByteArrayOutputStream byteArrayOutputStream = new ByteArrayOutputStream();
```
dougqh, line 277: This might prove rather allocation heavy.
My first thought was could this be done with HeapByteBuffer instead, but I suppose this is fine for now.

dd-java-agent/instrumentation/jdbc/src/main/java/datadog/trace/instrumentation/jdbc/JDBCDecorator.java
```diff
@@ -235,6 +242,73 @@ public String traceParent(AgentSpan span, int samplingPriority) {
...
+   * <p>Upsides: still "visible" in sub-queries, does not bust caches based on full query text
+   * Downsides: takes time.
+   *
+   * @param connection The same connection as the one that will be used for the actual statement
+   * @param dbInfo dbInfo of the instrumented database
+   * @return spanID pre-created spanID
+   */
+  public long setContextInfo(Connection connection, DBInfo dbInfo) {
+    final byte VERSION = 0;
+    final long spanID = Config.get().getIdGenerationStrategy().generateSpanId();
```
dougqh, line 265: It occurred to me that we could avoid having to generate the span ID ahead of time IF set context can come after the query rather than before.

That might be a better option long term, since that would also work with OTel where I don't believe there's a way to generate the span ID in advance.

nenadnoveljic: > It occurred to me that we could avoid having to generate the span ID ahead of time IF set context can come after the query rather than before.

Unfortunately, `set context_info` has to execute prior to the instrumented SQL statement, to instrument the session which will run the SQL statement.

dd-java-agent/instrumentation/jdbc/src/main/java/datadog/trace/instrumentation/jdbc/JDBCDecorator.java
```diff
@@ -235,6 +242,73 @@ public String traceParent(AgentSpan span, int samplingPriority) {
...
+          e);
+      DECORATE.onError(instrumentationSpan, e);
+    } finally {
+      if (instrumentationStatement != null) {
+        try {
+          instrumentationStatement.close();
+        } catch (SQLException e) {
+        }
+      }
+      instrumentationSpan.finish();
```
dougqh, line 307: Ideally, finish needs to be in a finally block.
If close raises anything other than a SQLException, the span isn't being finished.

dougqh (approved):
An issue remains where the span may not be properly closed, that should be fixed before this change is checked in.

There's also room for improving the allocation to be lighter, but that's less crucial.

I've gone ahead and approved with the expectation that the span finish problem will be fixed.

dd-java-agent/instrumentation/jdbc/src/main/java/datadog/trace/instrumentation/jdbc/JDBCDecorator.java
nenadnoveljic: Can I change `} catch (SQLException e) {` to `} catch (Exception e) {`, and leave `instrumentationSpan.finish();` where it is, instead?

dd-java-agent/instrumentation/jdbc/src/main/java/datadog/trace/instrumentation/jdbc/JDBCDecorator.java
nenadnoveljic: Do we need then another try/catch/finally block, as the code prior to `instrumentationStatement = connection.prepareStatement(instrumentationSql);` also can throw exceptions we would like to spool into log. That would lead to even more verbosity. What's your preference?

dd-java-agent/instrumentation/jdbc/src/main/java/datadog/trace/instrumentation/jdbc/JDBCDecorator.java
nenadnoveljic: I implemented the solution with `ByteBuffer`.

dd-java-agent/instrumentation/jdbc/src/main/java/datadog/trace/instrumentation/jdbc/JDBCDecorator.java
nenadnoveljic: fixed as described above

dd-java-agent/instrumentation/jdbc/src/main/java/datadog/trace/instrumentation/jdbc/JDBCDecorator.java
dougqh: My preference is usually that there's a tight try / finally with the resource acquisition just before the try and the resource disposed of by itself in the finally.  This code is mixing the handling of two different resources into the one try / finally structure which can be hard to follow.

dd-java-agent/instrumentation/jdbc/src/main/java/datadog/trace/instrumentation/jdbc/JDBCDecorator.java
nenadnoveljic: I opened a [separate PR](https://github.com/DataDog/dd-trace-java/pull/7405) for this.

PR #7197, opened 2024-06-14
Fix Protobuf schema sampling logic

piochelepiotr:
# What Does This Do

When `canSample` returns false, `trySample` was not called, so the weight was not incremented.
This leads to wrong weights being calculated.
This PR updates the code to increment the weight in `canSample`, that should be called exactly once.

# Motivation

# Additional Notes

Jira ticket: [PROJ-IDENT]


dd-java-agent/instrumentation/protobuf/src/main/java/datadog/trace/instrumentation/protobuf_java/AbstractMessageInstrumentation.java
```diff
@@ -36,7 +36,7 @@ public String hierarchyMarkerType() {
   public ElementMatcher<TypeDescription> hierarchyMatcher() {
     return declaresMethod(named("writeTo"))
         .and(extendsClass(named(hierarchyMarkerType())))
-        .and(not(nameStartsWith("com.google.protobuf")));
+        .and(not(nameStartsWith("com.google.protobuf")).or(named("com.google.protobuf.DynamicMessage")));
```
piochelepiotr, line 39: This is another change, to track dynamic message writes. I can remove it from this PR if needed.

kr-igor (approved):

PerfectSlayer (approved):
LTGM

dd-java-agent/instrumentation/protobuf/src/main/java/datadog/trace/instrumentation/protobuf_java/AbstractMessageInstrumentation.java
PerfectSlayer: This is fine to me. You can add it to the `additional notes` section for later, if someone wonders why it was added.

dd-java-agent/instrumentation/protobuf/src/main/java/datadog/trace/instrumentation/protobuf_java/AbstractMessageInstrumentation.java
```diff
@@ -36,7 +36,9 @@ public String hierarchyMarkerType() {
   public ElementMatcher<TypeDescription> hierarchyMatcher() {
     return declaresMethod(named("writeTo"))
         .and(extendsClass(named(hierarchyMarkerType())))
-        .and(not(nameStartsWith("com.google.protobuf")));
+        .and(
+            not(nameStartsWith("com.google.protobuf"))
+                .or(named("com.google.protobuf.DynamicMessage")));
```
vandonr, line 41: wouldn't `com.google.protobuf.DynamicMessag` be included in `nameStartsWith("com.google.protobuf")` ? I feel like this second check is redundant

dd-trace-core/src/test/groovy/datadog/trace/core/datastreams/DefaultDataStreamsMonitoringTest.groovy
```diff
@@ -77,11 +77,14 @@ class DefaultDataStreamsMonitoringTest extends DDCoreSpecification {
...
     dataStreams.trySampleSchema("schema1") == 1
+    dataStreams.canSampleSchema("schema2")
     dataStreams.trySampleSchema("schema2") == 1
-    dataStreams.trySampleSchema("schema1") == 0
-    dataStreams.trySampleSchema("schema1") == 0
+    !dataStreams.canSampleSchema("schema1")
+    !dataStreams.canSampleSchema("schema1")
     timeSource.advance(30*1e9 as long)
+    dataStreams.canSampleSchema("schema1")
     dataStreams.trySampleSchema("schema1") == 3
```
vandonr, line 88: I think a future maintainer would be grateful if there was one or two comments explaining why we expect those results

dd-java-agent/instrumentation/protobuf/src/main/java/datadog/trace/instrumentation/protobuf_java/AbstractMessageInstrumentation.java
piochelepiotr: the first one is `not`, it's to exclude a lot of the small internal classes (for example, exclude simple type serialization).

vandonr: oh ok I read too fast

vandonr (approved):
lessgo

PR #7220, opened 2024-06-19
Preserve original types before passing data to the WAF

smola:
# What Does This Do
* When converting parsed body or grpc messages, preserve types to the extent that they are supported by libddwaf. Rules that apply only to strings can be skipped for numbers, booleans, etc.
* For anything that is not supported by libddwaf, such as non-string keys in maps, just convert to strings.
* When a key cannot be converted, use `invalid_key:$i` string
* Skip fields that are known to be useless, such as `memoizedHashCode` and `memoizedSize` in protobuf.

# Motivation
Originally motivated by spurious matches of internal protobuf fields, then also realized there are several cases where we are doing unnecessary work. This turns out to be the first of a possible series of performance optimizations.

# Additional Notes

Jira ticket: [APPSEC-53469](https://datadoghq.atlassian.net/browse/APPSEC-53469)

Future work:
* When an object cannot be converted correctly, do not use anything that will be extraneous to the WAF (such as `MyClass@1234`). Fall back should to be an invalid object, which is a well defined type in the WAF. We'll need to change libddwaf-java a bit to expose the invalid object type.
* Emit further metrics or logs when we hit this sort of edge case.

[APPSEC-53469]: https://datadoghq.atlassian.net/browse/APPSEC-53469?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ

### DIFF
--- a/dd-java-agent/appsec/src/main/java/com/datadog/appsec/event/data/ObjectIntrospection.java
+++ b/dd-java-agent/appsec/src/main/java/com/datadog/appsec/event/data/ObjectIntrospection.java
@@ -1,7 +1,11 @@
 package com.datadog.appsec.event.data;
 
 import datadog.trace.api.Platform;
-import java.lang.reflect.*;
+import java.lang.reflect.Array;
+import java.lang.reflect.Field;
+import java.lang.reflect.InvocationTargetException;
+import java.lang.reflect.Method;
+import java.lang.reflect.Modifier;
 import java.util.ArrayList;
 import java.util.HashMap;
 import java.util.List;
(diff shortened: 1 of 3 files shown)


dd-java-agent/appsec/src/test/groovy/com/datadog/appsec/event/data/ObjectIntrospectionSpecification.groovy
```diff
@@ -1,35 +1,87 @@
...
-    convert(charBuffer) == 'hello'
+    convert(true) === true
   }
 
-  void 'numbers are converted to strings'() {
+  void 'bytes are preserved'() {
     expect:
-    convert(5L) == '5'
-    convert(0.33G) == '0.33'
+    convert(1 as byte) === 1 as byte
```
manuel-alvarez-alvarez, line 23: I think it's better to test it with a direct `(byte) 1` cast to prevent groovy magic from screwing the test (kind of affecting the experiment when measuring :smile:)

dd-java-agent/appsec/src/main/java/com/datadog/appsec/event/data/ObjectIntrospection.java
```diff
@@ -32,38 +36,89 @@ public final class ObjectIntrospection {
...
+  private static Object doConversion(Object obj, int depth, State state) {
+    state.elemsLeft--;
+    if (state.elemsLeft <= 0 || obj == null || depth > MAX_DEPTH) {
       return null;
     }
 
-    // char sequences / numbers
-    if (obj instanceof CharSequence || obj instanceof Number) {
+    // strings, booleans and numbers are preserved
+    if (obj instanceof String || obj instanceof Boolean || obj instanceof Number) {
```
manuel-alvarez-alvarez, line 115: We are missing primitives here (not sure if we can safely ignore them)

manuel-alvarez-alvarez (approved):

dd-java-agent/appsec/src/main/java/com/datadog/appsec/event/data/ObjectIntrospection.java
smola: You mean other types, or just the primitive counterparts? Since we're dealing with `Object` here, these will be necessarily autoboxed.

dd-java-agent/appsec/src/test/groovy/com/datadog/appsec/event/data/ObjectIntrospectionSpecification.groovy
smola: Improved these tests to be more explicit with input and output types (and parameterized them).

PR #7221, opened 2024-06-19
Apply appsec rate limiter on event instead of when request end

jandro996:
# What Does This Do

 - check rate limiter once per request
 - only waf events are rate limited

# Motivation
We need to propagate sampling decision on appsec event,  not on request end

# Additional Notes

Jira ticket: [PROJ-IDENT]

PR #7222, opened 2024-06-19
Ensure user tracking request headers are not rate limited

manuel-alvarez-alvarez:
# What Does This Do
It makes sure that request headers are always set in the span when there are user tracking events.

# Motivation
User tracking events are disjoint from waf events so rating limitations should not apply.

# Additional Notes

Jira ticket: [PROJ-IDENT]

### DIFF
--- a/dd-java-agent/appsec/src/main/java/com/datadog/appsec/gateway/GatewayBridge.java
+++ b/dd-java-agent/appsec/src/main/java/com/datadog/appsec/gateway/GatewayBridge.java
@@ -50,6 +50,7 @@ import java.util.Map;
 import java.util.Set;
 import java.util.function.Function;
 import java.util.regex.Pattern;
+import javax.annotation.Nullable;
 import org.slf4j.Logger;
 import org.slf4j.LoggerFactory;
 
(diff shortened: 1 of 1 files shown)


dd-java-agent/appsec/src/main/java/com/datadog/appsec/gateway/GatewayBridge.java
```diff
@@ -66,7 +67,9 @@ public class GatewayBridge {
 
   /** User tracking tags that will force the collection of request headers */
   private static final String[] USER_TRACKING_TAGS = {
-    "appsec.events.users.login.success.track", "appsec.events.users.login.failure.track"
+    "appsec.events.users.login.success.track",
+    "appsec.events.users.login.failure.track",
+    "appsec.events.users.signup.track"
```
manuel-alvarez-alvarez, line 72: I missed this event in the initial PR

smola (approved):

jandro996 (approved):
If we confirm that user tracking events are out of rate limiting, LGTM

manuel-alvarez-alvarez:
Not required after https://github.com/DataDog/dd-trace-java/pull/7221

PR #7224, opened 2024-06-20
enable timeline events by default when ddprof disabled

richardstartin:
# What Does This Do

Enables timeline events by default when JFR profiler is used, which allows for better integration into the timeline.

# Motivation

# Additional Notes

Jira ticket: [PROF-10003]

[PROF-10003]: https://datadoghq.atlassian.net/browse/PROF-10003?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ

### DIFF
--- a/dd-trace-api/src/main/java/datadog/trace/api/config/ProfilingConfig.java
+++ b/dd-trace-api/src/main/java/datadog/trace/api/config/ProfilingConfig.java
@@ -216,7 +216,7 @@ public final class ProfilingConfig {
 
   public static final String PROFILING_TIMELINE_EVENTS_ENABLED =
       "profiling.timeline.events.enabled";
-  public static final boolean PROFILING_TIMELINE_EVENTS_ENABLED_DEFAULT = false;
+  public static final boolean PROFILING_TIMELINE_EVENTS_ENABLED_DEFAULT = true;
 
   public static final String PROFILING_DETAILED_DEBUG_LOGGING = "profiling.detailed.debug.logging";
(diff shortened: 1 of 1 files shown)


dd-trace-api/src/main/java/datadog/trace/api/config/ProfilingConfig.java
```diff
@@ -216,7 +216,7 @@ public final class ProfilingConfig {
 
   public static final String PROFILING_TIMELINE_EVENTS_ENABLED =
       "profiling.timeline.events.enabled";
-  public static final boolean PROFILING_TIMELINE_EVENTS_ENABLED_DEFAULT = false;
+  public static final boolean PROFILING_TIMELINE_EVENTS_ENABLED_DEFAULT = true;
```
jbachorik, line 219: This will always enable thoee event, not only when ddprof is disabled, correct?

richardstartin: Only when ddprof is disabled (or DatadogProfilingIntegration fails to load) according to the logic in Agent#createProfilingContextIntegration

jbachorik: Ok. Can you add a comment for the default, please?

jbachorik (approved):

PR #7227, opened 2024-06-20
Track JVM RSS in JDK 21+

MattAlp:
# What Does This Do
This is a new PR following an extension of #7107's original changes + my botching of the commit history.

This PR is responsible for reading from`procfs`' `smaps` entries for JVMs running on Linux, and optionally using existing NMT region annotations in JDK22+ (for now) to identify where the RSS is coming from.

# Motivation
NMT within the JVM does not account for tracking of _all_ memory used by the process itself, which means that the NMT timeseries metrics we report are underestimating the real memory usage of the JVM (i.e. from the JVM itself, or native libraries used via JNI). See the discrepancy in reported memory vs actual memory usage below; this is particularly problematic when running within containers, where the heap may be sized correctly relative to the available memory, but the overall usage exceeds what is available leading to an OOMKill.

image
image
(Note around 1GB of memory out of 4 that is not tracked via NMT).

# Additional Notes

- The pattern matching used for parsing `System.map` output and the MBean interactions are a little hacky, but prevent us from needing to dive into VMStructs or working on OpenJDK to start collecting this data. Open to feedback for cleaning this up.
- Integration testing for this feature is nonexistent, open to pushback here as well.

Jira ticket: [PROF-9851](https://datadoghq.atlassian.net/browse/PROF-9851)

[PROF-9851]: https://datadoghq.atlassian.net/browse/PROF-9851?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ

### DIFF
--- a/dd-java-agent/agent-profiling/profiling-controller-openjdk/src/main/java/com/datadog/profiling/controller/openjdk/events/SmapEntryEvent.java
+++ b/dd-java-agent/agent-profiling/profiling-controller-openjdk/src/main/java/com/datadog/profiling/controller/openjdk/events/SmapEntryEvent.java
@@ -1,6 +1,5 @@
 package com.datadog.profiling.controller.openjdk.events;
 
-import de.thetaphi.forbiddenapis.SuppressForbidden;
 import jdk.jfr.Category;
 import jdk.jfr.DataAmount;
 import jdk.jfr.Description;
@@ -241,7 +240,6 @@ public class SmapEntryEvent extends Event {
     this.nmtCategory = nmtCategory;
   }
 
-  @SuppressForbidden // split with one-char String use a fast-path without regex usage
   public static void emit() {
     SmapEntryFactory.collectEvents().forEach(Event::commit);
   }
--- a/dd-java-agent/agent-profiling/profiling-controller-openjdk/src/main/java/com/datadog/profiling/controller/openjdk/events/SmapEntryFactory.java
+++ b/dd-java-agent/agent-profiling/profiling-controller-openjdk/src/main/java/com/datadog/profiling/controller/openjdk/events/SmapEntryFactory.java
@@ -112,6 +112,7 @@ public class SmapEntryFactory {
     }
   }
 
+  @SuppressForbidden // split with one-char String use a fast-path without regex usage
   static List<? extends Event> collectEvents() {
     if (!SMAP_ENTRY_EVENT.isEnabled()) {
       return Collections.emptyList();


dd-java-agent/agent-profiling/profiling-controller-openjdk/src/main/java/com/datadog/profiling/controller/openjdk/events/SmapEntryEvent.java
```diff
@@ -0,0 +1,512 @@
...
+
+        String[] emptyStringArgs = {};
+        Object[] dcmdArgs = {emptyStringArgs};
+        String[] signature = {String[].class.getName()};
+
+        String[] lines =
+            ((String) mbs.invoke(objectName, "systemMap", dcmdArgs, signature)).split("\n");
+        HashMap<Long, String> annotatedRegions = new HashMap<>();
+        Pattern pattern =
+            Pattern.compile(
```
jpbempel, line 237: could put this into a static field, no?

richardstartin (approved):
Please make the pattern a static final, otherwise LGTM

jbachorik (approved):

richardstartin (approved):

dd-java-agent/agent-profiling/profiling-controller-openjdk/src/main/java/com/datadog/profiling/controller/openjdk/OpenJdkController.java
```diff
@@ -239,6 +248,10 @@ && isEventEnabled(recordingSettings, "jdk.NativeMethodSample")) {
 
     // Register periodic events
     AvailableProcessorCoresEvent.register();
+
+    if (Platform.isLinux() && isEventEnabled(this.recordingSettings, "datadog.SmapEntry")) {
+      SmapEntryEvent.register();
+    }
```
jbachorik, line 254: This is messing with JMX - the registration needs to be delayed after JMX is ready.

You could try copying what is done for the deadlock detection - https://github.com/DataDog/dd-trace-java/blob/97e2337ea116e9f61c151dc2b54bb4ba09b3b779/dd-java-agent/agent-bootstrap/src/main/java/datadog/trace/bootstrap/Agent.java#L648 - it also requires access to JMX so the initialization must take place only when it is safe so.

PR #7231, opened 2024-06-21
Exploit prevention for SQL injection (blocking support)

ValentinZakharov:
# What Does This Do
Implemented blocking functionality for the SQL-injection Exploit Prevention.

# Motivation

# Additional Notes
This is part of Exploit prevention initiative (RASP)

Jira ticket: [APPSEC-46818]

[APPSEC-46818]: https://datadoghq.atlassian.net/browse/APPSEC-46818?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ

### DIFF
--- a/dd-java-agent/agent-bootstrap/src/main/java/datadog/trace/bootstrap/instrumentation/decorator/DatabaseClientDecorator.java
+++ b/dd-java-agent/agent-bootstrap/src/main/java/datadog/trace/bootstrap/instrumentation/decorator/DatabaseClientDecorator.java
@@ -3,9 +3,12 @@ package datadog.trace.bootstrap.instrumentation.decorator;
 import static datadog.trace.api.gateway.Events.EVENTS;
 import static datadog.trace.bootstrap.instrumentation.api.Tags.DB_TYPE;
 
+import datadog.appsec.api.blocking.BlockingException;
 import datadog.trace.api.Config;
 import datadog.trace.api.cache.DDCache;
 import datadog.trace.api.cache.DDCaches;
+import datadog.trace.api.gateway.BlockResponseFunction;
+import datadog.trace.api.gateway.Flow;
 import datadog.trace.api.gateway.RequestContext;
 import datadog.trace.api.gateway.RequestContextSlot;
 import datadog.trace.api.naming.NamingSchema;
@@ -14,7 +17,7 @@ import datadog.trace.bootstrap.instrumentation.api.AgentSpan;
 import datadog.trace.bootstrap.instrumentation.api.AgentTracer;
 import datadog.trace.bootstrap.instrumentation.api.Tags;
 import datadog.trace.bootstrap.instrumentation.api.UTF8BytesString;
-import java.util.function.BiConsumer;
+import java.util.function.BiFunction;
 
 public abstract class DatabaseClientDecorator<CONNECTION> extends ClientDecorator {
   protected static class NamingEntry {
@@ -116,14 +119,27 @@ public abstract class DatabaseClientDecorator<CONNECTION> extends ClientDecorato
    */
   public void onRawStatement(AgentSpan span, String sql) {
     if (Config.get().isAppSecRaspEnabled() && sql != null && !sql.isEmpty()) {
-      BiConsumer<RequestContext, String> sqlQueryCallback =
+      BiFunction<RequestContext, String, Flow<Void>> sqlQueryCallback =
           AgentTracer.get()
               .getCallbackProvider(RequestContextSlot.APPSEC)
               .getCallback(EVENTS.databaseSqlQuery());
       if (sqlQueryCallback != null) {
         RequestContext ctx = span.getRequestContext();
         if (ctx != null) {
-          sqlQueryCallback.accept(ctx, sql);
+          Flow<Void> flow = sqlQueryCallback.apply(ctx, sql);
+          Flow.Action action = flow.getAction();
+          if (action instanceof Flow.Action.RequestBlockingAction) {
+            BlockResponseFunction brf = ctx.getBlockResponseFunction();
+            if (brf != null) {
+              Flow.Action.RequestBlockingAction rba = (Flow.Action.RequestBlockingAction) action;
+              brf.tryCommitBlockingResponse(
+                  ctx.getTraceSegment(),
+                  rba.getStatusCode(),
+                  rba.getBlockingContentType(),
+                  rba.getExtraHeaders());
+            }
+            throw new BlockingException("Blocked request (for SQL query)");
+          }
         }
       }
     }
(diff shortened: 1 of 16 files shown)


dd-java-agent/instrumentation/jdbc/src/main/java/datadog/trace/instrumentation/jdbc/StatementInstrumentation.java
```diff
@@ -112,11 +114,16 @@ public static AgentScope onEnter(
...
         DECORATE.onStatement(span, copy);
-        return activateSpan(span);
       } catch (SQLException e) {
         // if we can't get the connection for any reason
         return null;
+      } catch (BlockingException e) {
+        // re-throw blocking exceptions
+        throw e;
+      } catch (Throwable e) {
+        // suppress anything else
```
smola, line 124: Needs to be logged (debug level), just as it happens with suppress = Throwable.

smola (changes requested):

dd-java-agent/instrumentation/jdbc/src/main/java/datadog/trace/instrumentation/jdbc/StatementInstrumentation.java
ValentinZakharov: Good point, added log message!

PerfectSlayer (approved):
LGTM for `internal-api` 👍

dd-java-agent/instrumentation/jdbc/src/main/java/datadog/trace/instrumentation/jdbc/StatementInstrumentation.java
```diff
@@ -55,7 +57,10 @@ public Map<String, String> contextStore() {
   @Override
   public String[] helperClassNames() {
     return new String[] {
-      packageName + ".JDBCDecorator", packageName + ".SQLCommenter",
+      packageName + ".JDBCDecorator",
+      packageName + ".SQLCommenter",
+      "org.slf4j.LoggerFactory",
```
amarziali, line 62: You shouldn't inject log4j classes

amarziali (changes requested):
slf4j  classes should never be injected

dd-java-agent/agent-bootstrap/src/main/java/datadog/trace/bootstrap/instrumentation/decorator/DatabaseClientDecorator.java
```diff
@@ -135,14 +151,27 @@ protected void processDatabaseType(AgentSpan span, String dbType) {
...
-      BiConsumer<RequestContext, String> connectDbCallback =
+      BiFunction<RequestContext, String, Flow<Void>> connectDbCallback =
           AgentTracer.get()
               .getCallbackProvider(RequestContextSlot.APPSEC)
               .getCallback(EVENTS.databaseConnection());
       if (connectDbCallback != null) {
         RequestContext ctx = span.getRequestContext();
         if (ctx != null) {
-          connectDbCallback.accept(ctx, dbType);
+          Flow<Void> flow = connectDbCallback.apply(ctx, dbType);
```
manuel-alvarez-alvarez, line 161: Can we block during the connection phase? (seems weird since we do not have a SQL query yet)

dd-java-agent/instrumentation/jdbc/src/main/java/datadog/trace/instrumentation/jdbc/StatementInstrumentation.java
ValentinZakharov: Thanks, fixed

dd-java-agent/instrumentation/jdbc/src/main/java/datadog/trace/instrumentation/jdbc/StatementInstrumentation.java
```diff
@@ -70,7 +73,7 @@ public void methodAdvice(MethodTransformer transformer) {
   }
 
   public static class StatementAdvice {
-    @Advice.OnMethodEnter(suppress = Throwable.class)
```
amarziali, line 73: why the suppression has been removed?

amarziali: ok found the answer below. :)

amarziali (approved):

dd-java-agent/instrumentation/jdbc/src/main/java/datadog/trace/instrumentation/jdbc/StatementInstrumentation.java
ValentinZakharov: That was in purpose to let blocking exception bubble-up from `StatementAdvice` to http server advices.
The idea is - to suppress everything except `BlockingException`. So, instead of `suppress = Throwable.class`, I used custom `try-catch` to suppress exceptions.

dd-java-agent/agent-bootstrap/src/main/java/datadog/trace/bootstrap/instrumentation/decorator/DatabaseClientDecorator.java
ValentinZakharov: We are checking query statement right before it's executed (not when it's built). That's why, we have database connection first and sql query execution, after. 
I think, blocking on connection phase can increase code complexity.

manuel-alvarez-alvarez (approved):

dd-java-agent/appsec/src/main/java/com/datadog/appsec/gateway/GatewayBridge.java
```diff
@@ -453,45 +453,34 @@ public void init() {
...
-            }
-            DataBundle bundle = new SingletonDataBundle<>(KnownAddresses.DB_TYPE, dbType);
-            try {
-              producerService.publishDataEvent(subInfo, ctx, bundle, false);
-              return;
-            } catch (ExpiredSubscriberInfoException e) {
-              dbConnectionSubInfo = null;
-            }
-          }
+          ctx.setDbType(dbType);
```
manuel-alvarez-alvarez, line 456: Multiple database calls (with potentially different providers) can coexist in a single request and adding this to the context might trigger race conditions. Since we always have the type in the decorator (it's saved in a context store linked to the jdbc Connection), whey don't we pass the type as an argument with the sql callback?

smola: As discussed offline, needs to be addressed at libddwaf-level, since it does not seem clear what would be the correct way to pass different pairs of query/dbtype.

dd-java-agent/instrumentation/jdbc/src/main/java/datadog/trace/instrumentation/jdbc/StatementInstrumentation.java
```diff
@@ -112,11 +116,18 @@ public static AgentScope onEnter(
...
         return null;
+      } catch (BlockingException e) {
+        // re-throw blocking exceptions
+        throw e;
+      } catch (Throwable e) {
+        // suppress anything else
+        InstrumentationLogger.debug(
+            "datadog.trace.instrumentation.jdbc.StatementInstrumentation", statement.getClass(), e);
       }
+      return activateSpan(span);
```
smola, line 130: I'm not sure this is equivalent to the previous code. If an exception was thrown at `DECORATE.onStatement(span, copy)`, `activateSpan(span)` would not be called.

ValentinZakharov: Indeed, return value is wrong.
Talked to @mcculls to confirm we need return `null` in case of suppressed `Throwable`.
Fixed

smola (approved):

PR #7233, opened 2024-06-23
Rename Dynamic Instrumentation system tests run name

shurivich:
# What Does This Do
Rename run from `Run APM Integrations tests` to `Run Dynamic Instrumentation system tests`


PerfectSlayer (approved):
Thanks for the fix!

PR #7237, opened 2024-06-25
Update ZIO to latest `2.0.x` version

guizmaii:
# What Does This Do

Update ZIO

# Motivation

The ZIO integration doesn't seem to work properly in my project

### DIFF
--- a/dd-java-agent/instrumentation/zio/zio-2.0/build.gradle
+++ b/dd-java-agent/instrumentation/zio/zio-2.0/build.gradle
@@ -1,4 +1,4 @@
-def zioVersion = '2.0.0'
+def zioVersion = '2.0.22'
 def scalaVersion = '2.12'
 
 muzzle {


amarziali:
👋 Thanks for having opened this PR. Having increased the version number in that gradle file won't have any effect on the integration since we are supposed to instrument from 2.0.0 to < 2.1. Hence this change won't make the desired effect.

What is the version of zio you are running on your project?

guizmaii:
@amarziali Thanks for your help! We're using ZIO `2.0.22` and Scala `3.4.2`

amarziali:
> 3.4.2

@guizmaii thanks for the answer. The tests we have for zio already passes for that version. I can suggest to close that PR (because won't fix the issue you are experiencing) and instead reach out the datadog support by explaining the issue. If you have debug logs you can provide them and ask if they can escalate the issue to eng. Thanks in advance

PR #7242, opened 2024-06-26
Capture all spark conf parameter

paul-laffon-dd:
# What Does This Do

Capture all spark conf properties. Use the `spark.redaction.regex` pattern to redact secret in the same way spark is doing

Also changed capturing the job parameters to only capture those for local root spans

# Motivation

Only capturing a subset meant that important properties for users could be missed

# Additional Notes

https://datadoghq.atlassian.net/browse/DJM-150


yiliangzhou (approved):
LGTM

PR #7244, opened 2024-06-26
Do not report code coverage for skipped tests

nikita-tkachenko-datadog:
# What Does This Do

Updates code coverage processing logic so that code coverage is not processed/reported for skipped tests.

# Motivation
If a test is skipped, code coverage is likely to be empty or incomplete. The backend has no use for this coverage, at least as of now, so we can save some resources by not processing it and not reporting it.

Jira ticket: [SDTEST-493]

[SDTEST-493]: https://datadoghq.atlassian.net/browse/SDTEST-493?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ

### DIFF
--- a/dd-java-agent/agent-ci-visibility/src/main/java/datadog/trace/civisibility/domain/TestImpl.java
+++ b/dd-java-agent/agent-ci-visibility/src/main/java/datadog/trace/civisibility/domain/TestImpl.java
@@ -220,11 +220,15 @@ public class TestImpl implements DDTest {
     InstrumentationTestBridge.fireBeforeTestEnd(context);
 
     CoverageBridge.removeThreadLocalCoverageProbeStore();
-    boolean coveragesGathered =
-        context.getCoverageProbeStore().report(sessionId, suiteId, span.getSpanId());
-    if (!coveragesGathered && !TestStatus.skip.equals(span.getTag(Tags.TEST_STATUS))) {
-      // test is not skipped, but no coverages were gathered
-      metricCollector.add(CiVisibilityCountMetric.CODE_COVERAGE_IS_EMPTY, 1);
+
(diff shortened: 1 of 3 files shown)


dd-java-agent/agent-ci-visibility/src/main/java/datadog/trace/civisibility/domain/TestImpl.java
```diff
@@ -220,11 +220,15 @@ public void end(@Nullable Long endTime) {
...
 
     CoverageBridge.removeThreadLocalCoverageProbeStore();
-    boolean coveragesGathered =
-        context.getCoverageProbeStore().report(sessionId, suiteId, span.getSpanId());
-    if (!coveragesGathered && !TestStatus.skip.equals(span.getTag(Tags.TEST_STATUS))) {
-      // test is not skipped, but no coverages were gathered
-      metricCollector.add(CiVisibilityCountMetric.CODE_COVERAGE_IS_EMPTY, 1);
+
+    // do not process coverage reports for skipped tests
+    if (span.getTag(Tags.TEST_STATUS) != TestStatus.skip) {
```
anmarchenko, line 225: should we add a test for that?

anmarchenko (approved):

dd-java-agent/agent-ci-visibility/src/main/java/datadog/trace/civisibility/domain/TestImpl.java
nikita-tkachenko-datadog: yup, that makes sense, added

PR #7252, opened 2024-06-28
Use spark application name when service is set to spark

paul-laffon-dd:
# What Does This Do

`DD_SPARK_APP_NAME_AS_SERVICE` also use `spark.app.name` as the default service if service is set to `spark`

# Motivation

USM can auto-infer the service to spark, in this case we still want to use the spark app name as the service

# Additional Notes

https://datadoghq.atlassian.net/browse/DJM-30

`DD_SPARK_APP_NAME_AS_SERVICE` was introduced in this PR: https://github.com/DataDog/dd-trace-java/pull/6460

### DIFF
--- a/dd-java-agent/instrumentation/spark/src/main/java/datadog/trace/instrumentation/spark/AbstractDatadogSparkListener.java
+++ b/dd-java-agent/instrumentation/spark/src/main/java/datadog/trace/instrumentation/spark/AbstractDatadogSparkListener.java
@@ -1118,13 +1118,24 @@ public abstract class AbstractDatadogSparkListener extends SparkListener {
   }
 
   private static String getSparkServiceName(SparkConf conf, boolean isRunningOnDatabricks) {
-    if (Config.get().isServiceNameSetByUser()
-        || !Config.get().useSparkAppNameAsService()
-        || isRunningOnDatabricks) {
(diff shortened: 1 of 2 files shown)


dd-java-agent/instrumentation/spark/src/main/java/datadog/trace/instrumentation/spark/AbstractDatadogSparkListener.java
```diff
@@ -1118,13 +1118,24 @@ private static String getDatabricksServiceName(SparkConf conf, String databricks
...
-        || isRunningOnDatabricks) {
+    // If config is not set or running on databricks, not changing the service name
+    if (!Config.get().useSparkAppNameAsService() || isRunningOnDatabricks) {
       return null;
     }
 
-    return conf.get("spark.app.name", null);
+    // Keep service set by user, except if it is only "spark" that can be set by USM
+    String serviceName = Config.get().getServiceName();
+    if ((Config.get().isServiceNameSetByUser() && !"spark".equals(serviceName))) {
```
yiliangzhou, line 1128: nit: there is extra pair of `()` around the boolean condition.

yiliangzhou (approved):

PR #7259, opened 2024-06-28
Switch to 128-bit trace ID tracking

spikat:
# What Does This Do

It switches used trace IDs to 128 bits.

# Motivation

Be compliant with the new tracer

# Additional Notes

Jira ticket: [PROJ-IDENT]

### DIFF
--- a/dd-java-agent/cws-tls/src/main/java/datadog/cws/tls/TlsScopeListener.java
+++ b/dd-java-agent/cws-tls/src/main/java/datadog/cws/tls/TlsScopeListener.java
@@ -25,10 +25,7 @@ public class TlsScopeListener implements ExtendedScopeListener {
     Deque<Span> stack = spanStack.get();
 
     Span top = stack.peek();
-    if (top == null
-        || top.getTraceId().toLong() != traceId.toLong()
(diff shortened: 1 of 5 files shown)


dd-java-agent/cws-tls/src/main/java/datadog/cws/tls/TlsScopeListener.java
```diff
@@ -25,8 +26,10 @@ void push(DDTraceId traceId, long spanId) {
     Deque<Span> stack = spanStack.get();
 
     Span top = stack.peek();
-    if (top == null || top.getTraceId().toLong() != traceId.toLong() || top.getSpanId() != spanId) {
-
+    if (top == null
+        || top.getTraceId().toLong() != traceId.toLong()
+        || top.getTraceId().toHighOrderLong() != traceId.toHighOrderLong()
```
PerfectSlayer, line 31: What about using `equals()` from `DDTraceId` instances?

PerfectSlayer (approved):

PerfectSlayer (commented):
Only a minor comment.
It might be nice to have some tests mixing 64 and 128b TID 🤷

ygree (approved):

dougqh (approved):

PR #7260, opened 2024-06-28
remove dependency on abstract message in schema extractor

piochelepiotr:
# What Does This Do

# Motivation

Trying to solve this error:

```
java.lang.IllegalAccessError: class datadog.trace.instrumentation.protobuf_java.SchemaExtractor (in unnamed module @xxxxxx) cannot access class com.google.protobuf.AbstractMessage (in module xxxx.yyyy) because module xxxx.yyyy does not export com.google.protobuf to unnamed module @xxxxxxxx
```

My theory is that the dynamic message instrumentation calls this class, but dynamic message has nothing to do with `AbstractMessage`, so maybe in some cases DynamicMessage exists, but not AbstractMessage?
But that doesn't seem to be 100% correlated with the error.

# Additional Notes

I'm not confident this will work. Since the error says the whole `protobuf` module is not exported.
But I only instrument classes extending protobuf classes, so I don't understand how that can happen.

Jira ticket: [PROJ-IDENT]


vandonr (approved):

manuel-alvarez-alvarez (approved):
LGTM

PR #7266, opened 2024-07-02
JMS: trace Queue and Topic producers when destination is explicit

amarziali:
# What Does This Do

The java agent today is not instrumenting jms send for signatures `QueueProducer.send(Queue, Message, ..)` or `TopicPublisher.send(Topic, Message)`.
Instead, we trace `MessageProducer.send(Destination, Message, ..)` only. 

This is generally OK when specialized classes are delegating to superclasses. In fact for ActiveMQ we have

```
public class ActiveMQQueueSender extends ActiveMQMessageProducer implements QueueSender {
   ....

    public void send(Queue queue, Message message) throws JMSException {
        super.send(queue, message);
    }
```

But it's not OK when the specialized class does not delegate to super. Hence we lost the tracing on the producer.

# Motivation

# Additional Notes

Jira ticket: [PROJ-IDENT]


mcculls (approved):
Good catch! The existing call-depth checks should avoid duplicate spans

Issue #7254, opened 2024-06-28
Unable to capture gRPC list field in dynamic instrumentation


robmillersoftware:
I've got a Java gRPC service running in our k8s cluster, and I'm trying to validate the message data we're receiving using a dynamic probe to check the values in field of type `com.google.protobuf.DoubleArrayList`. However, when I go to check the logs from the probe I'm running into the following error:

`java.lang.RuntimeException: Unsupported Collection type: com.google.protobuf.DoubleArrayList`

Below is a screenshot of the Stacktrace section
![image](https://github.com/DataDog/dd-trace-java/assets/21186815/11870caa-16f3-4b81-9b0c-b84e001803db)

jpbempel:
Hi @robmillersoftware , the `DoubleArrayList` is considered as a collection (because implements Collection interface) but we are restricting the access of the collection outside of the JDK classes because an external collection could implement the interface and notably the size method by doing something that could be very expensive (SQL call for example).
However in that case we should fallback to regular object to capture fields instead.

PR #7274, opened 2024-07-03
Serialize restricted collections as regular object

jpbempel:
# What Does This Do
We specially serialize collections/maps into snapshot using interfaces and calling methods like `size()` but we restrict this for JDK collections that are guarantee to be only in-memory. For all others we are raising an exception. Instead we should just serialize then as regular object with fields.

# Motivation

https://github.com/DataDog/dd-trace-java/issues/7254

# Additional Notes

Jira ticket: [DEBUG-2502]

[DEBUG-2502]: https://datadoghq.atlassian.net/browse/DEBUG-2502?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ


shatzi (approved):
I like this change. Make this more complete.

PR #7276, opened 2024-07-03
Fetch uri information correctly when httphost argument is null

abagavat:
# What Does This Do

Fixes the issue described in https://github.com/DataDog/dd-trace-java/issues/7275

# Motivation

# Additional Notes

Jira ticket: [PROJ-IDENT]


amarziali:
👋 Thank you for your contribution. There are some test failing so I've to understand the impact and perhaps do a commit in order to have the CI pass. Afterwards I will be able to approve this PR

amarziali:
@abagavat can you share the way you reproduce the issue? Perhaps the way the request is done with httpclient? I will add a test for it

abagavat:
@amarziali I didn't specifically try to reproduce it as it was happening at a large scale across many of our applications. This started happening after a recent SpringBoot upgrade. We use Spring RestTemplate which internally uses apache httpclient to make the http request.

Between the two versions the signature of the execute method to httpclient has changed.
**Prior** - https://github.com/spring-projects/spring-framework/blob/ac667a1e2b24baa432653603aba8ae81218f7006/spring-web/src/main/java/org/springframework/http/client/HttpComponentsClientHttpRequest.java#L93
**After upgrade** - https://github.com/spring-projects/spring-framework/blob/194b4cedfbf9f947797d97af851c8e718c2aa91e/spring-web/src/main/java/org/springframework/http/client/HttpComponentsClientHttpRequest.java#L99

As you can see after the upgrade Spring is explicitly setting the host to null.

I think you can repro this in the tests by using a different variant of this test fixture where the host is set to null.
https://github.com/DataDog/dd-trace-java/blob/23a8164e90701d92c3a513620f28accfa8b33979/dd-java-agent/instrumentation/apache-httpclient-5/src/test/groovy/ApacheHttpClientTest.groovy#L101

amarziali:
> I think you can repro this in the tests by using a different variant of this test fixture where the host is set to null.
@abagavat this is a great detailed answer. Thank you. I was able to reproduce the issue (exactly the `hostname` was missing) and to add a test case for it. I'll let the CI run and afterwards we're ready to go . Thank you again for this contribution

vandonr (approved):
👍

amarziali (approved):

PR #7280, opened 2024-07-04
Move snapshot UUID generation at serialization

jpbempel:
# What Does This Do
UUID generation is not cheap, and therefore should be performed only when the snapshot is about to be sent.
Moving UUID generation at seriazliation time makes it out of the critical path.

# Motivation
Performance

# Additional Notes

Jira ticket: [DEBUG-2346]

[DEBUG-2346]: https://datadoghq.atlassian.net/browse/DEBUG-2346?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ

### DIFF
--- a/.gitlab/benchmarks.yml
+++ b/.gitlab/benchmarks.yml
@@ -8,6 +8,7 @@ variables:
   timeout: 1h
   tags: ["runner:apm-k8s-tweaked-metal"]
   image: $BASE_CI_IMAGE
+  needs: [ "build" ]
   script:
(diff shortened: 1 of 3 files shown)


cimi (approved):
:+1: Good idea, nice improvement!

Curious to see the benchmark results after this.

PR #7284, opened 2024-07-05
Inject coverage store field into Thread class

nikita-tkachenko-datadog:
# What Does This Do

Optimises line-level per-test code coverage.
Coverage store lookup is optimised: instead of using a `ThreadLocal`, a dedicated field is injected into `java.lang.Thread` class.

# Motivation

The original coverage store thread local showed up in the performance profiles.

# Additional Notes

As thread is one of the core classes and likely cannot be safely retransformed in a running JVM, the injection is done the following way:
- Maven/Gradle parent process determines which JVM is used for running the tests for a specific module
- the home folder for the JVM is determined
- compiled `java.lang.Thread` class is extracted from that JVM's files (depending on whether Java 8 or newer is used, it is either extracted from `rt.jar` or from the `java.base` module archive)
- the class is patched (dedicated coverage store field is injected) and is saved to a temporary folder on disk
- Maven/Gradle parent patches the command that is used for forking JVM that runs the test: depending on whether Java 8 or newer is used, the patched class is either prepended to bootstrap classpath or is used to patch the `java.base` module.

A static final `MethodHandle` is used for accessing the field. This is more practical than making the thread class implement an interface, which requires adding the interface to the `java.base` module and updating the module's list of exported packages.

Jira ticket: [SDTEST-533]

[SDTEST-533]: https://datadoghq.atlassian.net/browse/SDTEST-533?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ

### DIFF
--- a/dd-java-agent/agent-ci-visibility/src/main/java/datadog/trace/civisibility/config/JvmInfo.java
+++ b/dd-java-agent/agent-ci-visibility/src/main/java/datadog/trace/civisibility/config/JvmInfo.java
@@ -1,25 +1,60 @@
 package datadog.trace.civisibility.config;
 
 import datadog.trace.civisibility.ipc.Serializer;
+import de.thetaphi.forbiddenapis.SuppressForbidden;
 import java.nio.ByteBuffer;
+import java.nio.file.Path;
+import java.nio.file.Paths;
 import java.util.Objects;
+import org.slf4j.Logger;
+import org.slf4j.LoggerFactory;
 
 public class JvmInfo {
+
+  private static final Logger LOGGER =
+      LoggerFactory.getLogger(ModuleExecutionSettingsFactoryImpl.class);
 
   public static final JvmInfo CURRENT_JVM =
       new JvmInfo(
           System.getProperty("java.runtime.name"),
           System.getProperty("java.version"),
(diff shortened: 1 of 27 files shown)


mcculls (changes requested):
This seems like a fragile solution - I would want to see benchmarks for the same workload before and after to see how much benefit this solution brings vs the added complexity. What happens with virtual threads in Java 22?

Also I notice that previously you were adding and removing the thread-local entry.

This would have been constantly mutating the thread-local map which increases the overhead. It is sometimes better to store a holder object, such as a one-element array, in the thread-local and update that than call `ThreadLocal.remove`.

ie. have
```
ThreadLocal<CoverageProbeStore[]> COVERAGE_PROBE_STORE = COVERAGE_PROBE_STORE = ThreadLocal.withInitial(() -> new CoverageProbeStore[1]);

// ...

COVERAGE_PROBE_STORE.get()[0] = myStore;

// ...

COVERAGE_PROBE_STORE.get()[0] = null;
```

and then run that simpler holder solution through the same benchmark for comparison.

cc @dougqh

PR #7290, opened 2024-07-08
Spring scheduling: do not wrap runnables on each schedule

amarziali:
# What Does This Do

This PR rollbacks instrumenting spring scheduling `TaskScheduler` which was wrapping the `Runnable` in each schedule.
This was done as part of #6249  in order to have the right Runnable names displayed on actuator's endpoints. 

Wrapping the runnable on each schedule can cause side effects if a custom task scheduler is used and if inside the `schedule` method `instanceof` is used to check whenever the Runnable is of a certain type since we're wrapping it with `SpringSchedulingRunnableWrapper`. 

In order to be less invasive and, at same time, preserve the right labelling on actuator, I've restored the old approach instrumenting `Task` instead of `TaskScheduler` which was causing that bad behaviour. I've also overridden the `toString` method of `SpringSchedulingRunnableWrapper` in order to target the wrapped runnable instead. This will work for spring boot actuator since the displayed name is obtained calling `toString` (https://github.com/spring-projects/spring-boot/blob/69630bba37caa3a092b19392949705a41234d703/spring-boot-project/spring-boot-actuator/src/main/java/org/springframework/boot/actuate/scheduling/ScheduledTasksEndpoint.java#L285)
# Motivation

# Additional Notes

Jira ticket: [APMS-12578]

[APMS-12578]: https://datadoghq.atlassian.net/browse/APMS-12578?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ

### DIFF
--- a/dd-java-agent/instrumentation/spring-scheduling-3.1/src/main/java/datadog/trace/instrumentation/springscheduling/SpringSchedulingInstrumentation.java
+++ b/dd-java-agent/instrumentation/spring-scheduling-3.1/src/main/java/datadog/trace/instrumentation/springscheduling/SpringSchedulingInstrumentation.java
@@ -1,30 +1,24 @@
 package datadog.trace.instrumentation.springscheduling;
 
-import static datadog.trace.agent.tooling.bytebuddy.matcher.HierarchyMatchers.implementsInterface;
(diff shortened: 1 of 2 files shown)


mcculls (approved):
Much better, thanks!

PR #7294, opened 2024-07-09
Use spark application name when service is set to hadoop

paul-laffon-dd:
# What Does This Do

`DD_SPARK_APP_NAME_AS_SERVICE` also use spark.app.name as the default service if service is set to `hadoop`

# Motivation

USM can auto-infer the service to hadoop, in this case we still want to use the spark app name as the service

# Additional Notes

Follow up of: https://github.com/DataDog/dd-trace-java/pull/7252


yiliangzhou (approved):
LGTM

Issue #7291, opened 2024-07-08
Graphql-java 22 Support Still Not Working


surecloud-meason:
We are running the latest spring boot 3.3.1 & graphql 22. We have upgraded to the latest Datadog agent 1.36 and still we are unable to view GraphQL spans.

Issue was apparently fixed here https://github.com/DataDog/dd-trace-java/issues/7153 but it isnt working.

amarziali:
👋  @surecloud-meason thanks for having reported the issue and sorry for that. I opened another PR to fix it.

amarziali:
The PR fixing the issue has been merged into master. The target date for the next release should be beginning of August. In the meanwhile, if you'd like to have an early testing, I can provide the link of a CI build. Just let me know

surecloud-meason:
@amarziali please can you provide a CI build / docker image.

amarziali:
You can use the circle CI build of the PR [here](https://output.circle-artifacts.com/output/job/3f37fd25-d936-427f-a6bc-792d6797ab25/artifacts/0/libs/dd-java-agent-1.38.0-SNAPSHOT.jar)

Please do not hesitate to share a feedback

PR #7295, opened 2024-07-09
GraphQl: Fix advices for 22+

amarziali:
# What Does This Do

This PR fixes the missing tracing for 22+. The cause was due to advices that were not correctly applying hence the instrumentation was not installed. Also, I updated the lockfile because tests were executed with a wrong version and I also added latestDep for 20 and 21

# Motivation

# Additional Notes

See #7291 

Jira ticket: [PROJ-IDENT]


vandonr (approved):
👍 🚢

PerfectSlayer (approved):

PR #7299, opened 2024-07-09
Update URI and URL call sites for precise taint tracking

manuel-alvarez-alvarez:
# What Does This Do
Updates the URL and URI call sites to ensure that we keep tracking of the different ranges on a best effort basis.

# Motivation
SSRF vulnerability has different scores depending on the part of the url that is coming from an external source (e.g. the host is the most important part to track as it gives big room for attacks).

# Additional Notes

Jira ticket: [APPSEC-53838]

[APPSEC-53838]: https://datadoghq.atlassian.net/browse/APPSEC-53838?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ

### DIFF
--- a/dd-java-agent/agent-iast/src/main/java/com/datadog/iast/propagation/CodecModuleImpl.java
+++ b/dd-java-agent/agent-iast/src/main/java/com/datadog/iast/propagation/CodecModuleImpl.java
@@ -68,20 +68,39 @@ public class CodecModuleImpl implements CodecModule {
 
   @Override
   public void onUriCreate(@Nonnull final URI result, final Object... args) {
-    taintUrlIfAnyTainted(result, args);
+    final IastContext ctx = IastContext.Provider.get();
+    if (ctx == null) {
+      return;
+    }
+    taintUrlIfAnyTainted(ctx.getTaintedObjects(), result, args);
   }
 
   @Override
   public void onUrlCreate(@Nonnull final URL result, final Object... args) {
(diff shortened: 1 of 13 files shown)


dd-java-agent/agent-iast/src/main/java/com/datadog/iast/propagation/CodecModuleImpl.java
```diff
@@ -0,0 +1,101 @@
...
+    }
+    final TaintedObjects to = ctx.getTaintedObjects();
+    final String toString = url.toString();
+    final RangeBuilder builder = new RangeBuilder();
+    for (final Object arg : args) {
+      final TaintedObject tainted = to.get(arg);
+      if (tainted != null) {
+        final int offset = toString.indexOf(arg.toString());
+        if (offset >= 0) {
+          builder.add(tainted.getRanges(), offset);
```
manuel-alvarez-alvarez, line 92: There is the case where some `args` are tainted but we cannot find them in the `toString` representation. Maybe we should taint the whole object in that particular case, WDYT?

amarziali:
Removed IDM from the reviewers since it only impact  ASM

smola (approved):

jandro996 (approved):

Issue #7300, opened 2024-07-10
Outgoing http requests are not instrumented with datadog tracer - Spring Boot / RestClient / Eclipse Jetty Client 12.0.5


lokesh1729:
Hi,

Outgoing http requests are not instrumented with datadog tracer. I am using java, spring boot, restclient with jetty 12.0.5 as http client. Here's my configuration of restclient...

We have other microservices where we were using OkHttp3 and the tracing is working fine. We pull the latest tracer in our deployment configuration and the docker files are same for every microservice.

Is jetty http client not supported?

```java
@Bean
public RestClient httpClient() {
    HttpClient httpClient = new HttpClient() {
        @Override
        public Request newRequest(URI uri) {
            Request request = super.newRequest(uri);
            request = loggingService.logRequestResponse(request);
            return middleware.handle(request);
        }
    };
    httpClient.setConnectTimeout(1000);
    httpClient.setIdleTimeout(15000);
    JettyClientHttpRequestFactory requestFactory = new JettyClientHttpRequestFactory(httpClient);
    requestFactory.setReadTimeout(Duration.ofMillis(60000));

    return RestClient.builder()
            .baseUrl(baseUrl)
            .requestFactory(requestFactory)
            .defaultHeader(HttpHeaders.CONTENT_TYPE, MediaType.APPLICATION_JSON_VALUE)
            .defaultHeader(HttpHeaders.ACCEPT, MediaType.APPLICATION_JSON_VALUE)
            .build();
}
```

amarziali:
👋 Hi @lokesh1729 thanks for having opened the issue. Indeed jetty 12 is not yet supported. However we're (right now) working at a PR to fix it so it will be available shortly

lokesh1729:
Hi Andrea,Thanks for the update. Any rough timelines you can share? so that we will plan our work accordingly.- LOKESH SANAPALLI- https://lokesh1729.com    ---- On Wed, 10 Jul 2024 21:48:05 +0530  ***@***.******@***.***> wrote ----  
👋 Hi @lokesh1729 thanks for having opened the issue. Indeed jetty 12 is not yet supported. However we're (right now) working at a PR to fix it so it will be available shortly

amarziali:
The PR to support this version has just been merged. It will be available in the next release (1.38.0) foresee for start of August

PR #7305, opened 2024-07-10
Support jetty client 12

amarziali:
# What Does This Do

Instrument jetty client 12+

# Motivation

See #7300 

# Additional Notes

Jira ticket: [PROJ-IDENT]


vandonr (commented):
This is a bit hard to review... I can diff those files locally with the 10.0 instrumentation, but maybe you can point to the things you had to change for this instrumentation ?

nayeem-kamal (approved):
Overall LGTM. As @vandonr 's comment states. Maybe you could just add a bit in the description or in a comment about what the changes are from previous versions.

amarziali:
> This is a bit hard to review... I can diff those files locally with the 10.0 instrumentation, but maybe you can point to the things you had to change for this instrumentation ?

The request/response classes have a package change as well the way the span completion listener is applied differs

vandonr:
For a future case like this, what we can do is do a first commit where the files are just copied, and then a commit where the changes are done. This allows reviewing the changes by looking at the second commit only.

PR #7310, opened 2024-07-11
Add high rate queue for log template snapshots

jpbempel:
# What Does This Do
We introduce a dedicated high rate queue with dedicated threads with better reactivity (max 100ms between polls). All non-capturing snapshots are added into this queue.

# Motivation
Historically, the Queue inside SnapshotSink was designed for a low rate number of snapshots (1 per second) and the consumer thread is the one shared with all other tasks (AgentTaskScheduler.INSTANCE). So the reactivity of the thread to consume the queue is not enough for handling up to 5000 snapshots/s allowed for log template probes. 

# Additional Notes

Jira ticket: [DEBUG-2499]

[DEBUG-2499]: https://datadoghq.atlassian.net/browse/DEBUG-2499?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ

### DIFF
--- a/dd-java-agent/agent-debugger/src/main/java/com/datadog/debugger/sink/DebuggerSink.java
+++ b/dd-java-agent/agent-debugger/src/main/java/com/datadog/debugger/sink/DebuggerSink.java
@@ -1,85 +1,60 @@
 package com.datadog.debugger.sink;
 
-import com.datadog.debugger.agent.DebuggerAgent;
 import com.datadog.debugger.instrumentation.DiagnosticMessage;
 import com.datadog.debugger.uploader.BatchUploader;
 import com.datadog.debugger.util.DebuggerMetrics;
 import datadog.trace.api.Config;
 import datadog.trace.bootstrap.debugger.DebuggerContext;
 import datadog.trace.bootstrap.debugger.ProbeId;
-import datadog.trace.core.DDTraceCoreInfo;
 import datadog.trace.util.AgentTaskScheduler;
-import datadog.trace.util.TagsHelper;
 import java.util.List;
 import java.util.concurrent.TimeUnit;
-import java.util.stream.Collectors;
+import java.util.concurrent.atomic.AtomicLong;
 import org.slf4j.Logger;
 import org.slf4j.LoggerFactory;
 
 /** Collects data that needs to be sent to the backend: Snapshots, metrics and statuses */
 public class DebuggerSink {
-  private static final Logger log = LoggerFactory.getLogger(DebuggerSink.class);
+  private static final Logger LOGGER = LoggerFactory.getLogger(DebuggerSink.class);
   private static final double FREE_CAPACITY_LOWER_THRESHOLD = 0.25;
   private static final double FREE_CAPACITY_UPPER_THRESHOLD = 0.75;
-  private static final int MIN_FLUSH_INTERVAL = 100;
-  private static final int MAX_FLUSH_INTERVAL = 2000;
-  private static final long INITIAL_FLUSH_INTERVAL = 1000;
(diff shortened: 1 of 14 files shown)


PerfectSlayer (approved):
Looks good for the core part

dd-java-agent/agent-debugger/src/main/java/com/datadog/debugger/sink/SnapshotSink.java
```diff
@@ -1,68 +1,183 @@
...
+  private void backOffHighRateFlush() {
+    long interval = currentHighRateFlushInterval;
+    currentHighRateFlushInterval =
+        Math.min(interval + HIGH_RATE_STEP_SIZE, HIGH_RATE_MAX_FLUSH_INTERVAL);
+    if (interval != currentHighRateFlushInterval) {
+      highRateReschedule();
+    }
+  }
+
+  private void reconsiderHighRateFlushInterval(int snapshotCount) {
```
shatzi, line 119: not sure I follow the logic here.

From what it seems this function try to have interval where snapshotCount is around 10%-25% of capacity. not sure why this is the goal.

The way I think about it is that uploadPayloads take snapshots and create batches of them. I think we want interval that send 1 batch per flush. 

```
ratio = 1 / (snapshotCount / averageSnapshotsPerAFullBatch)
newInterval = oldInterval * ratio.
```

this will try to get the interval to be the time it take to collect enough snapshots to emit them in a single batch.

We can estimate averageSnapshotsPerAFullBatch as we can take the average json size of a log message. or just calculate it after every batch was generated.

shatzi (approved):
I love the idea of having two simple queues with different logic on interval adjustment. Very NICE!

dd-java-agent/agent-debugger/src/main/java/com/datadog/debugger/sink/DebuggerSink.java
```diff
@@ -143,70 +103,72 @@ public SymbolSink getSymbolSink() {
...
 
-  private void reschedule() {
-    AgentTaskScheduler.Scheduled<DebuggerSink> localScheduled = this.scheduled;
-    if (localScheduled != null) {
-      localScheduled.cancel();
-    }
-    this.scheduled =
-        AgentTaskScheduler.INSTANCE.scheduleAtFixedRate(
-            this::flush, this, currentFlushInterval, currentFlushInterval, TimeUnit.MILLISECONDS);
+  private void lowRateReschedule() {
```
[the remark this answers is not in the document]
jpbempel: yeah it's a complicated beast as the flushing thread for low rate Q is also used for flushing probe status & symbols

dd-java-agent/agent-debugger/src/main/java/com/datadog/debugger/sink/SnapshotSink.java
jpbempel: no the logic is not like. 
We try to adjust the interval between flush to avoid filling the queue when we are doing the serialization of tall the snapshots.
it's a compromise between polling frequently the queue but not too much if the rate of arrival is low, but then if we have high rate we want to decrease the flush interval to avoid dropping snapshots with a full queue.

This is a heuristic so not trying to be exact or something.

jpbempel: log template snapshots are samll enough to consider taking the whole queue at once. Worst case we are splitting into 2 batches, not a big deal.

PR #7311, opened 2024-07-11
Reset call depth counter if query is blocked

ValentinZakharov:
# What Does This Do
Reset call depth counter in `StatementInstrumentation` if query with SQL-injection is blocked.

# Motivation
In case of multiple blocking of sql queries, an error occurred leading to blocking failure

# Additional Notes

Jira ticket: [APPSEC-47228]

[APPSEC-47228]: https://datadoghq.atlassian.net/browse/APPSEC-47228?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ

### DIFF
--- a/dd-java-agent/instrumentation/jdbc/src/main/java/datadog/trace/instrumentation/jdbc/StatementInstrumentation.java
+++ b/dd-java-agent/instrumentation/jdbc/src/main/java/datadog/trace/instrumentation/jdbc/StatementInstrumentation.java
@@ -120,6 +120,7 @@ public final class StatementInstrumentation extends InstrumenterModule.Tracing
         // if we can't get the connection for any reason
         return null;
       } catch (BlockingException e) {
+        CallDepthThreadLocalMap.reset(Statement.class);
         // re-throw blocking exceptions
         throw e;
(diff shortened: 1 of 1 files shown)


dd-java-agent/instrumentation/jdbc/src/main/java/datadog/trace/instrumentation/jdbc/StatementInstrumentation.java
```diff
@@ -120,6 +120,7 @@ public static AgentScope onEnter(
         // if we can't get the connection for any reason
         return null;
       } catch (BlockingException e) {
+        CallDepthThreadLocalMap.reset(Statement.class);
```
amarziali, line 123: I would have instead put `CallDepthThreadLocalMap.decrement(Statement.class)` at first line of `stopSpan` method for the simple reason that the code is way more readeable this way.

amarziali (approved):
LGTM. I left a suggestion to make it more readable

PR #7312, opened 2024-07-12
[PoC] AWS Payload observability

ygree:
# What Does This Do

# Motivation

# Additional Notes

Jira ticket: [PROJ-IDENT]


gradle/libs.versions.toml
```diff
@@ -26,6 +26,7 @@ commons = "3.2"
 mockito = '4.4.0'
 jctools = '3.3.0'
 moshi = '1.11.0'
+jsonPath = '2.9.0'
```
vandonr, line 29: jsonPath 2.9.0 has a dependency on slf4j v2.0.11 (see pom [here](https://repo1.maven.org/maven2/com/jayway/jsonpath/json-path/2.9.0/json-path-2.9.0.pom)), while the version that we are using (see line 2) is 1.7.30.
This is what makes a lot of tests fail.
I'm not sure how this should be handled.

dd-java-agent/instrumentation/aws-java-sdk-2.2/src/main/java/datadog/trace/instrumentation/aws/v2/AwsSdkClientDecorator.java
```diff
@@ -294,7 +307,20 @@ public AgentSpan onSdkResponse(
...
+      final ExecutionAttributes attributes,
+      Optional<InputStream> responseBody) {
+
+    if (responseBody.isPresent()) {
+      InputStream body = responseBody.get();
+      if (body instanceof ResponseBodyStreamWrapper) {
+        ResponseBodyStreamWrapper wrapper = (ResponseBodyStreamWrapper) body;
+        ByteArrayInputStream bodyStream = wrapper.asByteArrayInputStream();
+        if (bodyStream != null) {
+          AgentTracer.get().addTagsFromResponseBody(span, bodyStream, "aws.response.body");
```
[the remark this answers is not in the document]
ygree: According to the RFC, the idea was to generalize to any service (not just AWS) and be able to convert request and response body to tags for observability purposes.

I totally agree with your concerns about overhead, so I tried to limit parsing as much as possible by
- Avoiding materializing JSON and instead using event-based parsing to traverse in one go.
- Buffering only payloads that start with '{'.
- Having an upper limit of 256Kb for the buffer.
- Disabled by default.
- Limit the number of services it's enabled for (now only SQS, ApiGateway and EventBridge).

Also, so far we are going to release it internally for the Cloud Observability team and there is no intention to release it to GA at all.

Regarding your comment about services not using JSON, this is exactly what came up in testing I did recently. The format used depends not only on the service, but even on the minor version of the SDK. For example, SNS uses Http Form Params and XML, some earlier versions of the SDK for SQS also use XML instead of JSON. And as you mentioned, Kinesis uses CBOR. 

I appreciate your efforts to experiment with an alternative approach and provide a very detailed idea of how to implement it. I think we should have a specialized version for AWS SDK to use the provided POJOs as you suggested. That way we can optimize drastically for AWS SDK. And at the same time expand the visibility to AWS services that don't use JSON. 

Thank you!

dd-java-agent/instrumentation/aws-java-sdk-2.2/build.gradle
```diff
@@ -61,3 +77,15 @@ dependencies {
   latestDepTestImplementation group: 'software.amazon.awssdk', name: 'dynamodb', version: fixedSdkVersion
   latestDepTestImplementation group: 'software.amazon.awssdk', name: 'kinesis', version: fixedSdkVersion
 }
+
+tasks.withType(Test).configureEach {
+  usesService(testcontainersLimit)
+}
+
+tasks.named("test").configure {
```
amarziali, line 85: this (and the things below) should not be required

dd-java-agent/instrumentation/aws-java-sdk-2.2/src/main/java/datadog/trace/instrumentation/aws/v2/AwsSdkClientDecorator.java
```diff
@@ -110,11 +113,21 @@ public AgentSpan onSdkRequest(
...
     final String awsServiceName = attributes.getAttribute(SdkExecutionAttribute.SERVICE_NAME);
     final String awsOperationName = attributes.getAttribute(SdkExecutionAttribute.OPERATION_NAME);
     onOperation(span, awsServiceName, awsOperationName);
 
+    if (requestBody.isPresent()) {
+      Config config = Config.get();
+      if (config.isCloudRequestPayloadTaggingEnabled()
+          && config.isCloudPayloadTaggingEnabledFor(awsServiceName)) {
+        InputStream body = requestBody.get().contentStreamProvider().newStream();
+        AgentTracer.get().addTagsFromRequestBody(span, body, "aws.request.body");
```
amarziali, line 127: Theoretically we could have done it lazily in the agent thread (as tag postprocessor) so it would not affect the same thread of the aws request (i.e. only attach the buffer to the span and process on a TagPostProcessor). It would also have avoided the fact to expose those methods on the agent tracer

amarziali (approved):
There are improvements we can do specifically for AWS that would not require any json parsing that we should consider. Also, the json parsing can be do lazily. Since the feature is experimental and not enabled by default, I would not block from approval. However we should consider to implement such improvements.

dd-java-agent/instrumentation/aws-java-sdk-2.2/build.gradle
ygree: Without that it won't run `payloadTaggingTest` when `./gradlew :dd-java-agent:instrumentation:aws-java-sdk-2.2:test` is called.

amarziali: The CI runs `allTests` in any case and those will be executed

PR #7323, opened 2024-07-12
Fix session rewriting false positives

jandro996:
# What Does This Do

- Remove Session Tracking Modes check calls from servlet instrumentation
- Add IAST instrumentation to HttpSession#getSession with calls to Session Tracking Modes check

Solved for javax and jakarta implementations

# Motivation

The current implementation for detecting Session Rewriting is causing false positives because it only considers the servlet configuration. This may be insecure, but if the session is not being used, an Session Rewriting should not be reported.

# Additional Notes

Jira ticket: [APPSEC-53722]

[APPSEC-53722]: https://datadoghq.atlassian.net/browse/APPSEC-53722?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ


dd-java-agent/instrumentation/servlet/request-3/src/main/java/datadog/trace/instrumentation/servlet3/IastHttpServletRequest3Instrumentation.java
```diff
@@ -0,0 +1,110 @@
...
+    return implementsInterface(named(hierarchyMarkerType()))
+        // ignore wrappers that ship with servlet-api
+        .and(namedNoneOf("javax.servlet.http.HttpServletRequestWrapper"))
+        .and(not(extendsClass(named("javax.servlet.http.HttpServletRequestWrapper"))));
+  }
+
+  @Override
+  public void methodAdvice(MethodTransformer transformer) {
+    transformer.applyAdvice(
+        named("getSession").and(returns(named("javax.servlet.http.HttpSession"))).and(isPublic()),
```
jandro996, line 65: There are two het session methods 

javax.servlet.http.HttpSession#getSession()
javax.servlet.http.HttpSession#getSession(boolean)

In all the checked implementations getSession() calls getSession(true) so maybe it's better to only instrument the javax.servlet.http.HttpSession#getSession(boolean) 

Any way we are using the Instrumentation context to only call the iast module once per application

dd-java-agent/instrumentation/servlet/request-5/src/main/java/datadog/trace/instrumentation/servlet5/IastOptOutJakartaHttpServletRequestInstrumentation.java
```diff
@@ -0,0 +1,98 @@
...
+
+  @Override
+  public ElementMatcher<TypeDescription> hierarchyMatcher() {
+    return implementsInterface(named(hierarchyMarkerType()))
+        .and(not(WRAPPER_CLASS))
+        .and(not(extendsClass(WRAPPER_CLASS)));
+  }
+
+  @Override
+  protected boolean isOptOutEnabled() {
```
jandro996, line 50: Although we already have a JakartaHttpServletRequestInstrumentation for IAST we need another one due to the session rewriting is an opt-out feature

manuel-alvarez-alvarez: Isn't it already opt-out?

manuel-alvarez-alvarez: Isn't it already opt-out?

dd-java-agent/instrumentation/servlet/request-3/src/main/java/datadog/trace/instrumentation/servlet3/IastHttpServletRequest3Instrumentation.java
manuel-alvarez-alvarez: I'm worried about the possibility of a framework probing for the existence of a session (setting the flag to false) and we might trigger the vulnerability here. 

What do you think if:

- instrument only the variant taking a boolean
- if it's called with `false` and no session is created don't update the state so the instrumentation will enter next time

jandro996: As we are checking if the session exist at the begining of the advice this should not happens, but I will go for your approach as it would improve the performance in some cases

manuel-alvarez-alvarez (approved):

PR #7328, opened 2024-07-15
Fix EL function behavior for null values

jpbempel:
# What Does This Do
Change behavior to send evaluation error when value is null for
- len/count
- hasAny
- hasAll
- isEmpty
- index
- string predicates (startsWith, endsWith, contains, matches)

# Motivation
Conformance with other client libraries

# Additional Notes

Jira ticket: [DEBUG-2487]

[DEBUG-2487]: https://datadoghq.atlassian.net/browse/DEBUG-2487?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ

### DIFF
--- a/dd-java-agent/agent-debugger/debugger-el/src/main/java/com/datadog/debugger/el/expressions/HasAnyExpression.java
+++ b/dd-java-agent/agent-debugger/debugger-el/src/main/java/com/datadog/debugger/el/expressions/HasAnyExpression.java
@@ -1,5 +1,7 @@
 package com.datadog.debugger.el.expressions;
 
+import com.datadog.debugger.el.EvaluationException;
+import com.datadog.debugger.el.PrettyPrintVisitor;
 import com.datadog.debugger.el.Value;
 import com.datadog.debugger.el.Visitor;
 import com.datadog.debugger.el.values.ListValue;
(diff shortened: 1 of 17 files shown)


dd-java-agent/agent-debugger/debugger-el/src/main/java/com/datadog/debugger/el/expressions/HasAnyExpression.java
```diff
@@ -24,15 +26,17 @@ public HasAnyExpression(
   @Override
   public Boolean evaluate(ValueReferenceResolver valueRefResolver) {
     if (valueExpression == null) {
-      return Boolean.FALSE;
+      throw new EvaluationException(
+          "Cannot evaluate the expression for null value", PrettyPrintVisitor.print(this));
     }
     Value<?> value = valueExpression.evaluate(valueRefResolver);
     if (value.isUndefined()) {
```
shatzi, line 33: nit: we now have multiple locations that check for isUndefined and isNull and throw the same error. maybe we can create throwIfNullOrUndefined(value) - so we don't need to repeat this multiple times.

dd-java-agent/agent-debugger/debugger-el/src/test/java/com/datadog/debugger/el/ProbeConditionTest.java
```diff
@@ -265,7 +269,21 @@ class Obj {
     assertTrue(probeCondition.execute(ctx));
   }
 
-  private static ProbeCondition load(String resourcePath) throws IOException {
+  @Test
+  public void nullExpressions() {
```
shatzi, line 273: lets check undefined as well?

shatzi (approved):

dd-java-agent/agent-debugger/debugger-el/src/test/java/com/datadog/debugger/el/ProbeConditionTest.java
jpbempel: undefined is something I feel is now redundant. 
In the future I think I would like to remove it

dd-java-agent/agent-debugger/debugger-el/src/main/java/com/datadog/debugger/el/expressions/HasAnyExpression.java
jpbempel: will revisit when getting rid of undefined

PR #7333, opened 2024-07-16
Fix github artifact upload using explicit jar names

randomanderson:
# What Does This Do

Fixes the github artifact upload by using explicit file names

# Motivation
The previous way of using `*.jar` is broken because there are `-source.jar` and `-javadoc.jar` when using artifacts from the `deploy_to_sonotype` job. We know the exact name of the files because it is only run when there is an explicit `CI_COMMIT_TAG` set.

### DIFF
--- a/.gitlab-ci.yml
+++ b/.gitlab-ci.yml
@@ -236,11 +236,9 @@ deploy_artifacts_to_github:
   rules:
     - if: '$POPULATE_CACHE'
       when: never
-    - if: '$CI_COMMIT_TAG =~ /^v.*/'
+    - if: '$CI_COMMIT_TAG =~ /^v[0-9]+\.[0-9]+\.[0-9]+$/'
       when: on_success
-    - when: manual
-      allow_failure: true
-  # Requires the deploy_to_sonatype job to have run first as build cache is broken
+  # Requires the deploy_to_sonatype job to have run first the UP-TO-DATE gradle check across jobs is broken
   # This will deploy the artifacts built from the publishToSonatype task to the GitHub release
   needs:
     - job: deploy_to_sonatype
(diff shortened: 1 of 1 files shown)


.gitlab-ci.yml
```diff
@@ -236,11 +236,9 @@ deploy_artifacts_to_github:
   rules:
     - if: '$POPULATE_CACHE'
       when: never
-    - if: '$CI_COMMIT_TAG =~ /^v.*/'
+    - if: '$CI_COMMIT_TAG =~ /^v[0-9]+\.[0-9]+\.[0-9]+$/'
```
randomanderson, line 239: This needs to exactly match the `deploy_to_sonotype` condition above

.gitlab-ci.yml
```diff
@@ -236,11 +236,9 @@ deploy_artifacts_to_github:
...
       when: on_success
-    - when: manual
```
randomanderson, line 241: `manual` doesn't make sense because we use `CI_COMMIT_TAG` in the commands

.gitlab-ci.yml
```diff
@@ -250,10 +248,11 @@ deploy_artifacts_to_github:
     - aws ssm get-parameter --region us-east-1 --name ci.dd-trace-java.gh_release_token --with-decryption --query "Parameter.Value" --out text > github-token.txt
     - gh auth login --with-token < github-token.txt
     - gh auth status  # Maybe helpful to have this output in logs?
-    - cp workspace/dd-java-agent/build/libs/dd-java-agent-*.jar workspace/dd-java-agent/build/libs/dd-java-agent.jar # we upload two filenames
-    - gh release upload --clobber --repo DataDog/dd-trace-java $CI_COMMIT_TAG workspace/dd-java-agent/build/libs/*.jar
-    - gh release upload --clobber --repo DataDog/dd-trace-java $CI_COMMIT_TAG workspace/dd-trace-api/build/libs/*.jar
-    - gh release upload --clobber --repo DataDog/dd-trace-java $CI_COMMIT_TAG workspace/dd-trace-ot/build/libs/*.jar
+    - cp workspace/dd-java-agent/build/libs/dd-java-agent-${CI_COMMIT_TAG}.jar workspace/dd-java-agent/build/libs/dd-java-agent.jar # we upload two filenames
```
PerfectSlayer, line 251: Isn't CI commit tag `v1.37.0` where artifact name are `dd-java-agent-1.37.0`?
There is no `v` prefix in front of the version number.

.gitlab-ci.yml
```diff
@@ -250,10 +248,11 @@ deploy_artifacts_to_github:
...
+    - gh release upload --clobber --repo DataDog/dd-trace-java $CI_COMMIT_TAG workspace/dd-java-agent/build/libs/dd-java-agent.jar
+    - gh release upload --clobber --repo DataDog/dd-trace-java $CI_COMMIT_TAG workspace/dd-java-agent/build/libs/dd-java-agent-${CI_COMMIT_TAG}.jar
+    - gh release upload --clobber --repo DataDog/dd-trace-java $CI_COMMIT_TAG workspace/dd-trace-api/build/libs/dd-trace-api-${CI_COMMIT_TAG}.jar
+    - gh release upload --clobber --repo DataDog/dd-trace-java $CI_COMMIT_TAG workspace/dd-trace-ot/build/libs/dd-trace-ot-${CI_COMMIT_TAG}.jar
```
PerfectSlayer, line 255: `gh release upload` is able to take multiple files from the same command.
Is it better to make a single call or multiple ones? (I honestly don't know)

PerfectSlayer (commented):
There might be an issue with the version name.

.gitlab-ci.yml
randomanderson: Good catch

.gitlab-ci.yml
randomanderson: I think it's clearer to have multiple lines. I don't think there's a difference functionally

.gitlab-ci.yml
randomanderson: Fixed

PerfectSlayer (approved):
Looks good now!
Don't forget to squash on merge 👍

PR #7335, opened 2024-07-17
[🍒 7325] Fix tracer freeze when CI Visibility is enabled

nikita-tkachenko-datadog:
# What Does This Do

Changes `TraceProcessingWorker#serializerThread` to always be a daemon thread (before the changes the thread was started as non-daemon if CI Visibility was enabled).

# Motivation

This thread runs an infinite loop, polling traces that are ready to be processed.
If it is not daemon the JVM will never exit, unless the thread is interrupted or `System#exit` is called.
Maven and Gradle do the latter, which is why the builds do not hang.
However, if CI Visibility is enabled in an app that does not call `System#exit`, the tracer will never exit.
The easiest way to reproduce is tracing something simple like `java -version`.

# Additional Notes

The reason for making this thread non-daemon was to allow it to flush all the pending traces before the build finishes.
However, this is redundant because the tracer shutdown sequence flushes the traces and waits for them to finish:

- `CoreTracer` creates and registers `ShutdownHook`
- Calling `System.exit` will trigger `ShutdownHook`
- `ShutdownHook` calls `CoreTracer#close`
- `CoreTracer#close` calls `Writer#flush`
- `Writer#flush` calls `TraceProcessingWorker#flush`
- `TraceProcessingWorker#flush` creates a `FlushEvent` and waits up to `RemoteWriter#flushTimeout` for the flush to finish

Jira ticket: [SDTEST-569]

### DIFF
--- a/dd-trace-core/src/main/java/datadog/trace/common/writer/DDAgentWriter.java
+++ b/dd-trace-core/src/main/java/datadog/trace/common/writer/DDAgentWriter.java
@@ -39,6 +39,8 @@ public class DDAgentWriter extends RemoteWriter {
     Monitoring monitoring = Monitoring.DISABLED;
     boolean traceAgentV05Enabled = Config.get().isTraceAgentV05Enabled();
     boolean metricsReportingEnabled = Config.get().isTracerMetricsEnabled();
+    private int flushTimeout = 1;
+    private TimeUnit flushTimeoutUnit = TimeUnit.SECONDS;
     boolean alwaysFlush = false;
 
     private DDAgentApi agentApi;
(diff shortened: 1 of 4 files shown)


mcculls (approved):

dd-trace-core/src/main/java/datadog/trace/common/writer/DDAgentWriter.java
```diff
@@ -157,15 +165,23 @@ public DDAgentWriter build() {
...
 
   DDAgentWriter(
       TraceProcessingWorker worker,
       PayloadDispatcher dispatcher,
       HealthMetrics healthMetrics,
+      int flushTimeout,
+      TimeUnit flushTimeoutUnit,
       boolean alwaysFlush) {
-    super(worker, dispatcher, healthMetrics, alwaysFlush);
+    super(worker, dispatcher, healthMetrics, flushTimeout, flushTimeoutUnit, alwaysFlush);
```
PerfectSlayer, line 185: Would be easier to use a `Duration` here?

nikita-tkachenko-datadog: I'd keep it as is for now, given that this is a backport of what was already merged to master, and that the other writers use the same approach (timeout+timeUnit).
I'll try to remember this if I touch this code again.

PR #7337, opened 2024-07-17
Extend `contains` EL expression

jpbempel:
# What Does This Do
Add support for array/list/map/set of the `contains` EL expression when calling `contains` on array/list or set it will iterate on all elements to find it. On Maps we are using the `containsKey` method.

# Motivation

# Additional Notes

Jira ticket: [DEBUG-2488]

[DEBUG-2488]: https://datadoghq.atlassian.net/browse/DEBUG-2488?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ

### DIFF
--- a/dd-java-agent/agent-debugger/debugger-el/src/main/java/com/datadog/debugger/el/expressions/ContainsExpression.java
+++ b/dd-java-agent/agent-debugger/debugger-el/src/main/java/com/datadog/debugger/el/expressions/ContainsExpression.java
@@ -1,15 +1,61 @@
 package com.datadog.debugger.el.expressions;
 
+import com.datadog.debugger.el.EvaluationException;
+import com.datadog.debugger.el.PrettyPrintVisitor;
+import com.datadog.debugger.el.Value;
 import com.datadog.debugger.el.Visitor;
-import com.datadog.debugger.el.values.StringValue;
+import com.datadog.debugger.el.values.CollectionValue;
+import datadog.trace.bootstrap.debugger.el.ValueReferenceResolver;
 
-public class ContainsExpression extends StringPredicateExpression {
-  public ContainsExpression(ValueExpression<?> sourceString, StringValue str) {
-    super(sourceString, str, String::contains, "contains");
+public class ContainsExpression implements BooleanExpression {
+  private final ValueExpression<?> target;
+  private final ValueExpression<?> value;
+
+  public ContainsExpression(ValueExpression<?> target, ValueExpression<?> value) {
+    this.target = target != null ? target : ValueExpression.NULL;
+    this.value = value;
+  }
+
+  @Override
+  public Boolean evaluate(ValueReferenceResolver valueRefResolver) {
+    Value<?> targetValue = target.evaluate(valueRefResolver);
+    if (targetValue.isUndefined()) {
+      throw new EvaluationException(
+          "Cannot evaluate the expression for undefined value", PrettyPrintVisitor.print(this));
+    }
+    if (targetValue.isNull()) {
+      throw new EvaluationException(
+          "Cannot evaluate the expression for null value", PrettyPrintVisitor.print(this));
+    }
+    Value<?> val = value.evaluate(valueRefResolver);
+    if (val.isUndefined() || val.isNull()) {
+      return false;
+    }
+    if (targetValue.getValue() instanceof String) {
+      String targetStr = (String) targetValue.getValue();
+      if (val.getValue() instanceof String) {
+        String valStr = (String) val.getValue();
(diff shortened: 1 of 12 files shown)


dd-java-agent/agent-debugger/debugger-el/src/main/java/com/datadog/debugger/el/expressions/ContainsExpression.java
```diff
@@ -1,15 +1,61 @@
...
+    if (targetValue.isUndefined()) {
+      throw new EvaluationException(
+          "Cannot evaluate the expression for undefined value", PrettyPrintVisitor.print(this));
+    }
+    if (targetValue.isNull()) {
+      throw new EvaluationException(
+          "Cannot evaluate the expression for null value", PrettyPrintVisitor.print(this));
+    }
+    Value<?> val = value.evaluate(valueRefResolver);
+    if (val.isUndefined() || val.isNull()) {
```
shatzi, line 31: do we want to fail for undefined or null?

For example. maybe the user would like to do contains(array, null). 

I'm not sure about undefined...

dd-java-agent/agent-debugger/debugger-el/src/main/java/com/datadog/debugger/el/values/ListValue.java
```diff
@@ -130,6 +130,87 @@ public Value<?> get(int index) {
...
+    }
+    if (arrayHolder != null) {
+      int count = Array.getLength(arrayHolder);
+      if (arrayType.isPrimitive()) {
+        if (val.getValue() == null) {
+          return false;
+        }
+        if (arrayType == byte.class) {
+          for (int i = 0; i < count; i++) {
+            if (Array.getByte(arrayHolder, i) == (Byte) val.getValue()) {
```
shatzi, line 150: should we get (Byte) val.getValue() out of the loop?

dd-java-agent/agent-debugger/debugger-el/src/main/java/com/datadog/debugger/el/values/ListValue.java
```diff
@@ -130,6 +130,87 @@ public Value<?> get(int index) {
...
+            if (Array.getBoolean(arrayHolder, i) == (Boolean) val.getValue()) {
+              return true;
+            }
+          }
+        }
+        return false;
+      }
+      for (int i = 0; i < count; i++) {
+        Object cellValue = Array.get(arrayHolder, i);
+        if (cellValue != null && cellValue.equals(val.getValue())) {
```
shatzi, line 205: this introduce side effects. can we call "equals" at all?
Maybe we should limit contains to be on primitives only?

dd-java-agent/agent-debugger/debugger-el/src/main/java/com/datadog/debugger/el/values/MapValue.java
```diff
@@ -101,10 +101,24 @@ public Value<?> get(Object key) {
     return (Value<?>) mapHolder;
   }
 
+  @Override
+  public boolean contains(Value<?> val) {
+    if (mapHolder instanceof Map) {
+      if (WellKnownClasses.isSafe((Map<?, ?>) mapHolder)) {
+        Map<?, ?> map = (Map<?, ?>) mapHolder;
+        return map.containsKey(val.getValue());
```
shatzi, line 109: I assume that containsKey will also call "equals" on keys. This can introduce side effect. but maybe we should Ok with this as those collections works at production so calling equals should be ok?

shatzi (approved):

dd-java-agent/agent-debugger/debugger-el/src/main/java/com/datadog/debugger/el/expressions/ContainsExpression.java
jpbempel: added support for this

dd-java-agent/agent-debugger/debugger-el/src/main/java/com/datadog/debugger/el/values/ListValue.java
jpbempel: done

dd-java-agent/agent-debugger/debugger-el/src/main/java/com/datadog/debugger/el/values/ListValue.java
jpbempel: right, added checks

PR #7345, opened 2024-07-18
Add Untrusted Deserialization vulnerability

Mariovido:
# What Does This Do
Add Untrusted Deserialization vulnerability

# Motivation
Add a new vulnerability

# Additional Notes

Jira ticket: [APPSEC-17165]

[APPSEC-17165]: https://datadoghq.atlassian.net/browse/APPSEC-17165?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ

### DIFF
--- a/dd-java-agent/agent-iast/src/test/groovy/com/datadog/iast/sink/UntrustedDeserializationModuleTest.groovy
+++ b/dd-java-agent/agent-iast/src/test/groovy/com/datadog/iast/sink/UntrustedDeserializationModuleTest.groovy
@@ -1,0 +1,54 @@
+package com.datadog.iast.sink
+
+import com.datadog.iast.IastModuleImplTestBase
+import com.datadog.iast.Reporter
+import com.datadog.iast.model.Source
+import com.datadog.iast.model.Vulnerability
+import com.datadog.iast.model.VulnerabilityType
+import com.datadog.iast.taint.Ranges
+import datadog.trace.api.iast.SourceTypes
+import datadog.trace.api.iast.sink.UntrustedDeserializationModule
+
+class UntrustedDeserializationModuleTest extends IastModuleImplTestBase {
+
+  private UntrustedDeserializationModule module
+
+  def setup() {
+    module = new UntrustedDeserializationModuleImpl(dependencies)
+  }
+
+  @Override
+  protected Reporter buildReporter() {
+    return Mock(Reporter)
+  }
+
+  void 'test null value'() {
+    when:
+    module.onInputStream(null)
+
+    then:
+    0 * _
+  }
+
+  void 'test untrusted deserialization detection' () {
+    setup:
+    def inputStream = Mock(InputStream)
+
+    when:
+    module.onInputStream(inputStream)
+
+    then: 'without tainted input stream'
+    0 * reporter.report(_, _)
+
+    when:
+    taint(inputStream)
+    module.onInputStream(inputStream)
+
+    then: 'with tainted input stream'
+    1 * reporter.report(_, { Vulnerability vul -> vul.type == VulnerabilityType.UNTRUSTED_DESERIALIZATION})
+  }
+
+  private void taint(final Object value) {
+    ctx.getTaintedObjects().taint(value, Ranges.forObject(new Source(SourceTypes.REQUEST_BODY, 'name', value.toString())))
+  }
+}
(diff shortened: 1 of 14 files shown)


dd-java-agent/agent-iast/src/test/groovy/com/datadog/iast/sink/UntrustedDeserializationModuleTest.groovy
```diff
@@ -0,0 +1,54 @@
...
+    when:
+    taint(inputStream)
+    module.onInputStream(inputStream)
+
+    then: 'with tainted input stream'
+    1 * reporter.report(_, { Vulnerability vul -> vul.type == VulnerabilityType.UNTRUSTED_DESERIALIZATION})
+  }
+
+  private void taint(final Object value) {
+    ctx.getTaintedObjects().taint(value, Ranges.forObject(new Source(SourceTypes.REQUEST_BODY, 'name', value.toString())))
```
jandro996, line 52:
```suggestion
    ctx.getTaintedObjects().taint(value, [] as Range[])
```

jandro996 (approved):
LGTM

manuel-alvarez-alvarez (approved):
LGTM

I think we can expand the smoke testing in a separate PR to ensure we are not losing propagation in different scenarios:

1. Other smoke tests than spring boot as we might be losing propagation of the body (e.g. vertx, jetty, ...)
2. Multipart uploads with multiple binaries (using servlet or other APIs like commons file upload)

They are not directly related to untrusted deserialization but they really influence the rule :sweat: 

PS: Nice first PR!

dd-java-agent/instrumentation/java-io/src/main/java/datadog/trace/instrumentation/java/lang/ObjectInputStreamCallSite.java
```diff
@@ -0,0 +1,27 @@
...
+import datadog.trace.api.iast.IastCallSites;
+import datadog.trace.api.iast.InstrumentationBridge;
+import datadog.trace.api.iast.Sink;
+import datadog.trace.api.iast.VulnerabilityTypes;
+import datadog.trace.api.iast.sink.UntrustedDeserializationModule;
+import java.io.InputStream;
+
+@Sink(VulnerabilityTypes.UNTRUSTED_DESERIALIZATION)
+@CallSite(spi = IastCallSites.class)
+public class ObjectInputStreamCallSite {
```
anderruiz, line 13: Do we plan to fix this later? technically the vulnerability happens when readObject is executed. Shouldn't we follow the same pattern that @manuel-alvarez-alvarez started to taint the input stream here but then have the vulnerability when it is loaded. I am also worried about the posible FP that we may have, we were previously only launching the vulnerability on readObject and whenever a banned class was loaded or a process execution was reached

dd-smoke-tests/iast-util/src/main/java/datadog/smoketest/springboot/controller/IastWebController.java
```diff
@@ -394,6 +395,12 @@ public String headerInjectionRedaction(
     return "Ok";
   }
 
+  @GetMapping("/untrusted_deserialization")
+  public String untrustedDeserialization(HttpServletRequest request) throws IOException {
+    ObjectInputStream ois = new ObjectInputStream(request.getInputStream());
```
anderruiz, line 400: This for example should not cause a UD vulnerabilities

anderruiz (commented):
Left some comments

dd-java-agent/instrumentation/java-io/src/main/java/datadog/trace/instrumentation/java/lang/ObjectInputStreamCallSite.java
Mariovido: We have discussed it internally and have decided to follow the current approach and address it in another independent PR

dd-smoke-tests/iast-util/src/main/java/datadog/smoketest/springboot/controller/IastWebController.java
Mariovido: Same as above :)

PR #7366, opened 2024-07-29
Add Parameter to only inject data jobs for particular java commands

paul-laffon-dd:
# What Does This Do

Add parameter `dd.data.jobs.command.pattern` to only inject the tracer on specific command pattern for Data Jobs

In the context of Data Jobs Monitoring, change the Java tracer to become more selective about which Java process to instrument when Data Jobs is enabled and a Java command regex pattern is specified. In runtime environment where we are not sure the command pattern to use, the java tracer should just behave as before by continuing the Agent start.

# Motivation
With single step instrumentation, Java tracer is injected for all Java processes, this is causing some initial system overhead as Java tracer starts in those JVMs. In resource limited environment, this can cause the host to be replaced due to high system load. For Data Jobs Monitoring, the Java tracer only cares about Spark process running the driver and executors, this newly added command pattern allows us to exit the Agent early if we don't think the process is Spark related and saving the overhead of increased System load.

# Additional Notes
* The environment variable to specify the data jobs Java command pattern (regex):`DD_DATA_JOBS_COMMAND_PATTERN`. 
* `DD_DATA_JOBS_COMMAND_PATTERN` only takes effects if `DD_DATA_JOBS_ENABLED` is set to `true`.

# Contributor Checklist

- [ ] Format the title [according the contribution guidelines](/CONTRIBUTING.md#title-format)
- [ ] Assign the `type:` and (`comp:` or `inst:`) labels in addition to [any usefull labels](/CONTRIBUTING.md#labels)

Jira ticket: [PROJ-IDENT]

### DIFF
--- a/dd-java-agent/agent-bootstrap/src/main/java/datadog/trace/bootstrap/Agent.java
+++ b/dd-java-agent/agent-bootstrap/src/main/java/datadog/trace/bootstrap/Agent.java
@@ -204,11 +204,11 @@ public class Agent {
 
     boolean dataJobsEnabled = isFeatureEnabled(AgentFeature.DATA_JOBS);
     if (dataJobsEnabled) {
-      String javaCommand = System.getProperty("sun.java.command");
-      if (isDataJobsSupported(javaCommand)) {
-        log.error(
-            "Data Jobs Monitoring is not compatible with non-spark command {}. dd-trace-java will not be installed",
-            javaCommand);
+      if (!isDataJobsSupported()) {
+        log.warn(
+            "Data Jobs Monitoring is not compatible with non-spark command {} based on command pattern {}. dd-trace-java will not be installed",
+            System.getProperty("sun.java.command"),
+            Config.get().getDataJobsCommandPattern());
         return;
       }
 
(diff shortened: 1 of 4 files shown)


dd-java-agent/agent-bootstrap/src/main/java/datadog/trace/bootstrap/Agent.java
```diff
@@ -204,6 +204,14 @@ public static void start(final Instrumentation inst, final URL agentJarURL, Stri
 
     boolean dataJobsEnabled = isFeatureEnabled(AgentFeature.DATA_JOBS);
     if (dataJobsEnabled) {
+      if (!isDataJobsSupported()) {
```
dougqh, line 207: We need to make sure that that we update this to work with the new bootstrap telemetry reporting, too.
If this goes in before my PR, I can make the necessary changes.

dd-java-agent/agent-bootstrap/src/main/java/datadog/trace/bootstrap/Agent.java
```diff
@@ -1307,4 +1315,17 @@ private static boolean isJFRSupported() {
...
+    String javaCommand = System.getProperty("sun.java.command");
+    String dataJobsCommandPattern = Config.get().getDataJobsCommandPattern();
+
+    if (null == javaCommand || null == dataJobsCommandPattern) {
+      // if sun.java.command somehow is not set or data jobs command pattern is not
+      // set, assume it's supported due to lack of info.
+      return true;
+    }
+
+    return javaCommand.matches(dataJobsCommandPattern);
```
dougqh, line 1329: Normally, I'd want to see the regex compiled, but since this is only called it is fine.

dougqh (approved):

dd-java-agent/agent-bootstrap/src/main/java/datadog/trace/bootstrap/Agent.java
yiliangzhou: hi @dougqh, feel free making the necessary changes to include your bootstrap telemetry reporting. Thanks!

PR #7405, opened 2024-08-09
Fix exception handling for SQL Server full mode

nenadnoveljic:
# What Does This Do

Finish span in the `finally` block to avoid leaks after exceptions.

# Motivation

# Additional Notes

# Contributor Checklist

Jira ticket: [PROJ-IDENT]

### DIFF
--- a/dd-java-agent/instrumentation/jdbc/src/main/java/datadog/trace/instrumentation/jdbc/JDBCDecorator.java
+++ b/dd-java-agent/instrumentation/jdbc/src/main/java/datadog/trace/instrumentation/jdbc/JDBCDecorator.java
@@ -298,17 +298,13 @@ public class JDBCDecorator extends DatabaseClientDecorator<DBInfo> {
           e);
       DECORATE.onError(instrumentationSpan, e);
     } finally {
-      try {
-        instrumentationSpan.finish();
-      } catch (Exception e) {
-      } finally {
-        if (instrumentationStatement != null) {
-          try {
-            instrumentationStatement.close();
-          } catch (SQLException e) {
-          }
+      if (instrumentationStatement != null) {
+        try {
+          instrumentationStatement.close();
+        } catch (Throwable e) {
         }
       }
+      instrumentationSpan.finish();
     }
     return spanID;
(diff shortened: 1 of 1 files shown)


PerfectSlayer (changes requested):
Hi @nenadnoveljic 

I don't follow your changes compared to the PR description "Finish span in the finally block to avoid leaks after exceptions.". Can you elaborate a bit more but here is what I understand:

First, if there is an exception from statement closure, it will be caught and the span will still finish.
Then, you reduce the exception catch scope from `Exception` to `SQLException` so you will be catching fewer kinds of issues. It is expected?
Finally, you are no more closing the span if the `instrumentedStatement` is `null`, so you will be leaking more spans.

vandonr (approved):

PerfectSlayer:
@vandonr As you approved the changes, can you help me with the concerns I raise in the comment above?

vandonr:
> @vandonr As you approved the changes, can you help me with the concerns I raise in the comment above?

Hmm, yeah, so this PR is stemming from [this comment](https://github.com/DataDog/dd-trace-java/pull/7186#discussion_r1707600723) on a previous PR.

Actually I read the code a bit too fast without seeing your comment, but you are right, this is too high, I think we need yet another finally block a bit lower :(

nenadnoveljic:
> Hi @nenadnoveljic
> 
> I don't follow your changes compared to the PR description "Finish span in the finally block to avoid leaks after exceptions.". Can you elaborate a bit more but here is what I understand:
> 
> First, if there is an exception from statement closure, it will be caught and the span will still finish. Then, you reduce the exception catch scope from `Exception` to `SQLException` so you will be catching fewer kinds of issues. It is expected? Finally, you are no more closing the span if the `instrumentedStatement` is `null`, so you will be leaking more spans.

Thanks for catching it. I implemented the nested exception handling now.

vandonr (approved):
lgtm now

PerfectSlayer (approved):
Thanks for the fix and the discussion about `Exception`/`Throwable`! 👍

PR #7408, opened 2024-08-09
Support addEvent Otel API

mtoffl01:
# What Does This Do
Adds support for span events via `Span.addEvent` and `Span.recordException`.

# Motivation
https://datadoghq.atlassian.net/browse/APMAPI-199

# Additional Notes
Passing all system tests enabled here: https://github.com/DataDog/system-tests/pull/3027

# Contributor Checklist

Jira ticket: [PROJ-IDENT]


dd-java-agent/agent-otel/otel-shim/src/main/java/datadog/opentelemetry/shim/trace/OtelSpanEvent.java
```diff
@@ -0,0 +1,54 @@
...
+  public static String toTag(List<OtelSpanEvent> events) {
+    StringBuilder builder = new StringBuilder("[");
+    for (int i = 0; i < events.size(); i++) {
+      if (i > 0) {
+        builder.append(",");
+      }
+      builder.append(events.get(i).toString());
+    }
+    return builder.append("]").toString();
+  }
```
PerfectSlayer, line 53: I haven't check at the RFC yet but this looks like JSON.
Rather than try to avoid any conversion issue, like escaping double quote from attribute name, what about reusing the JsonBuffer that @dougqh is building in bootstrap?

https://github.com/DataDog/dd-trace-java/pull/7287/files#diff-519b444cb8473a52b363e23131006a599b0fc5c0d9968d0934ef9b59a785872d

dd-java-agent/agent-otel/otel-shim/build.gradle
```diff
@@ -6,6 +6,8 @@ minimumBranchCoverage = 0.0
 dependencies {
   // minimum OpenTelemetry API version this shim is compatible with
   compileOnly group: 'io.opentelemetry', name: 'opentelemetry-api', version: '1.4.0'
+  implementation 'org.jetbrains:annotations:24.0.0'
+  implementation 'com.squareup.moshi:moshi:1.15.0'
```
mtoffl01, line 10: is there any reason to avoid tacking on dependencies in the otel-shim?

dd-java-agent/agent-otel/otel-shim/src/main/java/datadog/opentelemetry/shim/trace/SpanEvent.java
```diff
@@ -0,0 +1,95 @@
+package datadog.opentelemetry.shim.trace;
+
+import com.squareup.moshi.FromJson;
+import com.squareup.moshi.JsonAdapter;
+import com.squareup.moshi.Moshi;
+import com.squareup.moshi.ToJson;
+import io.opentelemetry.api.common.Attributes;
+import java.util.List;
+import java.util.concurrent.TimeUnit;
+import org.jetbrains.annotations.NotNull;
```
mcculls, line 10: We use `edu.umd.cs.findbugs.annotations.NonNull` elsewhere in the codebase

dd-java-agent/agent-otel/otel-shim/build.gradle
mcculls: This shim ends up on the boot class-path, so we cannot have any direct dependencies to third-party libraries because they will end up exposed to the user's application - and could break that application if it expects a different version of those libraries.

Use `edu.umd.cs.findbugs.annotations.NonNull` instead in the code, which will avoid the need to import `org.jetbrains:annotations` - the JSR305 dependency for the `@NonNull` annotation is already added as a `compileOnly` dependency (via spotbugs) which means it will be used for compilation but won't end up in the final jar. This is ok because the `@NonNull` annotation is only used for compile-time checks in the IDE.

Unfortunately moshi cannot be used in the shim.

You could use a kind of "holder" pattern in `internal-api` where the implementation is registered during startup, which would let you use moshi from the boot class-path without exposing it. Or you could implement a simple JSON writer method that's specific to span events - TBH the span event structure is very simple, so writing event data out as JSON would be very straightforward and avoid the extra dependency.

dd-java-agent/agent-otel/otel-shim/src/main/java/datadog/opentelemetry/shim/trace/SpanEvent.java
```diff
@@ -0,0 +1,95 @@
...
+  private static long timeNano(long timestamp, TimeUnit unit) {
+    return TimeUnit.NANOSECONDS.convert(timestamp, unit);
+  }
+
+  @NotNull
+  public static String toTag(List<SpanEvent> events) {
+    StringBuilder builder = new StringBuilder("[");
+    int index = 0;
+    while (index < events.size()) {
+      String linkAsJson = getEncoder().toJson(events.get(index));
```
mcculls, line 41: As mentioned above we can avoid the `moshi` dependency if we wrote out own method(s) to write span event data as JSON. The structure is simple, so this shouldn't involve much code. Also we only need to handle writing JSON, not parsing it.

dd-java-agent/agent-otel/otel-shim/src/main/java/datadog/opentelemetry/shim/trace/SpanEvent.java
mcculls: `org.jetbrains.annotations.NotNull` is used in 2 files and requires the addition of a dependency:
```
compileOnly('org.jetbrains:annotations:24.0.0')
``` 

`edu.umd.cs.findbugs.annotations.NonNull` is used in 11 files and doesn't require any dependency changes

Alternatively don't use either annotation, because we already assume things are non-`null` by default - instead we use `@javax.annotation.Nullable` to mark parameters and return values that may be `null`

dd-java-agent/agent-otel/otel-shim/src/main/java/datadog/opentelemetry/shim/trace/SpanEvent.java
PerfectSlayer: One question I have is what the RFC states about encoding when reaching tag max length?
Should the tracer take care of it? Should we limit event count or drop attributes?
Because if the tag value is cropped, I don’t know what will happen on the backend side with invalid JSON object 🤷

dd-java-agent/agent-otel/otel-shim/src/main/java/datadog/opentelemetry/shim/trace/OtelSpanEvent.java
PerfectSlayer: Yeah, that’s what I though and why I propose Doug’s code in the first place.
We can find a proper place and add some tests to promote it as a resuable piece of code 👍

mtoffl01: @PerfectSlayer , Stuart and I discussed just using the manual parsing method I had initially, and to migrate the code to use Doug's parser once that's been merged (so as not to block this PR)

dd-java-agent/agent-otel/otel-shim/src/main/java/datadog/opentelemetry/shim/trace/SpanEvent.java
mtoffl01: The RFC does not specify any limits, however:

1. Span Links enforces a character limit with `TAG_MAX_LENGTH`
2. [Otel Span Events API](https://opentelemetry.io/docs/specs/otel/trace/api/#add-events) does not specify any limits on span event count
3. However, Otel does enforce a [default limit of 128 span events per span and a maximum of 128 attributes per span event](Span Events, they include a default of 128 maximum span events per span and a maximum of 128 attributes per span event) within the SDK

dd-java-agent/agent-otel/otel-shim/src/main/java/datadog/opentelemetry/shim/trace/OtelConventions.java
```diff
@@ -269,4 +281,43 @@ public static AgentSpan.Attributes convertAttributes(Attributes attributes) {
...
+   * @param additionalAttributes Attributes provided by the user
+   * @return Attributes collection that combines defaultAttributes and additionalAttributes
+   */
+  public static Attributes processExceptionAttributes(
+      Throwable exception, Attributes additionalAttributes) {
+    // "exception.escaped" should be true if exception will "escape" the scope of the span. I'm not
+    // sure how to check that from within here. And it doesn't even look like Otel dynamically sets
+    // this field:
+    // https://github.com/open-telemetry/opentelemetry-java/blob/v1.41.0/sdk/trace/src/main/java/io/opentelemetry/sdk/trace/internal/data/ImmutableExceptionEventData.java#L22
+    boolean escaped = false;
```
mtoffl01, line 302: Need assistance with this one.

dd-java-agent/agent-otel/otel-shim/src/main/java/datadog/opentelemetry/shim/trace/SpanEvent.java
```diff
@@ -0,0 +1,60 @@
...
+      builder
+          .append(",\"attributes\":")
+          .append(SpanAttributes.JSONParser.toJson(this.attributes.asMap()));
+    }
+    return builder.append("}").toString();
+  }
+
+  @NonNull
+  public static String toTag(List<SpanEvent> events) {
+    // TODO: Do we want to enforce a maximum tag size, like TAG_MAX_LENGTH in DDSpanLink?
```
mtoffl01, line 50: Worth discussing this.

dd-java-agent/instrumentation/opentelemetry/opentelemetry-1.4/src/test/groovy/OpenTelemetry14Test.groovy
```diff
@@ -175,29 +185,101 @@ class OpenTelemetry14Test extends AgentTestRunner {
...
+  //
+  //      then:
+  //      long endBlock = System.nanoTime();
+  //
+  //      // I'd like to grab the span's `time_unix_nano` field under `events` tag is not null, and is more than beforeBlock and less than afterBlock
+  //      // But I don't know how to access this without asserting the span tags look a specific way using assertTraces.
+  //
+  //      where:
+  //      long startBlock = System.nanoTime();
+  //    }
```
mtoffl01, line 281: Need assistance with this.
Either we need a way to control the time such that SpanEvent.timeNano() always returns a static value that we define here (therefore we know what to compare time_unix_nano to), or I need a way to access time_unix_nano as a key-value pair (not currently possible as it's just nested inside the larger events string)

mtoffl01:
The last commit for this branch produced this response (below) that references code I did not touch (at least not knowingly). Should I address this?
`Compilation failed for [SourceFile[/Users/mikayla.toffler/datadogrepos/dd-trace-java/dd-java-agent/instrumentation/opentelemetry/opentelemetry-1.4/src/test/groovy/opentelemetry14/context/propagation/W3cPropagatorTracestateTest.groovy]]; org.codehaus.groovy.control.MultipleCompilationErrorsException: startup failed:
None: 32: illegal colon after argument expression;
   solution: a complex label expression before a colon must be parenthesized @ line 32, column 24.
         .toArray(String[]::new)
                          ^

1 error
.
Compilation failed for [SourceFile[/Users/mikayla.toffler/datadogrepos/dd-trace-java/dd-java-agent/instrumentation/opentelemetry/opentelemetry-1.4/src/test/groovy/OpenTelemetry14Test.groovy]]; org.codehaus.groovy.control.MultipleCompilationErrorsException: startup failed:
None: 115: expecting '{', found '(' @ line 115, column 9.
       try (def scope = context.makeCurrent()) {
           ^

1 error
.
Compilation failed for [SourceFile[/Users/mikayla.toffler/datadogrepos/dd-trace-java/dd-java-agent/instrumentation/opentelemetry/opentelemetry-1.4/src/test/groovy/OpenTelemetry14ConventionsTest.groovy]]; org.codehaus.groovy.control.MultipleCompilationErrorsException: startup failed:
None: 104: unexpected token: String @ line 104, column 19.
       def keyFor = (String key) -> useAttributeKey ? stringKey(key) : key
                     ^

1 error
.

Deprecated Gradle features were used in this build, making it incompatible with Gradle 9.0.`

dd-java-agent/instrumentation/opentelemetry/opentelemetry-1.4/src/test/groovy/OpenTelemetry14Test.groovy
PerfectSlayer: First, I am not sure you can rely on `nanoTime` for Event time.
Have a look at `TimeSource` from DD Span context: `context.getTraceCollector().getTimeSource().getCurrentTimeMicros()`

What's wrong with testing the span tags to check if the event time is within the two timestamps? (before and after adding the event)

mtoffl01: @PerfectSlayer , so should I make Span Events use TimeSource instead (and then the test would make this change as well)?
Also, the issue is just that I don't know how to grab the `time_unix_nano` field from within the tags (in order to compare it) as a key-value pair -- As far as I can tell, I can only get the whole String value for "events" and don't know how to treat all that nested JSON as key-value pairs. Not sure if it's worth it for me to add a parsing method to do so?

dd-java-agent/instrumentation/opentelemetry/opentelemetry-1.4/src/test/groovy/OpenTelemetry14Test.groovy
```diff
@@ -175,29 +180,103 @@ class OpenTelemetry14Test extends AgentTestRunner {
...
+    "evt2" | TIME_NANO   | TimeUnit.NANOSECONDS  | Attributes.builder().put("string-key", "string-value").put("long-key", 123456789L).put("double-key", 1234.5678D).put("boolean-key-true", true).put("boolean-key-false", false).build() | '{ string-key: "string-value", long-key: 123456789, double-key: 1234.5678, boolean-key-true: true, boolean-key-false: false }'
+    "evt3" | TIME_NANO   | TimeUnit.NANOSECONDS  | Attributes.builder().put("string-key-array", "string-value1", "string-value2", "string-value3").put("long-key-array", 123456L, 1234567L, 12345678L).put("double-key-array", 1234.5D, 1234.56D, 1234.567D).put("boolean-key-array", true, false, true).build() | '{ string-key-array: [ "string-value1", "string-value2", "string-value3" ], long-key-array: [ 123456, 1234567, 12345678 ], double-key-array: [ 1234.5, 1234.56, 1234.567], boolean-key-array: [true, false, true] }'
+    //    "evt4" | TIME_NANO   | TimeUnit.NANOSECONDS  | Attributes.builder().put("string-array", "1", "2").build() | '{ string-array: [ "1", "2"]}'
+    // evt4 is failing because for some reason string arrays whose values are numbers are treated like number (int) arrays.
```
mtoffl01, line 223: TODO: Don't yet know why this occurs.

dd-java-agent/agent-otel/otel-shim/src/main/java/datadog/opentelemetry/shim/trace/OtelConventions.java
```diff
@@ -269,4 +291,53 @@ public static AgentSpan.Attributes convertAttributes(Attributes attributes) {
...
+    }
+    return attrsBuilder.build();
+  }
+
+  private static Map<String, Object> getAttributesFromException(Throwable exception) {
+    // "exception.escaped" should be true if exception will "escape" the scope of the span. I'm not
+    // sure how to check that from within here. And it doesn't even look like Otel dynamically sets
+    // this field:
+    // https://github.com/open-telemetry/opentelemetry-java/blob/v1.41.0/sdk/trace/src/main/java/io/opentelemetry/sdk/trace/internal/data/ImmutableExceptionEventData.java#L22
+    boolean escaped = false;
```
mtoffl01, line 333: Need assistance here.

dd-java-agent/agent-otel/otel-shim/src/main/java/datadog/opentelemetry/shim/trace/OtelSpanEvent.java
dougqh: Right now, my JsonBuffer is only in the bootstrap, but we could make it available elsewhere, too.

I'm also open to modifying it to having a template mode, so we can prebuild most of the structure and reuse.
It also already has a mechanism to put multiple buffers together if that would be of use here.

internal-api/src/main/java/datadog/trace/bootstrap/instrumentation/api/SpanAttributes.java
```diff
@@ -53,11 +54,89 @@ public String toString() {
     return "SpanAttributes{" + this.attributes + '}';
   }
 
+  // JSONParser is a helper class for turning the Map<String,String> that holds the attributes into
+  // a JSON string
+  public static class JSONParser {
```
amarziali, line 59: can we use libraries that already manipulate efficiently some map to json instead having to maintain (and test) this code?

mtoffl01: @amarziali , is it ok to add third party dependencies here? See related discussion on the opentelemetry shim path: https://github.com/DataDog/dd-trace-java/pull/7408/#discussion_r1715111876

amarziali: ok I was thinking to the same (using moshi) since we bundle in other packages. Thanks for having shared that

dd-java-agent/agent-otel/otel-shim/src/main/java/datadog/opentelemetry/shim/trace/OtelConventions.java
mtoffl01: note: do not set exception.escaped by default

PerfectSlayer:
Following on the topic on tag max size, did you get clarification from the RFC owner about how event should be encoded to avoid hitting max length?

mtoffl01:
Note: Test coverage is failing for SpanAttributes.Builder and SpanAttributes.JSONParser, but I'm confused by the results. The Opentelemetry14Test tests for adding events do utilize the builder and JSONParser, and when I run a test with coverage, we can see green on `convertAttributes` functionality that calls on SpanAttributes.Builder functionality, but the relevant SpanAttributes.Builder blocks are labeled with red.

dd-java-agent/agent-otel/otel-shim/src/main/java/datadog/opentelemetry/shim/trace/SpanEvent.java
```diff
@@ -0,0 +1,73 @@
...
+import datadog.trace.api.time.SystemTimeSource;
+import datadog.trace.api.time.TimeSource;
+import datadog.trace.bootstrap.instrumentation.api.AgentSpan;
+import datadog.trace.bootstrap.instrumentation.api.SpanAttributes;
+import edu.umd.cs.findbugs.annotations.NonNull;
+import io.opentelemetry.api.common.Attributes;
+import java.util.List;
+import java.util.concurrent.TimeUnit;
+
+public class SpanEvent {
```
PerfectSlayer, line 12: Can you rename it OtelSpanEvent?

dd-java-agent/agent-otel/otel-shim/src/main/java/datadog/opentelemetry/shim/trace/OtelSpan.java
```diff
@@ -71,13 +75,23 @@ public <T> Span setAttribute(AttributeKey<T> key, T value) {
 
   @Override
   public Span addEvent(String name, Attributes attributes) {
-    // Not supported
+    if (this.recording) {
+      if (this.events == null || this.events.isEmpty()) {
```
PerfectSlayer, line 79: Only null check is needed here

dd-java-agent/agent-otel/otel-shim/src/main/java/datadog/opentelemetry/shim/trace/OtelSpan.java
```diff
@@ -71,13 +75,23 @@ public <T> Span setAttribute(AttributeKey<T> key, T value) {
...
+      this.events.add(new SpanEvent(name, attributes));
+    }
     return this;
   }
 
   @Override
   public Span addEvent(String name, Attributes attributes, long timestamp, TimeUnit unit) {
-    // Not supported
+    if (this.recording) {
+      if (this.events == null || this.events.isEmpty()) {
```
PerfectSlayer, line 90: Only null check is needed here

dd-java-agent/agent-otel/otel-shim/src/main/java/datadog/opentelemetry/shim/trace/OtelSpan.java
```diff
@@ -97,10 +111,23 @@ public Span setStatus(StatusCode statusCode, String description) {
...
+   * href="https://javadoc.io/doc/io.opentelemetry/opentelemetry-api-trace/latest/io/opentelemetry/api/trace/Span.html#recordException(java.lang.Throwable)">...</a>
+   *
+   * @param exception the Throwable to record
+   * @param additionalAttributes the additional attributes to record
+   */
   @Override
   public Span recordException(Throwable exception, Attributes additionalAttributes) {
     if (this.recording) {
-      // Store exception as span tags as span events are not supported yet
+      if (this.events == null || this.events.isEmpty()) {
```
PerfectSlayer, line 126: Only null check is needed here

dd-java-agent/agent-otel/otel-shim/src/main/java/datadog/opentelemetry/shim/trace/SpanEvent.java
```diff
@@ -0,0 +1,73 @@
...
+    if (!this.attributes.isEmpty()) {
+      builder
+          .append(",\"attributes\":")
+          .append(SpanAttributes.JSONParser.toJson(this.attributes.asMap()));
+    }
+    return builder.append("}").toString();
+  }
+
+  @NonNull
+  public static String toTag(List<SpanEvent> events) {
```
PerfectSlayer, line 62: This can have package scope as it is implementation details

dd-java-agent/agent-otel/otel-shim/src/main/java/datadog/opentelemetry/shim/trace/SpanEvent.java
```diff
@@ -0,0 +1,73 @@
...
+    } else {
+      return timeSource.getCurrentTimeNanos();
+    }
+  }
+
+  public static void setTimeSource(TimeSource newTimeSource) {
+    timeSource = newTimeSource;
+  }
+
+  public String toString() {
```
PerfectSlayer, line 49: Why `toString()` is using a JSON encoding instead of a default Java object representation?

dd-java-agent/agent-otel/otel-shim/src/main/java/datadog/opentelemetry/shim/trace/SpanEvent.java
```diff
@@ -0,0 +1,73 @@
...
+    return builder.append("}").toString();
+  }
+
+  @NonNull
+  public static String toTag(List<SpanEvent> events) {
+    // TODO: Do we want to enforce a maximum tag size, like TAG_MAX_LENGTH in DDSpanLink?
+    StringBuilder builder = new StringBuilder("[");
+    for (int i = 0; i < events.size(); i++) {
+      if (i > 0) {
+        builder.append(",");
```
PerfectSlayer, line 67: ',' instead of "," (char instead of string)

dd-java-agent/agent-otel/otel-shim/src/main/java/datadog/opentelemetry/shim/trace/SpanEvent.java
```diff
@@ -0,0 +1,73 @@
...
+      }
+      builder.append(events.get(i).toString());
+    }
+    return builder.append("]").toString();
```
PerfectSlayer, line 71: Same for ']'

dd-java-agent/agent-otel/otel-shim/src/main/java/datadog/opentelemetry/shim/trace/OtelConventions.java
```diff
@@ -269,4 +291,53 @@ public static AgentSpan.Attributes convertAttributes(Attributes attributes) {
...
+  /**
+   * generate default list of attribtues about a Throwable
+   *
+   * @param exception the Throwable about which to generate attribtues
+   * @return Map of attributes about the Throwable
+   */
+  private static Map<String, Object> getAttributesFromException(Throwable exception) {
+    return new HashMap<String, Object>() {
+      {
+        put("exception.message", exception.getMessage());
```
PerfectSlayer, line 337: Is is expected to use different attributes than the past tags? (`error.xxx`)
Are they coming from OTel semantic conventions?

dd-java-agent/agent-otel/otel-shim/src/main/java/datadog/opentelemetry/shim/trace/OtelConventions.java
```diff
@@ -269,4 +291,53 @@ public static AgentSpan.Attributes convertAttributes(Attributes attributes) {
...
+   * additionalAttributes provided by the user
+   *
+   * <p>If the same key exists in defaultAttributes and additionalAttributes, the latter always
+   * wins.
+   *
+   * @param exception The Throwable from which to build default attributes
+   * @param additionalAttributes Attributes provided by the user
+   * @return Attributes collection that combines defaultAttributes and additionalAttributes
+   */
+  public static Attributes processExceptionAttributes(
```
PerfectSlayer, line 306: This feel more complex (and sometime requires more resources) than needed. I would:
* Cache all 3 attributes key (if not reused from OTel SEMCONV)
* Check if additionalAttritubes already has all the 3 keys
* If it has, exit and don't alter additionalAttributes
* If it doesn't, get an attribute builder from the additionalAttributes, check every missing key and set computed value to it.

This would:
* Avoid creating a new Map, a new Class, a String array, the stack trace computation, an attribute builder if elements are already present
* Only compute needed values

The whole method should not be big, the implementation can go into the `recordException` method.

dd-java-agent/agent-otel/otel-shim/src/main/java/datadog/opentelemetry/shim/trace/OtelConventions.java
```diff
@@ -269,4 +291,53 @@ public static AgentSpan.Attributes convertAttributes(Attributes attributes) {
...
+   *
+   * @param exception the Throwable about which to generate attribtues
+   * @return Map of attributes about the Throwable
+   */
+  private static Map<String, Object> getAttributesFromException(Throwable exception) {
+    return new HashMap<String, Object>() {
+      {
+        put("exception.message", exception.getMessage());
+        put("exception.type", exception.getClass().getName());
+        put("exception.stacktrace", Arrays.toString(exception.getStackTrace()));
```
PerfectSlayer, line 339: Are you sure it is the expected encoding? 

Here is what is output:
```java
jshell> Arrays.toString(new Exception().getStackTrace());
$1 ==> "[do_it$(java:7), java.base/jdk.internal.reflect.NativeMethodAccessorImpl.invoke0(Native Method), java.base/jdk.internal.reflect.NativeMethodAccessorImpl.invoke(NativeMethodAccessorImpl.java:77), java.base/jdk.internal.reflect.DelegatingMethodAccessorImpl.invoke(DelegatingMethodAccessorImpl.java:43), java.base/java.lang.reflect.Method.invoke(Method.java:569), jdk.jshell/jdk.jshell.execution.DirectExecutionControl.invoke(DirectExecutionControl.java:209), jdk.jshell/jdk.jshell.execution.RemoteExecutionControl.invoke(RemoteExecutionControl.java:116), jdk.jshell/jdk.jshell.execution.DirectExecutionControl.invoke(DirectExecutionControl.java:119), jdk.jshell/j ... cution.ExecutionControlForwarder.commandLoop(ExecutionControlForwarder.java:262), jdk.jshell/jdk.jshell.execution.Util.forwardExecutionControl(Util.java:76), jdk.jshell/jdk.jshell.execution.Util.forwardExecutionControlAndIO(Util.java:137), jdk.jshell/jdk.jshell.execution.RemoteExecutionControl.main(RemoteExecutionControl.java:70)]"

```

Where I previously used:
```java
jshell> var writer = new StringWriter()
writer ==>

jshell> new Exception().printStackTrace(new PrintWriter(writer))

jshell> writer.toString()
$6 ==> "java.lang.Exception\n\tat do_it$(java:5)\n\tat java.base/jdk.internal.reflect.NativeMethodAccessorImpl.invoke0(Native Method)\n\tat java.base/jdk.internal.reflect.NativeMethodAccessorImpl.invoke(NativeMethodAccessorImpl.java:77)\n\tat java.base/jdk.internal.reflect.DelegatingMethodAccessorImpl.invoke(DelegatingMethodAccessorImpl.java:43)\n\tat java.base/java.lang.reflect.Method.invoke(Method.java:569)\n\tat jdk.jshell/jdk.jshell.execution.DirectExecutionControl.invoke(DirectExecutionControl.java:209)\n\tat jdk.jshell/jdk.jshell.execution.RemoteExecutionControl.invoke(RemoteExecutionControl.java:116)\n\tat jdk.jshell/jdk.jshell.execution.DirectExecutionC ... ControlForwarder.commandLoop(ExecutionControlForwarder.java:262)\n\tat jdk.jshell/jdk.jshell.execution.Util.forwardExecutionControl(Util.java:76)\n\tat jdk.jshell/jdk.jshell.execution.Util.forwardExecutionControlAndIO(Util.java:137)\n\tat jdk.jshell/jdk.jshell.execution.RemoteExecutionControl.main(RemoteExecutionControl.java:70)\n"
```
The related logic can be found in `datadog.trace.core.DDSpan.addThrowable(java.lang.Throwable, byte)`

PerfectSlayer (commented):
I mostly focused on the OTel part and left a bunch of comments (sorry for that 😓 ).

In addition, I don't think we should reuse the _span_ attributes for the _event_ attributes.
Encoding and handling are not the same, and the introduced `format` impact nearly every single behavior (all except than String handling).

As span events are not a thing outside OpenTelemetry, they should not impact the `Span.Attributes` class nor the core part. I would advocate for creating a `SpanEvent.Attribute` class to handle this specifically for OTel and revert any change to `internal-api` and `core`.

How do you feel about it?

dd-java-agent/agent-otel/otel-shim/src/main/java/datadog/opentelemetry/shim/trace/OtelConventions.java
mtoffl01: Correct. https://opentelemetry.io/docs/specs/semconv/exceptions/exceptions-spans/#attributes

dd-java-agent/agent-otel/otel-shim/src/main/java/datadog/opentelemetry/shim/trace/SpanEvent.java
mtoffl01: To ensure that the final String product contains nested quotes in the expected places. Like the keys need to have quotes around them, and only String values need to have quotes around them. E.g, `"{\"name\":\"Mikayla\",\"age\":27}"`.
Also, there is all the specific Attributes manipulation to do, so just using the native toString() method wasn't sufficient (in my experience testing it?).

dd-java-agent/agent-otel/otel-shim/src/main/java/datadog/opentelemetry/shim/trace/OtelConventions.java
mtoffl01: I'm not sure. WDYT? 🤔
Looks like the latter contains formatting characters and some more information. Do you think that's a better representation of a Java stacktrace?

dd-java-agent/agent-otel/otel-shim/src/main/java/datadog/opentelemetry/shim/trace/SpanEvent.java
PerfectSlayer: > To ensure that the final String product contains nested quotes in the expected places

It sounds like you are trying to enforce business logic related to the events serialisation into the human readable String representation of the object instance, used for debugging. There is no "expected" quotes for debug logging.

About your example, I would expect to have "{Makayla, 27yo}". That's way more readable for debugging and not tied to an existing format (so people won't try to rely and on it and parse it).

If you need a method to return the JSON-encoded instance, simply create a `toJson()` method and leave the `toString()` for debugging instead.

dd-java-agent/agent-otel/otel-shim/src/main/java/datadog/opentelemetry/shim/trace/OtelConventions.java
PerfectSlayer: I don't know 🤷  What's best is what the backend will be happy with.

I am just referring there are some other parts of the tracer that handle it differently, so we would have to make sure of the expected format here as your PR introduces a different one.

dd-java-agent/agent-otel/otel-shim/src/main/java/datadog/opentelemetry/shim/trace/OtelConventions.java
PerfectSlayer: How does the RFC evolved about this part? Can we send those tags and agent or backend will make the translation? Or should the client do it?

mtoffl01: Good point, I overlooked this. See RecordException section of [RFC](https://docs.google.com/document/d/1b8GLnbC4oPrHoksmUOhy66iEJto1FjnVivr_jsZlSNw/edit#heading=h.9dsyr07huek):
> attribute[“exception.message”] => meta[“error.msg”]
attribute[“exception.type”] => meta[“error.type”]
attribute[“exception.stacktrace”] => meta[“error.stack”]

I'll go back and add this.

PerfectSlayer: So related to the stacktrace encoding comment, it seems to me you would better use the one already in place to stay consistent with how the `error.stack` tag is populated.

PerfectSlayer (approved):
Current limitations are:
* TimeSource is not fetch from CoreTracer
* JSON encoding is manual and not Moshi based (as currently not opened to instrumentations)

PR #7414, opened 2024-08-12
Tibco BW 5: fix child process parentship

amarziali:
# What Does This Do

This PR fixes the trace hierarchy of `tibco.process` spans generated by businessworks 5 applications.

In particular, a couple of situation where badly reported:

* Subprocesses spawned as Jobs: The process was not linked to the caller ancestor
* Subprocesses called directly: The process was not reported. The child activities were reported either child of current span context either as a trace root

# Motivation

# Additional Notes

# Contributor Checklist

Jira ticket: [PROJ-IDENT]


dd-java-agent/instrumentation/tibco-businessworks/tibcobw-5/src/main/java/datadog/trace/instrumentation/tibcobw5/JobInstrumentation.java
```diff
@@ -0,0 +1,50 @@
...
+      int suffixIdx = workflowName.indexOf(".process");
+      if (suffixIdx > 0) {
+        workflowName = workflowName.substring(0, suffixIdx);
+      }
+      AgentSpan span = startSpan(TIBCO_PROCESS_OPERATION);
+      DECORATE.afterStart(span);
+      DECORATE.onProcessStart(span, workflowName);
+      Map<String, AgentSpan> map =
+          InstrumentationContext.get(ProcessContext.class, Map.class)
+              .putIfAbsent(processContext, HashMap::new);
```
PerfectSlayer, line 46: Should we have a bounded map implementation?

PerfectSlayer (approved):
Looks good.
A quick comment about a collection as I am not well versed in the framework (and possible workflow cardinality).

dd-java-agent/instrumentation/tibco-businessworks/tibcobw-5/src/main/java/datadog/trace/instrumentation/tibcobw5/JobInstrumentation.java
amarziali: I don't think we can in this case because this map will contain the correspondence between an activity and a span (during the span lifetime). If I miss inserting things I will miss which span I should activate or use as parent.
However those activities are a finite (small) numbers since they represent the thing to do during a business process. I have no clues also about a right sizing of this map so I used a default one

dougqh: Hmm, typically, we'd still want a bounded map in these cases.  Otherwise, we can end up consuming a lot of memory and causing the JVM to GC a lot or even bring down the JVM.

While we might lose some tracing, bringing down the application is worse.

That said, I do appreciate the difficulty in sizing especially in this case.
I think what we need to do is create a BoundedMap that we can use in instrumentation that also reports metrics on capacity, so we can size better in the future.

Although to be clear, I'm not saying that we should do that right now for this PR.

PR #7442, opened 2024-08-14
Fix SymDB upload dropped requests

jpbempel:
# What Does This Do
SymbolSink was emitting an upload requests for every scope added into the sink queue. As the queue was 1024 entries in capacity we could easily end up in a situation where more than 20 requests are made when draining the Q. to avoid that we are batching the scope together in a single larger upload request and ensure that the sink queue is flushed when full before adding new element and not losing any symbols.

# Motivation
avoid dropping SymDB requests

# Additional Notes

# Contributor Checklist

Jira ticket: [DEBUG-2685]

[DEBUG-2685]: https://datadoghq.atlassian.net/browse/DEBUG-2685?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ

### DIFF
--- a/dd-java-agent/agent-debugger/src/main/java/com/datadog/debugger/sink/SymbolSink.java
+++ b/dd-java-agent/agent-debugger/src/main/java/com/datadog/debugger/sink/SymbolSink.java
@@ -9,7 +9,6 @@ import datadog.trace.api.Config;
 import datadog.trace.util.TagsHelper;
 import java.nio.charset.StandardCharsets;
 import java.util.ArrayList;
-import java.util.Collections;
 import java.util.List;
 import java.util.concurrent.ArrayBlockingQueue;
 import java.util.concurrent.BlockingQueue;
@@ -35,7 +34,7 @@ public class SymbolSink {
   private final String version;
   private final BatchUploader symbolUploader;
   private final BatchUploader.MultiPartContent event;
-  private final BlockingQueue<ServiceVersion> scopes = new ArrayBlockingQueue<>(CAPACITY);
+  private final BlockingQueue<Scope> scopes = new ArrayBlockingQueue<>(CAPACITY);
   private final Stats stats = new Stats();
 
   public SymbolSink(Config config) {
@@ -59,14 +58,12 @@ public class SymbolSink {
   }
 
   public void addScope(Scope jarScope) {
-    ServiceVersion serviceVersion =
-        new ServiceVersion(serviceName, env, version, "JAVA", Collections.singletonList(jarScope));
-    boolean added = scopes.offer(serviceVersion);
+    boolean added = scopes.offer(jarScope);
     int retries = 10;
     while (!added) {
       // Q is full, flushing synchronously
       flush();
-      added = scopes.offer(serviceVersion);
+      added = scopes.offer(jarScope);
       retries--;
       if (retries < 0) {
         throw new IllegalStateException("Scope cannot be enqueued after 10 retries" + jarScope);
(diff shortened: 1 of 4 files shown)


dd-java-agent/agent-debugger/src/main/java/com/datadog/debugger/sink/SymbolSink.java
```diff
@@ -59,37 +58,51 @@ public void stop() {
...
   }
 
   public void flush() {
     if (scopes.isEmpty()) {
       return;
     }
     List<ServiceVersion> scopesToSerialize = new ArrayList<>();
+    // ArrayBlockingQueue makes drainTo atomic, so it is safe to call flush from different and
+    // concurrent threads
     scopes.drainTo(scopesToSerialize);
```
shatzi, line 84: maybe we should check scopesToSerialize is empty just return

shatzi (approved):

dd-java-agent/agent-debugger/src/main/java/com/datadog/debugger/sink/SymbolSink.java
```diff
@@ -59,37 +58,51 @@ public void stop() {
...
-        stats.updateStats(classScopeCount, json.length());
-        symbolUploader.uploadAsMultipart(
-            "",
-            event,
-            new BatchUploader.MultiPartContent(
-                json.getBytes(StandardCharsets.UTF_8), "file", "file.json"));
-      } catch (Exception e) {
-        ExceptionHelper.logException(LOGGER, e, "Error during scope serialization:");
-      }
+      allScopes.addAll(serviceVersion.getScopes());
```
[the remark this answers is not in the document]
jpbempel: removing ServiceVersion make it simpler.
simplify by adding the getScopeCount method, but updateStats method stays

dd-java-agent/agent-debugger/src/main/java/com/datadog/debugger/sink/SymbolSink.java
```diff
@@ -59,37 +58,51 @@ public void stop() {
...
-  public boolean addScope(Scope jarScope) {
+  public void addScope(Scope jarScope) {
     ServiceVersion serviceVersion =
         new ServiceVersion(serviceName, env, version, "JAVA", Collections.singletonList(jarScope));
-    return scopes.offer(serviceVersion);
+    boolean added = scopes.offer(serviceVersion);
+    int retries = 10;
+    while (!added) {
+      // Q is full, flushing synchronously
+      flush();
```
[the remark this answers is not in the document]
jpbempel: the flush make upload http requests, but http requests are sent async in OkHttp component.

dd-java-agent/agent-debugger/src/main/java/com/datadog/debugger/sink/SymbolSink.java
jpbempel: done

dd-java-agent/agent-debugger/src/main/java/com/datadog/debugger/sink/SymbolSink.java
```diff
@@ -59,37 +57,47 @@ public void stop() {
     symbolUploader.shutdown();
   }
 
-  public boolean addScope(Scope jarScope) {
-    ServiceVersion serviceVersion =
-        new ServiceVersion(serviceName, env, version, "JAVA", Collections.singletonList(jarScope));
-    return scopes.offer(serviceVersion);
+  public void addScope(Scope jarScope) {
+    boolean added = scopes.offer(jarScope);
+    int retries = 10;
```
ojung, line 62: Why might we need to retry more than once if `flush()` is synchronous?

ojung (commented):
The way I understand these changes, we could still end up with more than 20 concurrent requests: When there are (20 + 1) * 1024 classes loaded at roughly the same time. Please correct me if I'm wrong or this isn't a concern.

jpbempel:
> The way I understand these changes, we could still end up with more than 20 concurrent requests: When there are (20 + 1) * 1024 classes loaded at roughly the same time. Please correct me if I'm wrong or this isn't a concern.

That's almost true, except that top level scopes are aggregated by jar name, en so each entry in the Queue is a different jar. So you need (20+1) *1024 different jars for that :)

dd-java-agent/agent-debugger/src/main/java/com/datadog/debugger/sink/SymbolSink.java
jpbempel: concurrency at the best: this flushing process is in background thread, and still (but unlikely) threads are loading a lot of classes/jars and we filling up again the Q while flushing the previous batch, we are trying again to flush.

Flush is synchronous, but we are still allowing to add to the queue once drained.

PR #7464, opened 2024-08-18
Run CI only for tasks affected by git changes

smola:
# What Does This Do

* Add a new property `-PgitBaseRef` that can be set to a git reference (or branch, tag, commit hash, etc) to compare the current `HEAD` against. This will compute which files changed. Changed our Circle CI configuration to pass `-PgitBaseRef=<base branch>` for PRs.
* Aggregate tasks like `:baseTests`, `:instTests`, `:baseCheck`, etc, now do not trigger tasks that are unaffected by the computed git changes.
* The logic to find affected tasks is as follows:
  * Get changes from git (e.g. `git diff --names-only origin/master..HEAD`).
  * Ignore certain files, such as `README.md`. Check the definition of `ignoredFiles` for the current patterns.
  * Detect global effect files, such as `gradle/**`. If any of these changes, the rest of the change tracking logic will be skipped and all tasks will be triggered. Check the definition of `globalEffectFiles`.
  * For each file, infer the project and task it should trigger directly. A file will be assigned to the project with the longest path which is a prefix of the file. If this cannot be found (e.g. root project changes), change tracking is disabled, and all tasks will be triggered.
  * Once a project is identified, a file will affect a set of tasks. If it's in `src/test`, it'll affect `testClasses`. If it's in `src/testFixtures`, it'll affect `testFixturesClasses`. If it's in `src/jmh`, it'll affect `jmhCompileGeneratedClasses`. Otherwise, it'll affect all tasks in the project.
  * Given an aggregate task (e.g. `:baseTests`), get all relevant tasks, after applying partitioning if necessary. And for each task, that task and all its transitive task dependencies will be checked against the previous collected list of affected tasks. If an affected task is found, then the task will be triggered.

Note that logging here is quite verbose, using warning level. The intent is making any decision around change tracking as transparent and debuggeable as possible.

Some minor notes around CI definition:

* Previous logic to skip instrumentation tests for profiler-only changes is kept intact, since there is an actual dependency across these modules which would always trigger all instrumentation tests. Ideally we would do some further changes to use a single skip mechanism for both cases.
* `dd-java-agent/agent-iast` does not skip instrumentation tests anymore, since there's a few instrumentation modules that depend on test fixtures from `dd-java-agent/agent-iast`.
* `.circleci/collect_results.sh` was changed to not fail when no tests were run in a job and no JUnit XML reports are available. This is necessary when there are no affected task for a given job partition.

## How to test it?

To test change detection, you can use `-PgitNewRef` to simulate change detection as if another reference was checked out. For example, to see which tasks would be run for `:baseTest` on a given branch when compared to master, the following command can be used:

```
./gradlew :baseTest -PgitBaseRef=origin/master -PgitNewRef=my-branch
```

# Motivation
Reduce PR turnaround time, and reduce the chance of hitting a flaky test in a PR.

# Additional Notes
* This is the second attempt, after https://github.com/DataDog/dd-trace-java/pull/6610 which was reverted at https://github.com/DataDog/dd-trace-java/pull/6699.
* The previous attempt used the [dropbox/AffectedModuleDetector](https://github.com/dropbox/AffectedModuleDetector/tree/main/affectedmoduledetector) plugin. However, it did not see changes throw custom configurations (e.g. it missed changes in `testFixtures`), which caused the revert.
* The current approach is simpler, yet more robust. The logic is quite concise, and being defined in our repository, we can tweak it easily.

# Contributor Checklist

- ~[ ] Update the [public documentation](https://docs.datadoghq.com/tracing/trace_collection/library_config/java/) in case of new configuration flag or behavior~


gradle/ci_jobs.gradle
```diff
@@ -0,0 +1,211 @@
...
+  final String newRef = rootProject.hasProperty("gitNewRef") ? rootProject.property("gitNewRef") : "HEAD"
+
+  rootProject.ext {
+    it.changedFiles = getChangedFiles(baseRef, newRef)
+    useGitChanges = true
+  }
+
+  // The ignoredFiles FileTree selects any file that should not trigger any tasks.
+  final ignoredFiles = fileTree(rootProject.projectDir) {
+    include '.gitingore', '.editorconfig'
```
PerfectSlayer, line 92:
```suggestion
    include '.gitignore', '.editorconfig'
```

PerfectSlayer (approved):
That looks nice! 
I only reviewed it "on paper" (did not run it).

I am curious about if we should shadow this change, like running the change detection while still running everything and keeping a close eye to its output, or just deploy it and refine it as we go? -- I expect there will be few tasks that might not be caught from the first try 😇

PerfectSlayer:
Muzzle failure? 🤔  
> :dd-java-agent:instrumentation:couchbase:couchbase-3.2:muzzle-AssertPass-com.couchbase.client-java-client-3.6.3 

It might come from master though...

smola:
> @PerfectSlayer I am curious about if we should shadow this change, like running the change detection while still running everything and keeping a close eye to its output, or just deploy it and refine it as we go?

Adding some sort of dry run sounds good in theory, the problem is that we'll probably be looking at results and think "ok, looks fine", because if it goes wrong, it'll probably be something that catches us off guard and not be obvious until it fails :thinking:

PR #7468, opened 2024-08-19
Do not leak call depth threadlocal in jdbc instrumentation

amarziali:
# What Does This Do

If the query info is missing, the jdbc advice onEnter is returning a null scope but the threadlocal for its calldepth is not reset.
That means that subsequent queries done on the same thread will not be traced because the calldepth threadlocal is dirty

# Motivation

# Additional Notes

Jira ticket: [APMS-12988]

[APMS-12988]: https://datadoghq.atlassian.net/browse/APMS-12988?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ


PerfectSlayer (approved):
Thanks for the fix

mcculls (approved):

PR #7496, opened 2024-08-23
Skip telemetry heartbeat and add app closing event on shutdown flush

smola:
# What Does This Do
* Skip heartbeat on last telemetry message batch, on message flush. If a heartbeat was needed, it would have been scheduled already.
* Include the app-closing message in the same batch as the rest of messages. This avoids wasting resources by sending 2 HTTP requests instead of 1 during shutdown. I did not add a special case for when there's no pending data to avoid the message-batch wrapper. I doubt it's worth it.

# Motivation
Review originally motivated by flaky system test that pointed that we sent heartbeats too fast. It turns out it was the heartbeat of the shutdown flush.

# Additional Notes

# Contributor Checklist

- ~[ ] Update the [public documentation](https://docs.datadoghq.com/tracing/trace_collection/library_config/java/) in case of new configuration flag or behavior~


telemetry/src/main/java/datadog/telemetry/TelemetryService.java
```diff
@@ -202,7 +190,14 @@ public boolean sendTelemetryEvents() {
       request =
           new TelemetryRequest(
               eventSource, eventSink, messageBytesSoftLimit, RequestType.MESSAGE_BATCH, debug);
-      request.writeHeartbeat();
+      if (isClosing) {
+        // Write early, if moved, ensure it does not get skipped because of size limits.
+        request.writeAppClosing();
```
manuel-alvarez-alvarez, line 195: As a side effect, this changes the request type for the app closing event, it won't be `app-closing` anymore but a `message-batch`. 

Can this cause any issues further down the line in the backend (probably not)?

smola: Not really. The only problem is that we won't be optimizing the message size for the case where there is only an app-closing message. I think that will usually not be the case because of metrics, and anyway it's not that much overhead.

manuel-alvarez-alvarez (approved):

PR #7573, opened 2024-09-06
Add watches support through probe tags

jpbempel:
# What Does This Do
Watches collect specific (deep) fields to add them exclusively into
the snapshot instead of locals/arguments/statics/...
Watches are defined using probe definition predefined tags:
 - `dd_watches_dsl`
 - `dd_watches_json`
 
When you just want to reference deep fields just `a.b.c` you can use the
dsl syntax (like Expression Language). But if you need more advanced
(accessing map or list, using filter, any, ... functions) you need to
use `dd_watches_json` and put the Json AST for Expression Language
(like log message template with segments).
Once the probe is executed, it interprets watches tags to evaluate the
expression then add it in the snapshot as special attributes `watches`.

# Motivation

# Additional Notes

# Contributor Checklist

Jira ticket: [DEBUG-2829]

[DEBUG-2829]: https://datadoghq.atlassian.net/browse/DEBUG-2829?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ

### DIFF
--- a/dd-java-agent/agent-debugger/src/main/java/com/datadog/debugger/probe/LogProbe.java
+++ b/dd-java-agent/agent-debugger/src/main/java/com/datadog/debugger/probe/LogProbe.java
@@ -7,6 +7,7 @@ import com.datadog.debugger.agent.Generated;
 import com.datadog.debugger.agent.StringTemplateBuilder;
 import com.datadog.debugger.el.EvaluationException;
 import com.datadog.debugger.el.ProbeCondition;
+import com.datadog.debugger.el.Value;
 import com.datadog.debugger.el.ValueScript;
 import com.datadog.debugger.instrumentation.CapturedContextInstrumentor;
 import com.datadog.debugger.instrumentation.DiagnosticMessage;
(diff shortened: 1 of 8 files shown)


dd-java-agent/agent-debugger/src/main/java/com/datadog/debugger/probe/LogProbe.java
```diff
@@ -759,6 +836,7 @@ public static class Builder extends ProbeDefinition.Builder<Builder> {
     private String template;
     private List<Segment> segments;
     private boolean captureSnapshot;
+    private List<ValueScript> watches;
```
shatzi, line 839: I think should be similar to EvaluationError format. to store "expr" and "value" for each watch. that way in the future the UI can help the user to understand which watch created witch value.

Maybe we should also add "error" if the watch evaluated as error - topics for the RFC I assume

dd-java-agent/agent-debugger/src/main/java/com/datadog/debugger/probe/LogProbe.java
```diff
@@ -390,16 +430,53 @@ public void evaluate(
...
-    if (logStatus.isSampled() && logStatus.getCondition()) {
-      StringTemplateBuilder logMessageBuilder = new StringTemplateBuilder(segments, LIMITS);
-      String msg = logMessageBuilder.evaluate(context, logStatus);
-      if (msg != null && msg.length() > LOG_MSG_LIMIT) {
-        StringBuilder sb = new StringBuilder(LOG_MSG_LIMIT + 3);
-        sb.append(msg, 0, LOG_MSG_LIMIT);
-        sb.append("...");
-        msg = sb.toString();
+    processMsgTemplate(context, logStatus);
+    processWatches(context, logStatus);
```
shatzi, line 434: Do we want watches get processed so normal logs probe as well? neat stuff.

shatzi (approved):
nice!

dd-java-agent/agent-debugger/src/main/java/com/datadog/debugger/probe/LogProbe.java
jpbempel: `ValueScript` already wraps string representation of the expr and the expression root AST. I am using the string DSL to add it to the snapshot that way we know which watch it is

PR #7585, opened 2024-09-09
Inject trace context into AWS Step Functions input

DylanLovesCoffee:
# What Does This Do
Adds an instrumentation for AWS SDK Step Functions. This enables tracing for when a Lambda function invokes a Step Function. Trace context is injected into the Step Function's `StartExecutionRequest/StartSyncExecutionRequest.Input` object. 

Example of traced step function in the Serverless org: [app](https://ddserverless.datadoghq.com/functions?cloud=aws&entity_view=step_functions&fromUser=false&graphType=flamegraph&panel_end=1727207040000&panel_paused=false&panel_start=1727203440000&shouldShowLegend=true&sort=time&sp=%5B%7B%22p%22%3A%7B%22entityId%22%3A%22aws-step-functions%2Barn%3Aaws%3Astates%3Asa-east-1%3A425362996713%3Astatemachine%3Adyl-state-express%22%7D%2C%22i%22%3A%22step-function-panel%22%7D%5D&spanID=9019823743937099326&text_search=dylan&traceID=2898078647503882730&traceQuery=&start=1727202972630&end=1727206572630&paused=false)

The [Logs to Traces Reducer](https://github.com/DataDog/logs-backend/blob/4ec54e460925fbec831cbf5528094740c2b1907b/domains/serverless/apps/logs-to-traces-reducer/src/main/java/com/dd/logs/serverless/logs_to_traces/reducer/output/stepfunction/trace/spanbuilder/control/RootSpanBuilder.java#L55-L62) will read the trace context from the Step Function logs and create a span for the Step Function. 

# Motivation
Continues the work done in [Python](https://github.com/DataDog/dd-trace-py/pull/7514) and [NodeJS](https://github.com/DataDog/dd-trace-js/pull/4069).

# Additional Notes
Screenshots of feature:
sfn-exec
sfn-sync-exec
Screenshot 2025-02-06 at 4 36 43 PM

# Contributor Checklist

Jira ticket: [[SVLS-5249]](https://datadoghq.atlassian.net/browse/SVLS-5249)

[SVLS-5249]: https://datadoghq.atlassian.net/browse/SVLS-5249?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ


dd-java-agent/instrumentation/aws-java-sdk-2.2/src/main/java/datadog/trace/instrumentation/aws/v2/AwsSdkClientDecorator.java
```diff
@@ -398,6 +403,37 @@ public AgentSpan onSdkResponse(
...
 
+  public static void injectTraceToStepFunctionInput(SdkRequest request, AgentSpan span) {
+    Class<?> clazz = request.getClass();
+    if ((clazz.getSimpleName().equals("StartExecutionRequest"))
+        || clazz.getSimpleName().equals("StartSyncExecutionRequest")) {
+      try {
+        Method method = clazz.getMethod("input");
+        String input = (String) method.invoke(request);
+
+        ObjectMapper mapper = new ObjectMapper();
```
amarziali, line 414: This object can be created once since it's thread safe

dd-java-agent/instrumentation/aws-java-sdk-2.2/src/main/java/datadog/trace/instrumentation/aws/v2/AwsSdkClientDecorator.java
```diff
@@ -398,6 +403,37 @@ public AgentSpan onSdkResponse(
     return span;
   }
 
+  public static void injectTraceToStepFunctionInput(SdkRequest request, AgentSpan span) {
+    Class<?> clazz = request.getClass();
+    if ((clazz.getSimpleName().equals("StartExecutionRequest"))
```
amarziali, line 408: May `instanceof` be used here?

dd-java-agent/instrumentation/aws-java-sdk-2.2/src/main/java/datadog/trace/instrumentation/aws/v2/AwsSdkClientDecorator.java
```diff
@@ -398,6 +403,37 @@ public AgentSpan onSdkResponse(
...
+          traceContext.put("x-datadog-trace-id", span.getTraceId().toString());
+          traceContext.put("x-datadog-parent-id", String.valueOf(span.getSpanId()));
+          traceContext.put("x-datadog-sampling-priority", span.getSamplingPriority().toString());
+          traceContext.put("x-datadog-tags", mapper.writeValueAsString(span.getTags()));
+
+          ObjectNode ddInputNode = (ObjectNode) inputJsonNode;
+          ddInputNode.put("_datadog", traceContext);
+        }
+
+        Field inputField = clazz.getDeclaredField("input");
```
amarziali, line 427: This is pretty inefficient and is discouraged. Instead you might obtain (once a `MethodHandle`) and unreflect it in order to avoid paying that extra cost

amarziali (changes requested):
Using reflection is not efficient and should be avoided. Instead, MethodHandles can be used.
The pull request lacks also of a minimum test coverage. Test must be added

dd-java-agent/instrumentation/aws-java-sdk-2.2/src/main/java/datadog/trace/instrumentation/aws/v2/AwsSdkClientDecorator.java
mcculls: Main issue is that not all AWS types will be available on the classpath - it depends which parts of the SDK the user deploys. So you can't use types directly like with `instanceof`.

However, I'd prefer to see this use the fully qualified class name, because there is a cost to calling `getSimpleName` and we know the package(s) ahead of time.

dd-java-agent/instrumentation/aws-java-sdk-2.2/src/main/java/datadog/trace/instrumentation/aws/v2/AwsSdkClientDecorator.java
mcculls: Agreed - I would prefer this to be moved to a step-function specific instrumentation module, like the SNS specific instrumentation module(s) added in https://github.com/DataDog/dd-trace-java/pull/6908 - you can still intercept the request and modify it as appropriate, but you'll have access to the right types and can directly target the appropriate part of the AWS SDK

Trying to do this in the generic AWS SDK instrumentation will make this code convoluted and hard to maintain.

I'd also recommend avoiding adding a dependency to `com.fasterxml.jackson.databind` in the instrumentation, because that might not be accessible in certain deployments - ideally take the same approach as `SnsInterceptor` and just build up the simple string (the JSON isn't that complex)

In fact it's probably a good idea to refactor that `_datadog` JSON building logic out so it can be shared across all the different AWS SDK interceptors that want to add `_datadog` to their requests...

DylanLovesCoffee: Ah thanks! That helps a ton. I'll give that a shot to try to clean things up + add test coverage

dd-java-agent/instrumentation/aws-java-sfn-2.0/src/main/java/datadog/trace/instrumentation/aws/v2/sfn/InputAttributeInjector.java
```diff
@@ -0,0 +1,44 @@
...
+        String.format(
+            "\"_datadog\": { \"x-datadog-trace-id\": \"%s\",\"x-datadog-parent-id\":\"%s\", \"x-datadog-tags\": %s }",
+            span.getTraceId().toString(), span.getSpanId(), spanTagsJSON);
+
+    return ddTraceContextJSON;
+  }
+
+  public static StringBuilder getModifiedInput(String request, String ddTraceContextJSON) {
+    StringBuilder modifiedInput = new StringBuilder(request);
+    int startPos = modifiedInput.indexOf("{");
```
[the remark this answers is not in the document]
DylanLovesCoffee: Hmm StringBuilder.indexOf() does not take a char type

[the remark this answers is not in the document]
PerfectSlayer: According to the Javadoc:

dd-java-agent/instrumentation/aws-java-sfn-2.0/build.gradle
```diff
@@ -0,0 +1,29 @@
...
+addTestSuiteForDir('latestDepTest', 'test')
+addTestSuiteExtendingForDir('latestDepForkedTest', 'latestDepTest', 'test')
+
+dependencies {
+  compileOnly group: 'software.amazon.awssdk', name: 'sfn', version: '2.15.35'
+
+  // Include httpclient instrumentation for testing because it is a dependency for aws-sdk.
+  testImplementation project(':dd-java-agent:instrumentation:apache-httpclient-4')
+  testImplementation project(':dd-java-agent:instrumentation:aws-java-sdk-2.2')
+  testImplementation 'software.amazon.awssdk:sfn:2.27.2'
```
PerfectSlayer, line 21: It mismatches muzzle requirements. Is it expected?

dd-java-agent/instrumentation/aws-java-sfn-2.0/build.gradle
```diff
@@ -0,0 +1,29 @@
...
+  testImplementation 'org.testcontainers:localstack:1.19.7'
```
PerfectSlayer, line 22: Can you pick the version from `lib.versions` instead to keep it aligned with `testcontainers`?

dd-java-agent/instrumentation/aws-java-sfn-2.0/src/main/java/datadog/trace/instrumentation/aws/v2/sfn/InputAttributeInjector.java
```diff
@@ -0,0 +1,44 @@
...
+    int endPos = modifiedInput.lastIndexOf("}");
+    String inputContent = modifiedInput.substring(startPos + 1, endPos);
+    if (inputContent.isEmpty()) {
+      modifiedInput.insert(endPos, ddTraceContextJSON);
+    } else {
+      modifiedInput.insert(
+          endPos, String.format(", %s", ddTraceContextJSON)); // prepend comma to existing input
```
PerfectSlayer, line 40: What about using`String.insert()` twice, or simply using string concat?

dd-java-agent/instrumentation/aws-java-sfn-2.0/src/main/java/datadog/trace/instrumentation/aws/v2/sfn/InputAttributeInjector.java
```diff
@@ -0,0 +1,44 @@
...
+public class InputAttributeInjector {
+  public static String buildTraceContext(AgentSpan span) {
+    // Extract span tags
+    StringBuilder spanTagsJSON = new StringBuilder();
+    spanTagsJSON.append('{');
+    span.getTags()
+        .forEach(
+            (tagKey, tagValue) ->
+                spanTagsJSON
+                    .append("\"")
```
PerfectSlayer, line 14: append('"')

dd-java-agent/instrumentation/aws-java-sfn-2.0/src/main/java/datadog/trace/instrumentation/aws/v2/sfn/InputAttributeInjector.java
```diff
@@ -0,0 +1,44 @@
...
+    // Build DD trace context object
+    String ddTraceContextJSON =
+        String.format(
+            "\"_datadog\": { \"x-datadog-trace-id\": \"%s\",\"x-datadog-parent-id\":\"%s\", \"x-datadog-tags\": %s }",
+            span.getTraceId().toString(), span.getSpanId(), spanTagsJSON);
+
+    return ddTraceContextJSON;
+  }
+
+  public static StringBuilder getModifiedInput(String request, String ddTraceContextJSON) {
```
PerfectSlayer, line 31: Result always used with `.toString()`. Better return `String` here.

dd-java-agent/instrumentation/aws-java-sfn-2.0/src/main/java/datadog/trace/instrumentation/aws/v2/sfn/SfnInterceptor.java
```diff
@@ -0,0 +1,55 @@
...
+      StartExecutionRequest request = (StartExecutionRequest) context.request();
+      if (request.input() == null) {
+        return request;
+      }
+      String ddTraceContextJSON = InputAttributeInjector.buildTraceContext(span);
+      // Inject the trace context into the Step Function input
+      StringBuilder modifiedInput =
+          InputAttributeInjector.getModifiedInput(request.input(), ddTraceContextJSON);
+
+      return request.toBuilder().input(modifiedInput.toString()).build();
```
PerfectSlayer, line 36: Can be deduplicated using a dedicated method:
```java
SdkRequest injectTraceContext(request, span) {
      String traceContext = InputAttributeInjector.buildTraceContext(span);
      // Inject the trace context into the Step Function input
      String modifiedInput = InputAttributeInjector.getModifiedInput(request.input(), traceContext);
      return request.toBuilder().input(modifiedInput).build()
}

File: dd-java-agent/instrumentation/aws-java-sfn-2.0/src/test/groovy/SfnClientTest.groovy [hunk 11]
```diff
@@ -0,0 +1,137 @@
...
+import software.amazon.awssdk.auth.credentials.AwsBasicCredentials
+import spock.lang.Shared
+
+import java.time.Duration
+
+import static datadog.trace.agent.test.utils.TraceUtils.basicSpan
+
+
+abstract class SfnClientTest extends VersionedNamingTestBase {
+  static final LOCALSTACK = new GenericContainer(DockerImageName.parse("localstack/localstack"))
```
PerfectSlayer, line 21: Why not using a shared resource like clients and initialization into `setupSpec`?

dd-java-agent/instrumentation/aws-java-sfn-2.0/src/test/groovy/SfnClientTest.groovy
```diff
@@ -0,0 +1,137 @@
...
+  .withExposedPorts(4566)
+  .withEnv("SERVICES", "stepfunctions")
+  .withReuse(true)
+  .withStartupTimeout(Duration.ofSeconds(120))
+
+  @Shared SfnClient sfnClient
```
PerfectSlayer, line 27: `close()` missing on `cleanupSpec`

dd-java-agent/instrumentation/aws-java-sfn-2.0/src/test/groovy/SfnClientTest.groovy
```diff
@@ -0,0 +1,137 @@
...
+    StartExecutionResponse response
+    TraceUtils.runUnderTrace('parent', {
+      response = sfnClient.startExecution { builder ->
+        builder.stateMachineArn(testStateMachineARN)
+          .input("{\"key\": \"value\"}")
+          .build()
+      }
+    })
+
+    def endPoint = "http://" + LOCALSTACK.getHost() + ":" + LOCALSTACK.getMappedPort(4566)
```
PerfectSlayer, line 63: Can be define as fixture field to avoid duplication (already computed at setup)

dd-java-agent/instrumentation/aws-java-sfn-2.0/src/test/groovy/SfnClientTest.groovy
```diff
@@ -0,0 +1,137 @@
...
+
+
+    then:
+    assertTraces(1) {
+      trace(2) {
+        basicSpan(it, "parent")
+        span {
+          serviceName "java-aws-sdk"
+          operationName "aws.http"
```
PerfectSlayer, line 72: Use `service()` and `operation()` instead

PerfectSlayer (commented):
I left a bunch of comments about refactoring and testing

dd-java-agent/instrumentation/aws-java-sfn-2.0/src/main/java/datadog/trace/instrumentation/aws/v2/sfn/SfnInterceptor.java
DylanLovesCoffee: Since I had to handle both `StartExecutionRequest` and `StartSyncExecutionRequest` param types I ended up doing method overloading for `injectTraceContext()`

DylanLovesCoffee:
@PerfectSlayer I think I covered all of the comments/corrections, but for some reason the CI's muzzle check is continuing to fail :/

dd-java-agent/instrumentation/aws-java-sfn-2.0/src/main/java/datadog/trace/instrumentation/aws/v2/sfn/SfnInterceptor.java
```diff
@@ -0,0 +1,63 @@
...
+public class SfnInterceptor implements ExecutionInterceptor {
+
+  public static final ExecutionAttribute<AgentSpan> SPAN_ATTRIBUTE =
+      InstanceStore.of(ExecutionAttribute.class)
+          .putIfAbsent("DatadogSpan", () -> new ExecutionAttribute<>("DatadogSpan"));
+
+  public SfnInterceptor() {}
+
+  @Override
+  public SdkRequest modifyRequest(
```
dougqh, line 22: We should probably guard against any exceptions escaping from our listeners.
I'd suggest moving the logic into a modifyRequestImpl and then have modifyRequest wrap modifyRequestImpl and catch any exceptions.

dougqh (dismissed):
I'd like to see the Json construction done in a safer way.
I also think we should guard against exceptions propagating out of the listener.

dd-java-agent/instrumentation/aws-java-sfn-2.0/src/main/java/datadog/trace/instrumentation/aws/v2/sfn/InputAttributeInjector.java
```diff
@@ -0,0 +1,44 @@
+package datadog.trace.instrumentation.aws.v2.sfn;
+
+import datadog.trace.bootstrap.instrumentation.api.AgentSpan;
+
+public class InputAttributeInjector {
+  public static String buildTraceContext(AgentSpan span) {
+    // Extract span tags
+    StringBuilder spanTagsJSON = new StringBuilder();
```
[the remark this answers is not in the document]
DylanLovesCoffee: Ah I updated this with the JsonBuffer but when I run muzzle locally it fails with 
```
error: cannot find symbol
import datadog.trace.bootstrap.JsonBuffer;
                              ^
  symbol:   class JsonBuffer
  location: package datadog.trace.bootstrap
```
Any ideas on if I did something very obviously wrong?

dd-java-agent/instrumentation/aws-java-sfn-2.0/src/main/java/datadog/trace/instrumentation/aws/v2/sfn/InputAttributeInjector.java
```diff
@@ -0,0 +1,46 @@
...
+import datadog.trace.bootstrap.instrumentation.api.AgentSpan;
+
+public class InputAttributeInjector {
+  public static String buildTraceContext(AgentSpan span) {
+    // Extract span tags
+    JsonBuffer spanTagsJSON = new JsonBuffer();
+    spanTagsJSON.beginObject();
+    span.getTags()
+        .forEach((tagKey, tagValue) -> spanTagsJSON.name(tagKey).value(tagValue.toString()));
+    spanTagsJSON.endObject();
```
avedmala, line 13: Is this JSON format for the `x-datadog-tags` expected? Not urgent but need to know at some point if we need to handle this format or not.

Usually we expect something like `_dd.p.dm=-0,_dd.p.tid=658b547000000000`

dd-java-agent/instrumentation/aws-java-sdk-2.2/src/main/java/datadog/trace/instrumentation/aws/v2/AwsSdkClientDecorator.java
nhulston: Hey @amarziali, I'm working on this PR now since Dylan is out. Marking this as resolved since it appears we don't use reflection anymore.

dd-java-agent/instrumentation/aws-java-sfn-2.0/src/main/java/datadog/trace/instrumentation/aws/v2/sfn/InputAttributeInjector.java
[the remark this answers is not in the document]
nhulston: I ran into the same muzzle issue trying to use a JSON parsing library when doing the EventBridge instrumentation. I was basically told we can't use additional libraries in instrumentations and to do the JSON parsing manually with stringbuilder.

There are obviously security concerns with this, so it looks like we're blocked until there is a standardized way to parse JSONs in instrumentations in this repo.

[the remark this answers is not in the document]
PerfectSlayer: I just merge #7973 It should provide API to build JSON payload from instrumentations (see `JsonWriter` and `JsonMapper`).

amarziali:
@nhulston is this PR ready to be reviewed ? I see tests are failing. Otherwise it can be put in draft for now

dd-java-agent/instrumentation/aws-java-sfn-2.0/src/main/java/datadog/trace/instrumentation/aws/v2/sfn/InputAttributeInjector.java
DylanLovesCoffee: Clarifying note here: `x-datadog-tags` should actually be comma separated key values 📝

nhulston:
Failures are unrelated (looks like AWS never published EventBridge 2.30.15 to Maven?)
https://mvnrepository.com/artifact/software.amazon.awssdk/eventbridge/2.30.15

PerfectSlayer (approved):
Looking good from Platform PoV. I will let IDM review instrumentation.

About the failing tests, you can retry them following: https://github.com/DataDog/dd-trace-java/blob/master/docs/how_to_test.md#flaky-tests

dd-java-agent/instrumentation/aws-java-sfn-2.0/src/main/java/datadog/trace/instrumentation/aws/v2/sfn/InputAttributeInjector.java
```diff
@@ -0,0 +1,38 @@
...
+    int endPos = modifiedInput.lastIndexOf("}");
+
+    String inputContent = modifiedInput.substring(startPos + 1, endPos).trim();
+    if (inputContent.isEmpty()) {
+      modifiedInput.insert(endPos, String.format("\"%s\":%s", DATADOG_KEY, ddTraceContextJSON));
+    } else {
+      // Prepend comma to separate from existing content
+      modifiedInput.insert(endPos, String.format(",\"%s\":%s", DATADOG_KEY, ddTraceContextJSON));
+    }
+    return modifiedInput.toString();
```
mcculls, line 36: This method could be made more performant with the following approach:
* skip trimming the string before finding `{` and `}` to avoid creating an intermediate string
* that would also let you use `request.indexOf('{')` etc. which is faster
* compare the numerical positions of `{` and `}` 
  * if they're next to each other, no need to create/trim anything and you can return `"{" + DATADOG_KEY + ":" + ddTraceContextJSON + "}"`
  * otherwise get the substring between `{` and `}` trim it and store it in `existingJSON`
  * if `existingJSON` is empty you can return `"{" + DATADOG_KEY + ":" + ddTraceContextJSON + "}"`
  * otherwise return `"{" + existingJSON + "," + DATADOG_KEY + ":" + ddTraceContextJSON + "}"`

you could also extract `DATADOG_KEY + ":" + ddTraceContextJSON` to a local variable to avoid repeating it

mcculls: Also I assume we know the original string will always be JSON?

nhulston: Yes, the original string must be a JSON string (if empty, it must be `{}`) in order for AWS to accept the request. However, the user could end up putting an incorrect json string; in that case, we can leave the request unchanged so that the user will see the error on from AWS, not from the tracer.

dd-java-agent/instrumentation/aws-java-sfn-2.0/src/main/java/datadog/trace/instrumentation/aws/v2/sfn/InputAttributeInjector.java
```diff
@@ -0,0 +1,49 @@
...
+      writer.endObject();
+      return writer.toString();
+    } catch (Exception e) {
+      return "{}";
+    }
+  }
+
+  public static String getModifiedInput(String request, String ddTraceContextJSON) {
+    if (request == null || ddTraceContextJSON == null) {
+      throw new IllegalArgumentException(); // leave request unmodified
```
mcculls, line 26:
```suggestion
      return request;
```
ideally we shouldn't be throwing exceptions in our code in case they escape the advice and affect the application

exceptions are also not cheap

dd-java-agent/instrumentation/aws-java-sfn-2.0/src/main/java/datadog/trace/instrumentation/aws/v2/sfn/InputAttributeInjector.java
```diff
@@ -0,0 +1,49 @@
...
+    }
+
+    final String traceContextProperty = "\"" + DATADOG_KEY + "\":" + ddTraceContextJSON;
+    int startPos = request.indexOf('{');
+    int endPos = request.lastIndexOf('}');
+
+    if (startPos < 0 || endPos < startPos) {
+      throw new IllegalArgumentException(); // leave request unmodified
```
mcculls, line 34:
```suggestion
      return request;
```
ideally we shouldn't be throwing exceptions in our code in case they escape the advice and affect the application

exceptions are also not cheap

dd-java-agent/instrumentation/aws-java-sfn-2.0/src/main/java/datadog/trace/instrumentation/aws/v2/sfn/InputAttributeInjector.java
```diff
@@ -0,0 +1,49 @@
...
+    String tagsJson = JsonMapper.toJson(span.getTags());
+    try (JsonWriter writer = new JsonWriter()) {
+      writer.beginObject();
+      writer.name("x-datadog-trace-id").value(span.getTraceId().toString());
+      writer.name("x-datadog-parent-id").value(String.valueOf(span.getSpanId()));
+      writer.name("x-datadog-tags").jsonValue(tagsJson);
+      writer.endObject();
+      return writer.toString();
+    } catch (Exception e) {
+      return "{}";
```
mcculls, line 20:
```suggestion
      return null;
```
to match the check in `getModifiedInput` to leave the request unmodified

mcculls (approved):
Thanks for the work on this - I've suggested a couple of final improvements

amarziali (approved):

PR #7613, opened 2024-09-12
Inject trace context into EventBridge detail

nhulston:
# What Does This Do
This creates a new instrumentation for EventBridge and intercepts `PutEventsRequest` to inject trace context. This allows the agent to combine spans from a distributed (serverless) architecture into a single trace.

This PR only injects trace context. I'm working on [PR 1](https://github.com/DataDog/datadog-agent/pull/29414) and [PR 2](https://github.com/DataDog/datadog-agent/pull/29551) to update the Lambda extension to use this trace context to create EventBridge spans.

# Motivation
SNS and SQS are already supported, and the tracer currently injects trace context into message attributes fields for them. However, EventBridge wasn't supported, and this PR aims to fix this problem.

# Additional Notes
Overall, AWS's EventBridge API is lacking some features, so we have to do some hacky solutions.
- SNS and SQS call their custom input field `messageAttributes`, and EventBridge calls it `detail`
- Unlike SNS and SQS, the `detail` field is given as a raw string. Therefore, we have to manually modify the `detail` string using StringBuilder.
- The agent has no reliable way of getting the start time of the EventBridge span, so the tracer has to put the current time into `detail[_datadog]` as `x-datadog-start-time`
- The EventBridge API has no way of getting the EventBridge bus name, so the tracer has to put the bus name (which is used to create the span resource name) into `detail[_datadog]` as `x-datadog-resource-name`

# Traces before these changes
**Lambda --> EventBridge --> Lambda**
Two different traces. Second trace is missing an EventBridge span
Screenshot 2024-09-12 at 4 50 56 PM

**Lambda --> EventBridge --> SQS --> Lambda**
Missing EventBridge span
Screenshot 2024-09-12 at 4 51 14 PM

**Lambda --> EventBridge --> SNS --> Lambda**
Missing EventBridge span
Screenshot 2024-09-12 at 4 51 06 PM

# Traces after these (and agent's) changes
**Lambda --> EventBridge --> Lambda**
Screenshot 2024-09-12 at 4 53 26 PM

**Lambda --> EventBridge --> SQS --> Lambda**
Screenshot 2024-09-12 at 4 53 18 PM

**Lambda --> EventBridge --> SNS --> Lambda**
Screenshot 2024-09-12 at 4 53 22 PM

# Contributor Checklist

Jira ticket: [SVLS-5666]

[SVLS-5666]: https://datadoghq.atlassian.net/browse/SVLS-5666?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ


dd-java-agent/instrumentation/aws-java-eventbridge-2.0/src/main/java/datadog/trace/instrumentation/aws/v2/eventbridge/EventBridgeInterceptor.java
```diff
@@ -0,0 +1,115 @@
...
+    propagate().inject(span, jsonBuilder, SETTER, TracePropagationStyle.DATADOG);
+    if (traceConfig().isDataStreamsEnabled()) {
+      propagate().injectPathwayContext(span, jsonBuilder, SETTER, getTags(eventBusName));
+    }
+    jsonBuilder.setLength(jsonBuilder.length() - 1); // Remove the last comma
+    jsonBuilder.append("}");
+    return jsonBuilder.toString();
+  }
+
+  private LinkedHashMap<String, String> getTags(String eventBusName) {
```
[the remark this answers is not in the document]
nhulston: `injectPathwayContext()` takes a parameter of `LinkedHashMap`, so we have to call have it return a LinkedHashMap instead of Map

dd-trace-core/src/main/java/datadog/trace/core/datastreams/TagsProcessor.java
```diff
@@ -33,9 +33,14 @@ public String apply(String key) {
...
   private static final Function<String, String> DIRECTION_TAG_PREFIX =
       new StringPrefix("direction:");
+  // SNS Topic
   public static final String TOPIC_TAG = "topic";
   private static final DDCache<String, String> TOPIC_TAG_CACHE = DDCaches.newFixedSizeCache(32);
   private static final Function<String, String> TOPIC_TAG_PREFIX = new StringPrefix("topic:");
+  // EventBridge Bus
+  public static final String BUS_TAG = "bus";
+  private static final DDCache<String, String> BUS_TAG_CACHE = DDCaches.newFixedSizeCache(32);
+  private static final Function<String, String> BUS_TAG_PREFIX = new StringPrefix("bus:");
```
nhulston, line 43: I'm actually not 100% sure what this does, but I was copying what the implementation for SNS did. Are these tags to be used by users for filtering purposes, or does it do something else?

dd-trace-core/src/main/java/datadog/trace/lambda/MoshiJsonAdapter.java
```diff
@@ -0,0 +1,32 @@
...
+          .add(SkipUnsupportedTypeJsonAdapter.newFactory())
+          .build()
+          .adapter(Object.class);
+
+  @Override
+  public Map<String, Object> fromJson(String json) throws IOException {
+    if (json == null || json.isEmpty()) {
+      return new HashMap<>();
+    }
+    @SuppressWarnings("unchecked")
```
nhulston, line 23: not sure if this line is necessary. Just stops your IDE from yelling at you

internal-api/src/main/java/datadog/trace/bootstrap/instrumentation/api/AgentTracer.java
```diff
@@ -525,6 +528,11 @@ public AgentHistogram newHistogram(double relativeAccuracy, int maxNumBins) {
     public void updatePreferredServiceName(String serviceName) {
       // no ops
     }
+
+    @Override
+    public JsonAdapter getJsonAdapter() {
+      return null;
+    }
```
nhulston, line 535: What's the point of the NoopTracerAPI?

dd-java-agent/instrumentation/aws-java-eventbridge-2.0/src/main/java/datadog/trace/instrumentation/aws/v2/eventbridge/EventBridgeClientInstrumentation.java
duncanista, line 1: Does this also cover Async client?
Do other AWS instrumentation do Async client instrumentation?

nhulston: Yes, this targets `software.amazon.awssdk.core.client.builder.SdkDefaultClientBuilder` which is the builder for synchronous and async clients.

This file is practically the same as the instrumentation for SNS: https://github.com/DataDog/dd-trace-java/blob/master/dd-java-agent/instrumentation/aws-java-sns-2.0/src/main/java/datadog/trace/instrumentation/aws/v2/sns/SnsClientInstrumentation.java

duncanista: Let's add integration testing for Async clients too

dd-java-agent/instrumentation/aws-java-eventbridge-2.0/src/main/java/datadog/trace/instrumentation/aws/v2/eventbridge/EventBridgeInterceptor.java
```diff
@@ -0,0 +1,104 @@
...
+public class EventBridgeInterceptor implements ExecutionInterceptor {
+  private final JsonAdapter jsonAdapter;
+  private static final Logger log = LoggerFactory.getLogger(EventBridgeInterceptor.class);
+
+  public static final ExecutionAttribute<AgentSpan> SPAN_ATTRIBUTE =
+      InstanceStore.of(ExecutionAttribute.class)
+          .putIfAbsent("DatadogSpan", () -> new ExecutionAttribute<>("DatadogSpan"));
+
+  private static final String SENT_TIMESTAMP_KEY = "SentTimestamp";
+  private static final String BUS_NAME_KEY = "BusName";
```
nhulston, line 39: `SentTimestamp` is already used as a key by the extension. `BusName` is new, should it be renamed?

dd-java-agent/instrumentation/aws-java-eventbridge-2.0/src/main/java/datadog/trace/instrumentation/aws/v2/eventbridge/EventBridgeInterceptor.java
```diff
@@ -0,0 +1,104 @@
...
+
+    return request.toBuilder().entries(modifiedEntries).build();
+  }
+
+  private String getTraceContextToInject(
+      ExecutionAttributes executionAttributes, String eventBusName) {
+    final AgentSpan span = executionAttributes.getAttribute(SPAN_ATTRIBUTE);
+    StringBuilder jsonBuilder = new StringBuilder();
+    jsonBuilder.append('{');
+    propagate().inject(span, jsonBuilder, SETTER, TracePropagationStyle.DATADOG);
```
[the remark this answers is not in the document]
duncanista: Yeah, it should use whatever the configuration is set to, not arbitrarily setting Datadog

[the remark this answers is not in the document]
nhulston: Is there a priority/ordering of propagation styles? Since the user can set multiple styles? I was thinking we just use the order in [this file](https://github.com/DataDog/dd-trace-java/blob/4684bd066e5d26fc95ba4f0b49b895fbd3366ebe/dd-trace-api/src/main/java/datadog/trace/api/TracePropagationStyle.java):

Datadog > B3Single > B3Multi > Haystack > XRay > TraceContext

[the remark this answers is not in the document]
nhulston: I can do this easily and inject trace context for whatever style the user picks. But the agent is hardcoded to accept `x-datadog-trace-id` and `x-datadog-parent-id`. Should we always include `TracePropagationStyle.DATADOG` anyway, or accept that if the user doesn't have `datadog` set in their `DD_TRACE_PROPAGATION_STYLE` env var that the inferred spans won't work?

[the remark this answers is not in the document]
purple4reina: Great question! Let's clear up a couple of things about how trace propagation works.

The agent is **not** hardcoded to accept only `x-datadog-trace-id` and `x-datadog-parent-id`. 🙂 To help clarify, I recommend checking out [this doc on trace propagation in the layer/extension](https://docs.google.com/document/d/1bLaMJPJt4-LomtReFPQv37nRIGbhRcGPxeYXSKi3NIM/edit).

It seems like there’s some confusion between two distinct scenarios where trace context is passed:

1. **From an upstream service to the Lambda**: This context comes in with the event payload and is configurable by the user.
2. **From the extension back to the tracing layer**: This is where we send trace data internally (via universal instrumentation), and it’s not user-configurable.

The key difference is that scenario 1 is customer-facing, while scenario 2 is internal. For scenario 1, the customer can indeed configure the trace propagation style using environment variables like `DD_TRACE_PROPAGATION_STYLE`. However, for scenario 2, we always use `x-datadog-trace-id` and `x-datadog-parent-id` because allowing changes here would break trace propagation entirely.

In your case, it seems like you’re dealing with scenario 1 —- propagating trace context from an upstream service to a downstream Lambda. Therefore, you _do_ need to strictly honor the customer’s configured trace propagation style environment variables. (Correct me if I’m misunderstanding the situation.)

With that in mind, I recommend deferring this decision to the tracer whenever possible. How does the Java tracer handle trace context injection elsewhere in the codebase? There’s likely existing logic that decides which propagation style to use based on the customer’s configuration —- tap into that.

Let me know if this makes sense!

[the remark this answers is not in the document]
nhulston: Yes that makes sense, thank you for the detailed explanation! I didn't find existing logic easily in the code, although it's very easy to implement. I was thinking of doing something like this:

```
    Set<TracePropagationStyle> propagationStyles = Config.get().getTracePropagationStylesToInject();
    if (propagationStyles.isEmpty()) {
      // handle this edge case somehow. Maybe hardcode to DATADOG style, since the user can just set their style to 'none' if they really don't want to inject trace context
    }
    for (TracePropagationStyle propagationStyle : propagationStyles) {
      propagate().inject(span, jsonBuilder, SETTER, propagationStyle);
    }
```

The current [SNS implementation](https://github.com/DataDog/dd-trace-java/blob/master/dd-java-agent/instrumentation/aws-java-sns-2.0/src/main/java/datadog/trace/instrumentation/aws/v2/sns/SnsInterceptor.java#L41) is also hardcoded to just inject using Datadog trace propagation style.

If my solution looks good to you, I can update EventBridge and SNS in this PR
cc @duncanista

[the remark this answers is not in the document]
purple4reina: I _think_ you'll want to do something [like this](https://github.com/DataDog/dd-trace-java/blob/2d2c02962a2751033511bff23e46660c5b1c47f3/docs/how_instrumentations_work.md#injectadapters--custom-getterssetters) to custom inject the headers... I'm not finding much documentation on the available apis in the java tracers, so I recommend talking to someone in #apm-java for help here.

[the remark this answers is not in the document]
purple4reina: And yeah, once you can figure out how to properly inject the correct header types, then we should probably do the same thing for SNS.

dd-java-agent/instrumentation/aws-java-sns-2.0/src/main/java/datadog/trace/instrumentation/aws/v2/sns/SnsInterceptor.java
```diff
@@ -38,7 +37,7 @@ private SdkBytes getMessageAttributeValueToInject(
     final AgentSpan span = executionAttributes.getAttribute(SPAN_ATTRIBUTE);
     StringBuilder jsonBuilder = new StringBuilder();
     jsonBuilder.append("{");
-    propagate().inject(span, jsonBuilder, SETTER, TracePropagationStyle.DATADOG);
+    propagate().inject(span, jsonBuilder, SETTER);
```
purple4reina, line 40: Can we be sure to add a test for this change too?

nhulston: I tested manually, and it works as expected and injects any and all trace propagation styles set by the user. But I can't figure out how to modify my tests to support this change--something to do with the environment variable not changing between tests. I also tried recreating a new EventBridgeClient during each loop of the test, but it doesn't work. 

I now modify the env var to have all propagation styles in `configurePreAgent()`: `'datadog,b3single,b3multi,haystack,xray,tracecontext'`. Not as thorough as my original tests, but it should still test the functionality we're going for. Thoughts on this solution?

But I think testing this isn't necessary in my instrumentation. I think this logic should be being tested in `AgentPropagation.java`, but it's not.

https://github.com/DataDog/dd-trace-java/blob/9f7d0c27cdad1d9387ad2e5979f6e533b3a1ecdc/dd-trace-core/src/test/groovy/datadog/trace/core/propagation/CorePropagationTest.groovy#L82-L112

Looks like it only tests for the default `datadog` propagation style and for when the propagation style is passed in as the 4th parameter to the `inject() method`

purple4reina (approved):
This looks good to me!

PerfectSlayer:
Quick question about the new label: is EventBridge falls under _AWS SDK_?

dd-java-agent/instrumentation/aws-java-eventbridge-2.0/src/main/java/datadog/trace/instrumentation/aws/v2/eventbridge/EventBridgeInterceptor.java
```diff
@@ -0,0 +1,121 @@
...
+import software.amazon.awssdk.services.eventbridge.model.PutEventsRequestEntry;
+
+public class EventBridgeInterceptor implements ExecutionInterceptor {
+  private static final Logger log = LoggerFactory.getLogger(EventBridgeInterceptor.class);
+
+  public static final ExecutionAttribute<AgentSpan> SPAN_ATTRIBUTE =
+      InstanceStore.of(ExecutionAttribute.class)
+          .putIfAbsent("DatadogSpan", () -> new ExecutionAttribute<>("DatadogSpan"));
+
+  private static final String DATADOG_KEY = "_datadog";
```
ygree, line 33: Why wouldn't you rather use `PathwayContext.DATADOG_KEY`?

ygree (approved):
LGTM

nhulston:
> Quick question about the new label: is EventBridge falls under _AWS SDK_?

Yes it does

duncanista (approved):

PR #7620, opened 2024-09-13
Parse 128 bit trace Id retuned by lambda extension

agocs:
# What Does This Do

I've updated the Lambda extension so it is capable of returning a 128 bit trace ID when a tracer calls the `/lambda/start-invocation` endpoint [in this PR](https://github.com/DataDog/datadog-agent/pull/27988)

As per [the RFC](https://datadoghq.atlassian.net/wiki/spaces/RUMP/pages/3545630931/RFC+Support+128+bit+trace+IDs+in+RUM+SDKs#:~:text=For%20Datadog%20headers%2C%20the%20128%20bit%20trace%20id%20is%20sent%20in%20two%20parts%2C%20lower%2064%20bits%20in%20x%2Ddatadog%2Dtrace%2Did%20(decimal)%20and%20the%20higher%2064%20bits%20in%20x%2Ddatadog%2Dtags%20header%20under%20_dd.p.tid%20(hex)%20tag), the

> lower 64 bits in `x-datadog-trace-id` (decimal) and the higher 64 bits in `x-datadog-tags` header under `_dd.p.tid` (hex) tag.

This change modifies the function that calls `/lambda/start-invocation`, allowing it to pick out the upper 64 bits of the trace ID and set the resulting 128-bit trace ID in the extracted context.

I labelled this with `tag: no release notes` because it just changes behavior between the tracer and the Lambda extension.

# Motivation

Step Function traces create 128 bit trace IDs. Soon everything else will too.

# Additional Notes

# Contributor Checklist

Jira ticket: [SVLS-5583]

[SVLS-5583]: https://datadoghq.atlassian.net/browse/SVLS-5583?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ


PerfectSlayer (commented):
I don't have the full context so I might ask a dumb question here (but I'm going to ask nonetheless 😇 ): why don't you use context propagation api with Datadog style here?

agocs:
Hi @PerfectSlayer, I'm extending code that was written two years ago. Can you point me to an example of the context propagation API?

PerfectSlayer:
Sure, if pass the `CoreTracer` instance as a `notifyStartInvocation()` parameter, you can then use:
```java
        Extracted context = tracer.propagate().extract(response.headers(), (carrier, classifier) -> {
          for (String headerName : carrier.names()) {
            classifier.accept(headerName, carrier.get(headerName));
          }
        });
```
`propagate()` gets the context propagation API and `extract()` extract context from a carrier.
Here, your carrier are HTTP headers, and you need to have a quick glue to explain it how to iterate over headers key-values (the lambda as 2nd argument).

Using the context propagation API will ensure you will get the same behavior as any other instrumentations, respect customer config and benefits from any future evolution.

agocs:
Thanks! I'll look into the failing checks and remove the old implementation of `notifyStartInvocation()` in the morning.

dd-trace-core/src/main/java/datadog/trace/lambda/LambdaHandler.java
```diff
@@ -73,6 +75,35 @@ public class LambdaHandler {
...
+            .extract(
+                response.headers(),
+                (carrier, classifier) -> {
+                  for (String headerName : carrier.names()) {
+                    classifier.accept(headerName, carrier.get(headerName));
+                  }
+                });
+      }
+    } catch (Throwable ignored) {
+      log.error("could not reach the extension");
```
PerfectSlayer, line 102: Is it expected to show an error to the customer in this case?

agocs: Yes, this means the user is starting a Java Lambda function instrumented with dd-trace-js, but it doesn't have the Lambda Extension, which means it can't send metrics etc. back to Datadog. It should leave an error in the logs so we can diagnose.

agocs:
What's the right way to new up a CoreTracer in the Groovy tests? [This](https://github.com/DataDog/dd-trace-java/pull/7620/files#diff-e229343be4838f9578edb1c6fd2ca2c8581f60e782fc95b1fd88a5a782ffe888R69) seems to be giving me some problems.

PerfectSlayer:
Using `DDCoreSpecification` and `tracerBuilder().build()` is the right way (plus `tracer.close()` on `cleanUp`).

Few things though:

1. You might want to pass the custom `Config` instance to the tracer builder using `config()`
2. Why are you trying to set xDatadogMaxLength to 512? It's already the default value 🤔

agocs:
Thanks @PerfectSlayer. `CoreTracer ct = tracerBuilder().config(config).build()` was giving me null pointer exceptions, so I opted to get rid of `Config config = Mock(Config)`. I honestly don't know why we were mocking config; I'm just doing my best to modify two year old code.

PerfectSlayer:
> I'm just doing my best to modify two year old code.

And tests look in good shape right now :wink:

Is there anything you have trouble with or want to improve as part of your PR?

agocs:
@PerfectSlayer No, I'm ready to merge if you are

nayeem-kamal (approved):

PerfectSlayer (approved):
Looks good to me! Thanks for upgrading the code

I include the change to the release note as it is useful change to be aware of.

PR #7548, opened 2024-09-02
Add collection of local var for method probe

jpbempel:
# What Does This Do
Collecting local var was disabled for method probe/unhanded exception because all local vars are not accessible/defined. But for Exception Debugging/Replay this is very useful to have. The solution is to hoist local variable at the top level of the method and initialize them to be able to capture them at any point of the method.
we are handling also the case of duplicate local var in different scope but with incompatible types

# Motivation
Collecting local vars even if uncaught exception

# Additional Notes

# Contributor Checklist

- [ ] Format the title [according the contribution guidelines](https://github.com/DataDog/dd-trace-java/blob/master/CONTRIBUTING.md#title-format)
- [ ] Assign the `type:` and (`comp:` or `inst:`) labels in addition to [any usefull labels](https://github.com/DataDog/dd-trace-java/blob/master/CONTRIBUTING.md#labels)
- [ ] Squash your commits prior merging or merge using GitHub's [Squash and merge](https://docs.github.com/en/pull-requests/collaborating-with-pull-requests/incorporating-changes-from-a-pull-request/about-pull-request-merges#squash-and-merge-your-commits)
- [ ] Don't use `close`, `fix` or any [linking keywords](https://docs.github.com/en/issues/tracking-your-work-with-issues/linking-a-pull-request-to-an-issue#linking-a-pull-request-to-an-issue-using-a-keyword) when referencing an issue.  
  Use `solves` instead, and assign the PR [milestone](https://github.com/DataDog/dd-trace-java/milestones) to the issue
- [ ] Update the [public documentation](https://docs.datadoghq.com/tracing/trace_collection/library_config/java/) in case of new configuration flag or behavior

Jira ticket: [DEBUG-2778]

PR #7624, opened 2024-09-16
Fix local var hoisting

jpbempel:
# What Does This Do
Rewrite the algorithm for hoisting local vars:
Go through all local vars defined in the method and determine by name all local var that are hoistable.
local vars are hoistable if there is no conflict by slot or by name with another.
If conflict by slot and same name we cancel hoisting for this name. if
 name is different we allocating a new slot for one of the local.
If conflict by name and type are not compatible we cancel hoisting for this name. If type compatible, we can add in hoistable list. Then we are considering all hoistable variables by name and if there are multiple occurrence for a name, we are merging those variables in a new slot and rewrite var instructions to use this new slot. if only one variable we just assign a new slot and rewrite the instructions in the range.

# Motivation
Fixing issues since #7548 

# Additional Notes
Tested against exploration tests

# Contributor Checklist

Jira ticket: [DEBUG-2841]

[DEBUG-2841]: https://datadoghq.atlassian.net/browse/DEBUG-2841?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ

### DIFF
--- a/dd-java-agent/agent-debugger/src/main/java/com/datadog/debugger/instrumentation/CapturedContextInstrumentor.java
+++ b/dd-java-agent/agent-debugger/src/main/java/com/datadog/debugger/instrumentation/CapturedContextInstrumentor.java
@@ -54,6 +54,7 @@ import org.objectweb.asm.tree.AbstractInsnNode;
 import org.objectweb.asm.tree.ClassNode;
 import org.objectweb.asm.tree.FieldInsnNode;
 import org.objectweb.asm.tree.FieldNode;
+import org.objectweb.asm.tree.IincInsnNode;
 import org.objectweb.asm.tree.InsnList;
 import org.objectweb.asm.tree.InsnNode;
 import org.objectweb.asm.tree.JumpInsnNode;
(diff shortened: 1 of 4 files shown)


dd-java-agent/agent-debugger/src/main/java/com/datadog/debugger/instrumentation/CapturedContextInstrumentor.java
```diff
@@ -459,75 +460,175 @@ private void instrumentMethodEnter() {
...
+    for (AbstractInsnNode insn = localVar.start;
+        insn != null && insn != localVar.end;
+        insn = insn.getNext()) {
+      if (insn instanceof VarInsnNode) {
+        VarInsnNode varInsnNode = (VarInsnNode) insn;
+        if (varInsnNode.var == oldSlot) {
+          varInsnNode.var = newSlot;
+        }
+      }
+      if (insn instanceof IincInsnNode) {
```
shatzi, line 625: does this cover all the rewrite needed be done?
for example, while searching what iinc does, I found [wide](https://docs.oracle.com/javase/specs/jvms/se7/html/jvms-6.html#jvms-6.5.wide) - which also can get variable index.

dd-java-agent/agent-debugger/src/main/java/com/datadog/debugger/instrumentation/CapturedContextInstrumentor.java
```diff
@@ -459,75 +460,175 @@ private void instrumentMethodEnter() {
...
+        newVarNode.end = findLatestLabel(methodNode.instructions, endLabels);
+        // remove previous local variables
+        methodNode.localVariables.removeIf(hoistableVars::contains);
+        methodNode.localVariables.add(newVarNode);
+      } else {
+        // hoist the single variable and rewrite all its local var instructions
+        newVarNode = hoistableVars.get(0);
+        int oldIndex = newVarNode.index;
+        newVarNode.index = newVar(getType(newVarNode.desc)); // new slot for the local variable
+        rewriteLocalVarInsn(newVarNode, oldIndex, newVarNode.index);
```
shatzi, line 507: I'm not sure that matter. but it seems we are rewrite (assign new slot) for variables that we should not.
For every hoist-able variable, even if there is no collision we rewrite it and sign it to a new slot.

Maybe add a logic that would create find new slots only for variable only if there is a conflict to be solved?

dd-java-agent/agent-debugger/src/test/java/com/datadog/debugger/agent/CapturedSnapshotTest.java
```diff
@@ -1810,12 +1811,23 @@ public void methodProbeLocalVarsDeepScopes() throws IOException, URISyntaxExcept
...
+    //       13      64     3 localVarL1   I
+    //        0      79     0  this   Lcom/datadog/debugger/CapturedSnapshot31;
+    //        0      79     1   arg   Ljava/lang/String;
+    //        2      77     2 localVarL0   I
+    assertEquals(5, snapshot.getCaptures().getReturn().getLocals().size());
     assertCaptureLocals(snapshot.getCaptures().getReturn(), "localVarL0", "int", "0");
     assertCaptureLocals(snapshot.getCaptures().getReturn(), "localVarL1", "int", "1");
     assertCaptureLocals(snapshot.getCaptures().getReturn(), "localVarL2", "int", "2");
     assertCaptureLocals(snapshot.getCaptures().getReturn(), "localVarL3", "int", "3");
-    assertCaptureLocals(snapshot.getCaptures().getReturn(), "localVarL4", "int", "4");
```
shatzi, line 1818: why localVarL4 cannot be hoisted?

dd-java-agent/agent-debugger/src/test/java/com/datadog/debugger/agent/CapturedSnapshotTest.java
```diff
@@ -1830,26 +1842,40 @@ public void methodProbeExceptionLocalVars() throws IOException, URISyntaxExcepti
...
+      value = "datadog.trace.api.Platform#isJ9",
+      disabledReason = "we cannot get local variable debug info")
+  public void duplicateLocalDifferentScope() throws IOException, URISyntaxException {
+    final String CLASS_NAME = "com.datadog.debugger.CapturedSnapshot31";
+    LogProbe probe =
+        createProbeAtExit(PROBE_ID, CLASS_NAME, "duplicateLocalDifferentScope", "(String)");
+    TestSnapshotListener listener = installProbes(CLASS_NAME, probe);
+    Class<?> testClass = compileAndLoadClass(CLASS_NAME);
+    int result = Reflect.onClass(testClass).call("main", "duplicateLocalDifferentScope").get();
+    assertEquals(28, result);
```
shatzi, line 1877: should we also asset 'ch' is not hoisted?

shatzi (approved):

dd-java-agent/agent-debugger/src/main/java/com/datadog/debugger/instrumentation/CapturedContextInstrumentor.java
jpbempel: the LocalVariableTable is first optional, and then non-exhaustive. It does not prevent slot to be reused outside of what is declared in the LVT. By hoisting we extending the scope of the slot and we may conflict with another usage that we could not anticipated just by looking at the LVT. Here we are conservative and assign in any case a new slot that way we are isolating the variable for possible conflict outside of the LVT.

dd-java-agent/agent-debugger/src/main/java/com/datadog/debugger/instrumentation/CapturedContextInstrumentor.java
jpbempel: wide is very special instruction as this is a composite one. it modifies an existing instruction like `iload` or `iinc` to handle index more than 255 or incrementation > 128.
in ASM library decoding this instruction will result in either a `VarInsnNode` or `IIncInsnNode` because index are represented with `int` anyway. It means that we are covering all instructions that way.
Here the handling of `wide` instruction by ASM: https://github.com/llbit/ow2-asm/blob/a695934043d6f8f0aee3f6867c8dd167afd4aed8/src/org/objectweb/asm/ClassReader.java#L1066-L1073

dd-java-agent/agent-debugger/src/test/java/com/datadog/debugger/agent/CapturedSnapshotTest.java
jpbempel: fixed

dd-java-agent/agent-debugger/src/test/java/com/datadog/debugger/agent/CapturedSnapshotTest.java
jpbempel: fixed

PR #7632, opened 2024-09-17
Improved stack trace reporting via telemetry

ValentinZakharov:
# What Does This Do
Added support for wrapped exceptions for error reporting via telemetry (including exceptions thrown from WAF). 

# Motivation
We encountered a difficulty while investigating a recent customer's issue. There was an error in the logs, but it only contained a top-level exception and it was completely unclear what the root cause of the exception was. For future cases, we have improved the error report so that it contains full information about all nested exceptions.

# Additional Notes
image

# Contributor Checklist

Jira ticket: [PROJ-IDENT]


telemetry/src/main/java/datadog/telemetry/log/LogPeriodicAction.java
```diff
@@ -45,43 +53,79 @@ public void doIteration(TelemetryService service) {
...
+      if (t != null) {
+        result.append("Caused by: ");
+      }
     }
-    writePendingRedacted(result, pendingRedacted);
 
     return result.toString();
   }
 
+  private static int countCommonFrames(
```
smola, line 114: In the test, can we cover a few more cases? For example, when the nested exception has no frames in common with the parent, when all of them are common, etc. This looks like the kind of code where an off-by-one error will easily slip in, or where could get some unexpected textual representation in some cases.

smola (approved):

PR #7644, opened 2024-09-19
Add proper context propagation for reactive streams

amarziali:
# What Does This Do

This PR aims definitely addressing context propagation for reactivestreams `Publisher` and `Subscriber`. It addresses also patterns (`Mono`, `Flux`, `Processor`) provided by project reactor.

The old advices we had were either not always applied (`TerminalSubscriberInstrumentation`) either not applied by default (`ReactorHooks`). In particular `TerminalSubscriberInstrumentation` was only applied when the Mono or Flux was peeking elements (i.e. methods `doOnXXXX` or `log`).
Also it had no effects on propagation. The propagation was only handled by our `java_concurrent` advices that was missing use cases like subscription on unhandled thread pools or work queues.

This PR also unify the way traced methods finish spans when the return type is a Producer. We had two different extensions:
* One generic for Publisher
* One for Mono or Flux

Considering this use case
```
@WithSpan
public Mono doSomething() {
   ....
}
```
In that case there it may happen that `ReactiveStreamsAsyncResultSupportExtension` is used (since Mono implements Publisher). This one was wrapping the publisher in an internal types causing class casts when bytebuddy replaces the return type since our wrapped type cannot be casted back to Mono

Generally speaking this PR avoid rewrapping Publisher and Subscriber in order to save a lot of object allocations (but transforms more types). 

# Motivation

# Additional Notes

# Contributor Checklist

Jira ticket: [AIDM-331]

[AIDM-331]: https://datadoghq.atlassian.net/browse/AIDM-331?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ


dd-java-agent/instrumentation/reactive-streams/src/main/java/datadog/trace/instrumentation/reactivestreams/PublisherInstrumentation.java
```diff
@@ -0,0 +1,116 @@
...
+import java.util.HashMap;
+import java.util.Map;
+import net.bytebuddy.asm.Advice;
+import net.bytebuddy.description.type.TypeDescription;
+import net.bytebuddy.matcher.ElementMatcher;
+import org.reactivestreams.Publisher;
+import org.reactivestreams.Subscriber;
+
+/**
+ * This instrumentation is responsible for capturing the span when {@link
```
ygree, line 30: Thanks for this description

ygree (approved):
LGTM!

dd-java-agent/instrumentation/lettuce-5/src/main/java/datadog/trace/instrumentation/lettuce5/rx/RedisSubscriptionSubscribeAdvice.java
```diff
@@ -34,8 +34,11 @@ public static State beforeSubscribe(
 
     AgentScope parentScope = null;
     RedisSubscriptionState state =
-        InstrumentationContext.get(Subscription.class, RedisSubscriptionState.class)
-            .get(subscription);
+        (RedisSubscriptionState)
+            InstrumentationContext.get(
+                    "io.lettuce.core.RedisPublisher$RedisSubscription",
+                    "datadog.trace.instrumentation.lettuce5.rx.RedisSubscriptionState")
```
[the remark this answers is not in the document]
amarziali: The class is (package) private so I cannot reference directly using class literals

dd-java-agent/instrumentation/mule-4/src/test/groovy/mule4/MuleForkedTest.groovy
```diff
@@ -81,7 +80,6 @@ class MuleForkedTest extends WithHttpServer<MuleTestContainer> {
     container.stop()
   }
 
-  @Flaky("SocketTimeoutException / https://github.com/DataDog/dd-trace-java/issues/4690")
```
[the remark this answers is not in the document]
amarziali: Yes I'll close as soon as this PR will be merged thanks!

dd-java-agent/instrumentation/lettuce-5/src/main/java/datadog/trace/instrumentation/lettuce5/rx/RedisSubscriptionSubscribeAdvice.java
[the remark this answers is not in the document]
ygree: I got it! I didn't know it was either/or.

dd-java-agent/instrumentation/reactive-streams/src/main/java/datadog/trace/instrumentation/reactivestreams/ReactiveStreamsAsyncResultSupportExtension.java
```diff
@@ -23,13 +23,15 @@ public static void initialize() {}
...
+    return ret;
   }
 
   @Override
   public Object apply(Object result, AgentSpan span) {
-    if (result instanceof Publisher) {
-      return new WrappedPublisher<>((Publisher) result, span);
+    if (result != null) {
+      // the span will be closed then the subscriber span will finish.
+      return new WrappedPublisher<>((Publisher<?>) result, span);
```
richardstartin, line 34: Why is this cast now safe without checking?

dd-java-agent/instrumentation/reactive-streams/src/main/java/datadog/trace/instrumentation/reactivestreams/ReactiveStreamsAsyncResultSupportExtension.java
```diff
@@ -60,7 +62,7 @@ public WrappedSubscriber(Subscriber<T> delegate, AgentSpan span) {
 
     @Override
     public void onSubscribe(Subscription s) {
-      this.delegate.onSubscribe(s);
+      this.delegate.onSubscribe(new WrappedSubscription(s, this.span));
```
richardstartin, line 65: How do we know `s` is not a `WrappedSubscription`?

dd-java-agent/instrumentation/reactive-streams/src/main/java/datadog/trace/instrumentation/reactivestreams/SubscriberInstrumentation.java
```diff
@@ -0,0 +1,106 @@
...
+   * Propagates the span captured onSubscribe when the onComplete method is called. We do not let to
+   * propagate the active span if different from the span captured on subscribe because we need to
+   * ensure that late subscriptions that kicks onComplete have the right context.
+   */
+  public static class SubscriberCompleteAdvice {
+    @Advice.OnMethodEnter(suppress = Throwable.class)
+    public static AgentScope before(@Advice.This final Subscriber self) {
+      final AgentSpan span =
+          InstrumentationContext.get(Subscriber.class, AgentSpan.class).get(self);
+      return span == null || span == activeSpan() ? null : activateSpan(span);
```
richardstartin, line 96: We already have reference counting at the scope level to guard against creating a new scope for the active span, so this check will just double the number of scope stack accesses we need to do.

dd-java-agent/instrumentation/reactor-core-3.1/src/test/groovy/SubscriptionTest.groovy
```diff
@@ -1,28 +1,92 @@
...
-import static datadog.trace.agent.test.utils.TraceUtils.runUnderTrace
-
 class SubscriptionTest extends AgentTestRunner {
 
-  def "subscription test"() {
+  @Override
+  boolean useStrictTraceWrites() {
+    // there are continuations we are not handling today on reactor internals.
+    false
+  }
```
richardstartin, line 25: Do we intend to track these down or are the cases impossible to handle?

dd-java-agent/instrumentation/trace-annotation/src/main/java/datadog/trace/instrumentation/trace_annotation/DoNotTraceAdvice.java
```diff
@@ -17,10 +18,11 @@ public static AgentScope before() {
   @Advice.OnMethodExit(suppress = Throwable.class, onThrowable = Throwable.class)
   public static void after(
       @Advice.Enter final AgentScope scope,
+      @Advice.Origin final MethodType methodType,
```
richardstartin, line 21: I know the `Method` origin is expensive, it looks like this works in a similar way. Do we know how often this will be called?

dd-java-agent/instrumentation/reactive-streams/src/main/java/datadog/trace/instrumentation/reactivestreams/ReactiveStreamsAsyncResultSupportExtension.java
amarziali: If we have two nested methods we should wrap twice the subscription. Subscription mechanism is mostly recursive so I don't think we can skip wrapping if it's already wrapped because there are use cases (I added a test for it) that should support it

dd-java-agent/instrumentation/reactive-streams/src/main/java/datadog/trace/instrumentation/reactivestreams/SubscriberInstrumentation.java
amarziali: ok thanks I will remove this kind of checks in other similar places so.

dd-java-agent/instrumentation/reactive-streams/src/main/java/datadog/trace/instrumentation/reactivestreams/ReactiveStreamsAsyncResultSupportExtension.java
amarziali: This is done by `AsyncResultDecorator` that calls `supports` before.  `supports` already do this kind of 'cast-ability' check so I removed from there

dd-java-agent/instrumentation/reactive-streams/src/main/java/datadog/trace/instrumentation/reactivestreams/SubscriberInstrumentation.java
amarziali: @richardstartin I was thinking that perhaps I cannot do it since I need to know on the onEnter if the scope was already active in order to know if I have to close or not in the onExit. In fact, if I let flow the already active scope I should not close on the onExit otherwise I would break things

dd-java-agent/instrumentation/trace-annotation/src/main/java/datadog/trace/instrumentation/trace_annotation/DoNotTraceAdvice.java
amarziali: This is called every time a method is annotated with `@Trace` (or `@DoNotTrace` in this case). Actually that method call frequency depend on the user code since it is linked to the trace annotations (traced methods in general). What would be an alternative way to obtain the same info that's cheaper and that I can use here?

dd-java-agent/instrumentation/reactive-streams/src/main/java/datadog/trace/instrumentation/reactivestreams/SubscriberInstrumentation.java
richardstartin: Can we expand the the `AgentTracer` interface to yield a new scope unless the span is already active instead?

dd-java-agent/instrumentation/trace-annotation/src/main/java/datadog/trace/instrumentation/trace_annotation/DoNotTraceAdvice.java
richardstartin: We might be able to add a field to store the method's return type, then we could get it with a field access instead. I don't think we need to block on this, but maybe we should keep this instrumentation disabled by default until we know how high the overhead is.

richardstartin (approved):
I have some performance concerns and think we can improve the tracer API to support this better, but don't want to block the PR, it's better to have something that works than the current state.

dd-java-agent/instrumentation/reactive-streams/src/main/java/datadog/trace/instrumentation/reactivestreams/SubscriberInstrumentation.java
amarziali: I was wrong. There is actually a refcount on scope activation so activating twice does not hurt. I simplified those things accordingly then thank you

dd-java-agent/instrumentation/reactor-core-3.1/src/test/groovy/SubscriptionTest.groovy
amarziali: I removed it and actually excluded context propagation on some runnable used internally by the reactor eventloop that's in my mind a better solution

Issue #7152, opened 2024-06-10
Use hikari pool name as service name


mwisnicki:
I have 2 postgresql connections and 2 redshift connections.

In order to differentiate them one can set `DD_TRACE_DB_CLIENT_SPLIT_BY_INSTANCE` but in my case I have database of the same name in postgresql and redshift which would create a conflict.

Another option is `DD_TRACE_DB_CLIENT_SPLIT_BY_HOST` but that requires more complicated service mappings and would be problematic if someone has 2 databases on the same server.

We already have to set unique pool names for Hikari so it would be great if dd could just pick them up as service names.

jordan-wong:
Should be resolved with [Add Hikari Pool Name tag #7672](https://github.com/DataDog/dd-trace-java/pull/7672), thanks for reporting!

PR #7672, opened 2024-09-23
Add Hikari Pool Name tag 

jordan-wong:
# What Does This Do
Adds instrumentation for Hikari Pool Name. If using a Hikari pool, adds a `db.pool.name` tag to span.

# Motivation
Helps with issue: https://github.com/DataDog/dd-trace-java/issues/7152

# Additional Notes

# Contributor Checklist

Jira ticket: [PROJ-IDENT]


jordan-wong:
mistake ping, messed up rebase

dd-java-agent/instrumentation/jdbc/src/main/java/datadog/trace/instrumentation/jdbc/DataSourceInstrumentation.java
```diff
@@ -61,12 +62,17 @@ public static AgentScope start(@Advice.This final DataSource ds) {
         // Don't want to generate a new top-level span
         return null;
       }
-
       final AgentSpan span = startSpan(DATABASE_CONNECTION);
       DECORATE.afterStart(span);
 
       span.setResourceName(DECORATE.spanNameForMethod(ds.getClass(), "getConnection"));
 
+      if (ds instanceof HikariDataSource) {
```
amarziali, line 70: You cannot assume that the customer has this dependency on the classpath. Making that instrumentation depending to hikari jars will make that unusable for people not having it because muzzle blocks this when not all the classes needed are available at runtime (reason why the muzzle task is failing on the CI).
I strongly suggest to do a separate instrumentation for this (i.e. `HikariDataSourceInstrumentation`) instrumenting `HikariDataSource` only. You should also have a separate `muzzleDirective` for it if you want to test muzzle correctly

dd-java-agent/instrumentation/jdbc/src/main/java/datadog/trace/instrumentation/jdbc/DataSourceInstrumentation.java
```diff
@@ -61,7 +61,6 @@ public static AgentScope start(@Advice.This final DataSource ds) {
         // Don't want to generate a new top-level span
         return null;
       }
-
```
nayeem-kamal, line 64: We can revert this line back for cleanliness

dd-java-agent/instrumentation/jdbc/src/test/groovy/RemoteJDBCInstrumentationTest.groovy
```diff
@@ -183,7 +183,7 @@ abstract class RemoteJDBCInstrumentationTest extends VersionedNamingTestBase {
     mysql.start()
     PortUtils.waitForPortToOpen(mysql.getHost(), mysql.getMappedPort(MySQLContainer.MYSQL_PORT), 5, TimeUnit.SECONDS)
     jdbcUrls.put(MYSQL, "${mysql.getJdbcUrl()}")
-    sqlserver = new MSSQLServerContainer().acceptLicense().withPassword(jdbcPasswords.get(SQLSERVER))
+    sqlserver = new MSSQLServerContainer(MSSQLServerContainer.IMAGE).acceptLicense().withPassword(jdbcPasswords.get(SQLSERVER))
```
nayeem-kamal, line 186: I think we can revert the changes to this file since it seems unrelated to your testing

dd-java-agent/instrumentation/jdbc/src/main/java/datadog/trace/instrumentation/jdbc/HikariDataSourceInstrumentation.java
```diff
@@ -0,0 +1,46 @@
...
+  public void methodAdvice(MethodTransformer transformer) {
+    transformer.applyAdvice(
+        named("getConnection"),
+        HikariDataSourceInstrumentation.class.getName() + "$HikariGetConnectionAdvice");
+  }
+
+  public static class HikariGetConnectionAdvice {
+    @Advice.OnMethodEnter(suppress = Throwable.class)
+    public static void start(@Advice.This final HikariDataSource ds) {
+      if (activeSpan() == null) {
```
amarziali, line 36: This won't work because it relies on the `DatasourceInstrumentation` that's disabled by default. The test passes because `injectSysConfig("dd.integration.jdbc-datasource.enabled", "true")`
Also there are two advices kickin in on the same method enter and here we assume an execution order.
Also, in order to do a split-by-tag on the `jdbc.query` spans you should add this tag on the statment span and not on the connection span.

An approach might be here to store that pool information on the `DBInfo` state that's stored when the connection is created. In fact `DriverInstrumentation` already link 'java.sql.Connection` to `DbInfo`. So here you can use the `InstrumentationContext` to retrieve that dbInfo value and enrich it with the hikari pool. Then, you shall ensure that that the tag is added when the statement spans are created

dd-java-agent/agent-bootstrap/src/main/java/datadog/trace/bootstrap/instrumentation/jdbc/DBInfo.java
```diff
@@ -141,6 +147,11 @@ public Builder port(Integer port) {
       return this;
     }
 
+    public Builder hikariPoolName(String poolName) {
```
amarziali, line 150: I would rather use a generic name (i.e. `poolName`) not linked to `hikari` since we might add this also for other pool implementations in the future

internal-api/src/main/java/datadog/trace/bootstrap/instrumentation/api/Tags.java
```diff
@@ -47,6 +47,7 @@ public class Tags {
   public static final String DB_HOST = "db.host";
   public static final String DB_SCHEMA = "db.schema";
   public static final String MESSAGE_BUS_DESTINATION = "message_bus.destination";
+  public static final String HIKARI_POOL_NAME = "hikari.poolname";
```
amarziali, line 50: may we use generic names like `db.pool.name`?

dd-java-agent/instrumentation/jdbc/src/main/java/datadog/trace/instrumentation/jdbc/HikariDataSourceInstrumentation.java
```diff
@@ -0,0 +1,74 @@
...
+      try {
+        if (con.isWrapperFor(Connection.class)) {
+          unwrapped = con.unwrap(Connection.class);
+        }
+      } catch (Throwable t) {
+        return;
+      }
+
+      String hikariPoolname = ds.getPoolName();
+      DBInfo dbInfo = InstrumentationContext.get(Connection.class, DBInfo.class).get(unwrapped);
```
amarziali, line 67: nit: check for null before using `get()`

amarziali (approved):
LGTM. Hikari 6 is not yet tested hence the lockfile should be bumped before merging

nayeem-kamal (approved):
LGTM

internal-api/src/main/java/datadog/trace/bootstrap/instrumentation/api/Tags.java
```diff
@@ -47,6 +47,7 @@ public class Tags {
   public static final String DB_HOST = "db.host";
   public static final String DB_SCHEMA = "db.schema";
   public static final String MESSAGE_BUS_DESTINATION = "message_bus.destination";
+  public static final String DB_POOL_NAME = "db.pool.name";
```
bouwkast, line 50: Just to check :)

The PR description is outdated I think as it states that it'd add `hikari.poolname` as the tag, but it appears to be `db.pool.name`. Might be worth to update the PR description

PR #7677, opened 2024-09-24
Add support for `TRACE_HTTP_CLIENT_TAG_QUERY_STRING` and change default value of `HTTP_CLIENT_TAG_QUERY_STRING`

mhlidd:
# What Does This Do
Changes the default value of the `HTTP_CLIENT_TAG_QUERY_STRING` to have the value of `true` instead of `false` to be consistent with other language defaults. Additionally, we add support to handle 
the `TRACE_HTTP_CLIENT_TAG_QUERY_STRING` tag as well as the original `HTTP_CLIENT_TAG_QUERY_STRING` in efforts to be more consistent with the tags from other languages as well.
# Motivation
Our goal is to make the implementation of configuration variables consistent for all languages as part of the config consistency effort listed in the following [RFC](https://docs.google.com/document/d/1kI-gTAKghfcwI7YzKhqRv2ExUstcHqADIWA4-TZ387o/edit).
# Additional Notes
In Release Notes: Mention the change of the default value and state that the tag can be manually set to false
# Contributor Checklist

Jira ticket: [PROJ-IDENT]


internal-api/src/main/java/datadog/trace/bootstrap/config/provider/ConfigProvider.java
```diff
@@ -463,6 +467,10 @@ private static Properties loadConfigurationFile(ConfigProvider configProvider) {
   public abstract static class Source {
     public final String get(String key, String... aliases) {
       String value = get(key);
+      if(key == "trace.http.client.tag.query-string"){
+        System.out.println("internal get call: ");  
+        System.out.println(value);
+      }
```
PerfectSlayer, line 473: Leftover?

internal-api/src/main/java/datadog/trace/bootstrap/config/provider/ConfigProvider.java
```diff
@@ -194,6 +194,10 @@ private <T> T get(String key, T defaultValue, Class<T> type, String... aliases)
       try {
         String sourceValue = source.get(key, aliases);
         T value = ConfigConverter.valueOf(sourceValue, type);
+        if(key == "trace.http.client.tag.query-string"){
+          System.out.println("VALUE");
+          System.out.println(value);
+        }
```
PerfectSlayer, line 200: Leftover?

internal-api/src/main/java/datadog/trace/api/Config.java
```diff
@@ -1293,9 +1293,13 @@ private Config(final ConfigProvider configProvider, final InstrumenterConfig ins
         configProvider.getBoolean(
             HTTP_SERVER_ROUTE_BASED_NAMING, DEFAULT_HTTP_SERVER_ROUTE_BASED_NAMING);
 
+    System.out.println("HTTP_CLIENT_TAG_QUERY_STRING: ");
+    System.out.println(HTTP_CLIENT_TAG_QUERY_STRING);
```
PerfectSlayer, line 1297: Leftover?

internal-api/src/main/java/datadog/trace/api/Config.java
```diff
@@ -1293,9 +1293,13 @@ private Config(final ConfigProvider configProvider, final InstrumenterConfig ins
...
     httpClientTagQueryString =
         configProvider.getBoolean(
-            HTTP_CLIENT_TAG_QUERY_STRING, DEFAULT_HTTP_CLIENT_TAG_QUERY_STRING);
+            "trace." + HTTP_CLIENT_TAG_QUERY_STRING, DEFAULT_HTTP_CLIENT_TAG_QUERY_STRING, HTTP_CLIENT_TAG_QUERY_STRING);
+    System.out.println("httpClientTagQueryString: ");
+    System.out.println(httpClientTagQueryString);
```
PerfectSlayer, line 1302: Leftover?

dd-java-agent/instrumentation/opentracing/api-0.31/src/main/java/datadog/trace/instrumentation/opentracing31/OTTracer.java
```diff
@@ -41,6 +41,7 @@ public Span activeSpan() {
 
   @Override
   public SpanBuilder buildSpan(final String operationName) {
+    System.out.println("BUILDING A SPAN");
```
PerfectSlayer, line 44: Leftover?

PerfectSlayer (commented):
Is the PR ready for review? I can see a bunch of leftovers 🤔

mhlidd:
> Is the PR ready for review? I can see a bunch of leftovers 🤔
> 
> If not, thank for using draft as mentioned in contribution guidelines 🙏

Accidentally marked as ready earlier. Code has been cleaned up and is ready for review! 🙇 
@PerfectSlayer

internal-api/src/main/java/datadog/trace/api/Config.java
```diff
@@ -1295,7 +1295,9 @@ private Config(final ConfigProvider configProvider, final InstrumenterConfig ins
 
     httpClientTagQueryString =
         configProvider.getBoolean(
-            HTTP_CLIENT_TAG_QUERY_STRING, DEFAULT_HTTP_CLIENT_TAG_QUERY_STRING);
+            "trace." + HTTP_CLIENT_TAG_QUERY_STRING,
```
PerfectSlayer, line 1298: Can you create a dedicated constant near to the legacy one, and deprecate the older one using`@Deprecated` and some Javadoc comment?

PerfectSlayer (commented):
About test changes, it feels this duplicate the `test url handling for #url` test twice without adding new behavior test. If the goal is to test the setting is well applied, it is already covered through config name alias tests.

If you still want have a test for this behavior, I would recommend creating a parametric tests with all possible user config combinations and only checking at the value the config returns.

internal-api/src/main/java/datadog/trace/api/Config.java
mhlidd: This was a conversation I had with @mcculls, who proposed that we go through with this solution as it is already used in other config scenarios. Do you think that we should change this implementation to have a dedicated constant that does this? 
`public static final String TRACE_HTTP_SERVER_TAG_QUERY_STRING = "trace.http.server.tag.query-string";`

PerfectSlayer: > Do you think that we should change this implementation to have a dedicated constant that does this?

Exactly. The main rationale is that we usually do a text search for a configuration key to find out how it is used in the code base. When using `"trace." + HTTP_CLIENT_TAG_QUERY_STRING`, we won't be able to find it 😞

dd-java-agent/agent-bootstrap/src/test/groovy/datadog/trace/bootstrap/instrumentation/decorator/HttpClientDecoratorTest.groovy
```diff
@@ -98,6 +100,110 @@ class HttpClientDecoratorTest extends ClientDecoratorTest {
     req = [url: url == null ? null : new URI(url)]
   }
 
+  def "test url handling for #url : trace env var"() {
```
mtoffl01, line 103: Rather than differentiating these tests with "default env var" and "trace env var," I'd suggest actually using the env var names (Because "default" didn't mean anything to me until I read "trace"): DD_HTTP_CLIENT_TAG_QUERY_STRING and DD_TRACE_HTTP_CLIENT_TAG_QUERY_STRING.

Not sure if this conflicts with Java standards, but just my two cents from a generic code reader

PerfectSlayer: Spock test methods does not have to respect Java standards, so we're good here.

But if we really want to test the behavior with all the different config values (I started a discussion with @mhlidd already as it feels more like testing config rather than testing the HTTP client decorator), I would recommend creating test child classes, having the config setting / retrieval in the parent abstract method. -- and here, the class name will have to respect the standards 😉

dd-trace-api/src/main/java/datadog/trace/api/config/TraceInstrumentationConfig.java
```diff
@@ -50,6 +50,8 @@ public final class TraceInstrumentationConfig {
       "http.server.decoded.resource.preserve-spaces";
   public static final String HTTP_SERVER_ROUTE_BASED_NAMING = "http.server.route-based-naming";
   public static final String HTTP_CLIENT_TAG_QUERY_STRING = "http.client.tag.query-string";
```
PerfectSlayer, line 52: Can you `@Deprecated` the previous constant and document the one to use instead?

PerfectSlayer (commented):
If the new default behavior is to have `TRACE_HTTP_CLIENT_TAG_QUERY_STRING` enabled, you would have to update all the tests that break with this new behavior rather than disabling it with `injectSysConfig(TRACE_HTTP_CLIENT_TAG_QUERY_STRING, "false")` -- I'm referring to the integration test updates.

dd-java-agent/instrumentation/aws-java-sns-1.0/src/test/groovy/SnsClientTest.groovy
```diff
@@ -147,7 +148,7 @@ abstract class SnsClientTest extends VersionedNamingTestBase {
           tags {
             "$Tags.COMPONENT" "java-aws-sdk"
             "$Tags.SPAN_KIND" Tags.SPAN_KIND_CLIENT
-            "$Tags.HTTP_URL" endPoint+'/'
+            //            "$Tags.HTTP_URL" endPoint+'/'
```
PerfectSlayer, line 151: Leftover?

dd-java-agent/instrumentation/aws-common/src/main/java/datadog/trace/instrumentation/aws/ExpectedQueryParams.java
```diff
@@ -0,0 +1,60 @@
+package datadog.trace.instrumentation.aws;
+
+import java.util.Arrays;
+import java.util.List;
+
+public class ExpectedQueryParams {
```
PerfectSlayer, line 6: You can set the class as `final` and add a private constructor (`private ExpectedQueryParams() {}`).
This usually denotes _utility_ classes, letting developers know this class only holds utility methods, and should not be instantiated.

dd-java-agent/instrumentation/aws-java-sns-2.0/src/test/groovy/SnsClientTest.groovy
```diff
@@ -137,7 +136,7 @@ abstract class SnsClientTest extends VersionedNamingTestBase {
           tags {
             "$Tags.COMPONENT" "java-aws-sdk"
             "$Tags.SPAN_KIND" Tags.SPAN_KIND_CLIENT
-            "$Tags.HTTP_URL" endPoint+'/'
+            //            "$Tags.HTTP_URL" endPoint+'/'
```
PerfectSlayer, line 139: Leftover?

dd-java-agent/instrumentation/aws-java-sdk-2.2/build.gradle
```diff
@@ -23,6 +23,7 @@ def fixedSdkVersion = '2.20.33' // 2.20.34 is missing and breaks IDEA import
 
 dependencies {
   compileOnly group: 'software.amazon.awssdk', name: 'aws-core', version: '2.2.0'
+  implementation project(':dd-java-agent:instrumentation:aws-common')
```
PerfectSlayer, line 26: This should be `testImplementation` rather than `implementation`.
`implementation` adds dependency for both the instrumentation code and the test classes, while `testImplementation` only adds dependency for the test classes.

dd-java-agent/instrumentation/aws-java-sns-1.0/build.gradle
```diff
@@ -14,6 +14,7 @@ addTestSuiteExtendingForDir('latestDepForkedTest', 'latestDepTest', 'test')
 
 dependencies {
   compileOnly group: 'com.amazonaws', name: 'aws-java-sdk-sns', version: '1.12.710'
+  implementation project(':dd-java-agent:instrumentation:aws-common')
```
PerfectSlayer, line 17: Same as the an upper comment: this should be `testImplementation` rather than `implementation`

dd-java-agent/instrumentation/aws-java-sns-2.0/build.gradle
```diff
@@ -14,6 +14,7 @@ addTestSuiteExtendingForDir('latestDepForkedTest', 'latestDepTest', 'test')
 
 dependencies {
   compileOnly group: 'software.amazon.awssdk', name: 'sns', version: '2.25.40'
+  implementation project(':dd-java-agent:instrumentation:aws-common')
```
PerfectSlayer, line 17: Same as the an upper comment: this should be `testImplementation` rather than `implementation`

dd-java-agent/instrumentation/aws-java-sqs-2.0/build.gradle
```diff
@@ -19,6 +19,7 @@ addTestSuiteExtendingForDir('latestDepForkedTest', 'latestDepTest', 'test')
 dependencies {
   compileOnly group: 'software.amazon.awssdk', name: 'sqs', version: '2.2.0'
   compileOnly group: 'com.amazonaws', name: 'amazon-sqs-java-messaging-lib', version: '2.0.0'
+  implementation project(':dd-java-agent:instrumentation:aws-common')
```
PerfectSlayer, line 22: Same as the an upper comment: this should be `testImplementation` rather than `implementation`

dd-java-agent/instrumentation/aws-java-sqs-2.0/src/test/groovy/SqsClientTest.groovy
```diff
@@ -286,6 +288,7 @@ abstract class SqsClientTest extends VersionedNamingTestBase {
   @IgnoreIf({instance.isDataStreamsEnabled()})
   def "trace details propagated from SQS to JMS"() {
     setup:
+    injectSysConfig(TRACE_HTTP_CLIENT_TAG_QUERY_STRING, "false")
```
PerfectSlayer, line 291: Is it disabled on purpose for this test case?

dd-java-agent/instrumentation/mule-4/build.gradle
```diff
@@ -72,6 +72,7 @@ tasks.named("compileTestGroovy").configure {
 
 dependencies {
   compileOnly group: 'org.mule.runtime', name: 'mule-core', version: muleVersion
+  implementation project(':dd-java-agent:instrumentation:aws-common')
```
PerfectSlayer, line 75: Same as the an upper comment: this should be `testImplementation` rather than `implementation`

dd-java-agent/instrumentation/spring-messaging-4/build.gradle
```diff
@@ -22,6 +22,7 @@ addTestSuiteForDir('latestDepTest', 'test')
 
 dependencies {
   compileOnly group: 'org.springframework', name: 'spring-messaging', version: '4.0.0.RELEASE'
+  implementation project(':dd-java-agent:instrumentation:aws-common')
```
PerfectSlayer, line 25: Same as the an upper comment: this should be `testImplementation` rather than `implementation`

dd-java-agent/testing/src/main/groovy/datadog/trace/agent/test/asserts/TagsAssert.groovy
```diff
@@ -131,6 +131,41 @@ class TagsAssert {
...
+            spanQueryParams.add(pair)
+          }
+        }
+        for(String param : queryParams){
+          if (!spanQueryParams.contains(param)){
+            return false
+          }
+        }
+      }
+      return true
```
PerfectSlayer, line 165: This will return `true` if the `http.query.string` tag is not set (`null` or empty) even if the test assert as `queryParams` set and non empty. Is it expected?

For example: `urlTags("http://localhost/", Arrays.asList("param1", "param2")` will be valid for the following tags: `http.query.string` = `http://localhost/` and `http.query.string` = null (or `""`)

dd-java-agent/instrumentation/aws-common/src/main/java/datadog/trace/instrumentation/aws/ExpectedQueryParams.java
PerfectSlayer: In addition, as it is a test class, it would be better to put it under `src/test/java` rather than `src/main/java`.

PerfectSlayer (commented):
That's great improvements!  🙌
I left a bunch of comments as there are some changes needed but the hardest part is done 😉

dd-java-agent/testing/src/main/groovy/datadog/trace/agent/test/asserts/TagsAssert.groovy
mhlidd: > `urlTags("http://localhost/", Arrays.asList("param1", "param2")` will be valid for the following tags: `http.query.string` = `http://localhost/` and `http.query.string` = null (or `""`)

I'm not sure what you mean by `http.query.string` = `http://localhost/`, because the `http.query.string` portion would only check the query params. We go through the array of `["param1", "param2"]` and because they don't exist as part of `"http://localhost/"`,  we will return `false` here.

dd-java-agent/instrumentation/aws-java-sqs-2.0/src/test/groovy/SqsClientTest.groovy
mhlidd: Fixed

dd-java-agent/instrumentation/aws-common/src/main/java/datadog/trace/instrumentation/aws/ExpectedQueryParams.java
PerfectSlayer: I got it wrong, it is well `src/main/java` if you would like to be able to reuse it in some other modules 🙇

PerfectSlayer (approved):
Looks good to me! Thanks for keep improving it PR 👍

PR #7719, opened 2024-10-03
Fix exploration tests

jpbempel:
# What Does This Do
update the registry for pushing the image
install jdk 21
use the jsoup library and patch it for running test with instrument-the-world mode
add include file to specify which packages/classes we want to include into the instrumentation instead of instrumenting everything (like build tools)
do not collect static inherited fields to avoid duplicate class definition for enums.

# Motivation

# Additional Notes

# Contributor Checklist

Jira ticket: [PROJ-IDENT]


dd-java-agent/agent-debugger/exploration-tests/Dockerfile.exploration-tests
```diff
@@ -16,25 +17,36 @@ RUN bash -c "source $HOME/.sdkman/bin/sdkman-init.sh && \
     yes | sdk install java $JAVA_8_VERSION && \
     yes | sdk install java $JAVA_11_VERSION && \
     yes | sdk install java $JAVA_17_VERSION && \
```
evanchooly, line 19: if you commit a `.sdkmanrc` file this all can be one line.

jpbempel: ok thanks will improve that later

dd-java-agent/agent-debugger/src/main/java/com/datadog/debugger/agent/DebuggerTransformer.java
```diff
@@ -277,15 +304,32 @@ private byte[] transformTheWorld(
...
+          log.debug("Skipping method: {}", fqnMethod);
+          continue;
+        }
         if (methodNames.add(methodNode.name)) {
           LogProbe probe =
               LogProbe.builder()
                   .probeId(UUID.randomUUID().toString(), 0)
                   .where(classNode.name, methodNode.name)
-                  .captureSnapshot(true)
+                  .captureSnapshot(false)
```
shatzi, line 332: isn't that reduce the power of the exploration test? as capture Snapshot add logic that can break.

dd-java-agent/agent-debugger/src/main/java/com/datadog/debugger/agent/DebuggerTransformer.java
```diff
@@ -328,6 +374,16 @@ private boolean isExcludedFromTransformation(String classFilePath) {
     return false;
   }
 
+  private boolean isIncludedForTransformation(String classFilePath) {
+    if (includeClasses.contains(classFilePath)) {
```
shatzi, line 378: includeClasses can be null... how this works?

dd-java-agent/agent-debugger/src/main/java/com/datadog/debugger/agent/DebuggerTransformer.java
```diff
@@ -328,6 +374,16 @@ private boolean isExcludedFromTransformation(String classFilePath) {
     return false;
   }
 
+  private boolean isIncludedForTransformation(String classFilePath) {
```
shatzi, line 377: you don't check includeMethods? not sure how this works...

dd-java-agent/agent-debugger/src/main/java/com/datadog/debugger/agent/DebuggerTransformer.java
```diff
@@ -328,6 +374,16 @@ private boolean isExcludedFromTransformation(String classFilePath) {
...
+    if (includeClasses.contains(classFilePath)) {
+      return true;
+    }
+    if (includeTrie.hasMatchingPrefix(classFilePath)) {
```
shatzi, line 381: can be null right?

dd-java-agent/agent-debugger/src/main/java/com/datadog/debugger/instrumentation/CapturedContextInstrumentor.java
```diff
@@ -1167,7 +1170,12 @@ private static List<FieldNode> extractStaticFields(
         }
       }
     }
-    addInheritedStaticFields(classNode, classLoader, limits, results, fieldCount);
+    if (!Config.get().isDebuggerInstrumentTheWorld()) {
+      // Collects inherited static fields only if the ITW mode is not enabled
+      // because it can lead to LinkageError: attempted duplicate class definition
+      // for example, when a probe is located in method overridden in enum element
```
shatzi, line 1176: can we just avoid and mark those probes as "blocked'? 
as this might be something the user can do as well (very rare agree)... but it can crash the application

dd-java-agent/agent-debugger/src/main/java/com/datadog/debugger/instrumentation/CapturedContextInstrumentor.java
```diff
@@ -1185,17 +1193,22 @@ private static void addInheritedStaticFields(
...
+            results.add(fieldNode);
+            log.debug("Adding static inherited field {}", fieldNode.name);
+            fieldCount++;
+            if (fieldCount > limits.maxFieldCount) {
+              return;
+            }
           }
         }
+      } catch (ClassCircularityError ex) {
+        break;
```
shatzi, line 1211: is that recoverable? or the process has crashed and class failed to load because of us?
not sure we should just break, maybe we should report it to telemetry?

dd-java-agent/agent-debugger/src/main/java/com/datadog/debugger/agent/DebuggerTransformer.java
```diff
@@ -277,15 +304,32 @@ private byte[] transformTheWorld(
           loader,
           location);
       ClassNode classNode = parseClassFile(classFilePath, classfileBuffer);
+      if (classNode.superName.equals("java/lang/ClassLoader")
```
shatzi, line 307: maybe move this into the isExcludedFromTransformation method?

dd-java-agent/agent-debugger/src/main/java/com/datadog/debugger/agent/DebuggerTransformer.java
```diff
@@ -277,15 +304,32 @@ private byte[] transformTheWorld(
...
+      }
       List<ProbeDefinition> probes = new ArrayList<>();
       Set<String> methodNames = new HashSet<>();
       for (MethodNode methodNode : classNode.methods) {
+        if (methodNode.name.equals("<clinit>")) {
+          // skip static class initializer
+          continue;
+        }
+        String fqnMethod = classNode.name + "::" + methodNode.name;
+        if (excludeMethods.contains(fqnMethod)) {
```
shatzi, line 323: first, this can be null right?

second, you don't check includedMethods here? 

third, maybe refactor it a method like isMethodIncludedForTransformation

fourth, maybe we should refactor both include and exclude logs into one - shouldTransformMethod()

shatzi (commented):
added few comments - I might miss some context here.

dd-java-agent/agent-debugger/src/main/java/com/datadog/debugger/agent/DebuggerTransformer.java
jpbempel: no, the instrumentation is the same (except try/catch block but we don't use them anymore) but we don't serialize so it helps to perform faster the test

dd-java-agent/agent-debugger/src/main/java/com/datadog/debugger/agent/DebuggerTransformer.java
jpbempel: right should handle that, but basically starting an exploration test without an include is very bad idea. 
In practice we specify the include that is the package of the tests we want to instrument. no need to instrument the build system, that's wasteful in time

dd-java-agent/agent-debugger/src/main/java/com/datadog/debugger/instrumentation/CapturedContextInstrumentor.java
jpbempel: I am not sure I can detect this in advance, very tricky

dd-java-agent/agent-debugger/src/main/java/com/datadog/debugger/instrumentation/CapturedContextInstrumentor.java
jpbempel: yes we don't capture the problematic field in that case. not sure it is worth a telemetry but I should maybe report it as not captured

dd-java-agent/agent-debugger/src/main/java/com/datadog/debugger/agent/DebuggerTransformer.java
jpbempel: correction, no they cannot be null in the case of InstrumentTheWorld mode

dd-java-agent/agent-debugger/src/main/java/com/datadog/debugger/agent/DebuggerTransformer.java
jpbempel: no they cannot be null in the case of InstrumentTheWorld mode

dd-java-agent/agent-debugger/src/main/java/com/datadog/debugger/agent/DebuggerTransformer.java
shatzi: Doesn't this break the case for running normally? They are set to null.

I guess I'm missing something

shatzi (approved):

dd-java-agent/agent-debugger/src/main/java/com/datadog/debugger/agent/DebuggerTransformer.java
jpbempel: Nope because I need to parse the classfile before performing the check.
isExcludedFromTransformation is purely based on the class name which more efficient than parsing every class before checking if we exclude it or not

dd-java-agent/agent-debugger/src/main/java/com/datadog/debugger/agent/DebuggerTransformer.java
jpbempel: done

PR #7741, opened 2024-10-07
Add experimental taint propagation to the String replace, replaceFirst, replaceAll methods

Mariovido:
# What Does This Do
This adds the instrumentation to propagate the taint values through the methods of `String` called `replace`, `replaceFirst`, `replaceAll`.
The approach is using a `CallSite.Around` to avoid iterating all the string to check the tainted values. This will add an overhead to the method as the algorithm is not as efficient as the original, but will increase the efficiency in as it will not iterate a second time through the string.

# Motivation
Increase propagation of `String` methods

# Additional Notes

# Contributor Checklist

Jira ticket: [APPSEC-5763]

[APPSEC-5763]: https://datadoghq.atlassian.net/browse/APPSEC-5763?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ


dd-java-agent/instrumentation/java-lang/src/main/java/datadog/trace/instrumentation/java/lang/StringCallSite.java
```diff
@@ -281,4 +282,74 @@ public static String[] afterSplitWithLimit(
...
+  public static String afterReplaceCharSeq(
+      @CallSite.This final String self,
+      @CallSite.Argument(0) final CharSequence oldCharSeq,
+      @CallSite.Argument(1) final CharSequence newCharSeq) {
+    final StringModule module = InstrumentationBridge.STRING;
+    if (module != null) {
+      try {
+        return module.onStringReplace(self, oldCharSeq, newCharSeq);
+      } catch (final Throwable e) {
+        module.onUnexpectedException("afterReplaceCharSeq threw", e);
```
jandro996, line 314:
```suggestion
        module.onUnexpectedException("aroundReplaceCharSeq threw", e);
```

dd-java-agent/instrumentation/java-lang/src/main/java/datadog/trace/instrumentation/java/lang/StringCallSite.java
```diff
@@ -281,4 +282,74 @@ public static String[] afterSplitWithLimit(
...
+      @CallSite.This @Nonnull final String self,
+      @CallSite.Argument(0) final char oldChar,
+      @CallSite.Argument(1) final char newChar,
+      @CallSite.Return @Nonnull final String result) {
+    final StringModule module = InstrumentationBridge.STRING;
+    if (module != null) {
+      try {
+        module.onStringReplace(self, oldChar, newChar, result);
+      } catch (final Throwable e) {
+        module.onUnexpectedException("afterReplaceChar threw", e);
```
jandro996, line 297:
```suggestion
        module.onUnexpectedException("aroundReplaceChar threw", e);
```

dd-java-agent/instrumentation/java-lang/src/test/groovy/datadog/trace/instrumentation/java/lang/StringCallSiteTest.groovy
jandro996, line 1: It will be nice to also test for the arounds that if the module throws an exception the original method it's executed

dd-java-agent/agent-iast/src/test/groovy/com/datadog/iast/taint/RangesTest.groovy
```diff
@@ -356,6 +356,55 @@ class RangesTest extends DDSpecification {
...
 
+  void 'test splitRanges method'() {
+    when:
+    final result = Ranges.splitRanges(start, end, newLength, range as Range, offset, diffLength)
+
+    then:
+    final expectedArray = expected as Range[]
+    result == expectedArray
+
+    where:
```
jandro996, line 367: IMHO it's better to use Range range(final int start, final int length), instead of Range rangeWithSource(final byte source, final String name = 'name', final String value = 'value') as you don't need the source an the test will look cleaner

dd-java-agent/agent-iast/src/main/java/com/datadog/iast/taint/Ranges.java
```diff
@@ -409,12 +409,11 @@ private static int updateRangesWithIndentation(
         if (range.getStart() + range.getLength() > end) {
           newLength -= lineOffset;
         }
-        newRanges[i] = new Range(newStart, newLength, range.getSource(), range.getMarks());
+        newRanges[i] = copyWithPosition(range, newStart, newLength);
```
jandro996, line 412: Good catch

dd-java-agent/agent-iast/src/main/java/com/datadog/iast/propagation/StringModuleImpl.java
```diff
@@ -631,6 +633,110 @@ public void onIndent(@Nonnull String self, int indentation, @Nonnull String resu
...
+    if (rangesSelf.length == 0) {
+      return;
+    }
+
+    taintedObjects.taint(result, rangesSelf);
+  }
+
+  @Override
+  @SuppressFBWarnings("ES_COMPARING_PARAMETER_STRING_WITH_EQ")
+  public String onStringReplace(
```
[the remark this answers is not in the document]
Mariovido: Added

dd-java-agent/agent-iast/src/test/groovy/com/datadog/iast/taint/RangesTest.groovy
Mariovido: I have changed it. It's true that it is not really relevant doing it this way. It was done this way because in the first iteration all the logic was in this tests, and I forgot to remove it.

dd-java-agent/instrumentation/java-lang/src/main/java/datadog/trace/instrumentation/java/lang/StringCallSite.java
Mariovido: Changed

dd-java-agent/instrumentation/java-lang/src/main/java/datadog/trace/instrumentation/java/lang/StringCallSite.java
```diff
@@ -281,4 +282,74 @@ public static String[] afterSplitWithLimit(
...
+      }
+    }
+    return result;
+  }
+
+  @CallSite.Around(
+      "java.lang.String java.lang.String.replace(java.lang.CharSequence, java.lang.CharSequence)")
+  public static String afterReplaceCharSeq(
```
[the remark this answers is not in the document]
Mariovido: Changed

dd-java-agent/instrumentation/java-lang/src/test/groovy/datadog/trace/instrumentation/java/lang/StringCallSiteTest.groovy
Mariovido: I added two test where we test if the exception is thrown

dd-java-agent/agent-iast/src/main/java/com/datadog/iast/util/StringUtils.java
```diff
@@ -48,4 +54,171 @@ public static int leadingWhitespaces(@Nonnull final String value, int start, int
...
+   * @param taintedObjects the ctx object to save the range of the tainted values
+   * @param target the string to be replaced
+   * @param pattern the pattern to be replaced
+   * @param replacement the replacement string
+   * @param ranges the ranges of the string to be replaced
+   * @param rangesInput the ranges of the input string
+   * @param numOfReplacements the number of replacements to be made
+   */
+  @Nonnull
+  public static String replaceAndTaint(
```
jandro996, line 70: It would be nice if you could invest some time in reduce the cognitive and cyclomatic complexity :sweatsmile:
For instance extract code to function could help in terms of readability and it's a quick win.

Mariovido: I have reduced the complexity by reusing a function of the RangeBuilder which adds the ranges with an offset. Apart from that, it's hard to reduce more the complexity as I am replacing the `replace` method and at the same time tainting the values.

dd-java-agent/instrumentation/java-lang/src/main/java/datadog/trace/instrumentation/java/lang/StringCallSite.java
```diff
@@ -281,4 +282,74 @@ public static String[] afterSplitWithLimit(
...
+      @CallSite.Argument(1) final CharSequence newCharSeq) {
+    final StringModule module = InstrumentationBridge.STRING;
+    if (module != null) {
+      try {
+        return module.onStringReplace(self, oldCharSeq, newCharSeq);
+      } catch (final Throwable e) {
+        module.onUnexpectedException("aroundReplaceCharSeq threw", e);
+      }
+    }
+    return self.replace(oldCharSeq, newCharSeq);
```
manuel-alvarez-alvarez, line 317: We don't want to be in the stack if the original function fails:

```suggestion
    try {
        return self.replace(oldCharSeq, newCharSeq);
    } catch (Throwable e) {
        throw StackUtils.filterDatadog(e);
    }     
```

manuel-alvarez-alvarez:
Rolling out this feature like this really makes me feel nervous, I think we should:
* Always forward the call to the JDK for the replaced methods (those annotated with `@Around`) and them compare their result with the one we've computed, if they don't match:
  * Throw away the computed tainted value
  *  Return the JDK computed method
  * Generate a telemetry log with all the info we can get (without causing PI issues)
* Feature gate those instrumentations so customers can disable them in case of issues.

Once we are more confident with the solution we could return it to the most optimal state without having to rely on the JDK.

CC @smola

dd-java-agent/agent-iast/src/test/groovy/com/datadog/iast/propagation/StringModuleTest.groovy
```diff
@@ -1081,19 +1081,170 @@ class StringModuleTest extends IastModuleImplTestBase {
...
+    0           | "==>123\r\n<====>123<=="         | "==>123\n<====>123<=="
+    0           | "==>123<==\r==>123<=="           | "==>123<==\n==>123<=="
+    0           | "==>123\r<====>123<=="           | "==>123\n<====>123<=="
+    -4          | "    ==>123<==\n    12==>3<=="   | "==>123<==\n12==>3<=="
+    -4          | "    ==>123<==\r\n    12==>3<==" | "==>123<==\n12==>3<=="
+    -4          | "    ==>123\n    1<==2==>3<=="   | "==>123\n1<==2==>3<=="
+    -4          | "    ==>123\r\n    1<==2==>3<==" | "==>123\n1<==2==>3<=="
+  }
+
+  void 'test replace with a single char and make sure IastRequestContext is called'() {
```
manuel-alvarez-alvarez, line 1099: For the replacing methods did you have look to [Regex.java](https://github.com/openjdk/jdk/blob/master/test/jdk/java/lang/String/Regex.java) and [LiteralReplace.java](https://github.com/openjdk/jdk/blob/master/test/jdk/java/lang/String/LiteralReplace.java), there are many cases including ones that produce out of memory errors that we should contemplate for this PR.

dd-java-agent/agent-iast/src/main/java/com/datadog/iast/util/StringUtils.java
```diff
@@ -48,4 +54,149 @@ public static int leadingWhitespaces(@Nonnull final String value, int start, int
...
+      int newRangesSize = ranges.length * 2;
+      if (rangesInput != null) {
+        newRangesSize += rangesInput.length;
+      }
+      RangeBuilder newRanges = new RangeBuilder(newRangesSize);
+
+      int firstRange = 0;
+      int newLength = replacement.length();
+
+      StringBuffer sb = new StringBuffer();
```
manuel-alvarez-alvarez, line 95: Shouldn't we use `StringBuilder`?, `StringBuffer` comes with thread safety that we don't really need here,

dd-java-agent/agent-iast/src/main/java/com/datadog/iast/taint/Ranges.java
```diff
@@ -426,4 +425,41 @@ private static int updateRangesWithIndentation(
...
+    start += offset;
+    end += offset;
+    int rangeStart = range.getStart() + offset;
+    int rangeEnd = rangeStart + range.getLength() + diffLength;
+
+    int firstLength = start - rangeStart;
+    int secondLength = range.getLength() - firstLength - newLength + diffLength;
+    if (rangeStart > end || rangeEnd <= start) {
+      if (firstLength <= 0) {
+        return new Range[0];
```
manuel-alvarez-alvarez, line 454:
```suggestion
        return Ranges.EMPTY;
```

dd-java-agent/agent-iast/src/main/java/com/datadog/iast/util/StringUtils.java
```diff
@@ -48,4 +54,149 @@ public static int leadingWhitespaces(@Nonnull final String value, int start, int
...
+    Matcher matcher = pattern.matcher(target);
+    boolean result = matcher.find();
+    if (result) {
+      int offset = 0;
+
+      int newRangesSize = ranges.length * 2;
+      if (rangesInput != null) {
+        newRangesSize += rangesInput.length;
+      }
+      RangeBuilder newRanges = new RangeBuilder(newRangesSize);
```
manuel-alvarez-alvarez, line 90: This is probably not what you want, `new RangeBuilder(newRangesSize);` means that the builder will accept at most `newRangesSize` ranges. The builder is limited by default to at most `dd.iast.max-range-count` that is 10 by default. For this particular case is better to use the empty constructor.

dd-java-agent/instrumentation/java-lang/src/main/java/datadog/trace/instrumentation/java/lang/StringCallSite.java
Mariovido: Done!

dd-java-agent/agent-iast/src/main/java/com/datadog/iast/util/StringUtils.java
```diff
@@ -48,4 +54,149 @@ public static int leadingWhitespaces(@Nonnull final String value, int start, int
...
+          Range range = ranges[firstRange];
+          int rangeStart = range.getStart();
+          int rangeEnd = rangeStart + range.getLength();
+          // If the replaced value is between one range
+          if (rangeStart <= start && rangeEnd >= end) {
+            Range[] splittedRanges =
+                Ranges.splitRanges(start, end, newLength, range, offset, diffLength);
+
+            if (splittedRanges.length > 0 && splittedRanges[0].getLength() > 0) {
+              newRanges.add(splittedRanges[0]);
```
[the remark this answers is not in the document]
Mariovido: Added a variable that checks the return boolean to break the loop and stops generating ranges.

dd-java-agent/agent-iast/src/main/java/com/datadog/iast/util/StringUtils.java
Mariovido: The reason is that methods inside the Matcher that accepts `StringBuilder` are since Java 9

dd-java-agent/agent-iast/src/main/java/com/datadog/iast/util/StringUtils.java
Mariovido: Changed to empty constructor

dd-java-agent/instrumentation/java-lang/src/test/groovy/datadog/trace/instrumentation/java/lang/StringCallSiteTest.groovy
jandro996, line 1: As the around instrumentation it's "dangerous" it will be nice to test that with IAST_EXPERIMENTAL_PROPAGATION_ENABLED = false (default) we don't do anything or at least that we don't call the propagation module

dd-java-agent/instrumentation/java-lang/src/main/java/datadog/trace/instrumentation/java/lang/StringCallSite.java
jandro996, line 1: Just a thought... 😅  What do you think about wrap every thing on the `if(Config.get().isIastExperimentalPropagationEnabled()) ` clauses into a try catch instead of only the call to the module to prevent unexpected behaviour?
cc @manuel-alvarez-alvarez

dd-java-agent/agent-iast/src/main/java/com/datadog/iast/util/RangeBuilder.java
```diff
@@ -47,11 +47,15 @@ public RangeBuilder(final int maxSize, final int arrayChunkSize) {
   }
 
   public boolean add(final Range range) {
+    return add(range, 0);
+  }
+
+  public boolean add(final Range range, final int offset) {
```
manuel-alvarez-alvarez, line 53: Since you added this operation I think it will make sense to add
```java
    if (ranges.length == 1) {
      return add(ranges[0], offset);
    }
```

when adding ranges with an array of just one element (it will improve performance)

dd-java-agent/instrumentation/java-lang/src/main/java/datadog/trace/instrumentation/java/lang/StringCallSite.java
manuel-alvarez-alvarez: This one was solved by adding a new experimental call site that won't be used if the flag is disabled.

dd-java-agent/agent-iast/src/main/java/com/datadog/iast/util/RangeBuilder.java
Mariovido: Modified!

manuel-alvarez-alvarez (approved):

PR #7749, opened 2024-10-10
Improve data streams performance

piochelepiotr:
# What Does This Do

We want to enable collection of Data Streams stats by default. To do that, we want to make sure that performance overhead is minimal.
This PR looks at all the biggest culprits for overhead (by using Datadog profiling) and removes as much overhead as possible.

With this PR, compared to APM overhead, DSM overhead will be small.

# Motivation

image
Setting schema name & schema type are taking a lot of CPU

Most of the DSM overhead is caused by the MspcBlockingConsumerArrayQueue:
image
Also, some time is spent setting the pathway hash tag on spans, which we don't use:
image

# Additional Notes

# Contributor Checklist

Jira ticket: [PROJ-IDENT]


dd-trace-core/src/main/java/datadog/trace/core/datastreams/DefaultDataStreamsMonitoring.java
```diff
@@ -318,7 +313,11 @@ public void run() {
       Thread currentThread = Thread.currentThread();
       while (!currentThread.isInterrupted()) {
         try {
-          InboxItem payload = inbox.take();
+          InboxItem payload = inbox.poll();
+          if (payload == null) {
+            Thread.sleep(10);
+            continue;
+          }
```
randomanderson, line 320: Is busy waiting really better than a parked thread?

piochelepiotr: From the profiles, it looks much better (overhead of this queue went to nearly 0)

piochelepiotr: I did something similar in Go: https://github.com/DataDog/dd-trace-go/pull/2455/files

ddyurchenko: > From the profiles, it looks much better (overhead of this queue went to nearly 0)

Mb. overhead metric calculated from profile is not right? What is the change in latency?
I second on this one, sounds strange that doing more work results in less overhead.

piochelepiotr: I don't know Java enough, but in Go, the overhead of doing Go routine synchronization was much more than the overhead of checking if there is data once every 10 ms.
The problem, is that at each write, a Go channel had to then try to wake up the consumer, and that was causing a lot of overhead. I expect it's about the same here.

richardstartin: Something similar was done in #4409 to address the same bottleneck - `pthread_cond_signal`. 

I don't think it's helpful to describe this as "strange" without consideration of the production and consumption rates of this queue. The production rate is equivalent to `rps`, and the consumption rate is a constant 100. Considering the overhead from the perspective of the producer, which is no longer responsible for unblocking threads and can just drop the item off in the queue (improving the _latency_ of the producer thread). On the other hand, the consumer is now waking up 100 times per second, either to do some work or to go back to sleep, incurring a useless context switch every 10ms. (However, when it was waiting on a condition, it might have woken up more than 100 times per second anyway if `rps` were greater than 100 and would only have woken up less often if the traffic were bursty.) So this change should reduce CPU overhead whenever `(items produced per second) * (cpu cost of pthread_cond_signal) > (cost of scheduling the thread) * 100`. Assuming `pthread_cond_signal` costs the same as scheduling the thread (we have to assume because the profiler only samples one of these) this is a win whenever rps > 100. I think it's more likely that the cost of scheduling the thread is 10% of `pthread_cond_signal` meaning this change would win at rps > 10.

If you aim to reduce latency impact on application threads, this change is a win because the application threads are no longer responsible for waking up one of our background threads. If you aim

richardstartin (approved):
Great use of the profiler, and the changes make sense to me.

dd-trace-core/src/main/java/datadog/trace/core/datastreams/DefaultDataStreamsMonitoring.java
piochelepiotr: Thanks for all the context!

PR #7763, opened 2024-10-14
Expand SSRF support in IAST to all APM-supported clients

Mariovido:
# What Does This Do
This will expand SSRF to be supported in all APM-supported clients. This will be done by putting the sink in the HttpClientDecorator which is common to all different http clients such as OkHttp, Apache Http, Commons Http...

Another thing done in this PR, is the removal of the redundant instrumentation made by the previous change.

# Motivation
The motivation of this PR is expanding the SSRF support for IAST similar to the one done in RASP.

# Additional Notes

# Contributor Checklist

Jira ticket: [APPSEC-55237]

[APPSEC-55237]: https://datadoghq.atlassian.net/browse/APPSEC-55237?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ


dd-java-agent/agent-bootstrap/src/main/java/datadog/trace/bootstrap/instrumentation/decorator/HttpClientDecorator.java
```diff
@@ -168,4 +173,14 @@ public long getResponseContentLength(final RESPONSE response) {
 
     return 0;
   }
+
+  private void ssrfIastCheck(final String networkConnection) {
```
jandro996, line 177: Maybe I'm missing something but I think we can keep this as an Object as we have `SsrfModule#onURLConnection(@Nullable Object url)`

dd-java-agent/agent-iast/src/test/groovy/com/datadog/iast/sink/SsrfModuleTest.groovy
jandro996, line 1: Not relevant bu IMHO it's better to avoid this kind of new line modifications 😄

dd-java-agent/agent-bootstrap/src/main/java/datadog/trace/bootstrap/instrumentation/decorator/HttpClientDecorator.java
jandro996: I'm not 100% sure, but I think that we need to annotate this method with @Sink(VulnerabilityTypes.SSRF) to keep the metrics as we are removing them deleting the IAST instrumentations. Would this work @manuel-alvarez-alvarez ?

dd-java-agent/instrumentation/apache-httpclient-4/src/main/java/datadog/trace/instrumentation/apachehttpclient/IastApacheHttpClientInstrumentation.java
```diff
@@ -1,161 +0,0 @@
...
-
-    @Advice.OnMethodExit(onThrowable = Throwable.class, suppress = Throwable.class)
-    public static void methodExit() {
-      IastHelperMethods.doMethodExit();
-    }
-  }
-
-  public static class RequestAdvice {
-    @Advice.OnMethodEnter(suppress = Throwable.class)
-    @Sink(VulnerabilityTypes.SSRF)
```
jandro996, line 146: We need to keep this `@Sink` annotations, check my comment in `HttpClientDecorator.java`

dd-java-agent/instrumentation/apache-httpclient-4/src/test/groovy/IastHttpClientInstrumentationTest.groovy
```diff
@@ -41,22 +41,20 @@ class IastHttpClientInstrumentationTest extends AgentTestRunner {
     httpClient.execute(*args)
 
     then:
-    if(classExpected == URI){
-      1 * ssrf.onURLConnection(_ as URI)
-    } else {
-      1 * ssrf.onURLConnection(_ as String, _ as String, _ as String)
+    if(classExpected == String) {
```
jandro996, line 44: IMHO we need to keep the tests as they were before this PR to ensure that we are not changing our previous behaviour.

For instance `1 * ssrf.onURLConnection(_ as String, _ as String, _ as String)` is checking one special case that I think we are missing with the new changes

dd-java-agent/instrumentation/apache-httpclient-4/src/main/java/datadog/trace/instrumentation/apachehttpclient/IastHelperMethods.java
```diff
@@ -1,40 +0,0 @@
...
-    if (callDepth > 0) {
-      return;
-    }
-    final SsrfModule module = InstrumentationBridge.SSRF;
-    if (module != null) {
-      module.onURLConnection(request.getURI());
-    }
-  }
-
-  public static void doMethodEnter(HttpHost httpHost, HttpRequest request) {
```
jandro996, line 24: I need to check the current implementation properly... but can you check that we are not missing coverage to this special case? this was done basically to check if the httpHost is tainted

dd-java-agent/instrumentation/okhttp-3/src/test/groovy/IastOkHttp3InstrumentationTest.groovy
jandro996, line 1: same here 😅

dd-java-agent/instrumentation/commons-httpclient-2/src/test/groovy/IastCommonsHttpClientInstrumentationTest.groovy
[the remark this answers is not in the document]
Mariovido: Depending the version of apache-httpclient it seems that the url will finish with "/" or not, so to avoid this I directly pass the url normalized to avoid this problem. What I mean is if I pass the `server.address.toString()`, that string doesn't end with "/" so for some versions of apache-httpclient it will work and for others it won't work as it will add this "/" at the end. I am doing this to make easier the testing

dd-java-agent/instrumentation/okhttp-2/src/test/groovy/IastOkHttp2InstrumentationTest.groovy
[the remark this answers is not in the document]
Mariovido: Answered in the previous comment :)

dd-java-agent/agent-iast/src/test/groovy/com/datadog/iast/sink/SsrfModuleTest.groovy
Mariovido: You're right I miss that one

Mariovido: Removed

dd-java-agent/agent-bootstrap/src/main/java/datadog/trace/bootstrap/instrumentation/decorator/HttpClientDecorator.java
manuel-alvarez-alvarez: No, since we are reusing APM instrumentation we probably don't want to annotate anything but report telemetry directly with:
```java
IastMetricCollector.add(EXECUTED_SINK, SSRF, span.getRequestContext());
```

PR #7798, opened 2024-10-17
Support service name overrides for DSM

kr-igor:
This PR adds a mechanism which allows overriding service name for DSM checkpoints per thread. 
Stats buckets will be split by service name, so per 1 time interval we may have several "named" buckets. 
This will be used to correctly capture service (application) names for spark / flink jobs.


dd-java-agent/instrumentation/spark-executor/src/main/java/datadog/trace/instrumentation/spark/SparkExecutorDecorator.java
```diff
@@ -26,12 +30,29 @@ protected CharSequence component() {
...
   }
 
-  public void onTaskStart(AgentSpan span, Executor.TaskRunner taskRunner) {
+  public void onTaskStart(AgentSpan span, Executor.TaskRunner taskRunner, Object taskDescription) {
     span.setTag("task_id", taskRunner.taskId());
     span.setTag("task_thread_name", taskRunner.threadName());
+
+    if (taskDescription != null) {
+      try {
+        Field prop = taskDescription.getClass().getDeclaredField("properties");
```
amarziali, line 39: I think you should optimize this by unreflecting a method handle for this getter. It will save access checking costs

dd-java-agent/instrumentation/spark-executor/src/main/java/datadog/trace/instrumentation/spark/SparkExecutorDecorator.java
```diff
@@ -26,12 +30,29 @@ protected CharSequence component() {
...
+        prop.setAccessible(true);
+        Properties props = (Properties) prop.get(taskDescription);
+        String appName = props.getProperty(propSparkAppName);
+        if (appName != null) {
+          AgentTracer.get()
+              .getDataStreamsMonitoring()
+              .setThreadServiceName(taskRunner.getThreadId(), appName);
```
amarziali, line 46: Is product wise coherent to have a different service name from what represented by DSM and what represented by the tracing?

dd-trace-core/src/main/java/datadog/trace/core/datastreams/DefaultDataStreamsMonitoring.java
```diff
@@ -363,21 +388,30 @@ private long currentBucket(long timestampNanos) {
   private void flush(long timestampNanos) {
     long currentBucket = currentBucket(timestampNanos);
 
-    List<StatsBucket> includedBuckets = new ArrayList<>();
-    Iterator<Map.Entry<Long, StatsBucket>> mapIterator = timeToBucket.entrySet().iterator();
+    // stats are grouped by time buckets and service names
+    Map<String, List<StatsBucket>> includedBuckets = new HashMap<>();
```
piochelepiotr, line 392: we could also put the service name in the statsPoint. What do you think? It would avoid doing this grouping logic here.

kr-igor: The service name is reported per bucket. You can put it inside the stats point but then you'll have to modify the writer to correctly split the bucket. Effectively doing similar thing in a different place.

dd-java-agent/instrumentation/spark-executor/src/main/java/datadog/trace/instrumentation/spark/SparkExecutorDecorator.java
kr-igor: It makes sense in case with Spark / Flink. It may not be possible to pass the service name from outsize in managed environments (for instance spans generated within Databricks for DJM may have service name == "databricks.all-purpose-cluster.dlt-execution-123"). At the same time customers expect to see the Spark Job name, not the cluster node name.

Also there a cases when a single running application (Task Manager in Flink) runs multiple independent stream processing tasks. Each task has it's own "service name", using node name is incorrect.

The better approach would be to dynamically set global service name depending on the context. This seems like a way large change. Happy to discuss this separately.

amarziali: actually you can do it by calling `updatePreferredServiceName` (https://github.com/DataDog/dd-trace-java/blob/00856e02d3478ae8b317ec1adc31b565423ef8d0/internal-api/src/main/java/datadog/trace/bootstrap/instrumentation/api/AgentTracer.java#L293). Is it what you were looking from?

amarziali: OK if it depends from the current context then it's not globally applying to the JVM.

dd-java-agent/instrumentation/kafka-clients-0.11/src/main/java/datadog/trace/instrumentation/kafka_clients/KafkaProducerInstrumentation.java
```diff
@@ -200,7 +200,8 @@ public static void onEnter(@Advice.Argument(value = 0) int estimatedPayloadSize)
                 saved.getTimestampNanos(),
                 saved.getPathwayLatencyNano(),
                 saved.getEdgeLatencyNano(),
-                estimatedPayloadSize);
+                estimatedPayloadSize,
+                saved.getServiceNameOverride());
```
amarziali, line 204: Are message produced in the same thread where the task is running? Also are we testing it? I did not see spark tests with kafka but perhaps I missed it

dd-trace-core/src/main/java/datadog/trace/core/datastreams/DefaultDataStreamsMonitoring.java
```diff
@@ -74,6 +75,8 @@ public class DefaultDataStreamsMonitoring implements DataStreamsMonitoring, Even
   private volatile boolean agentSupportsDataStreams = false;
   private volatile boolean configSupportsDataStreams = false;
   private final ConcurrentHashMap<String, SchemaSampler> schemaSamplers;
+  private static final ConcurrentHashMap<Long, String> threadServiceNames =
```
amarziali, line 78: I've some concerns adding unbounded maps here. It might be ok for the spark instrumentation but this has been added to a generic API on DataStreamMonitoring. It can be used in the future in a bad way and start leaking. If service names have to be pinned to a thread, possible a threadlocal variable should be used in order to ensure that that value will go away when the thread is collected.

amarziali (changes requested):
I think that addition is a good to have to improve the experience of DSM consumers. I've still some concerns especially around adding an unbounded map to store thread id to name correspondences. It should be done in a threadlocal and actually it might not work if the kafka producer (in this example) is called in a thread different from the spark task one. Also it will be nice to have more e2e testing with spark and kafka

amarziali (approved):

PR #7799, opened 2024-10-17
Updating Span Link creation due to header tag propagations for invalid spans

mhlidd:
# What Does This Do
Updates the keys in the Span Link creation to include the `attributes` field and to only include `tracestate` when the propagation style is `tracecontext`. Additionally, wrote a unit test to verify the creation of Span Links when spans passed down through header tags are "invalid" due to differing trace_ids. 
# Motivation
Currently the implementation of how Span Links are used when DD receives invalid traces are not consistent among libraries. This PR focuses on making the Java library's implementation of Span Links consistent with the following [RFC](https://docs.google.com/document/d/1xacBonCyuVk95D-L1STXGqdLtsOuf0_3jG-138uXHnQ/edit?pli=1&tab=t.0#bookmark=id.uwnmpxy1znny). 
# Additional Notes
The way that the Java library is sending the `traceFlags` field is unclear, as there is no real way to differentiate whether a sampling flag is unset, or manually set to 0. 
# Contributor Checklist

Jira ticket: [PROJ-IDENT]


dd-trace-core/src/main/java/datadog/trace/core/propagation/HttpCodec.java
```diff
@@ -12,18 +12,14 @@
...
 import java.io.UnsupportedEncodingException;
 import java.net.URLDecoder;
 import java.net.URLEncoder;
-import java.util.ArrayList;
-import java.util.EnumMap;
-import java.util.EnumSet;
-import java.util.List;
-import java.util.Map;
-import java.util.Set;
+import java.util.*;
```
PerfectSlayer, line 22: Could you make sure to setup your IDE to not use wildcard import?
https://github.com/DataDog/dd-trace-java/blob/master/CONTRIBUTING.md#intellij-idea

dd-trace-core/src/main/java/datadog/trace/core/propagation/HttpCodec.java
```diff
@@ -240,7 +236,12 @@ public <C> TagContext extract(
               }
             } else {
               // Terminate extracted context and add it as span link
-              context.addTerminatedContextLink(DDSpanLink.from((ExtractedContext) extracted));
+              Map<String, String> attributes = new HashMap<>();
+              attributes.put("reason", "terminated_context");
+              attributes.put("context_headers", extracted.getPropagationStyle().toString());
+              context.addTerminatedContextLink(
+                  DDSpanLink.from(
+                      (ExtractedContext) extracted, SpanAttributes.fromMap(attributes)));
```
PerfectSlayer, line 244: What about using `SpanAttributes.Builder` rather than creating an intermediate Java `Map`?

dd-trace-core/src/test/groovy/datadog/trace/core/propagation/HttpExtractorTest.groovy
```diff
@@ -166,4 +166,43 @@ class HttpExtractorTest extends DDSpecification {
...
+    DynamicConfig dynamicConfig = DynamicConfig.create().apply()
+    HttpCodec.Extractor extractor = HttpCodec.createExtractor(config, { dynamicConfig.captureTraceConfig() })
+
+    final Map<String, String> actual = [:]
+    actual.put(DatadogHttpCodec.TRACE_ID_KEY.toUpperCase(), datadogTraceId)
+    actual.put(DatadogHttpCodec.SPAN_ID_KEY.toUpperCase(), datadogSpanId)
+    actual.put(B3HttpCodec.TRACE_ID_KEY.toUpperCase(), b3TraceId)
+    actual.put(B3HttpCodec.SPAN_ID_KEY.toUpperCase(), b3SpanId)
+    actual.put(W3CHttpCodec.TRACE_PARENT_KEY.toUpperCase(), w3cTraceParent)
+    actual.put(W3CHttpCodec.TRACE_STATE_KEY.toUpperCase(), traceState)
```
PerfectSlayer, line 184:
```suggestion
    final Map<String, String> actual = [
      DatadogHttpCodec.TRACE_ID_KEY.toUpperCase(): datadogTraceId,
      DatadogHttpCodec.SPAN_ID_KEY.toUpperCase(): datadogSpanId,
      B3HttpCodec.TRACE_ID_KEY.toUpperCase(): b3TraceId,
      B3HttpCodec.SPAN_ID_KEY.toUpperCase(): b3SpanId,
      W3CHttpCodec.TRACE_PARENT_KEY.toUpperCase(): w3cTraceParent,
      W3CHttpCodec.TRACE_STATE_KEY.toUpperCase(): traceState
    ]
```

dd-trace-core/src/main/java/datadog/trace/core/propagation/HttpCodec.java
```diff
@@ -240,7 +236,12 @@ public <C> TagContext extract(
               }
             } else {
               // Terminate extracted context and add it as span link
-              context.addTerminatedContextLink(DDSpanLink.from((ExtractedContext) extracted));
+              Map<String, String> attributes = new HashMap<>();
+              attributes.put("reason", "terminated_context");
+              attributes.put("context_headers", extracted.getPropagationStyle().toString());
```
PerfectSlayer, line 241: Dumb question: where is the spec for this part? Because at some point, context won’t always come from "headers" but message properties too, etc...

dd-trace-core/src/main/java/datadog/trace/core/DDSpanLink.java
```diff
@@ -53,6 +54,7 @@ public static SpanLink from(ExtractedContext context, Attributes attributes) {
     byte traceFlags = context.getSamplingPriority() > 0 ? SAMPLED_FLAG : DEFAULT_FLAGS;
     String traceState =
         context.getPropagationTags() == null
+                || context.getPropagationStyle() != TracePropagationStyle.TRACECONTEXT
```
PerfectSlayer, line 57: Why do we want to drop trace state if the context is not issued from W3C Trace Context?
The API should be generic enough to allow any context as input.

PerfectSlayer (commented):
Nice additions. I left few questions during the review

dd-trace-core/src/main/java/datadog/trace/core/propagation/HttpCodec.java
mhlidd: We are only worried about extracting information from headers here and creating span links for those headers that have inconsistent trace ids. ([RFC](https://docs.google.com/document/d/1xacBonCyuVk95D-L1STXGqdLtsOuf0_3jG-138uXHnQ/edit?pli=1&tab=t.0#bookmark=id.uwnmpxy1znny))

dd-trace-core/src/main/java/datadog/trace/core/DDSpanLink.java
mhlidd: Each span link is tied to an inconsistent trace extracted with a specific propagator type. For a b3 header, for instance, it doesn't make sense to include the W3C tracestate. ([RFC](https://docs.google.com/document/d/1xacBonCyuVk95D-L1STXGqdLtsOuf0_3jG-138uXHnQ/edit?pli=1&tab=t.0#bookmark=id.uwnmpxy1znny))

mhlidd: Update: We will keep tracestates the way that they are currently implemented in Java

PerfectSlayer:
> I think the important distinction to be made here is that there is a difference between not sampling and sampling unset, since the DD agent may update the sampling state after the span is created if the sampling is unset. Thus, it would be nice to be able to differentiate whether a span is not sampled vs sampling unset. 

What you are describing is the `dropped` decision from the tracer. But such flag don't exist in tracecontext state, only `sampled` exist.

> If both are set to 0, there is no way to differentiate the two in the backend, which is also the reason that the bit-shifting was important for v04 encodings. This isn't a feature that is critical to this task, but is more of a nice to have.

The v0.4 protocol workaround should not impact the JSON encoding. If they ever come up with a drop flag, we will implement it :)

PerfectSlayer (approved):
LGTM 👍

PR #7820, opened 2024-10-21
Introduce aggregated smap events (enabled by default)

MattAlp:
# What Does This Do
This introduces a new event type that records smap events aggregated by NMT category rather than individual smap events for every single map reported by the OS. This feature is enabled by default but can be overridden for customers interested in seeing granular info.

# Motivation
Due to the usually-high number of maps created by the JVM, recording every single smap entry tends to bloat uploaded profile sizes. We currently only make use of the RSS property from the individual smap events, rather than everything scraped, so this feature restricts collection to only the data needed (unless overridden).

# Additional Notes
We may want to follow fast with a PSS property as well. Will squash and merge on approval.

# Contributor Checklist

Jira ticket: [PROF-10767]

[PROF-10767]: https://datadoghq.atlassian.net/browse/PROF-10767?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ

### DIFF
--- a/dd-java-agent/agent-profiling/profiling-controller-openjdk/src/main/java/com/datadog/profiling/controller/openjdk/OpenJdkController.java
+++ b/dd-java-agent/agent-profiling/profiling-controller-openjdk/src/main/java/com/datadog/profiling/controller/openjdk/OpenJdkController.java
@@ -227,7 +227,6 @@ public final class OpenJdkController implements Controller {
         ProfilingConfig.PROFILING_SMAP_COLLECTION_ENABLED,
         ProfilingConfig.PROFILING_SMAP_COLLECTION_ENABLED_DEFAULT)) {
       disableEvent(recordingSettings, "datadog.SmapEntry", "User disabled smaps collection");
-      System.out.println("User disabled smaps collection");
     } else if (!configProvider.getBoolean(
         ProfilingConfig.PROFILING_SMAP_AGGREGATION_ENABLED,
         ProfilingConfig.PROFILING_SMAP_AGGREGATION_ENABLED_DEFAULT)) {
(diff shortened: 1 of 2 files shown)


richardstartin:
Looks good, except there are some `System.out.println`s you'll need to remove.

richardstartin (approved):

dd-java-agent/agent-profiling/profiling-controller-openjdk/src/main/java/com/datadog/profiling/controller/openjdk/OpenJdkController.java
```diff
@@ -223,12 +223,17 @@ ProfilingConfig.PROFILING_ALLOCATION_ENABLED, isObjectAllocationSampleAvailable(
...
+    if (!configProvider.getBoolean(
         ProfilingConfig.PROFILING_SMAP_COLLECTION_ENABLED,
         ProfilingConfig.PROFILING_SMAP_COLLECTION_ENABLED_DEFAULT)) {
-      enableEvent(recordingSettings, "datadog.SmapEntry", "User enabled smaps collection");
-    } else {
       disableEvent(recordingSettings, "datadog.SmapEntry", "User disabled smaps collection");
+    } else if (!configProvider.getBoolean(
+        ProfilingConfig.PROFILING_SMAP_AGGREGATION_ENABLED,
+        ProfilingConfig.PROFILING_SMAP_AGGREGATION_ENABLED_DEFAULT)) {
+      disableEvent(
```
jbachorik, line 233: Does this work? The event is not marked as `@Enabled` in the event class, so I would assume it will not be auto-enabled? We are not enabling by default in `dd.jfp` neither.

jbachorik: 🤦 Thanks for checking this. The default enablement when `@Enabled` is not present is really confusing, to say the least ...

PR #7824, opened 2024-10-22
Add retry policy for uploading requests to agent

jpbempel:
# What Does This Do
Add a mechanism to retry on network failure or 500 response code. The retry policy store the call made to OkHttp to track number of retries made and the maximum retries allowed by failure categories: network or 500.
3 retries are allowed for snapshots, 10 for probe statuses & symdb

# Motivation

# Additional Notes

# Contributor Checklist

Jira ticket: [PROJ-IDENT]

### DIFF
--- a/dd-java-agent/agent-debugger/src/main/java/com/datadog/debugger/agent/DebuggerAgent.java
+++ b/dd-java-agent/agent-debugger/src/main/java/com/datadog/debugger/agent/DebuggerAgent.java
@@ -48,8 +48,6 @@ import org.slf4j.LoggerFactory;
 /** Debugger agent implementation */
 public class DebuggerAgent {
   private static final Logger LOGGER = LoggerFactory.getLogger(DebuggerAgent.class);
-  private static final BatchUploader.RetryPolicy SNAPSHOT_RETRY_POLICY =
-      new BatchUploader.RetryPolicy(3, 3);
   private static ConfigurationPoller configurationPoller;
   private static DebuggerSink sink;
   private static String agentVersion;
@@ -171,7 +169,8 @@ public class DebuggerAgent {
         new SnapshotSink(
             config,
             tags,
-            new BatchUploader(config, config.getFinalDebuggerSnapshotUrl(), SNAPSHOT_RETRY_POLICY));
+            new BatchUploader(
+                config, config.getFinalDebuggerSnapshotUrl(), SnapshotSink.RETRY_POLICY));
     SymbolSink symbolSink = new SymbolSink(config);
     return new DebuggerSink(
         config,
--- a/dd-java-agent/agent-debugger/src/main/java/com/datadog/debugger/agent/DebuggerAgent.java
+++ b/dd-java-agent/agent-debugger/src/main/java/com/datadog/debugger/agent/DebuggerAgent.java
@@ -48,6 +48,8 @@ import org.slf4j.LoggerFactory;
 /** Debugger agent implementation */
 public class DebuggerAgent {
   private static final Logger LOGGER = LoggerFactory.getLogger(DebuggerAgent.class);
+  private static final BatchUploader.RetryPolicy SNAPSHOT_RETRY_POLICY =
+      new BatchUploader.RetryPolicy(3, 3);
   private static ConfigurationPoller configurationPoller;
   private static DebuggerSink sink;
   private static String agentVersion;
@@ -167,7 +169,9 @@ public class DebuggerAgent {
     String tags = getDefaultTagsMergedWithGlobalTags(config);
     SnapshotSink snapshotSink =
         new SnapshotSink(
-            config, tags, new BatchUploader(config, config.getFinalDebuggerSnapshotUrl()));
+            config,
+            tags,
+            new BatchUploader(config, config.getFinalDebuggerSnapshotUrl(), SNAPSHOT_RETRY_POLICY));
     SymbolSink symbolSink = new SymbolSink(config);
     return new DebuggerSink(
         config,
--- a/dd-java-agent/agent-debugger/src/main/java/com/datadog/debugger/agent/DebuggerTransformer.java
+++ b/dd-java-agent/agent-debugger/src/main/java/com/datadog/debugger/agent/DebuggerTransformer.java
@@ -152,7 +152,12 @@ public class DebuggerTransformer implements ClassFileTransformer {
             DebuggerMetrics.getInstance(config),
             new ProbeStatusSink(config, config.getFinalDebuggerSnapshotUrl(), false),
             new SnapshotSink(
-                config, "", new BatchUploader(config, config.getFinalDebuggerSnapshotUrl())),
+                config,
+                "",
+                new BatchUploader(
+                    config,
+                    config.getFinalDebuggerSnapshotUrl(),
+                    new BatchUploader.RetryPolicy(3, 3))),
             new SymbolSink(config)));
   }
 
--- a/dd-java-agent/agent-debugger/src/main/java/com/datadog/debugger/agent/DebuggerTransformer.java
+++ b/dd-java-agent/agent-debugger/src/main/java/com/datadog/debugger/agent/DebuggerTransformer.java
@@ -155,9 +155,7 @@ public class DebuggerTransformer implements ClassFileTransformer {
                 config,
                 "",
                 new BatchUploader(
-                    config,
-                    config.getFinalDebuggerSnapshotUrl(),
-                    new BatchUploader.RetryPolicy(3, 3))),
+                    config, config.getFinalDebuggerSnapshotUrl(), SnapshotSink.RETRY_POLICY)),
             new SymbolSink(config)));
   }
 
--- a/dd-java-agent/agent-debugger/src/main/java/com/datadog/debugger/sink/DebuggerSink.java
+++ b/dd-java-agent/agent-debugger/src/main/java/com/datadog/debugger/sink/DebuggerSink.java
@@ -43,7 +43,10 @@ public class DebuggerSink {
         DebuggerMetrics.getInstance(config),
         probeStatusSink,
         new SnapshotSink(
-            config, null, new BatchUploader(config, config.getFinalDebuggerSnapshotUrl())),
+            config,
+            null,
+            new BatchUploader(
+                config, config.getFinalDebuggerSnapshotUrl(), new BatchUploader.RetryPolicy(3, 3))),
         new SymbolSink(config));
   }
 
--- a/dd-java-agent/agent-debugger/src/main/java/com/datadog/debugger/sink/DebuggerSink.java
+++ b/dd-java-agent/agent-debugger/src/main/java/com/datadog/debugger/sink/DebuggerSink.java
@@ -46,7 +46,7 @@ public class DebuggerSink {
             config,
             null,
             new BatchUploader(
-                config, config.getFinalDebuggerSnapshotUrl(), new BatchUploader.RetryPolicy(3, 3))),
+                config, config.getFinalDebuggerSnapshotUrl(), SnapshotSink.RETRY_POLICY)),
         new SymbolSink(config));
   }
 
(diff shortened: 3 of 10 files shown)


dd-java-agent/agent-debugger/src/main/java/com/datadog/debugger/uploader/BatchUploader.java
```diff
@@ -301,6 +369,11 @@ public void onResponse(final Call call, final Response response) {
                 response.message(),
                 response.code());
           }
+          if (response.code() >= 500) {
```
ojung, line 372: There are also `408` (client timeout) and `429` (too many requests) that should be retried according to the official logs http api docs: https://docs.datadoghq.com/api/latest/logs/

dd-java-agent/agent-debugger/src/main/java/com/datadog/debugger/uploader/BatchUploader.java
```diff
@@ -260,28 +289,67 @@ private boolean canEnqueueMoreRequests() {
...
+      handleRetry(call, retryPolicy.maxNetworkFailures);
+    }
+
+    private void handleRetry(Call call, int maxFailures) {
+      Integer failure = retryPolicy.failures.remove(call);
+      if (failure != null) {
+        int failureCount = failure + 1;
+        if (failureCount <= maxFailures) {
+          LOGGER.debug("Retrying upload {}/{}", failureCount, maxFailures);
+          enqueueCall(client, call.request(), this, retryPolicy, failureCount, inflightRequests);
```
ojung, line 336: Don't want to over-complicate things, but maybe we should add a backoff? If the requests fail fast, all retries might be done in a very short time interval.

dd-java-agent/agent-debugger/src/main/java/com/datadog/debugger/uploader/BatchUploader.java
```diff
@@ -260,28 +289,67 @@ private boolean canEnqueueMoreRequests() {
     return client.dispatcher().queuedCallsCount() < MAX_ENQUEUED_REQUESTS;
   }
 
+  private static void enqueueCall(
```
ojung, line 292: How are we handling back pressure? If network is unreliable and we retry the same symdb uploads multiple times with backoff, are we going to slow down serializing new classes into memory?

dd-java-agent/agent-debugger/src/main/java/com/datadog/debugger/agent/DebuggerAgent.java
```diff
@@ -48,6 +48,8 @@
 /** Debugger agent implementation */
 public class DebuggerAgent {
   private static final Logger LOGGER = LoggerFactory.getLogger(DebuggerAgent.class);
+  private static final BatchUploader.RetryPolicy SNAPSHOT_RETRY_POLICY =
+      new BatchUploader.RetryPolicy(3, 3);
```
ojung, line 52: Not sure if we should retry snapshots 🤔 SymDB uploads and probe statuses I agree, but for logs, it might be overkill?

ojung (commented):
Thanks 👍 Added couple of comments / questions.

dd-java-agent/agent-debugger/src/main/java/com/datadog/debugger/uploader/BatchUploader.java
```diff
@@ -260,28 +289,67 @@ private boolean canEnqueueMoreRequests() {
...
       ratelimitedLogger.warn("Failed to upload batch to {}", call.request().url(), e);
+      handleRetry(call, retryPolicy.maxNetworkFailures);
+    }
+
+    private void handleRetry(Call call, int maxFailures) {
+      Integer failure = retryPolicy.failures.remove(call);
+      if (failure != null) {
+        int failureCount = failure + 1;
+        if (failureCount <= maxFailures) {
+          LOGGER.debug("Retrying upload {}/{}", failureCount, maxFailures);
```
shatzi, line 335: add call.request().url() to better understand which sync is was coming from or add BatchUploader.name() ?

dd-java-agent/agent-debugger/src/main/java/com/datadog/debugger/uploader/BatchUploader.java
```diff
@@ -260,28 +289,67 @@ private boolean canEnqueueMoreRequests() {
...
+      RetryPolicy retryPolicy,
+      int failureCount,
+      Phaser inflightRequests) {
+    Call call = client.newCall(request);
+    retryPolicy.failures.put(call, failureCount);
+    call.enqueue(responseCallback);
+    inflightRequests.register();
+  }
+
   private static final class ResponseCallback implements Callback {
```
shatzi, line 305: suggest this design here, would be simpler to understand.
1. make it private inner class (not static) - so you don't need to pass ratelimitedLoger, inflightRequests ... they can just use the parent class references.
2. add two members: request + failures.
3. make new instances of this callback for every call of buildAndSendRequest - it only need to get a request object.
4. this can handle all response and can use the request for retry and simple count the failures.

this would enable to remove the failures map that look out of place and not really needed.

dd-java-agent/agent-debugger/src/test/java/com/datadog/debugger/uploader/BatchUploaderTest.java
```diff
@@ -48,6 +50,7 @@ public class BatchUploaderTest {
 
   private final MockWebServer server = new MockWebServer();
   private HttpUrl url;
+  private final BatchUploader.RetryPolicy retryPolicy = new BatchUploader.RetryPolicy(3, 3);
```
shatzi, line 53: for all those tests, maybe add back that BatchUploader ctor overload with default retryPolicy ? or do we really want to specify that retryPolicy everywhere?

shatzi (approved):

dd-java-agent/agent-debugger/src/main/java/com/datadog/debugger/agent/DebuggerAgent.java
jpbempel: ok let's turn it to 0. at least we have the mechanism in place

dd-java-agent/agent-debugger/src/main/java/com/datadog/debugger/agent/DebuggerAgent.java
```diff
@@ -167,7 +169,9 @@ private static DebuggerSink createDebuggerSink(Config config, ProbeStatusSink pr
     String tags = getDefaultTagsMergedWithGlobalTags(config);
     SnapshotSink snapshotSink =
         new SnapshotSink(
-            config, tags, new BatchUploader(config, config.getFinalDebuggerSnapshotUrl()));
+            config,
+            tags,
+            new BatchUploader(config, config.getFinalDebuggerSnapshotUrl(), SNAPSHOT_RETRY_POLICY));
```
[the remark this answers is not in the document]
jpbempel: > Snapshots are stream of events.
not quite agree with this view. some snapshots are more valuable than others specially if you put conditions.
But as answered to Oskar, let's turn it to 0. and see how it goes.

dd-java-agent/agent-debugger/src/main/java/com/datadog/debugger/agent/DebuggerTransformer.java
```diff
@@ -152,7 +152,12 @@ public DebuggerTransformer(Config config, Configuration configuration) {
             DebuggerMetrics.getInstance(config),
             new ProbeStatusSink(config, config.getFinalDebuggerSnapshotUrl(), false),
             new SnapshotSink(
-                config, "", new BatchUploader(config, config.getFinalDebuggerSnapshotUrl())),
+                config,
+                "",
+                new BatchUploader(
+                    config,
+                    config.getFinalDebuggerSnapshotUrl(),
+                    new BatchUploader.RetryPolicy(3, 3))),
```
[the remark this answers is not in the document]
jpbempel: done

dd-java-agent/agent-debugger/src/main/java/com/datadog/debugger/sink/DebuggerSink.java
```diff
@@ -43,7 +43,10 @@ public DebuggerSink(Config config, ProbeStatusSink probeStatusSink) {
         DebuggerMetrics.getInstance(config),
         probeStatusSink,
         new SnapshotSink(
-            config, null, new BatchUploader(config, config.getFinalDebuggerSnapshotUrl())),
+            config,
+            null,
+            new BatchUploader(
+                config, config.getFinalDebuggerSnapshotUrl(), new BatchUploader.RetryPolicy(3, 3))),
```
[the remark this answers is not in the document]
jpbempel: done

dd-java-agent/agent-debugger/src/main/java/com/datadog/debugger/uploader/BatchUploader.java
```diff
@@ -56,7 +58,18 @@ public String getFileName() {
     }
   }
 
-  private static final Logger log = LoggerFactory.getLogger(BatchUploader.class);
+  public static class RetryPolicy {
+    public final ConcurrentMap<Call, Integer> failures = new ConcurrentHashMap<>();
+    public final int maxNetworkFailures;
```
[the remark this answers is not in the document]
jpbempel: done

dd-java-agent/agent-debugger/src/main/java/com/datadog/debugger/uploader/BatchUploader.java
jpbempel: we have a MAX_ENQUEUED_REQUESTS limit placed at 20 (see above)
so we are dropping requests if reach this limit

dd-java-agent/agent-debugger/src/main/java/com/datadog/debugger/uploader/BatchUploader.java
jpbempel: on critical path I am always trying to reduce allocations. That's why I have tried to keep this ReponseCallback as it is because we are avoiding to allocate an instance for each request.
I can understand this is not most straightforward to handle the tracking of failures.
Not easy to place correctly the cursor about the perf impact and readability

dd-java-agent/agent-debugger/src/main/java/com/datadog/debugger/uploader/BatchUploader.java
jpbempel: done

dd-java-agent/agent-debugger/src/main/java/com/datadog/debugger/uploader/BatchUploader.java
jpbempel: I see your point. The problem is how I deal with this backoff? where I put the new request on hold, who is responsible for enqueue again after x time?
also, we are keeping for a longest period large buffers (serialized symbols up to 50MB)

dd-java-agent/agent-debugger/src/test/java/com/datadog/debugger/uploader/BatchUploaderTest.java
jpbempel: I am conflicted to add new overloaded constructor because sometimes it's complicated to maintain on the long run ( because we are accumulating and version for each new parameter)

dd-java-agent/agent-debugger/src/main/java/com/datadog/debugger/uploader/BatchUploader.java
ojung: How much effort would it be to change the way we extract symbols to:
1. Discover jars to parse
2. Put them (path only) on a queue
3. Only parse them as fast as we can (without dropping requests)

dd-java-agent/agent-debugger/src/main/java/com/datadog/debugger/uploader/BatchUploader.java
ojung: Well, we would have to keep them in a collection on the class level I guess. We could either have a scheduled thread that re-enqueues the requests or store a scheduled thread for each retry that has a given initial delay. As I said, it's making things a bit more complicated, but the more I think about it the less I like retrying immediately.

ojung (approved):

PR #7838, opened 2024-10-25
Split SymDB payload when too large

jpbempel:
# What Does This Do
SymDB payload cannot be larger than 50MB otherwise datadog agent will generate an error. When serializing the upload request we are now verifying that we are below this limit, otherwise we are splitting the payload first by uploading by jar scope, but if a jar scope is still large we are splitting by class scopes. If down to one class the payload is still too large we are dropping the upload.

# Motivation

# Additional Notes

# Contributor Checklist

Jira ticket: [DEBUG-3068]

[DEBUG-3068]: https://datadoghq.atlassian.net/browse/DEBUG-3068?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ

### DIFF
--- a/dd-java-agent/agent-debugger/src/main/java/com/datadog/debugger/sink/SymbolSink.java
+++ b/dd-java-agent/agent-debugger/src/main/java/com/datadog/debugger/sink/SymbolSink.java
@@ -140,18 +140,34 @@ public class SymbolSink {
             continue;
           }
           LOGGER.debug("Sending {} jar scope size={}", scope.getName(), json.length());
-          updateStats(Collections.singletonList(scope), json);
-          symbolUploader.uploadAsMultipart(
-              "",
-              event,
-              new BatchUploader.MultiPartContent(
-                  json.getBytes(StandardCharsets.UTF_8), "file", "file.json"));
+          doUpload(Collections.singletonList(scope), json);
         }
       } else {
-        LOGGER.warn(
-            "Too many jar scopes to send, {} jar scopes, max is {}",
-            scopesToSerialize.size(),
-            BatchUploader.MAX_ENQUEUED_REQUESTS);
+        // split the list of jar scope in 2 list jar scopes with half of the jar scopes
+        int half = scopesToSerialize.size() / 2;
+        List<Scope> firstHalf = scopesToSerialize.subList(0, half);
+        List<Scope> secondHalf = scopesToSerialize.subList(half, scopesToSerialize.size());
+        LOGGER.debug("split jar scope list in 2: {} and {}", firstHalf.size(), secondHalf.size());
+        String jsonFirstHalf =
+            SERVICE_VERSION_ADAPTER.toJson(
+                new ServiceVersion(serviceName, env, version, "JAVA", firstHalf));
+        if (jsonFirstHalf.length() > maxPayloadSize) {
+          LOGGER.warn(
+              "Cannot split jar scope list in 2, first half is too big: {}",
+              jsonFirstHalf.length());
+          return;
+        }
+        doUpload(firstHalf, jsonFirstHalf);
+        String jsonSecondHalf =
+            SERVICE_VERSION_ADAPTER.toJson(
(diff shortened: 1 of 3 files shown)


dd-java-agent/agent-debugger/src/main/java/com/datadog/debugger/sink/SymbolSink.java
```diff
@@ -87,13 +96,92 @@ public void flush() {
...
+      List<Scope> firstHalf = jarScope.getScopes().subList(0, half);
+      List<Scope> secondHalf = jarScope.getScopes().subList(half, jarScope.getScopes().size());
+      LOGGER.debug(
+          "split jar scope {} in 2 jar scopes: {} and {}",
+          jarScope.getName(),
+          firstHalf.size(),
+          secondHalf.size());
+      splitAndSend(
+          Arrays.asList(
+              createJarScope(jarScope.getName(), firstHalf),
```
shatzi, line 178: @ojung - how having two jar scope would work at the backend?

dd-java-agent/agent-debugger/src/main/java/com/datadog/debugger/sink/SymbolSink.java
```diff
@@ -87,13 +96,92 @@ public void flush() {
...
+          LOGGER.debug("Sending {} jar scope size={}", scope.getName(), json.length());
+          updateStats(Collections.singletonList(scope), json);
+          symbolUploader.uploadAsMultipart(
+              "",
+              event,
+              new BatchUploader.MultiPartContent(
+                  json.getBytes(StandardCharsets.UTF_8), "file", "file.json"));
+        }
+      } else {
+        LOGGER.warn(
```
shatzi, line 151: so, dropping all those jars? maybe we do the same splitting by two for jars case...

dd-java-agent/agent-debugger/src/main/java/com/datadog/debugger/sink/SymbolSink.java
```diff
@@ -87,13 +96,92 @@ public void flush() {
...
         SERVICE_VERSION_ADAPTER.toJson(
             new ServiceVersion(serviceName, env, version, "JAVA", scopesToSerialize));
-    LOGGER.debug("Sending {} jar scopes size={}", scopesToSerialize.size(), json.length());
-    updateStats(scopesToSerialize, json);
-    symbolUploader.uploadAsMultipart(
-        "",
-        event,
-        new BatchUploader.MultiPartContent(
-            json.getBytes(StandardCharsets.UTF_8), "file", "file.json"));
+    if (json.length() > maxPayloadSize) {
```
shatzi, line 99: because we already calculated json.length() -> we can tell how much we need to split...
aka, chunks = Math.ciel(json.length()/maxPayloadSize).

this can avoid the other bug when we try to upload too many small jars and drop all of them

shatzi (approved):
I think this solves it - and we need to ensure that duplicated jar scope are supported by the backend. so the backend would be smart enough to merge the same jar scope if it come from same runtime-id.

We are serializing things again and agin if we hit that max size limit, we should put some metric/diagnostics to see if that happen to often - if it does - we might want to refactor to split earlier.

dd-java-agent/agent-debugger/src/main/java/com/datadog/debugger/sink/SymbolSink.java
ojung: Yes, that is supported since a while ago. It's also supported for `dll` and `module`.

dd-java-agent/agent-debugger/src/main/java/com/datadog/debugger/sink/SymbolSink.java
jpbempel: done

PR #7846, opened 2024-10-29
Prevent errors linked to Security Manager when reading config

vandonr:
# What Does This Do

extends the logic implemented by @dougqh for the bootstrap in https://github.com/DataDog/dd-trace-java/blob/master/dd-java-agent/src/main/java/datadog/trace/bootstrap/SystemUtils.java to the rest of the codebase.

# Motivation

The Java Security Manager can prevent us from accessing some env variables or system properties, but we should fallback to default values rather than crash when this happens.

# Additional Notes

I added a warning log, because if we completely swallow those errors, it can get very confusing to the customers why their configs are not applied. We can also provide some guidance on how to fix the issue.

Also, I wasn't sure where to put the boundary between where we can fallback to a default value, and where we really expect to get something. For instance, there is still a bunch of unprotected `getProperty` in https://github.com/DataDog/dd-trace-java/blob/5622bdce6e5405a56e8fa7985d696e15ebb6dafb/internal-api/src/main/java/datadog/trace/api/Platform.java and also here:
https://github.com/DataDog/dd-trace-java/blob/5c54fec10d896860f0e234f23258f497a03fa6c1/internal-api/src/main/java/datadog/trace/api/Config.java#L3263-L3268

# Contributor Checklist

Jira ticket: [APMS-13069]

[APMS-13069]: https://datadoghq.atlassian.net/browse/APMS-13069?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ


PerfectSlayer (commented):
A good addition would be to add the wrapped method into the forbidden APIs to prevent devs to use them any more.

vandonr:
> I'll work on the tests for this as part of my APM Ecosystems mentorship, learning java!

oh I missed this comment ! Feel free to, but it's not the easiest task to start with ! Some tests are failing because we are using the logger "too early". We probably need to change the code to store a flag if we couldn't access the settings we wanted, and check it later, when the initialization is done, to log something.

internal-api/src/main/java/datadog/trace/api/Config.java
```diff
@@ -3956,7 +3957,22 @@ private static String getProp(String name, String def) {
               : ConfigProvider.getInstance(),
           InstrumenterConfig.get());
 
+  private static boolean configErrorsLogged = false;
+
   public static Config get() {
+    // checking if there are errors that we couldn't log earlier to report
+    if (!configErrorsLogged && (SystemUtils.hasEnvError || SystemUtils.hasPropertyError)) {
```
vandonr, line 3964: reeeeaaally not sure this is a good idea, but it's the best place I found to do something "once after warmup"

dd-java-agent/agent-bootstrap/src/main/java/datadog/trace/bootstrap/Agent.java
```diff
@@ -370,6 +371,18 @@ public void run() {
...
+      log.warn(
+          "The Java Security Manager prevented the Datadog Tracer from accessing at least one environment variable. "
+              + "Consider granting AllPermission to the dd-java-agent jar.");
+    }
+    if (SystemUtils.hasPropertyError) {
+      log.warn(
+          "The Java Security Manager prevented the Datadog Tracer from accessing at least one system property. "
+              + "Consider granting AllPermission to the dd-java-agent jar.");
+    }
+
```
PerfectSlayer, line 385: Can you move the reporting logic to the `SystemUtils` class to keep the `Agent` as simple as possible?

vandonr: After some discussion [here](https://dd.slack.com/archives/C047YAMGZ7T/p1730822107288379) we decided to go for a different approach

vandonr:
> Should we move it to a draft state as there are still things to address, such as logging too early and marking old APIs as forbidden?

Ah yes maybe I need to edit the description, but we're not logging too early anymore.

PR #7865, opened 2024-10-31
Fix stack trace inconsistency between excluded frames in vulnerability location and metastruct stack trace

jandro996:
# What Does This Do

- Parametrize `StackUtils#generateUserCodeStackTrace` with a predicate
- Add `AbstractStackWalker#isNotDatadogTraceStackElement` as default filter
- Update isNotDatadogTraceStackElement to also filter `com.datadog.appsec`

# Motivation

We are filtering different stack frames according to their class to calculate the vulnerability location and the vulnerability stack trace that is reported via meta struct

# Additional Notes

# Contributor Checklist

Jira ticket: [PROJ-IDENT]


jandro996:
I'm not sure if we need more test to validate this change, keep in mind that it would be also tested in system tests as right now it would fail due to the vulnerability location is filtered from the reported stack trace as the location is in a `com.datadog` package

dd-java-agent/agent-debugger/build.gradle
```diff
@@ -7,7 +7,7 @@ apply from: "$rootDir/gradle/java.gradle"
 apply from: "$rootDir/gradle/version.gradle"
 
 minimumInstructionCoverage = 0.1
-minimumBranchCoverage = 0.6
+minimumBranchCoverage = 0.5
```
jandro996, line 10: Moving CodeOriginTest to package foo.bar makes jacoco fail as the coverage decrease to 0.5 😓

smola (approved):
Looks good to me for RASP and IAST, but needs a review by the DI team.

smola:
Also note the performance regression in debugger benchmarks.

internal-api/src/main/java/datadog/trace/util/stacktrace/AbstractStackWalker.java
```diff
@@ -18,6 +18,6 @@ final Stream<StackTraceElement> doFilterStack(Stream<StackTraceElement> stream)
 
   static boolean isNotDatadogTraceStackElement(final StackTraceElement el) {
     final String clazz = el.getClassName();
-    return !clazz.startsWith("datadog.trace.") && !clazz.startsWith("com.datadog.iast.");
+    return !clazz.startsWith("datadog.trace.") && !clazz.startsWith("com.datadog.");
```
jpbempel, line 21: This is problematic. this is too broad.
debugger backend services uses `com.datadog.debugger` as package and therefore will be excluded where we. may don't want to.

dd-java-agent/agent-debugger/src/test/java/foo/bar/debugger/origin/CodeOriginTest.java
```diff
@@ -1,4 +1,4 @@
-package com.datadog.debugger.origin;
+package foo.bar.debugger.origin;
```
evanchooly: It's weird to me that this is the *only* thing that needed to be changed which makes me question if it was the right fix to begin with.

internal-api/src/main/java/datadog/trace/util/stacktrace/AbstractStackWalker.java
evanchooly: Agreed.  And at such a fundamental level, the ripple effects of this one would be near impossible to see.

jandro996: Thanks for the advice, as seems that there are many problems with this method I will keep it as it is and create a new one to filter what we need in iast/appsec

dd-java-agent/agent-debugger/build.gradle
jandro996: I usually try to avoid these kinds of changes, but that drop of coverage caught my attention due to a package change.

dd-java-agent/agent-debugger/src/test/java/foo/bar/debugger/origin/CodeOriginTest.java
jandro996: I drop this change as `com.datadog.debugger` is not filtered any more :)

internal-api/src/test/java/datadog/trace/util/stacktrace/StackUtilsTest.java
```diff
@@ -92,13 +97,30 @@ public void test_filter_until() {
     assertThat(noRemoval.getStackTrace()).isEqualTo(stack);
   }
 
-  @Test
-  public void test_generateUserCodeStackTrace() {
-    List<StackTraceFrame> userCodeStack = StackUtils.generateUserCodeStackTrace();
+  private static Stream<Arguments> test_generateUserCodeStackTrace_Params() {
```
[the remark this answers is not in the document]
jandro996: My intention with this test is cover the StackUtils#generateUserCodeStackTrace parametrization, more than the AbstractStackWalker#isNotDatadogTraceStackElement that is covered by other tests. But I can try to add more specific if you feel that is necessary

smola (approved):
Looks good now, broader impact is gone.

manuel-alvarez-alvarez (approved):

PR #7867, opened 2024-10-31
delay queue time rate limiting until event is committed

richardstartin:
# What Does This Do

Applies rate limiting to the known outliers, not to every queue transaction.

# Motivation

# Additional Notes

# Contributor Checklist

Jira ticket: [PROJ-IDENT]

### DIFF
--- a/dd-java-agent/agent-bootstrap/src/main/java/datadog/trace/bootstrap/instrumentation/java/concurrent/QueueTimerHelper.java
+++ b/dd-java-agent/agent-bootstrap/src/main/java/datadog/trace/bootstrap/instrumentation/java/concurrent/QueueTimerHelper.java
@@ -36,7 +36,7 @@ public class QueueTimerHelper {
   }
 
   public static void startQueuingTimer(State state, Class<?> schedulerClass, Object task) {
-    if (Platform.isNativeImageBuilder()) {
+    if (Platform.isNativeImage()) {
       // explicitly not supported for Graal native image
       return;
     }
@@ -52,7 +52,7 @@ public class QueueTimerHelper {
   }
 
   public static void stopQueuingTimer(Timing timing) {
-    if (Platform.isNativeImageBuilder()) {
+    if (Platform.isNativeImage()) {
       // explicitly not supported for Graal native image
       return;
     }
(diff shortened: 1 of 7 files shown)


dd-java-agent/agent-bootstrap/src/main/java/datadog/trace/bootstrap/instrumentation/java/concurrent/QueueTimerHelper.java
```diff
@@ -41,15 +42,22 @@ public static void startQueuingTimer(State state, Class<?> schedulerClass, Objec
...
       QueueTiming timing =
           (QueueTiming) AgentTracer.get().getProfilingContext().start(Timer.TimerType.QUEUEING);
       timing.setTask(task);
       timing.setScheduler(schedulerClass);
       state.setTiming(timing);
     }
   }
+
+  public static void stopQueuingTimer(Timing timing) {
+    if (Platform.isNativeImageBuilder()) {
```
jbachorik, line 55: This is specific for when the native image is being built.
Do you want `Platform.isNativeImage()` instead?

richardstartin: I don't think they have a practical difference at runtime. One binds the value of `"org.graalvm.nativeimage.builder".equals(System.getProperty("jdk.module.main"))` to a constant in `Platform` - the other is an initialised-once constant via a holder. I just need to avoid `PerRecordingRateLimiter` being initialised in a native image builder, because then a good chunk of the tracer will also need to be initialised, so I don't think this makes any difference, does it?

jbachorik: `isNativeImageBuilder` is checking system property which turned out to be potentially problematic when running the native image.
But if this is just for avoiding initialization at the build time, this is fine.

richardstartin: What behaviour have I specified in a native image?

jbachorik: I meant that `isNativeImageBuilder` might not have been reliable in native image. 
If you need to have native image specific behavoir, `Platform.isNativeImage()` should be used.

richardstartin: I think it's correct the way it is, I just need to avoid initialising things at build time.

jbachorik (approved):

PR #7873, opened 2024-11-01
ByteBuddy Exception Suppression Checking

dougqh:
# What Does This Do

Introduces an annotation processor that checks ByteBuddy Advice.OnMethodEnter and Advice.OnMethodExit annotations to make sure they have the suppress property set

This annotation processor is now run automatically for all Java integrations.  Presently, the processor only emits warnings.
In a later pull request, this will likely be changed to error.

For situations where suppression isn't needed.  The check can be suppressed using a SuppressWarnings annotation.

# Motivation

Reduce cases where exception leak from our instrumentation into application / framework code.


amarziali (approved):
looks good! Can you please update the guide (https://github.com/DataDog/dd-trace-java/blob/master/docs/add_new_instrumentation.md) detailing how to use `@SuppressWarning` in case this check has to be shut down?

PerfectSlayer (approved):
Nice safety net 👍

PR #7886, opened 2024-11-05
Test support for reactor-kafka

amarziali:
# What Does This Do

This PR adds testing for reactor-kafka. There is no need for an instrumentation. The kafka one is already handling the observability also for the reactive version (since it uses `Flux.fromIterable(...)`)

# Motivation

# Additional Notes

# Contributor Checklist

Jira ticket: [PROJ-IDENT]


amarziali:
> I've noticed that the KafkaReactorForkedTests are identical in the latestDepTest and the test modules. It might be worth looking into why that is. Otherwise, it looks good to me.

Yes @ygree you are right. The latestDep are duplicated in another folder because of kafka client api being different on the latest version comparing to 0.11. This is not the same for the reactive api that are the same. however, to have the latest version, the test class has to be on latestDep hence the reason why is duplicated

PR #7905, opened 2024-11-07
Adjust crash upload timeout

dougqh:
# What Does This Do
Lowers the crash uploader default timeout

# Motivation
We don't want the crash tracker to block process shutdown for an extended period of time

# Additional Notes
While trying to make this change, I discovered that crash tracking isn't working on OS X because of the ancient version of bash that ships with OS X by default.

I've made the assertions a bit easier to debug and added some comments, so others won't spend time on this in the future.
In a later PR, it might be good to come back and make crash reporting work on OS X just to ease development.


PerfectSlayer:
> The current implementation of crash tracking doesn't work with ancient version of bash that ships with OS X by default.

Can you give a little more details about it?

PR #7913, opened 2024-11-07
Add peer service tag in dbm sql commenter

jordan-wong:
# What Does This Do
Adds a peer.service annotation to the SQL Commenter for DBM tracing. 

# Motivation
We want to add tags in line with the peer service model 

# Additional Notes

# Contributor Checklist

Jira ticket: [AIT-10029]

[AIT-10029]: https://datadoghq.atlassian.net/browse/AIT-10029?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ


jordan-wong:
Changed behavior according to discussion with Will. `peer.service` is inferred and is does not need to be manually set. We want to include the peer.service tag if a user enables peer.service tagging in their application, and as a result we see `peer.service` tag on our span. If we see a `peer.service` tag on our active span, we inject it into our SQL statement

dd-java-agent/instrumentation/jdbc/src/main/java/datadog/trace/instrumentation/jdbc/SQLCommenter.java
```diff
@@ -15,6 +19,7 @@ public class SQLCommenter {
   private static final String DATABASE_SERVICE = encode("dddbs");
   private static final String DD_HOSTNAME = encode("ddh");
   private static final String DD_DB_NAME = encode("dddb");
+  private static final String DD_DB_INSTANCE = "ddprs";
```
nayeem-kamal, line 22: I would rename this to dd_peer_service or something

dd-java-agent/instrumentation/jdbc/src/test/groovy/SQLCommenterTest.groovy
```diff
@@ -105,4 +111,39 @@ class SQLCommenterTest extends AgentTestRunner {
...
+        sqlWithComment = SQLCommenter.inject(query, dbService, dbType, host, dbName, traceParent, true, appendComment)
+      } else {
+        if (appendComment) {
+          sqlWithComment = SQLCommenter.append(query, dbService, dbType, host, dbName)
+        } else {
+          sqlWithComment = SQLCommenter.prepend(query, dbService, dbType, host, dbName)
+        }
+      }
+    }
+    sqlWithComment == expected
```
nayeem-kamal, line 137: This is already checked in the `then:` block.

dd-java-agent/instrumentation/jdbc/src/test/groovy/SQLCommenterTest.groovy
```diff
@@ -105,4 +111,39 @@ class SQLCommenterTest extends AgentTestRunner {
...
+    when:
+    String sqlWithComment = ""
+    runUnderTrace("testTrace"){
+      AgentSpan currSpan = AgentTracer.activeSpan()
+      currSpan.setTag(Tags.PEER_SERVICE, peerService)
+
+      if (injectTrace) {
+        sqlWithComment = SQLCommenter.inject(query, dbService, dbType, host, dbName, traceParent, true, appendComment)
+      } else {
+        if (appendComment) {
```
nayeem-kamal, line 130: This `if` branching structure is a bit hard to read. I think you could just do `else if(appendComment)` and then just a final `else` rather than having nested if statements like this

dd-java-agent/instrumentation/jdbc/src/test/groovy/SQLCommenterTest.groovy
```diff
@@ -105,4 +111,39 @@ class SQLCommenterTest extends AgentTestRunner {
...
+        }
+      }
+    }
+    sqlWithComment == expected
+
+    then:
+    sqlWithComment == expected
+
+    where:
+    query                                                                                                         | ddService      | ddEnv  | dbService    | dbType     | host | dbName | ddVersion     | injectTrace | appendComment | traceParent                                               | peerService | expected
```
nayeem-kamal, line 143: we can probably update this to only have the cases that include `ddprs` since that is what we want to test here. the other cases should already be tested in other tests so no need to repeat it.

dd-java-agent/instrumentation/jdbc/src/main/java/datadog/trace/instrumentation/jdbc/SQLCommenter.java
```diff
@@ -213,6 +230,34 @@ protected static boolean toComment(
...
+  }*/
+
+  protected static boolean toComment(
+      StringBuilder sb,
+      final boolean injectTrace,
+      final String parentService,
+      final String dbService,
+      final String hostname,
+      final String dbName,
+      final String dbInstance,
```
nayeem-kamal, line 242: lets rename this to peerService for cleanliness

nayeem-kamal (commented):
Nice work. Just a few comments to address.

nayeem-kamal (approved):
LGTM

dd-java-agent/instrumentation/jdbc/src/main/java/datadog/trace/instrumentation/jdbc/SQLCommenter.java
```diff
@@ -94,13 +99,21 @@ public static String inject(
       }
     }
 
+    AgentSpan currSpan = activeSpan();
+    Object peerServiceObj = null;
+    if (currSpan != null) {
+      peerServiceObj = currSpan.getTag(Tags.PEER_SERVICE);
```
amarziali, line 105: @jordan-wong is the goal to add the `peer.service` tag that the user manually set (this is rare) or the one that we are supposed to calculate in the tracer?

jordan-wong: The goal is to just add the `peer.service` tag that we are already supposed to calculate in the tracer

PR #7923, opened 2024-11-08
Replace scanner-based smap + annotations parsing with hand-crafted parsers

MattAlp:
# What Does This Do
Replaces the scanner-and-matcher-based NMT map annotation + `/proc/smaps` parsing with handcrafted parsers instead.

# Motivation
The usage of `Pattern`, `Matcher`, and `Scanner` by the smap JFR events allocates and reserves a certain amount of baseline memory that has been flagged by customers. This replaces that to use a minimal amount of regex and eliminate the heaviest allocations.

# Additional Notes
See PROF-10699 for an example of 'significant' memory usage.

# Contributor Checklist

Jira ticket: [PROF-10699]

[PROF-10699]: https://datadoghq.atlassian.net/browse/PROF-10699?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ

### DIFF
--- a/dd-java-agent/agent-profiling/profiling-controller-openjdk/src/main/java/com/datadog/profiling/controller/openjdk/events/SmapEntryFactory.java
+++ b/dd-java-agent/agent-profiling/profiling-controller-openjdk/src/main/java/com/datadog/profiling/controller/openjdk/events/SmapEntryFactory.java
@@ -3,10 +3,10 @@ package com.datadog.profiling.controller.openjdk.events;
 import datadog.trace.api.Platform;
 import datadog.trace.bootstrap.instrumentation.jfr.JfrHelper;
 import de.thetaphi.forbiddenapis.SuppressForbidden;
-import java.io.BufferedReader;
 import java.io.FileNotFoundException;
 import java.io.FileReader;
 import java.io.IOException;
+import java.io.PushbackReader;
 import java.lang.management.ManagementFactory;
 import java.util.ArrayList;
 import java.util.Arrays;
@@ -174,10 +174,10 @@ public class SmapEntryFactory {
     }
   }
 
-  private static long readLong(BufferedReader br, char delimiter, int base) throws IOException {
+  private static long readLong(PushbackReader reader, char delimiter, int base) throws IOException {
     long number = 0;
     char ch;
-    while ((ch = (char) br.read()) != delimiter) {
+    while ((ch = (char) reader.read()) != delimiter) {
       number *= base;
       number += Character.digit(ch, base);
     }
@@ -185,34 +185,34 @@ public class SmapEntryFactory {
     return number;
   }
 
-  private static int readInt(BufferedReader br, char delimiter, int base) throws IOException {
+  private static int readInt(PushbackReader reader, char delimiter, int base) throws IOException {
     int number = 0;
     char ch;
-    while ((ch = (char) br.read()) != delimiter) {
+    while ((ch = (char) reader.read()) != delimiter) {
       number *= base;
       number += Character.digit(ch, base);
     }
     return number;
   }
 
-  private static void readUntil(BufferedReader br, char target) throws IOException {
+  private static void readUntil(PushbackReader reader, char target) throws IOException {
     char ch = '\n';
     while (ch != target) {
-      ch = (char) br.read();
+      ch = (char) reader.read();
     }
   }
 
-  private static String readStringUntil(BufferedReader br, char target) throws IOException {
+  private static String readStringUntil(PushbackReader reader, char target) throws IOException {
     StringBuilder sb = new StringBuilder();
     char ch = '\n';
     while (ch != target) {
-      ch = (char) br.read();
+      ch = (char) reader.read();
       sb.append(ch);
     }
     return sb.toString();
   }
 
-  private static SmapHeader parseLine(BufferedReader br) throws IOException {
+  private static SmapHeader parseLine(PushbackReader reader) throws IOException {
     long startAddress;
     long endAddress;
     String perms;
@@ -221,19 +221,19 @@ public class SmapEntryFactory {
     int inode;
     String pathname;
 
-    startAddress = readLong(br, '-', 16);
+    startAddress = readLong(reader, '-', 16);
     if (startAddress == 0xffffffffff600000L) {
       startAddress = -0x1000 - 1;
       endAddress = -1;
-      readUntil(br, ' ');
+      readUntil(reader, ' ');
     } else {
-      endAddress = readLong(br, ' ', 16);
+      endAddress = readLong(reader, ' ', 16);
     }
-    perms = readStringUntil(br, ' ');
-    offset = readLong(br, ' ', 16);
-    dev = readStringUntil(br, ' ');
-    inode = readInt(br, ' ', 10);
-    pathname = readStringUntil(br, '\n').trim();
+    perms = readStringUntil(reader, ' ');
+    offset = readLong(reader, ' ', 16);
+    dev = readStringUntil(reader, ' ');
+    inode = readInt(reader, ' ', 10);
+    pathname = readStringUntil(reader, '\n').trim();
 
     return new SmapHeader(endAddress, startAddress, perms, offset, dev, inode, pathname);
   }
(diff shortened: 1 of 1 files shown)


dd-java-agent/agent-profiling/profiling-controller-openjdk/src/main/java/com/datadog/profiling/controller/openjdk/events/SmapEntryFactory.java
```diff
@@ -91,34 +86,49 @@ private static HashMap<Long, String> getAnnotatedRegions() {
...
+          } else {
+            if (line.startsWith("Total")) {
+              break;
             }
-            String description = matcher.group(6);
-            annotatedRegions.put(startAddress, description);
-            if (description.isEmpty()) {
-              annotatedRegions.put(startAddress, "UNDEFINED");
-            } else if (description.startsWith("STACK")) {
+          }
```
jbachorik, line 104: Could you add comments about the assumptions?
Eg. that there always must be line with 'size' etc.

dd-java-agent/agent-profiling/profiling-controller-openjdk/src/main/java/com/datadog/profiling/controller/openjdk/events/SmapEntryFactory.java
```diff
@@ -91,34 +86,49 @@ private static HashMap<Long, String> getAnnotatedRegions() {
...
+
+          String[] segments = line.split("\\s+");
+          long startAddress;
+          if (segments[0].equals("0x" + VSYSCALL_START_ADDRESS)) {
+            startAddress = -0x1000 - 1;
+          } else {
+            startAddress = Long.decode(segments[0]);
+          }
```
jbachorik, line 112: It would be good to add a defensive check if the split would not yield any elements.
Not very probable, but murphy's laws etc.

dd-java-agent/agent-profiling/profiling-controller-openjdk/src/main/java/com/datadog/profiling/controller/openjdk/events/SmapEntryFactory.java
```diff
@@ -91,34 +86,49 @@ private static HashMap<Long, String> getAnnotatedRegions() {
...
               annotatedRegions.put(startAddress, "STACK");
             } else if (description.startsWith("[") || description.startsWith("/")) {
               annotatedRegions.put(startAddress, "SYSTEM");
+            } else if (description.startsWith("CDS")) {
+              annotatedRegions.put(startAddress, "CDS");
+            } else if (description.startsWith("INTERN")) {
+              annotatedRegions.put(startAddress, "INTERN");
             } else {
-              annotatedRegions.put(startAddress, description.split("\\s+")[0]);
+              annotatedRegions.put(startAddress, description);
```
jbachorik, line 125: Nit - it seems like all descriptors are upper case
```suggestion
              annotatedRegions.put(startAddress, description.toUpperCase());
```

dd-java-agent/agent-profiling/profiling-controller-openjdk/src/main/java/com/datadog/profiling/controller/openjdk/events/SmapEntryFactory.java
```diff
@@ -170,137 +179,194 @@ static List<? extends Event> collectEvents() {
...
-          // address 0x1000 less than the end address to keep relative sizing correct
+
+        buffer.setLength(0);
+        char[] chars = line.toCharArray();
+        int i = 0;
+
+        while (chars[i] != '-') {
+          buffer.append(chars[i]);
+          i++;
+        }
```
jbachorik, line 200: Nit - I wonder how this compares to `indexOf()` + `substring()`
But no need to change this for now - probably the perf diff won't be dramatic

dd-java-agent/agent-profiling/profiling-controller-openjdk/src/main/java/com/datadog/profiling/controller/openjdk/events/SmapEntryFactory.java
```diff
@@ -170,137 +179,194 @@ static List<? extends Event> collectEvents() {
...
+        buffer.insert(0, new char[] {'0', 'x'});
```
jbachorik, line 201: `new char[] {'0', 'x'}` could be hoisted out of the loop so we do not keep allocation that array

dd-java-agent/agent-profiling/profiling-controller-openjdk/src/main/java/com/datadog/profiling/controller/openjdk/events/SmapEntryFactory.java
```diff
@@ -170,137 +179,194 @@ static List<? extends Event> collectEvents() {
...
+        startAddress = Long.decode(buffer.toString());
+
+        if (buffer.toString().equals("0x" + VSYSCALL_START_ADDRESS)) {
```
jbachorik, line 209: Hm, wouldn't the `startAddress == VSYSCALL_START_ADDRESS` if `VSYSCALL_START_ADDRESS` is a numeric value instead of String?

dd-java-agent/agent-profiling/profiling-controller-openjdk/src/main/java/com/datadog/profiling/controller/openjdk/events/SmapEntryFactory.java
```diff
@@ -170,137 +179,194 @@ static List<? extends Event> collectEvents() {
...
+        }
+        perms = buffer.toString();
+        buffer.setLength(0);
+        i++;
+
+        while (chars[i] != ' ') {
+          buffer.append(chars[i]);
+          i++;
+        }
+        buffer.insert(0, new char[] {'0', 'x'});
```
jbachorik, line 240: Hoisting

dd-java-agent/agent-profiling/profiling-controller-openjdk/src/main/java/com/datadog/profiling/controller/openjdk/events/SmapEntryFactory.java
jbachorik: Oh wait, this is just to make the String formatted like `0x...` ?
In that case I would strongly suggest computing the number on the fly and avoid this string construction only to get it parsed back a few moments later.

jbachorik: > I was wondering if it may be elided entirely by the compiler

Probably not - it is escaping the current scope since it is passed as an argument. This means that it must become a full allocation.

dd-java-agent/agent-profiling/profiling-controller-openjdk/src/main/java/com/datadog/profiling/controller/openjdk/events/SmapEntryFactory.java
jbachorik: Well, you could precompute its numeric value and use that for comparison here.
I mean, if you do something like `Long.parse("0x" + VSYSCALL_START_ADDRESS)` you should get the numeric representation of that string, right?

jbachorik (changes requested):
I think that removing the complexity of reading the hexa numbers char-by-char just to have them parsed back to numeric values can be avoided by computing the number on the fly. That will simplify the code and reduce string operations.

dd-java-agent/agent-profiling/profiling-controller-openjdk/src/main/java/com/datadog/profiling/controller/openjdk/events/SmapEntryFactory.java
jbachorik: 1. `buffer.toString()` is an extra allocation (new String is created)
2. String comparison is value-by-value whereas for the number you will get effectively one CPU instruction

jbachorik:
Any progress on this?

dd-java-agent/agent-profiling/profiling-controller-openjdk/src/main/java/com/datadog/profiling/controller/openjdk/events/SmapEntryFactory.java
```diff
@@ -44,6 +40,33 @@ private enum ErrorReason {
     VM_MAP_PARSING_ERROR,
   }
 
+  private static class SmapHeader {
+    long startAddress = 0;
```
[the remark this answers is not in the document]
jbachorik: This Datadog static code analyzer. It has been added to all DD repos recently.

dd-java-agent/agent-profiling/profiling-controller-openjdk/src/main/java/com/datadog/profiling/controller/openjdk/events/SmapEntryFactory.java
```diff
@@ -170,155 +280,143 @@ static List<? extends Event> collectEvents() {
...
+              thpEligible = buffer.toString().equals("1");
+            } else if ("ProtectionKey".equals(attributeName)) {
+              // Original event did not include protection key attribute, so skipping for now
               encounteredForeignKeys = true;
-              break;
+            } else {
+              System.out.println("ATTR: " + attributeName);
+              switch (attributeName) {
+                case "Size":
+                  size = Long.decode(buffer.toString()) * 1024;
```
jbachorik, line 332: I see that this is still using `string->stringBuffer->number` conversion.
I would propose extracting the logic from `parseLine` method for converting the stream of characters into a number.

Something like:
```
long readLong(String line, int from, char delimiter, int base) throws IndexOutOfBoundsException {
  long number = 0;
  for (int i = from; i < line.length() && (char c = line.charAt(i)) != delimiter; i++) {
    number *= base;
    number += Character.digit(char, base);
  }
  return number;
}
``` 

With this you should also unify the approach for reading the char at i-th position. Currently, one method is using `charAt(i)` while the other first converts the line into char array and then accesses that array elements.
I have no strong preference for one or the other - it would just be nice to keep to the same convention :)

dd-java-agent/agent-profiling/profiling-controller-openjdk/src/main/java/com/datadog/profiling/controller/openjdk/events/SmapEntryFactory.java
```diff
@@ -170,155 +280,143 @@ static List<? extends Event> collectEvents() {
...
+            }
+
+            if ("ThpEligible".equals(attributeName)) {
+              thpEligible = buffer.toString().equals("1");
+            } else if ("ProtectionKey".equals(attributeName)) {
+              // Original event did not include protection key attribute, so skipping for now
               encounteredForeignKeys = true;
-              break;
+            } else {
+              System.out.println("ATTR: " + attributeName);
```
jbachorik, line 329: Dangling debug output

jbachorik (changes requested):

dd-java-agent/agent-profiling/profiling-controller-openjdk/src/main/java/com/datadog/profiling/controller/openjdk/events/SmapEntryFactory.java
jbachorik: Let's try with `ByteBuffer`

dd-java-agent/agent-profiling/profiling-controller-openjdk/src/main/java/com/datadog/profiling/controller/openjdk/events/SmapEntryFactory.java
```diff
@@ -91,34 +115,54 @@ private static HashMap<Long, String> getAnnotatedRegions() {
...
-            } else if (description.startsWith("STACK")) {
+            continue;
+          } else {
+            // The final line contains the total number of mappings and their net vsize
+            if (line.startsWith("Total")) {
+              break;
+            }
+          }
+
+          String[] segments = line.split("\\s+");
```
jbachorik, line 137: This will expand to full regex - I think this should do the same thing
```suggestion
          String segment = "";
          int idx = line.indexOf(' ');
          if (idx > - 1) {
            segment = line.substring(0, idx);
          }
          ...
```

dd-java-agent/agent-profiling/profiling-controller-openjdk/src/main/java/com/datadog/profiling/controller/openjdk/events/SmapEntryFactory.java
```diff
@@ -130,21 +174,76 @@ private static HashMap<Long, String> getAnnotatedRegions() {
...
+
+  private static int readInt(PushbackReader reader, char delimiter, int base) throws IOException {
+    int number = 0;
+    char ch;
+    while ((ch = (char) reader.read()) != delimiter) {
+      number *= base;
+      number += Character.digit(ch, base);
+    }
+    return number;
+  }
```
jbachorik, line 196: I think you can safely have just `readLong()` and then (potentially) cast the result when you are using it. The parsing logic is 100% the same.

dd-java-agent/agent-profiling/profiling-controller-openjdk/src/main/java/com/datadog/profiling/controller/openjdk/events/SmapEntryFactory.java
```diff
@@ -130,21 +174,76 @@ private static HashMap<Long, String> getAnnotatedRegions() {
...
   }
 
-  @SuppressForbidden // split with one-char String use a fast-path without regex usage
-  static List<? extends Event> collectEvents() {
-    if (!SMAP_ENTRY_EVENT.isEnabled()) {
-      return Collections.emptyList();
+  private static long readLong(PushbackReader reader, char delimiter, int base) throws IOException {
+    long number = 0;
+    char ch;
+    while ((ch = (char) reader.read()) != delimiter) {
```
jbachorik, line 180: I think you need to check for the reader limit - eg. that `reader.read()` does not return `-1` 
If I read this correctly, you would start reading the EOF in an endless loop.
The same for the rest of the parsing functions.

I would also strongly suggest creating tests for the parsing part - I started playing with this conversion a while back, it is not finalized but it contains the test data (https://github.com/DataDog/dd-trace-java/compare/master...jb/smap) so you can just take this data and write your tests.

I would feel much more confident if we had the parsing covered by tests.

PR #7940, opened 2024-11-12
Add trace injection for prepared statements in Postgres

nenadnoveljic:
# What Does This Do

Adding trace injection for prepared statements in Postgres.

# Motivation

90% of APM/DBM correlation attempts for Postgres involve prepared statements, but we can't retrieve execution plans for these. Injecting trace context into prepared statements will greatly improve the success rate.

# Additional Notes

It's not possible to inject trace context into comments for prepared statements. Instead, the tracer sets the application_name with the trace parent, which the Agent samples via `pg_stat_activity.application_name`, and the backend parses.

Disadvantages:

- Setting application_name requires an extra database round trip.
- The original application_name is overwritten.

These drawbacks will be noted in the public documentation.

The feature is disabled by default and can be enabled by setting `dd.dbm.trace_prepared_statements` to `true`.

image

# Contributor Checklist

Jira ticket: [PROJ-IDENT]


internal-api/src/main/java/datadog/trace/bootstrap/instrumentation/api/InstrumentationTags.java
```diff
@@ -107,6 +107,7 @@ public class InstrumentationTags {
   public static final String TWILIO_STATUS = "twilio.status";
   public static final String TWILIO_PARENT_SID = "twilio.parentSid";
   public static final String DBM_TRACE_INJECTED = "_dd.dbm_trace_injected";
+  public static final String TIME_MS = "dd.instrumentation.time_ms";
```
mcculls, line 110: `TIME_MS` is too vague as a constant name, it should be `INSTRUMENTATION_TIME_MS`

dd-java-agent/instrumentation/jdbc/src/main/java/datadog/trace/instrumentation/jdbc/JDBCDecorator.java
```diff
@@ -312,6 +320,59 @@ public long setContextInfo(Connection connection, DBInfo dbInfo) {
...
+    final long startTime = System.currentTimeMillis();
+    try {
+
+      Integer priority = span.forceSamplingDecision();
+      if (priority == null) {
+        return;
+      }
+      final String traceParent = DECORATE.traceParent(span, priority);
+      if (traceParent == null
+          || !traceParent.matches("^00-[a-f0-9]{32}-[a-f0-9]{16}-[a-f0-9]{2}$")) {
```
amarziali, line 345: this can be precompiled since in this way it looks pretty inefficient

dd-java-agent/instrumentation/jdbc/src/main/java/datadog/trace/instrumentation/jdbc/JDBCDecorator.java
```diff
@@ -312,6 +320,59 @@ public long setContextInfo(Connection connection, DBInfo dbInfo) {
...
+          "Failed to set extra DBM data in application_name for trace {}. "
+              + "To disable this behavior, set trace_prepared_statements to 'false'. "
+              + "See https://docs.datadoghq.com/database_monitoring/connect_dbm_and_apm/ for more info.{}",
+          span.getTraceId().toHexString(),
+          e);
+      DECORATE.onError(span, e);
+    } finally {
+      span.setTag(DBM_TRACE_INJECTED, true);
+      final long elapsed = System.currentTimeMillis() - startTime;
+      span.setTag("dd.instrumentation.time_ms", elapsed);
```
amarziali, line 372: this should be in a constant (i.e `InstrumentationTags` can be a good place for it)

dd-java-agent/instrumentation/jdbc/src/main/java/datadog/trace/instrumentation/jdbc/JDBCDecorator.java
```diff
@@ -312,6 +320,59 @@ public long setContextInfo(Connection connection, DBInfo dbInfo) {
...
+
+      // SET doesn't work with parameters
+      StringBuilder sql = new StringBuilder();
+      sql.append("SET application_name = '");
+      sql.append(traceContext);
+      sql.append("';");
+
+      try (Statement statement = connection.createStatement()) {
+        statement.execute(sql.toString());
+      } catch (SQLException e) {
```
amarziali, line 358: this can be widen to `Throwable`

dd-java-agent/instrumentation/jdbc/src/main/java/datadog/trace/instrumentation/jdbc/JDBCDecorator.java
```diff
@@ -312,6 +320,59 @@ public long setContextInfo(Connection connection, DBInfo dbInfo) {
...
+        throw e;
+      }
+    } catch (Exception e) {
```
amarziali, line 361: This can be widen to Throwable

dd-java-agent/instrumentation/jdbc/src/main/java/datadog/trace/instrumentation/jdbc/JDBCDecorator.java
```diff
@@ -312,6 +320,59 @@ public long setContextInfo(Connection connection, DBInfo dbInfo) {
...
+      log.debug(
```
amarziali, line 362: If the statement execution above fails I think that will be catched by the `SQLException` case. Is this log never printed?

dd-java-agent/instrumentation/jdbc/src/main/java/datadog/trace/instrumentation/jdbc/JDBCDecorator.java
```diff
@@ -312,6 +320,59 @@ public long setContextInfo(Connection connection, DBInfo dbInfo) {
...
+      // SET doesn't work with parameters
+      StringBuilder sql = new StringBuilder();
+      sql.append("SET application_name = '");
+      sql.append(traceContext);
+      sql.append("';");
+
+      try (Statement statement = connection.createStatement()) {
+        statement.execute(sql.toString());
+      } catch (SQLException e) {
+        throw e;
```
amarziali, line 359: Do you really want to throw this by our code? I think it should silently swapped and something printed

dd-java-agent/instrumentation/jdbc/src/main/java/datadog/trace/instrumentation/jdbc/JDBCDecorator.java
```diff
@@ -312,6 +320,59 @@ public long setContextInfo(Connection connection, DBInfo dbInfo) {
...
+    try {
+
+      Integer priority = span.forceSamplingDecision();
+      if (priority == null) {
+        return;
+      }
+      final String traceParent = DECORATE.traceParent(span, priority);
+      if (traceParent == null
+          || !traceParent.matches("^00-[a-f0-9]{32}-[a-f0-9]{16}-[a-f0-9]{2}$")) {
+        throw new IllegalArgumentException("Invalid trace parent: " + traceParent);
```
vandonr, line 346: rather than throwing here just to catch it below on line 361, it'd better to write some logic to handle this without an exception.

dd-java-agent/instrumentation/jdbc/src/main/java/datadog/trace/instrumentation/jdbc/JDBCDecorator.java
vandonr: besides, this sequence of code
```
      } catch (SQLException e) {
        throw e;
      }
    } catch (Exception e) {
```
is a bit bloated, you can achieve the same result by not catching here since there is no specific treatment and the exception is re-caught just after.

dd-java-agent/instrumentation/jdbc/src/main/java/datadog/trace/instrumentation/jdbc/JDBCDecorator.java
vandonr: they are not different `catch` clauses of the same try block, the other one is a catch within the try :p

dd-java-agent/instrumentation/jdbc/src/main/java/datadog/trace/instrumentation/jdbc/JDBCDecorator.java
vandonr: you even created the constant, you only need to use it here :)

dd-java-agent/instrumentation/jdbc/src/main/java/datadog/trace/instrumentation/jdbc/JDBCDecorator.java
dougqh: Recompiling a regex repeatedly is expensive.  It would be better to compile the regex into a Pattern object.

dd-java-agent/instrumentation/jdbc/src/main/java/datadog/trace/instrumentation/jdbc/JDBCDecorator.java
```diff
@@ -312,6 +320,59 @@ public long setContextInfo(Connection connection, DBInfo dbInfo) {
...
+      }
+      final String traceContext = "_DD_" + traceParent;
+
+      // SET doesn't work with parameters
+      StringBuilder sql = new StringBuilder();
```
dougqh, line 351: Sizing the StringBuilder will avoid an extra underlying array allocation and array copy happening here

dd-java-agent/instrumentation/jdbc/src/main/java/datadog/trace/instrumentation/jdbc/JDBCDecorator.java
nenadnoveljic: The exception will be propagated to the outer exception handler. I think that works fine.

dd-java-agent/instrumentation/jdbc/src/main/java/datadog/trace/instrumentation/jdbc/JDBCDecorator.java
nenadnoveljic: > Do you really want to throw this by our code? I think it should silently swapped and something printed

The error is printed in the outer `catch`, which catches the `SQLException` thrown by `statement.execute`.

> is a bit bloated, you can achieve the same result by not catching here since there is no specific treatment and the exception is re-caught just after.

The inner `try-with-resources/catch` block ensures the statement is automatically closed, preventing leaks. Without it, we'd need a `finally` block to check if `statement != null` and close it explicitly. This approach is cleaner and less verbose.

vandonr: I think you can write a "resource try" without a catch or a finally 🤔

internal-api/src/main/java/datadog/trace/bootstrap/instrumentation/api/InstrumentationTags.java
nenadnoveljic: Changed now.

dd-java-agent/instrumentation/jdbc/src/main/java/datadog/trace/instrumentation/jdbc/JDBCDecorator.java
nenadnoveljic: Changed now.

dd-java-agent/instrumentation/jdbc/src/main/java/datadog/trace/instrumentation/jdbc/JDBCDecorator.java
nenadnoveljic: Changed now.

dd-java-agent/instrumentation/jdbc/src/main/java/datadog/trace/instrumentation/jdbc/JDBCDecorator.java
nenadnoveljic: Changed now.

dd-java-agent/instrumentation/jdbc/src/main/java/datadog/trace/instrumentation/jdbc/JDBCDecorator.java
nenadnoveljic: Removed `catch`, thanks!

dd-java-agent/instrumentation/jdbc/src/main/java/datadog/trace/instrumentation/jdbc/JDBCDecorator.java
nenadnoveljic: Changed now.

dd-java-agent/instrumentation/jdbc/src/main/java/datadog/trace/instrumentation/jdbc/JDBCDecorator.java
nenadnoveljic: Changed now.

dd-java-agent/instrumentation/jdbc/src/main/java/datadog/trace/instrumentation/jdbc/JDBCDecorator.java
vandonr: you need to compile the regex in a class field, otherwise it'll be recompiled every time, and there is no benefit to it

nenadnoveljic: > you need to compile the regex in a class field, otherwise it'll be recompiled every time, and there is no benefit to it

I changed it, thanks!

> rather than throwing here just to catch it below on line 361, it'd better to write some logic to handle this without an exception.

The logic matches the `catch` block: log a debug message and add a Throwable to the span. Should I extract it into a method and reuse it instead of throwing IllegalArgumentException?

dd-java-agent/instrumentation/jdbc/src/main/java/datadog/trace/instrumentation/jdbc/JDBCDecorator.java
```diff
@@ -312,6 +324,60 @@ public long setContextInfo(Connection connection, DBInfo dbInfo) {
...
+      final String setCommandEnd = "';";
+      StringBuilder sql =
+          new StringBuilder(
+              setCommandBegin.length() + traceContext.length() + setCommandEnd.length());
+      sql.append(setCommandBegin);
+      sql.append(traceContext);
+      sql.append(setCommandEnd);
+
+      try (Statement statement = connection.createStatement()) {
+        statement.execute(sql.toString());
```
vandonr, line 364: actually the use of a StringBuilder is not really necessary here, since we're not looping on values for instance. We could write this in a much simpler way with no performance impact as
```suggestion
      try (Statement statement = connection.createStatement()) {
        statement.execute("SET application_name = '" + traceContext + "';");
```
since the `+` operator is automatically converted to a StringBuilder by the compiler.

dd-java-agent/instrumentation/jdbc/src/main/java/datadog/trace/instrumentation/jdbc/JDBCDecorator.java
vandonr: yeah I saw that... given that this is a case that's not supposed to happen in the normal course of things, I think it's _okay_ to use throw as some kind of `goto` here

nenadnoveljic:
Tests unrelated to my changes, like `:dd-java-agent:instrumentation:grpc-1.5:forkedTest`, are failing. Should I merge master into my branch to fix them?

nenadnoveljic:
How can we fix the failing `:dd-java-agent:instrumentation:lettuce-5:latestDepTest`? (Unrelated to my changes).

nenadnoveljic:
Anything else I need to do on this PR?

dd-java-agent/instrumentation/jdbc/src/main/java/datadog/trace/instrumentation/jdbc/JDBCDecorator.java
```diff
@@ -312,6 +324,51 @@ public long setContextInfo(Connection connection, DBInfo dbInfo) {
...
+        throw new IllegalArgumentException("Invalid trace parent: " + traceParent);
+      }
+      final String traceContext = "_DD_" + traceParent;
+
+      // SET doesn't work with parameters
+      try (Statement statement = connection.createStatement()) {
+        statement.execute("SET application_name = '" + traceContext + "';");
+      }
+    } catch (Throwable e) {
+      log.debug(
```
amarziali, line 358: I would suggest to use `if log.isDebugEnabled()` in this case since it will avoid computing `toHexString` if fails and, if it fails, it can fail very often

dd-java-agent/instrumentation/jdbc/src/main/java/datadog/trace/instrumentation/jdbc/JDBCDecorator.java
```diff
@@ -312,6 +324,51 @@ public long setContextInfo(Connection connection, DBInfo dbInfo) {
...
+        return;
+      }
+      final String traceParent = DECORATE.traceParent(span, priority);
+      if (traceParent == null || !traceParentPattern.matcher(traceParent).matches()) {
+        throw new IllegalArgumentException("Invalid trace parent: " + traceParent);
+      }
+      final String traceContext = "_DD_" + traceParent;
+
+      // SET doesn't work with parameters
+      try (Statement statement = connection.createStatement()) {
```
amarziali, line 354: Have you tried to set is in the connection directly without the need of shooting a sql request for it? I remember that something like `connection.setClientInfo("ApplicationName", ...)` would also work. I'm curious if the result is the same and if we can benchmark the two approaches

nenadnoveljic: I switched to connection.setClientInfo, which offers the same performance as using SQL but is safer against SQL injection. This allowed me to remove trace format checks. The overhead is an additional roundtrip to the database.

Performance observations:

Remote database:

Without instrumentation: ~100ms (network time).
image

With instrumentation (using SQL or setClientInfo): Adds one roundtrip, no difference in performance.
image

Local database:

Both methods impose ~zero overhead.
image

amarziali: Yes the internal `AbstractJdbc4Connection` executes as well a statement (the same but with sql escaping). Thanks for the benchmarks

amarziali (approved):

internal-api/src/main/java/datadog/trace/bootstrap/instrumentation/api/InstrumentationTags.java
```diff
@@ -107,6 +107,7 @@ public class InstrumentationTags {
   public static final String TWILIO_STATUS = "twilio.status";
   public static final String TWILIO_PARENT_SID = "twilio.parentSid";
   public static final String DBM_TRACE_INJECTED = "_dd.dbm_trace_injected";
+  public static final String INSTRUMENTATION_TIME_MS = "_dd.instrumentation.time_ms";
```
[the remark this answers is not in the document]
nenadnoveljic: `public static final String DBM_TRACE_INJECTED = "_dd.dbm_trace_injected";` was already there. Can we leave the refactoring for a separate PR?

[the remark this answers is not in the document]
amarziali: yes of course!

dorongold:
@nenadnoveljic Can you please clarify if trace injection for prepared statements is also supported for MySQL?
The [public documentation](https://docs.datadoghq.com/database_monitoring/connect_dbm_and_apm/?tab=java) currently says:

> Tracer versions 1.44 and greater: Enable the prepared statements tracing using one of the following methods:
        Set the system property dd.dbm.trace_prepared_statements=true
        Set the environment variable export DD_DBM_TRACE_PREPARED_STATEMENTS=true

Because it does not mention for which databases this is supported, it is implied that it is supported for all of them (Postgres, MySQL, SQL Server, Oracle).

nenadnoveljic:
@dorongold We've updated the public docs. Thanks for letting us know!

PR #7986, opened 2024-11-20
Increase lambda extension request timeout

nhulston:
# What Does This Do
Increases timeout for communicating with the Lambda extension.

# Motivation
Customers are occasionally getting
```
[dd.trace 2024-11-20 18:26:27:926 +0900] [main] ERROR datadog.trace.lambda.LambdaHandler - could not reach the extension, not injecting the context
java.io.InterruptedIOException: timeout
at datadog.okhttp3.RealCall.timeoutExit(RealCall.java:108)
at datadog.okhttp3.RealCall.execute(RealCall.java:97)
at datadog.trace.lambda.LambdaHandler.notifyEndInvocation(LambdaHandler.java:162)
at datadog.trace.agent.core.CoreTracer.notifyExtensionEnd(CoreTracer.java:956)
```
This should resolve this.

# Additional Notes

# Contributor Checklist

Jira ticket: [SLES-1952]

[SLES-1952]: https://datadoghq.atlassian.net/browse/SLES-1952?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ

### DIFF
--- a/dd-trace-core/src/main/java/datadog/trace/lambda/LambdaHandler.java
+++ b/dd-trace-core/src/main/java/datadog/trace/lambda/LambdaHandler.java
@@ -44,7 +44,7 @@ public class LambdaHandler {
   private static final String START_INVOCATION = "/lambda/start-invocation";
   private static final String END_INVOCATION = "/lambda/end-invocation";
 
-  private static final Long REQUEST_TIMEOUT_IN_S = 1L;
+  private static final Long REQUEST_TIMEOUT_IN_S = 3L;
   private static final int MAX_IDLE_CONNECTIONS = 5;
   private static final Long KEEP_ALIVE_DURATION = 300L;
 


dd-trace-core/src/main/java/datadog/trace/lambda/LambdaHandler.java
```diff
@@ -44,7 +44,7 @@ public class LambdaHandler {
   private static final String START_INVOCATION = "/lambda/start-invocation";
   private static final String END_INVOCATION = "/lambda/end-invocation";
 
-  private static final Long REQUEST_TIMEOUT_IN_S = 1L;
+  private static final Long REQUEST_TIMEOUT_IN_S = 3L;
```
purple4reina, line 47: Seems really weird that a 1 second timeout wouldn't be enough. Have you been able to reproduce these errors?

If you were able to reproduce the errors, did this change solve it?

PerfectSlayer (commented):
> Customers are occasionally getting

If customers are seeing this kind of error, we might want to add this PR in the release notes (as bug fix or enhancement, whatever fits best)

dd-trace-core/src/main/java/datadog/trace/lambda/LambdaHandler.java
nhulston: See https://datadoghq.atlassian.net/browse/SLES-1952 for more details.

Maybe implementing a retry mechanism would be a better solution, WDYT?

purple4reina: I think this solution is fine 😄 Since we've been able to reproduce the error in self monitoring, in a perfect world we would build a layer from this branch and test it out for a bit. But since this change is otherwise harmless, I think just going ahead with it is fine. Though, for the customer's sake, I still think we should test this change out in self monitoring to ensure it indeed fixes the issue.

purple4reina (approved):

dd-trace-core/src/main/java/datadog/trace/lambda/LambdaHandler.java
nhulston: Yes that's a good point. This _might_ solve the problem, but maybe not. We only saw it in self monitoring once every few days, so we'll have to wait a while and see if the messages ever pops up again.

PR #7997, opened 2024-11-22
IAST security controls: handling custom validation and sanitization methods

jandro996:
# What Does This Do

Add secure marks to custom input validators and sanitisers via env variable 

[RFC](https://docs.google.com/document/d/1j1hp87-2wJnXUGADZxzLnvKJmaF_Gd6ZR1hPS3LVguQ/edit?tab=t.0#heading=h.p8x5e2cgs22w) (Milestone 0) 

# Motivation

IAST frequently identifies vulnerabilities that are reported despite the fact that the application executes custom validation mechanisms before reaching those points, effectively preventing any exploitation. As a result, security teams receive unnecessary alerts about potential vulnerabilities that are not actually exploitable. 

There is a need for a mechanism to recognise when these custom validation methods are in place, to avoid false positives.

# Additional Notes

**Security Control Configuration Format**

To configure a list of security controls, follow the format and field specifications below. This format uses specific separators to structure each security control entry.

Field Specifications:

**Security Control Type:** Defines the type of control
**Accepted values:** INPUT_VALIDATOR or SANITIZER
**Secure Marks:** List of vulnerability types to apply. Possible values are defined in [vulnerability_schema](https://github.com/DataDog/experimental/blob/a63eb3b86977577f3df1348c6314bd7c204ea3f9/teams/asm/vulnerability_schema/vulnerability_schema.json#L129). Optionally, use * to indicate applicability to all types.
**Class/File:** Fully qualified class or file implementing the security control.
**Method:** Name of the method implementing the security control.
**Parameters(Optional):** Fully qualified class parameters. Is used to discern between overloaded methods. If it’s not used and we have overloading security control will be applied to all the methods
**Parameters to Validate (Optional):** Zero-based list of parameter positions to validate, where the first parameter is position 0. This field applies only to INPUT_VALIDATOR types and is used when not all parameters require validation.

Separators:

**; (Semicolon)** - Separates each security control.
**: (Colon)** - Separates each field within a security control.
**, (Comma)** - Separates items within a field that accepts a list.

```
INPUT_VALIDATOR:SQLI,XSS:com.example.Validator:validateInput:java.lang.String,bar.foo.Something,java.lang.Object:0,1;
SANITIZER:Secure:*:com.example.Sanitizer:sanitizeInput
```

Note: Each security control is separated by ;, each field within a security control is separated by :, and items in lists (e.g., Secure Marks or Parameters to Validate) are separated by ,.

# Contributor Checklist

Jira ticket: [APPSEC-55329]

[APPSEC-55329]: https://datadoghq.atlassian.net/browse/APPSEC-55329?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ


dd-java-agent/agent-iast/src/main/java/com/datadog/iast/securitycontrol/SecurityControlMethodAdapter.java
```diff
@@ -0,0 +1,60 @@
...
+      "datadog/trace/api/iast/securitycontrol/SecurityControlHelper";
+  public static final String METHOD = "setSecureMarks";
+  public static final String DESCRIPTOR = "(Ljava/lang/Object;I)V";
+  private final MethodVisitor mv;
+  private final SecurityControl securityControl;
+  private final String desc;
+
+  public SecurityControlMethodAdapter(
+      final MethodVisitor mv, final SecurityControl securityControl, final String desc) {
+    super(Opcodes.ASM8, mv);
```
smola, line 21: This is possibly not valid if the method uses newer Java?

dd-java-agent/agent-iast/src/test/java/foo/bar/securitycontrol/SecurityControlTestSuite.java
```diff
@@ -0,0 +1,24 @@
...
+
+  public static boolean validate(String input) {
+    return true; // dummy implementation
+  }
+
+  public static boolean validate(Object o, String input, String input2) {
+    return true; // dummy implementation
+  }
+
+  public static String sanitize(String input) {
```
smola, line 21: Can we also test `sanitize(String input, Object something)`?
We should probably test a few more variants here, like having an `int` at the end, or having a large number of parameters...

Also, what happens if the return type is not `String`? or if validate result is not a `boolean`?

jandro996: I'll add more test cases!

> Can we also test sanitize(String input, Object something)?
> We should probably test a few more variants here, like having an int at the end, or having a large number of parameters...

According to the RFC the params in the sanitizer are not relevant as you only need to specify them if you w want to set explicit an overloaded method as security control

> Also, what happens if the return type is not String? 
We will check if its tainted

> or if validate result is not a boolean?
It doesn't matter as the input_validator only cares about inputs

dd-java-agent/agent-iast/src/main/java/com/datadog/iast/securitycontrol/SecurityControlMethodAdapter.java
jandro996: i've updated it to ASM9, that is the most higher version that our library version supports (there is an experimental ASM10). I think that this cover new bytecode introduced until java 17

dd-smoke-tests/iast-util/src/testFixtures/groovy/datadog/smoketest/AbstractIastSpringBootTest.groovy
```diff
@@ -1,15 +1,17 @@
 package datadog.smoketest
 
+import static datadog.trace.api.config.IastConfig.IAST_DEBUG_ENABLED
```
Mariovido, line 3: I don't think this change is needed

dd-smoke-tests/iast-util/src/main/java/datadog/smoketest/springboot/controller/XssController.java
```diff
@@ -190,4 +192,52 @@ public void format4(final HttpServletRequest request, final HttpServletResponse
...
+      InputValidator.validateAll(s);
+      response.getWriter().write(s);
+    } catch (IOException e) {
+      throw new RuntimeException(e);
+    }
+  }
+
+  @GetMapping("/validate2")
+  @SuppressFBWarnings
+  public void validate2(final HttpServletRequest request, final HttpServletResponse response) {
```
Mariovido, line 220: This is not really important, but shouldn't the name be `validateAll2`?

dd-java-agent/agent-iast/src/main/java/com/datadog/iast/securitycontrol/SecurityControlMethodAdapter.java
jandro996: Sorry, Santi, my previous comment wasn’t accurate. Let me provide some additional context with this response.

First of all, when we instantiate a new `org.objectweb.asm.ClassReader`, it internally checks whether the class version is supported by the ASM version. If it isn’t, an `IllegalArgumentException` is thrown. To handle this, I’ve added a safeguard to log that exception and avoid breaking the code.

Secondly, my earlier explanation about the opcode was incorrect. In this case, the opcode represents the ASM API version and, for the `MethodVisitor`, it determines the functionalities available for that specific API version.

To confirm, the entire tracer uses ASM version **9.7**, which can be verified in the `libs.versions.toml `file.

dd-java-agent/agent-iast/src/main/java/com/datadog/iast/securitycontrol/SecurityControlMethodAdapter.java
```diff
@@ -0,0 +1,103 @@
...
+    for (int i = 0; i < types.length; i++) {
+      Type type = types[i];
+      boolean isPrimitive = isPrimitive(type);
+      if (allParameters) {
+        if (!isPrimitive) {
+          callInputValidation(parametersCount);
+        }
+      } else if (securityControl.getParametersToMark().contains(i)) {
+        if (isPrimitive) {
+          LOGGER.warn(
```
smola, line 76: Just something to verify: if the user does not specified a number parameter, do we output this warning or not? It seems to me that we should not, as not specifying the exact param, and a method having an extra boolean should not be a problem?

jandro996: > if the user does not specified a number parameter, do we output this warning or not?

The warning only appears when the client specifies a number parameter, when SecurityControlMethodAdapter#visitInsn is executed there are only two possibilities:

- All parameters needs validation
- The user specified the number of the parameters that need to be validated

Just to clarify if the user defines the ParameterTypes we check previously if the method should be transformed in SecurityControlMethodClassVisitor#shouldBeAdapted

smola (approved):

Mariovido (approved):

dd-java-agent/agent-iast/src/main/java/com/datadog/iast/securitycontrol/IastSecurityControlTransformer.java
```diff
@@ -0,0 +1,61 @@
...
+import org.slf4j.Logger;
+import org.slf4j.LoggerFactory;
+
+public class IastSecurityControlTransformer implements ClassFileTransformer {
+
+  private static final Logger LOGGER =
+      LoggerFactory.getLogger(IastSecurityControlTransformer.class);
+
+  private final List<SecurityControl> securityControls;
+  private final Set<String> classFilter;
```
manuel-alvarez-alvarez, line 21: Why don't you keep them as a `Map<String, List<SecurityControl>>`?

dd-java-agent/agent-iast/src/main/java/com/datadog/iast/securitycontrol/IastSecurityControlTransformer.java
```diff
@@ -0,0 +1,61 @@
...
+    }
+    List<SecurityControl> match = getSecurityControl(className);
+    if (match == null || match.isEmpty()) {
+      return null; // Do not transform classes that do not have a security control
+    }
+    try {
+      ClassReader cr = new ClassReader(classfileBuffer);
+      ClassWriter cw = new ClassWriter(cr, ClassWriter.COMPUTE_FRAMES);
+      ClassVisitor cv = new SecurityControlMethodClassVisitor(cw, match);
+      cr.accept(cv, 0);
```
manuel-alvarez-alvarez, line 48:
```suggestion
      cr.accept(cv, SKIP_DEBUG | SKIP_FRAMES);
```

I think you just need to visit the code, but you can safely ignore debug info and frames.

internal-api/src/main/java/datadog/trace/api/iast/securitycontrol/SecurityControl.java
```diff
@@ -0,0 +1,86 @@
...
+
+  private int marks;
+
+  @Nonnull private String className;
+
+  @Nonnull private String method;
+
+  @Nullable private List<String> parameterTypes;
+
+  @Nullable private Set<Integer> parametersToMark;
```
manuel-alvarez-alvarez, line 20: A `BitSet` might be better suited for this than a `Set`

dd-java-agent/agent-iast/src/main/java/com/datadog/iast/securitycontrol/IastSecurityControlTransformer.java
jandro996: Much better! thanks!

dd-java-agent/agent-iast/src/main/java/com/datadog/iast/securitycontrol/SecurityControlMethodAdapter.java
```diff
@@ -0,0 +1,103 @@
...
+
+    super.visitInsn(opcode);
+  }
+
+  private void processSanitizer() {
+    Type returnType = Type.getReturnType(desc);
+    if (isPrimitive(returnType)) {
+      // no need to check primitives as we are not tainting them
+      LOGGER.warn(
+          "Sanitizers should not be used on non-primitive return types. Return type {}. Security control: {}",
```
manuel-alvarez-alvarez, line 50:
```suggestion
          "Sanitizers should not be used on primitive return types. Return type {}. Security control: {}",
```

dd-java-agent/agent-iast/src/main/java/com/datadog/iast/securitycontrol/SecurityControlMethodAdapter.java
```diff
@@ -0,0 +1,103 @@
...
+          callInputValidation(parametersCount);
+        }
+      }
+      parametersCount += types[i].getSize();
+    }
+  }
+
+  private void callInputValidation(int i) {
+    // Duplicate the parameter value on the stack
+    mv.visitVarInsn(Opcodes.ALOAD, i);
```
manuel-alvarez-alvarez, line 91: In instance methods the `ALOAD_0` actually loads the this instance, this is fine for static method.

dd-java-agent/agent-iast/src/test/java/foo/bar/securitycontrol/SecurityControlTestSuite.java
```diff
@@ -0,0 +1,100 @@
+package foo.bar.securitycontrol;
+
+public class SecurityControlTestSuite {
+
+  public static boolean validateAll(long input, String input2) {
```
manuel-alvarez-alvarez, line 5: You have duplicate all the tests using instance methods (non static)

dd-java-agent/agent-iast/src/main/java/com/datadog/iast/securitycontrol/IastSecurityControlTransformer.java
```diff
@@ -34,28 +33,22 @@ public byte[] transform(
       Class<?> classBeingRedefined,
       java.security.ProtectionDomain protectionDomain,
       byte[] classfileBuffer) {
-    if (!classFilter.contains(className)) {
+    if (!securityControls.containsKey(className)) {
```
manuel-alvarez-alvarez, line 36: This is no longer required, as it's also checked in the `securityControls.get(className);`

dd-java-agent/agent-iast/src/main/java/com/datadog/iast/securitycontrol/SecurityControlMethodClassVisitor.java
```diff
@@ -0,0 +1,111 @@
...
+
+  private MethodVisitor adaptSanitizer(
+      final MethodVisitor mv,
+      final SecurityControl control,
+      final int accessFlags,
+      final Type method) {
+    if (isPrimitive(method.getReturnType())) {
+      // no need to check primitives as we are not tainting them
+      LOGGER.warn(
+          "Sanitizers should not be used on non-primitive return types. Return type {}. Security control: {}",
```
manuel-alvarez-alvarez, line 90:
```suggestion
          "Sanitizers should not be used on primitive return types. Return type {}. Security control: {}",
```

manuel-alvarez-alvarez (approved):

PR #8045, opened 2024-12-02
Improve crash tracking install logging

PerfectSlayer:
# What Does This Do

This PR avoids logging stack trace when failing to clean up previous crash tracking configuration files.

# Motivation

Avoid showing exception message to the customers for clean up operations

# Additional Notes

# Contributor Checklist

Jira ticket: [PROJ-IDENT]

### DIFF
--- a/dd-java-agent/agent-crashtracking/src/main/java/com/datadog/crashtracking/OOMENotifierScriptInitializer.java
+++ b/dd-java-agent/agent-crashtracking/src/main/java/com/datadog/crashtracking/OOMENotifierScriptInitializer.java
@@ -124,7 +124,11 @@ public final class OOMENotifierScriptInitializer {
           Files.walkFileTree(dir, new ScriptCleanupVisitor());
         }
       } catch (IOException e) {
-        LOG.warn("Failed cleaning up process specific files in {}", dir, e);
+        if (LOG.isDebugEnabled()) {
+          LOG.info("Failed cleaning up process specific files in {}", dir, e);
+        } else {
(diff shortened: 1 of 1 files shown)


r1viollet:
I will approve as this probably hurts our experience. Though I have a few extra questions as I am not too familiar with java side of things.
I assume the info level is not shown to customers by default ? Hence do we need the separate logging strategies ?
Why are we failing to do this cleanup ? Is there something else we should consider behind this failure ?

r1viollet (approved):
LGTM, thanks for looking at this.

PerfectSlayer:
> I assume the info level is not shown to customers by default ?

The level is part of the log entry/event so customers are aware of it.
In addition, level is used for filtering and it helps when it is consistency across the whole client library.

>  Hence do we need the separate logging strategies ?

When debug is enabled (and it is usually enabled by the customers when support ask them to), you will have the full stack trace. When debug is disabled, you will only get the message / summary reason of the failure. (having a `Throwable` as last parameter of a log call implicitly capture the whole stack trace as part of the log event/entry)

> Why are we failing to do this cleanup ? Is there something else we should consider behind this failure ?

The specific error was: `java.nio.file.AccessDeniedException` from the `UnixFileSystemProvider`. It can comes from multiples reasons (file lock, maybe already deleted if 2 cleaners are running in parallel, etc...). Exceptions are not uncommon with file I/O. 
About failure recovery, I don’t think there is much to do. The cleaner will be run again next time an instrumented process will start. And worst case scenario, a bunch of cfg files will stay in the tmp folder.

PR #8051, opened 2024-12-03
New tag for DSM checkpoints - a bitmask representing enabled products / features

kr-igor:
# What Does This Do

# Motivation

# Additional Notes

# Contributor Checklist

Jira ticket: [PROJ-IDENT]

### DIFF
--- a/dd-trace-core/src/main/java/datadog/trace/core/datastreams/DefaultPathwayContext.java
+++ b/dd-trace-core/src/main/java/datadog/trace/core/datastreams/DefaultPathwayContext.java
@@ -1,5 +1,6 @@
 package datadog.trace.core.datastreams;
 
+import static datadog.trace.core.datastreams.TagsProcessor.PRODUCTS_MASK;
 import static java.nio.charset.StandardCharsets.ISO_8859_1;
 import static java.nio.charset.StandardCharsets.UTF_8;
 import static java.util.concurrent.TimeUnit.MILLISECONDS;
@@ -52,6 +53,7 @@ public class DefaultPathwayContext implements PathwayContext {
   // direction != current direction
   private long closestOppositeDirectionHash;
   private String previousDirection;
+  private String productMaskTag;
 
   private static final Set<String> hashableTagKeys =
       new HashSet<String>(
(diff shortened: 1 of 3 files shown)


dd-trace-core/src/main/java/datadog/trace/core/datastreams/DefaultPathwayContext.java
```diff
@@ -129,7 +140,9 @@ public void setCheckpoint(
     try {
       // So far, each tag key has only one tag value, so we're initializing the capacity to match
       // the number of tag keys for now. We should revisit this later if it's no longer the case.
-      List<String> allTags = new ArrayList<>(sortedTags.size());
+      List<String> allTags = new ArrayList<>(sortedTags.size() + 1);
+      allTags.add(getProductMaskTag());
```
piochelepiotr, line 144: What do you think of putting that in the "hash of known tags"?
That means we need to update the hashOfKnownsTags every time we enable / disable data streams via remote configuration, but it would give a boost in performance.

kr-igor: That will add one more field to hash, so I'm not sure about performance gains (the check for enabled products just checks a few variables). Also changing the configuration will change pathway hashes if we hash using this tag. Not critical, but it will create duplicating pathways during the overlapping time period.

internal-api/src/main/java/datadog/trace/api/Config.java
```diff
@@ -3244,6 +3244,28 @@ public boolean isDataJobsEnabled() {
...
+    }
+    if (isDataJobsEnabled()) {
+      enabledProducts |= DJM_PRODUCT;
+    }
+    if (isProfilingEnabled()) {
+      enabledProducts |= PROFILING_PRODUCT;
+    }
+
+    return enabledProducts;
+  }
```
PerfectSlayer, line 3267: This does not feel related to the core API but to the DSM product.
Can you move it near to where it is used, ie within `dd-trace-core/src/main/java/datadog/trace/core/datastreams/`, so we decouple it later more easily?

kr-igor: Sounds good, I'll move this logic to DSM.

dd-trace-core/src/main/java/datadog/trace/core/datastreams/DefaultPathwayContext.java
piochelepiotr: Good point. Let's just not use it for hashing.

piochelepiotr: What do you think about putting this mask as a new field on the message pack? It could be a top level field on the payload instead of being set on each checkpoint.
So we could call the config right before writing the message pack in the writer.

kr-igor: Good idea! Moved the logic

PerfectSlayer (approved):
Looks good! Thanks for the update 👍

